PR #3, opened 2015-10-20
More tests for _

danvk:
A bug in the declaration of `_.zip`, caught by a test!

### DIFF
--- a/underscore/test.js
+++ b/underscore/test.js
@@ -2,6 +2,38 @@
 var _ = require('underscore');
 
 /**
+ * _.find
+ */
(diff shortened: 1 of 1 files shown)


samwgoldman:
I think the old type of `_.zip` was correct. `_.zip([1], ['a']) === [[1, 'a']]`, which is `[number,string][]`.

danvk:
Ah, you're right! I changed the declaration back and updated the tests.

It would be nice to turn these into executable tests (e.g. using Mocha), though if FlowTyped gets to be anywhere near the size of DefinitelyTyped, that could have problems scaling.

PR #5, opened 2016-01-18
add types of verbal-expressions

nmn:


jeffmo:
Thanks, @nmn! Would you mind adding a few tests (similar to what underscore has)?

Issue #4, opened 2015-10-21
How should these definitions be consumed?


splodingsocks:
As mentioned by @danvk, we need to figure out how best to instruct others to obtain and use these definitions. I'll be back next week to discuss more.

jeffmo:
cc @gabelevi who's working on some related things WRT interface definitions that can easily be shipped alongside (but decoupled from) the library implementation

danvk:
I do like the idea of libraries distributing their own typings.

Some pros:
- typings are tied to the version of the library to which they apply
- Installation is simple (type definitions come with the node module)

Some cons:
- Owners of major libraries may not care to distribute type definitions (especially older ones which aren't updated often)
- When there's a new version of Flow, it's harder for the Flow owners to check for regressions against _all_ type definitions (which they could do with this repo).

jeffmo:
> When there's a new version of Flow, it's harder for the Flow owners to check for regressions against all type definitions (which they could do with this repo).

Yea agree with the problem, though I see this as a separate problem that we need to work on even independent of interface definitions.

One thing that we do with our codebases to address a part of this is to explicitly set the `version=` option in the config. Then we use a wrapper script for the `flow` command that reads this and then runs the appropriate version of flow (rather than a single global version of Flow installed). I really think this is the right direction for the command you should get through homebrew/opam/flow-cli.

It doesn't quite solve another problem, though: What to do when two libraries used by a single project dictate that they rely on disjoint versions of flow? For this I don't have a good answer yet...but it's probably something we need to consider soon. We haven't had to address it with our own projects yet because we don't sub-version project dependencies...but with npm and such there's no getting around it as a common case.

magicpat:
@jeffmo As a starter, since there will be a lot of unknowns how to do the versioning and similar stuff... would it be a good idea to start a new cli-tool to retrieve definitions from this repository, similar how DefinitelyTyped's cli-tool does? I would be keen to invest some time in this and get something going.

splodingsocks:
@magicpat I think it's for sure a good idea. In fact, I think @jeffmo has already done some thinking in this area.

hallettj:
In my opinion, the simplest useful option would be to push FlowTyped to npm as-is. Users could include it as a dependency, and add lines in `.flowconfig` to point to the type definitions that they want to use. For example:

```
[libs]
node_modules/FlowTyped/underscore/underscore.js
```

[...]

PR #11, opened 2016-03-09
Move to new file structure

splodingsocks:
Move to the file structure suggested by @jeffmo in #4


marudor:
Looks jut like I've understood the Proposal by @jeffmo

magicpat:
It would be nice to also see a specific flow-version, not only `flow-all`... 
Also I am not 100% sure which versions of flow should be tried on the `flow-all` directory... how is that defined?

Other than that, looks really promising! I am hyped :raised_hands:

splodingsocks:
@ryyppy I wasn't quite sure how we're going to manage having multiple versions of flow-bin. I figured we'd address that need as soon as someone wants to contribute a definition that only works on a specific version of flow.

jeffmo:
This looks good!

> Also I am not 100% sure which versions of flow should be tested on the flow-all directory

I suppose the test runner should check against all of them to keep "-all" honest

> I wasn't quite sure how we're going to manage having multiple versions of flow-bin

I imagine it might be easier/faster for the runner to grab/use the [flow binaries](https://github.com/facebook/flow/releases) directly -- but I suppose if it's more convenient to have the runner do an ephemeral `npm install flow-bin@VERSION` for each test, that works too

splodingsocks:
That would certainly work.

I think I'll merge this, and we can do that in a different branch.

PR #12, opened 2016-03-10
Updates to Mori definitions

hallettj:
I switched a project from Immutable to Mori, and fixed a lot of type errors that cropped up during the process. During that process I made a number of fixes and improvements to the Mori type definitions.


splodingsocks:
Awesome, thanks @hallettj

PR #15, opened 2016-03-16
Add a test runner

jeffmo:
`flow-typed run-tests [opt-test-name-pattern]`

Running tests means the following: 

1. For each Package/Version/FlowVersion directory, find all versions of Flow that satisfy the specified Flow version
1. Copy the libdef + all test files into a directory with a .flowconfig specifying a "$ExpectError" suppression comment
1. Run `flow check` for each resolved version of Flow on the test directory. If there are any errors, assert them.

This also adds `run-tests` to the travis command


jeffmo:
I'm noticing some strange issues in Travis with Flow 0.22, so will need to debug those a bit later before this is ready.

The code should be mostly good for review, though

npm/src/commands/runTests.js
```diff
@@ -0,0 +1,417 @@
...
+          libDefPath: path.join(flowVerDirPath, pkgName + "-" + pkgVer + ".js"),
+          flowVersion: flowVer,
+        });
+      }));
+    }))
+  }))
+  return testGroups;
+}
+
+function binSearchArray<T>(arr: Array<T>, testFn: (elem: T) => number): ?T {
```
splodingsocks: Looks like this might not be used?

splodingsocks:
Looks great @jeffmo! Way to churn out code. I'm :+1: for merging when you're ready and Travis is passing. Did you say there's some code cleanup to do, too?

npm/src/commands/runTests.js
jeffmo: good catch

PR #17, opened 2016-03-18
Compress `Package/Version` into `Package-Version` + cleanup

jeffmo:
By compressing `PackageName/Version` into `PackageName-Version` it's a little easier to grok which package versions are available (by just scanning the dirs in `/definitions/`) and the tooling code gets negligibly simpler.

This also cleans up and simplifies some code before moving on to the postinstall script.


ryyppy:
Since you guys do the reviews, I just wanted to say that I stalk this PRs... I have a question:
You are using alot of `await` syntax, which is quite bleeding-edge, but you are using a lot of string concatenation instead of template-strings... why's that? :-)

npm/src/lib/definitions.js
```diff
@@ -5,67 +5,97 @@ import {fs, https, path} from "./node.js";
...
-  itemName: string,
-  itemStats: fs.Stats
-): boolean {
+export function versionToString(ver: Version): string {
+  const rangeStr = ver.range ? ver.range : '';
   return (
-    itemStats.isFile()
-    && itemName.substr(0, 'test-'.length) === 'test-'
-    && path.extname(itemName) === '.js'
+    rangeStr + 'v' + String(ver.major) + '.' + String(ver.minor) + '.' +
```
ryyppy: Why not something like:
```
const { range, major, minor, patch } = ver;
const rangeStr = range ? range : '';

return `${rangeStr}v${major}.${minor}.${patch}`;
);
```

I mean, it's not really an issue, but since this tool relies a lot of path-building, I find this much easier to read...

splodingsocks:
@ryyppy Thanks for bringing in feedback!

@jeffmo I think it looks good to merge whenever you're ready.

npm/src/commands/validateDefinitions.js
```diff
@@ -3,151 +3,83 @@
...
         }
-
-        const itemPath = path.join(pkgName, pkgVer, itemName);
-        if (itemStats.isFile()) {
+      } else if (itemStats.isFile()) {
+        if (!isTestFileName(itemName)) {
           errs.push(
-            "'" + itemPath + "': Only 'test-*' files are permitted in this " +
-            "directory!"
+            itemPath + ": All files under a package-version dir must be test " +
```
ryyppy: Also here, template-strings span over newlines:

```
errs.push(`${itemPath}: All files under a package-version dir must be test
                 files named test-SOMETESTNAME.js`);
```

jeffmo:
> you are using a lot of string concatenation instead of template-strings... why's that?

Old habits die hard :) You're totally right, this does clean a bunch of stuff up. Update coming...

> Since you guys do the reviews, I just wanted to say that I stalk this PRs...

Your feedback is definitely welcome -- the more eyes the better!

npm/src/commands/validateDefinitions.js
jeffmo: Unfortunately they also seem to pick up all the whitespace after the newline -- so this doesn't come out very well formatted :(

```
asdf: Only 'test-*' files are permitted in this 
            directory!
```

PR #23, opened 2016-04-06
`flow-typed libdefs-for-pkg --flowVersion=FLOW_VERSION <PATH_TO_PKGJSON_FILE>`

jeffmo:
This new command will take a package.json file, extract it's complete list of dependencies, and prints a list of urls from the flow-typed repo that satisfy as many of the dependencies as possible.

The part most worth discussing is the `_chooseLibDefVersion()` function (and its tests). The goal is to choose the *lowest* compatible libdef version. This ensures that a project with a package versioned at, say, `^1.1.0` will always receive a library definition that works with `1.1.0` and never `1.2.0` -- even though `1.2.0` is backward compatible.

This is important to ensure that all users who `npm install`ed back at version `1.1.0` will receive the same Flow interface as anyone who installs after version `1.2.0` is published.

NOTE: Building this has made it clear to me that comparing version ranges is tricky and has many many edge cases. After writing/testing/thinking hard about `_chooseLibDefVersion()` I'm sort of tempted to suggest that we just start out simple and only support libdef versions at non-ranged versions (that is: only versions like `1.0.0` or `1.x`) for now. Thoughts?


splodingsocks:
@jeffmo What would be the disadvantage of choosing the simple route, and only supporting non-ranged libdefs? At the moment, I can't think of a downside. In fact, I'd prefer that as a user.

marudor:
I'm all in for simplicity to get it started.
If we really see a need for ranges we can come back to see how exactly we implement it.
But for now simple stuff is good stuff.

ryyppy:
I also think it's a good start!

splodingsocks:
@jeffmo I like the direction you're heading here. Would you prefer to revise the code for the simplified case discussed above before merging?

npm/src/lib/node.js
```diff
@@ -30,6 +30,13 @@ export const fs = {
       });
     });
   },
+  readFile: function(f: string, opts: Object): Promise<Buffer> {
```
splodingsocks: How come you chose to write this node file, instead of using something like promisifyAll from Bluebird? 

http://bluebirdjs.com/docs/api/promise.promisifyall.html

npm/src/lib/github.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import GitHub from "github";
+
+const CLIENT = new GitHub({version: "3.0.0"});
+if (process.env.GH_TOK) {
```
splodingsocks: Does the user have to have GH_TOK set en their env in order for the CLI to work? I've never used the node GH client before.

jeffmo:
> What would be the disadvantage of choosing the simple route, and only supporting non-ranged libdefs? 

I think this makes a lot of sense and would simplify a lot. I'll update today

npm/src/lib/node.js
jeffmo: Mostly because Flow doesn't have a libdef for things like promisifyAll (which I think could be a little tricky to write). This way the API stays well typed.

npm/src/lib/github.js
jeffmo: No, it's optional. The only reason I added this is because the GitHub API will rate limit you more aggressively if you don't have a token -- so I added such a token to Travis to hopefully keep test runs from getting too flaky

splodingsocks:
I've pulled down your branch and done a little bit of refactoring, @jeffmo, in order to be able to share some functionality between CLI commands and fetching info from GH. What would you say is the best way for me to get those changes up? Merge this one, and then you and I can both open PRs with our refactorings?

npm/src/lib/node.js
splodingsocks: That makes sense. :+1:

npm/src/lib/github.js
splodingsocks: Nice! :package:

splodingsocks:
You can see my changes here: https://github.com/flowtype/flow-typed/commit/f79a3e9cc3c46f59236975ba3b2867ded9befd80

jeffmo:
@splodingsocks: since this PR is coming from my fork, I made you a collaborator on that fork. Maybe you can just push to this branch there?

splodingsocks:
Done! But I didn't test any of my changes :astonished:. I'll run the script and see if anything breaks now.

splodingsocks:
Okay, things appear to be working alright.

jeffmo:
testinprod

npm/src/lib/github.js
```diff
@@ -10,3 +19,119 @@ if (process.env.GH_TOK) {
...
+/**
+ * Memoizes the operation of retrieving all package library definition info and
+ * organizing it.
+ *
+ * Note that this initial fetch only contains Package/PackageVersion info. It
+ * doesn't include FlowVersion info -- which must be fetched separately for each
+ * package libdef.
+ */
+let _libdefs = null;
+export async function getAllShallowLibDefs(
```
jeffmo: nit: Since this stuff specifically pertains to libdefs (and is less concerned with where the libdefs are fetched from), would this make more sense over in `lib/libDef.js`?

splodingsocks: Sure, I think that'd be fine. My reasoning here was that since the implementation is tied directly to GitHub, and the GitHub client, we'd put it with the GitHub stuff. But I think it would make sense in lib/libDef also.

splodingsocks:
I'm trying to run the jest tests locally, and all of the semver tests are failing. I tried checking out the commit where you first added the tests, but they still don't pass.

```
● semver › stringToVersion › it parses a little of everything
  - Expected: undefined toEqual: {
    | major: 1,
    | minor: 2,
    | patch: 'x',
    | range: '>='
    }
```

Are they passing for you currently, @jeffmo ?

splodingsocks:
I did some more fishing around this morning. Looks like Jest wasn't actually disabling automocking. Once I added the following to the package.json, under `jest`:

```json
    "unmockedModulePathPatterns": [
      "<rootDir>/src/.*",
      "<rootDir>/dist/.*"
    ]
``` 

Then the semver tests all started to pass.

Now I'm debugging a failure in the `libdefsForPkg`:

```
TypeError: The super constructor to `inherits` must have a prototype.
    at Object.exports.inherits (util.js:755:11)
    at Object.<anonymous> (/Users/murphy/projects/flow-typed/npm/node_modules/sshpk/lib/private-key.js:43:6)
    at Object.<anonymous> (/Users/murphy/projects/flow-typed/npm/node_modules/sshpk/lib/utils.js:15:18)
```

jeffmo:
(...still digging, just an update from my end...)

Looking into the `The super constructor to `inherits` must have a prototype.` issue... It seems related to simply `import`-ing the `"request"` module.

According to `npm ls`: `sshpk` -> `http-signature@1.1.1` -> `request@2.69.0`.

It looks like request@2.69 shipped in early february and 2.70 shipped last week -- but the issue repros with both. `http-signature` hasn't shipped a new version in 3 months, so this is a bit odd...

jeffmo:
This is out-dated and messy and should really be built on top of the `install` command. I still think we should build this, though, so I'm just going to close this for now and can come back to it later after the `install` command is up and running

PR #27, opened 2016-04-14
Search Command

splodingsocks:
I'm still in the middle of this, and I've got to run for this morning, but I wanted to get something up here for you to take a look at.


npm/flow-typed/jest.js.flow
```diff
@@ -2,6 +2,7 @@ declare function beforeEach(cb: () => void): void;
 declare function describe(suiteName: string, cb: () => void): void;
 declare function expect(value: any): Object;
 declare function it(testName: string, cb: (done: () => void) => void): void;
+declare function pit(testName: string, cb: () => Promise<any>): void;
```
jeffmo: Return type of the callback here should be `Promize<void>` I think

npm/src/lib/libDef.js
```diff
@@ -346,3 +354,48 @@ export async function getShallowGHLibDefs(
...
+    repo: 'flow-typed',
+    path: `/definitions/${def.libDef.pkgNameVersionStr}`
+  })
+  .flatMap(identity)
+  .filter(notNull)
+  .flatMap(identity)
+  .map(i => i.name)
+  .filter(name => name.indexOf('flow-') == 0)
+  .map(name => {
+    let withoutPrefix = name.match(/flow-(.*)/)[1]
```
jeffmo: nit: might as well be `const`, eh?

(PS: "nit" === I don't feel strongly, feel free to object/ignore)

jeffmo: Also, surprised Flow is ok with this... .match() returns nullable (null if it doesn't find any matching patterns)

jeffmo: (Ahh, prob because flow doesn't have a libdef for RX)

jeffmo:
Excite!

npm/src/lib/libDef.js
splodingsocks: Good points!

splodingsocks:
Oh, nice! It looks like flow 0.23 is out! And it broke our tests :)

splodingsocks:
@jeffmo Thanks for reviewing even while on PTO :). I've added a search command, now! Should only be a few short steps to finish that command up, and then add an install command.

I still need to make the search command check to see what the current flow version is for the project. Do you think I should do that by running `flow version` and hope they have it in the path? Or should I look in the package.json for a `flow-bin`? Or should I do both? First looking at the package.json?

I think it'd be best to do both.

splodingsocks:
K, I've done both. I think this could be merged now, unless you think it needs more automated tests?

npm/src/commands/search.js
```diff
@@ -0,0 +1,54 @@
...
+
+function findFlowVersion(): string {
+  const nodeModulesBin = path.resolve('node_modules', '.bin')
+  const _path = `${nodeModulesBin}:${process.env.PATH}`
+
+  function fail() {
+    console.error('Couldn\'t identify your flow version. Please make sure ' +
+    'flow is installed and in your PATH, or flow-bin is installed ' +
+    'in your project.')
+    process.exit(-1)
```
jeffmo: Probably better to throw an `Error` than to `process.exit()`. That way anyone who potentially ends up using this elsewhere could, say, wrap the error message with some context (etc).

npm/src/commands/search.js
```diff
@@ -0,0 +1,54 @@
+// @flow
+import {getGHLibsAndFlowVersions, filterDefs, formatDefTable}
+  from "../lib/libDef.js";
+import type {LibDefWithFlow} from "../lib/libDef.js";
+import {child_process, path} from '../lib/node'
+
+const P = Promise;
```
jeffmo: Looks like you're not using `P` or `Promise`

npm/src/commands/search.js
```diff
@@ -0,0 +1,54 @@
...
+  const matches = output.match(/.*version (.*)\n$/)
+  if (!matches || !matches[1]) {
+    fail()
+  } else {
+    return matches[1]
+  }
+}
+
+export async function run(args: {}): Promise<number> {
+  if (!args._ || !args._.length > 1) {
```
jeffmo: `!args._.length > 1` will always be `false` -- it's the same as `(!args._.length) > 1`

npm/src/commands/search.js
```diff
@@ -0,0 +1,54 @@
...
+    console.error('Please provide a term for which to search!')
+    return 1;
+  }
+
+  const flowVersion = args.flowVersion || findFlowVersion()
```
jeffmo: I wonder if we should require the user to opt-in to searching for Flow? Doing it implicitly could easily end up with, say, a global version they have installed (even though maybe they're using something like `npm run flow` for all of their project-specific needs)?

Maybe something like `flow-typed --flowVersion=auto`?

npm/src/lib/libDef.js
```diff
@@ -346,3 +356,80 @@ export async function getShallowGHLibDefs(
...
+      return stringToVersion(withoutPrefix)
+    } else {
+      return emptyVersion()
+    }
+  })
+  .toArray()
+  .toPromise()
+}
+
+export function getGHLibsAndFlowVersions(
```
jeffmo: Should probably call `_verifyCLIVersion()` in here first to be sure we're working with a version of the GH repo that we're compatible with

npm/src/commands/search.js
splodingsocks: D'oh ⚡

npm/src/commands/search.js
splodingsocks: What would you think of just omitting this functionality for now?

npm/src/lib/libDef.js
splodingsocks: Awesome, thanks!

npm/src/commands/search.js
jeffmo: That works too

splodingsocks:
@jeffmo I think this is ready to merge now. Want to take a look?

npm/src/lib/__tests__/libDef-test.js
```diff
@@ -0,0 +1,117 @@
...
+          pkgVersionStr: 'v0.3.x',
+          pkgNameVersionStr: 'mori_v0.3.x',
+          flowVersionStr: '>=v0.18.x',
+          flowVersion: { range: '>=', major: 0, minor: 18, patch: 'x' }
+        }
+      ]
+
+      const formatted = formatDefTable(fixture)
+      expect(formatted).toEqual(
+        "\nFound definitions:" + "\n" +
```
jeffmo: ooo...

npm/src/lib/libDef.js
```diff
@@ -388,3 +398,81 @@ export async function getShallowGHLibDefs(
   }
   return _libdefs
 }
+
+const identity = i => i
+
+function getGHFlowVersionsForDef(def: ShallowGHLibDef): Promise<Array<Version>> {
```
jeffmo: Need to call `_verifyCLIVersion()` in here too

(I'm going to send a commit to just add this in here myself -- just noting so I remember)

splodingsocks:
Woo hoo!!

PR #28, opened 2016-04-14
Limit the flow version of the Mori definition

splodingsocks:
Flow 0.23 became available, and the definition for Mori isn't compatible. Here's hoping it works to put space-separated versions in the folder name!


hallettj:
In that case I am going to need to update the Mori definitions.

splodingsocks:
Closing this in favor of a branch to support upper limits on libdefs

PR #29, opened 2016-04-16
Upper flow-version limits

splodingsocks:
At the moment, we don't support setting an upper limit on or flow versions. I started work on supporting it, but realized that I was getting distracted from work on the download command. So I pushed up what I've got here.

I'm not sure I like the direction I was heading with this. It feels messy.


npm/src/lib/libDef.js
```diff
@@ -148,7 +148,8 @@ export async function getLocalLibDefs(
   return libDefs;
 };
 
-const FLOW_DIR_NAME_RE = /^flow-(all|([><]?=)?v([0-9]+)\.([0-9]+|x)\.([0-9]+|x))$/;
+const FLOW_DIR_NAME_RE =
+  /^flow-(all|([><]?=)?v([0-9]+)\.([0-9]+|x)\.([0-9]+|x)(_([><]?=)?v([0-9]+)\.([0-9]+|x)\.([0-9]+|x))?)$/
```
jeffmo: Hehe, oh regexp.

nit: Maybe breaking this down with strings would be a little more readable?

```javascript
const FLOW_VER = '([><]?=)?v([0-9]+)\.([0-9]+|x)\.([0-9]+|x)';
const FLOW_DIR_NAME_RE = new RegExp(
  `^flow-(all|${FLOW_VER_RANGE}(_${FLOW_VER_RANGE})?)$`
);
```

jeffmo: Also I guess this leaves `([><]?=)` as optional in the upper bound... Should we make it required?

npm/src/lib/libDef.js
```diff
@@ -186,13 +189,23 @@ export async function getLocalLibDefFlowVersions(
...
       patch = _validateVersionPart(patch, "patch", itemPath, validationErrors);
 
-      const flowVersion = {range, major, minor, patch};
+      if (upMajor) {
+        upRange = _validateVersionRange(upRange, itemPath, validationErrors);
+        upMajor = _validateVersionNumPart(upMajor, "major", itemPath, validationErrors);
+        upMinor = _validateVersionPart(upMinor, "minor", itemPath, validationErrors);
+        upPatch = _validateVersionPart(upPatch, "patch", itemPath, validationErrors);
+      }
+
```
jeffmo: We should probably also validate that the range makes sense (i.e. something like `flow->=v0.23.x_<=v0.23.x` doesn't make much sense).

(Feel free to `// TODO: ` this, though -- probably not critical right now)

jeffmo:
Looks like the test runner needs to be updated as well (somewhere in [here](https://github.com/flowtype/flow-typed/blob/master/npm/src/commands/runTests.js#L252-L279), I think)

splodingsocks:
Thanks for the review!

After taking a bit to think about this, I'm wondering if we shouldn't include an upper bound, but just require all definitions to work with all flow versions above their lower bound?

But then what would we do with libraries that break from a flow upgrade (which will probably be many) until they are fixed?

jeffmo:
Hmm, yea I like this upper-bound idea and I think it makes a lot of sense for the reasons you stated.

jeffmo:
Also let me know if I can help out with any parts of this or anything else -- happy to work on whatever needs to be done next!

splodingsocks:
Great, thanks, Jeffmo!

Would you mind taking over this branch and getting the support for upper limits in?

jeffmo:
You betcha

splodingsocks:
Awesome, thanks! I'm making progress on the install command, but slowly.

jeffmo:
Ok, I think this is ready for review.

Notice that I also changed the delimiter between the parts of file/directory names to `_` instead of `-`. It just seemed a little less noisy now that some directories have several parts...

jeffmo:
Weird that the option to squash-on-merge isn't here... I see that the option is enabled in settings, I wonder what I'm missing?

npm/src/commands/runTests.js
```diff
@@ -250,10 +252,27 @@ async function runTestGroup(
...
-      return (
-        testGroup.flowVersion === "all"
-        || semver.satisfies(flowVer, testGroup.flowVersion)
-      );
+      if (testGroup.flowVersion === "all") {
+        return true;
+      }
+
+      const testGrpUpperBound = testGroup.flowVersion.upperBound;
+      testGroup.flowVersion.upperBound = undefined;
```
splodingsocks: Why set it to undefined here before setting it again on line 265?

splodingsocks:
Looks great, @jeffmo, 👍, Thanks!

splodingsocks:
As soon as you merge I'll rebase the search command branch.

npm/src/commands/runTests.js
jeffmo: If there's an upperbound, `versionToString()` would render a string like "v0.1.0_<=v0.2.0", but we want the two separate so we can use the `semver` module. So to handle this I'm just temporarily separating the two versions for string-rendering reasons.

I could also split on `'_'`, but it would just be a one-off place where we define the schema for how upper/lower bounds are formatted with each other (i.e. if we change this again we'd have to remember to change it here and in the utils modules)

splodingsocks: 👍

splodingsocks:
also, @jeffmo, I don't know if you have time to jump on IRC, but I'm trying to figure out automatic picking of a version range to install, and I'd love to discuss it with you.

jeffmo:
Just hopped on IRC

PR #31, opened 2016-04-22
classnames

marudor:


definitions/classnames_v2.x.x/flow_>=v0.23.x/classnames_v2.x.x.js
```diff
@@ -0,0 +1,6 @@
+declare module 'classnames' {
+  declare class Classnames {
+    (...classes: Array<string | Object | void>): string
```
jeffmo: If I'm reading the [readme](https://www.npmjs.com/package/classnames) correctly, `Object` here should be `{[class: string]: boolean}`, right?

Also, what's the purpose of the `void`?

marudor: Yes it should. Changed.
No Idea why I got a void in there. Ask past me.

PR #32, opened 2016-04-22
color

marudor:


definitions/color_v0.7.x/flow_>=v0.23.x/color_v0.7.x.js
```diff
@@ -0,0 +1,107 @@
+type RGBObject = {
```
jeffmo: I'm actually building support for un-exported type aliases inside `declare module`s as we speak, but for now could you namespace these with something like `$npm$color$RGBObject` (etc) since these will end up as globals for anyone who uses this?

marudor:
Prefixed everything.

PR #33, opened 2016-04-22
core-decorators

marudor:


jeffmo:
Adding "incomplete" until we can get the `any`s filled in.

(Also filed #49 to build a bot that will point out `any` types on PRs in the future -- we want to minimize these in our libdefs as much as possible)

PR #35, opened 2016-04-22
deep-diff

marudor:


definitions/deep-diff_v0.3.x/flow_>=v0.23.x/deep-diff_v0.3.x.js
```diff
@@ -0,0 +1,18 @@
+type Difference = {
```
jeffmo: Needs namespace

definitions/deep-diff_v0.3.x/flow_>=v0.23.x/deep-diff_v0.3.x.js
```diff
@@ -0,0 +1,18 @@
...
+  lhs: ?any,
+  rhs: ?any,
+  index?: number,
+  item?: Difference,
+}
+
+type PrefilterFn = (path: Array<string>, key: string) => bool|void;
+
+declare module 'deep-diff' {
+  declare function exports(lhs: any, rhs: any, prefilter?: PrefilterFn, acc?: Array<any>): ?Array<Difference>;
```
jeffmo: I think the mere presence of a variable called `exports` in a `declare module` will prevent other things from being exported right now...

As such, should probably define this as:

```javascript
declare var exports: {
  (lhs: mixed, rhs: mixed, ...): ?Array<Difference>;
  diff(lhs: mixed, rhs: mixed, ...): ?Array<Difference>;
  ...
};
```

Also note: For input types like (function param types) in a libdef, it's usually better to use `mixed` than `any` -- just because `mixed` is more strict. The difference is unlikely to be observed, but if ever the `mixed` type finds it's way out somehow, it won't be unexpectedly over-permissive.

definitions/deep-diff_v0.3.x/flow_>=v0.23.x/deep-diff_v0.3.x.js
jeffmo: Reading the docs -- it seems like this would be well-suited to be typed as a disjoint union (discriminated by the `kind` property) since `lhs` doesn't exist on `kind === "N"`, `rhs` doesn't exist on `kind === "D"`, `index` only exists when `kind === "A"`, etc..

```javascript
type Difference = {
};
```

PR #37, opened 2016-04-22
jquery

marudor:


jeffmo:
This one's a doozy! Probably warrants a good audit, but we shouldn't spin forever here since this one is probably pretty commonly requested.

Will come back to this later

emily-e:
the JQuery class seems to be missing the .remove() method: https://api.jquery.com/remove/

emily-e:
Also, all the fields of the JQueryAjaxSettings should be optional so the two AJAX methods on lines 642 and 649 work(I got this to work locally using $Shape: `ajax(settings: $Shape<JQueryAjaxSettings>): JQueryXHR;` ( http://api.jquery.com/jquery.ajax/ - "settings -Type: PlainObject - A set of key/value pairs that configure the Ajax request. All settings are optional.")

unhammer:
thanks @emily-e! I use this currently:
[shape.diff.txt](https://github.com/flowtype/flow-typed/files/489256/shape.diff.txt)
but of course it loses some info on the functions (I couldn't figure out where to put the `?` – https://flowtype.org/docs/objects.html#optional-properties only talks about non-function fields)

IanVS:
I see that this issue is closed.  Is it no longer planned to bring `jquery` into `flow-typed`?

marudor:
@IanVS it is.
But this was a conversion from typescript and very immature.

PR #40, opened 2016-04-22
isomorphic-fetch

marudor:


thejameskyle:
This is in the wrong directory `definitions/iso...` vs `definitions/npm/iso...`

marudor:
Again?... I'm sure I moved it... I sure fail at git quite often :/

PR #41, opened 2016-04-22
i18next

marudor:


bassettsj:
Seems good to go, what else is needed?

thejameskyle:
Was this intentionally not put within the `npm` directory?

marudor:
@thejameskyle 
Actually no - this was just such an old PR that it did not include the npm folder..
And hasty me moerged without checking twice
Thanks for pointing out.

PR #42, opened 2016-04-22
lodash

marudor:


ryyppy:
Would love to see `lodash/fp` declarations in there too!
https://github.com/lodash/lodash/wiki/FP-Guide

The auto-currying functions are hard to type, not sure how we tackle this, also I was looking for a way to make function composition with `compose` type-safe.. if someone has an idea for that, that would be really great!

vaukalak:
@marudor thanks a lot for this. I'd like to help you with tests to make it merged. But as it's quite a lot of work to do, and work of different people could overlap, I'd propose the following flow:

1 - every method is tested in an appropriate test file.
2 - before someone starts to write tests, he writes here methods he will cover.
3 - all tests go to `marudor:lodash`

Does this make sense?

mirek:
Considering a fact that flow-typed is in beta, wouldn't it be better to merge definitions as they are from this PR (and others) to speed up the process?

I'm sure people will "cherry-pick-commit" fixes much faster.

At the moment definitions live in different branches which makes it very hard to contribute.

Instead of waiting for perfect version, we could have something to work on, even if it's not confirmed 100% correct.

Later we can add some stability flag or coverage report but for now I think just merging not-100%-covered definitions would be a win for the project.

jeffmo:
@mirek: Only the CLI is in beta -- the definitions are real :)

Moreove, it's important that tests pass always -- otherwise we'll dig ourselves into a corner where only some definitions are high quality and others aren't.

Id say it's ok to submit *partial* libdefs with intent to come back and fill in more API surface area later, but tests and versioning and such if the submitted surface area are pretty important even now.

mirek:
What about creating some sort of "sandbox" directory which mimics "definitions"? The focus should be on minimising size of PRs and making contributions easy.

Without this kind of mechanism what are people doing? They look for things like `lodash`, etc. they won't find definition for it so they probably go back to reconsidering typescript because those were two alternatives they were looking at in the first place.

Some people will dig deeper and see lodash in PR. Hanging there for 3 months, not much activity. They'll copy paste it to their project. Comment out or fix things that are broken and wait for new version. They won't commit back their changes for sure.

The thing is that if people are able to install incomplete definitions (with `--sandbox` flag or whatever explicit intention) they will test it via their codebase. This is great leverage to use. If type checking fails against their project they will see exactly what kind of test + fix to write so it passes that single error. Just few lines of code, create PR - no problem.

On the flow-typed side - it's much easier as well, small patches that you can cache and run in your head can be accepted very quickly.

The whole thing will move forward.

jeffmo:
Also worth noting that we aim for quick turnaround on PRs that pass tests . In this case, @marudor had a large number of libdefs that were around before flow-typed...so getting through those takes some time. In the future it's more likely that we'll have a more steady stream of libdefs written with flow-typed in mind (and thus tests + versioning, etc) -- so hopefully those will land more steadily.

We should probably prioritize this particular one higher for turn-around than others tho (given than lodash is pretty popular)

jeffmo:
> What about creating some sort of "sandbox" directory which mimics "definitions"?

That's not a bad idea. We could move PRs there after some SLA of non-responsiveness rather than just closing them (this inevitably happens). I also like the idea of excluding them from search/install results in the CLI without a special `--include-sandbox` (or whatever) flag.

Feel like making an issue to track this?

BlooJeans:
PSA: The updated definitions are found at https://github.com/marudor/flowInterfaces/blob/master/packages/iflow-lodash/index.js.flow

marudor:
I've updated it and reused the underscore tests.
Lets get this merged - finally.

PR #61, opened 2016-04-23
redux

marudor:


definitions/redux_v3.x.x/flow_>=v0.23.x/redux_v3.x.x.js
```diff
@@ -0,0 +1,29 @@
+declare module 'redux' {
+  declare type State = any;
+  declare type Action = Object;
```
jeffmo: I'm a noob on the nuances of `redux`, so a question: Does it make sense to define at least a `type: string` property here?

i.e. `{type: string, [_: string]: any}`

jeffmo:
cc @gaearon as a SME

definitions/redux_v3.x.x/flow_>=v0.23.x/redux_v3.x.x.js
gaearon: Redux only enforces that it isn’t `undefined`. Some people really want to use Symbols 😄 .

gaearon:
From the cursory glance this appears right, but I’m no Flow expert.
cc @kyldvs, @leoasis

leoasis:
This is still missing all the polymorphic types for each of the functions. For instance, this won't type check that you're dispatching valid actions or valid things that the middleware chain can handle, because that type information is not in `createStore` and particularly it's not in the returned `Store` type. It should be polymorphic and depend on the types of both state and actions, and also how the middlewares augment that `Action` type. This is what I was trying to achieve here: https://github.com/reactjs/redux/pull/817/files#diff-3d784680b18b6d52c58df8665c93ae56R31, but couldn't figure out how to make Flow happy for that to work.

definitions/redux_v3.x.x/flow_>=v0.23.x/redux_v3.x.x.js
vaukalak: @gaearon, @jeffmo  if I understand correctly, various types of actions can be dispatched as an async action. So the interface of dispatch will be always `Action | AsyncAction`, and as `AsyncAction = any` it will become same as `Action | any` or simply `any`. So there's no reason at all to specify interface for actions. I think it would make more sense, to make the store generic:

For example:

```
declare type Store<S, A> = {
        dispatch: Dispatch<A>,
        getState: () => S,
        subscribe: (listener:() => void) => () => void,
        replaceReducer: (reducer:Reducer<S>) => void
};
```

marudor:
I've not had time to do a very detailed libdef.
This was more to get started.
Any Objections if we merge this and create an Issue to track that we need to improve them

gaearon:
I agree we can start with more relaxed types and refine later. Not sure how this works with regards to semver though.

marudor:
@gaearon well - As we only make it stricter to match the actual API we can probably just update this.
I don't think we really version the actual libdefs. Only what version of library they match.
So as long as this version of libdef can be used with redux and the future ones can as well we should be fine.

jeffmo:
Sounds like all parties are on board with the current state of these libdefs, so let's merge it.

PR #64, opened 2016-04-23
redux-actions

marudor:


definitions/redux-actions_v0.9.x/flow_>=v0.23.x/redux-actions_v0.9.x.js
```diff
@@ -0,0 +1,5 @@
+declare module 'redux-actions' {
+  declare function createAction(type: string, payloadCreator?: Function, metaCreator?: Function): Function;
```
jeffmo: Does it make sense to type `payloadCreator` as `(payload: mixed) => any`? (and if so, same for `metaCreator`?)

definitions/redux-actions_v0.9.x/flow_>=v0.23.x/redux-actions_v0.9.x.js
```diff
@@ -0,0 +1,5 @@
...
+  declare function handleAction(type: string, reducer: Object|Function): void;
+  declare function handleActions(reducerMap: Object, defaultState?: Object): void;
```
0xR: This doesn't return a void, now flow gives errors

PR #66, opened 2016-04-23
WIP - Install command

splodingsocks:
@ryyppy wants to take a stab at finishing up the install command, and time is growing short for me, so here's my WIP. I left off in the install command at the point of creating the flow-typed folder if it doesn't exist, and actually downloading the selected libdef file from github.

After that, it should just be figuring out how to write an automated test around the install command.

Thanks for being willing, @ryyppy. Please let me know if you need anything!

(Also, feel free to change up what I wrote in whatever way you feel is best).


splodingsocks:
Another note, this is based on PR https://github.com/flowtype/flow-typed/pull/27, which hasn't been merged yet. Once that's merged, this should be rebased, and only the last four commits or so should be kept.

ryyppy:
Hm, did you think about how we are going to prevent request-limits on the github API? Right now, I cannot really test the install, since I was apparently doing to many requests:

```
[ 'install', 'kefir' ]
UNCAUGHT ERROR: {
  "code": 403,
  "message": "{\"message\":\"API rate limit exceeded for X.X.X.X. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)\",\"documentation_url\":\"https://developer.github.com/v3/#rate-limiting\"}"
}
```

This error when the `https.get` was executed.. I tried appending a generated clientId / clientSecret for now, but this does not work... maybe because `raw.githubusercontent.com` does not implement the same authorization procedure as `api.github.com`?

Any ideas?

splodingsocks:
Hmm, maybe we can get around it by using https://rawgit.com?

ryyppy:
This is awsome! Works like a charm, thanks

(We still need to think about production data fetching in the near future tho)

ryyppy:
Okay I am done with my changes, but I think I cannot directly push on this remote branch, since I don't have write access :-)

splodingsocks:
oh, right. What would you think about forking, and opening a PR from your fork?

I'm excited to see your changes!

ryyppy:
PR for merging with this branch can be found in #68

splodingsocks:
See #68

PR #67, opened 2016-04-23
pluralize

danez:
Should I use a class instead? Because I saw some definitions use classes. What is the difference when using a class?


marudor:
Problem is: Once you declare var exports nothing else will export.
So you should export an Object or a class containing the functions.

danez:
I see, good to know. I changed that.

PR #68, opened 2016-04-23
Install command

ryyppy:
Finished @splodingsocks install command... it's downloading the file and puts it in the CWD's `flow-typed` directory

### DIFF
--- a/npm/src/commands/install.js
+++ b/npm/src/commands/install.js
@@ -1,76 +1,90 @@
 // @flow
-import mkdir from 'mkdirp'
-import {getGHLibsAndFlowVersions, filterDefs, formatDefTable}
-  from "../lib/libDef.js";
-import type {LibDefWithFlow} from "../lib/libDef.js";
+
+import mkdirp from 'mkdirp'
+import {
+  getGHLibsAndFlowVersions,
+  filterDefs,
+  formatDefTable
+} from '../lib/libDef';
+
+import type {LibDefWithFlow} from '../lib/libDef';
 import {child_process, path, https, fs} from '../lib/node'
 
-export const name = "install";
-export const description =
-  "Installs a libdef";
+export const name = 'install';
+export const description = 'Installs a libdef to the ./flow-typed directory';
+
+export function setup(yargs: Object): Object {
+  return yargs
+    .usage(`$0 ${name} - ${description}`)
+    .options({
+      flowVersion: {
+        demand: true,
+        describe: 'The version of Flow fetched libdefs must be compatible with',
+        type: 'string',
+      },
+    });
+};
+
+function failWithMessage(message: ?string) {
+  if (message != null) {
+    console.error(message);
+  }
+  return 1;
+}
 
 type Args = {
   _: Array<string>;
   flowVersion?: string;
 }
+
 export async function run(args: Args): Promise<number> {
+  const { _ } = args;
 
-  function errorNeedDefName() {
-    console.error('Please provide a libdef name ' +
-      '(example: lodash, or lodash@4.2.1)')
+  if (_ == null || !(_.length > 1)) {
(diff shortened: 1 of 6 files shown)


npm/src/commands/install.js
```diff
@@ -1,76 +1,90 @@
...
-    return 1;
+  if (_ == null || !(_.length > 1)) {
+    return failWithMessage('Please provide a libdef name (example: lodash, or lodash@4.2.1)');
   }
 
-  if (!args.flow) {
-    errorNeedFlowVersion()
-    return 1;
+  if (args.flowVersion == null) {
+    return failWithMessage('Please provide a flow version (example: --flow 0.23)');
```
splodingsocks: If the arg is changed to `flowVersion` as on line, 53, the help string should be `--flowVersion 0.23`

npm/src/commands/install.js
```diff
@@ -1,76 +1,90 @@
...
-  console.log('DEBUG: About to install', def.pkgName, def.pkgVersionStr)
+  await new Promise((resolve, reject) => {
+    const flowTypedDir = path.resolve('./flow-typed')
+
+    console.log(`File will be downloaded to '${flowTypedDir}'`);
+    mkdirp.sync(flowTypedDir)
 
-  await new Promise((res, rej) => {
-    const flowTypedFolder = path.resolve('./flow-typed')
+    const url = `https://raw.githubusercontent.com/flowtype/flow-typed/master/definitions/${def.pkgName}_${def.pkgVersionStr}/flow_${def.flowVersionStr}/${def.pkgName}_${def.pkgVersionStr}.js`;
```
splodingsocks: Did you not end up using rawgit.com after all? How'd you make it work?

splodingsocks:
Thanks for finishing this up, @ryyppy! I just had a few questions / comments. It looks like you're babel change did end up breaking ./runtests.sh (according to Travis).

Afterward I think it looks good to merge. (After the search PR is merged, right?)

npm/src/commands/install.js
ryyppy: I realized that actually the fetching of the libDefs was making problems and I never reached the point where I queried `raw.githubusercontent.com` ... I added the GH-client / secret environment variable and used my own application authorization data for now.

splodingsocks: Oh right, okay. Sounds good!

ryyppy:
I just realized that I didn't cover following test-case: `flow-typed install --flowVersion=0.23.0 kefir@3.0.0` ... in that case nothing will be found. Want to fix that.

For the `runtests.sh` I need to check what could go wrong here...

ryyppy:
I fixed the regenerator error now, ... also I fixed a problem where the pkgName / version regex didn't split properly...

The `install kefir@3.0.0` command does not work yet, since there is a problem with the function `filterDefs`... it seems like `semver.satisfy` does not fit the filter use-case?

Not sure if we should merge back to master if this feature is not yet working.

jeffmo:
Just wanted to drop by with an update on my end:

I'm in the midst of catching up on things with my team and shipping the 0.24 Flow deploy this week, so I've been kind of MIA. I should have some more time next week to help get this landed if @splodingsocks doesn't beat me to it!

splodingsocks:
Thanks for the update @jeffmo. I've been distracted myself with work, presentation prep, and other things.

@ryyppy are you still wanting to finish this? Or are you handing back off to me?

ryyppy:
Well I would love to do it, but I would need to touch the semver matching code, which I don't feel confident changing without breaking other code ... the state of now is clean, you can merge that in your feature branch and add functionality for e.g. matching `3.X.X`

jeffmo:
Going to merge this into the flowtype branch, check that out, and finish this up.

Thanks a ton for your work here, @ryyppy!

PR #69, opened 2016-04-25
Axios

marudor:


definitions/axios_v0.11.x/test_axios-v0.11.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import axios from 'axios';
+(axios.get('foo'): Promise);
```
jeffmo: Would you mind specifying `Promise<*>` here instead?

In 0.25, Flow is going to start requiring that parameterized types have all of their parameters explicitly instantiated. Using `*` just means "I explicitly want Flow to infer the type"

definitions/axios_v0.11.x/flow_>=v0.23.x/axios_v0.11.x.js
```diff
@@ -0,0 +1,57 @@
...
+    use(
+      successHandler: ?(response: AxiosXHR<T>) => mixed,
+      errorHandler: ?(error: mixed) => mixed,
+    ): AxiosInterceptorIdent;
+    eject(ident: AxiosInterceptorIdent): void;
+  }
+  declare class Axios {
+    constructor<T>(config: AxiosXHRConfigBase<T>): Promise<T>;
+    <T>(config: AxiosXHRConfigBase<T>): Promise<T>;
+    [key: $Enum<Axios>]: Function;
```
jeffmo: Intentional?

marudor: What exactly?

jeffmo: The indexer? If so -- what does it do in the presence of all the other methods?

marudor: Ah.
Well... 
Thats prolly from work. I'm overwriting the get/post methods by index.
I'm fine removing it.

jeffmo: Ok, I'm not too familiar with this library so you understand this libdef best -- but if this isn't strictly necessary I think we should remove it. It seems like it could have potential to stir Flow in the wrong direction when mixed with the other members here

marudor: I agree. I removed it.

PR #72, opened 2016-04-26
Make `validate-defs` require that at least one test applies to every definition file

jeffmo:
Per https://github.com/flowtype/flow-typed/issues/70, we want to be sure that all libdefs are tested.
By adding this requirement to `validate-defs`, we ensure that Travis checks this requirement for us


jeffmo:
(this is rebased on top of https://github.com/flowtype/flow-typed/pull/71 to make tests pass -- so that will need to land before this)

PR #74, opened 2016-04-28
Add highlight.js

danez:
This adds highlight.js with tests.

the only thing which I'm not sure is the `HighlightJs$Continuation`, what I wanted to achieve is that whatever is returned in `HighlightResult.top` is valid as `continuation` argument to `highlight()`, and nothing else. The object that is returned in `HighlightResult.top` is an internal token to be passed to highlight(), but is not described in their API docs.

### DIFF
--- a/definitions/highlight.js_v8.x.x/flow_>=v0.13.x/highlight.js_v8.x.x.js
+++ b/definitions/highlight.js_v8.x.x/flow_>=v0.13.x/highlight.js_v8.x.x.js
@@ -1,0 +1,51 @@
+declare class HighlightJs$Continuation {}
+
+type HighlightJs$HighlightResult = {
+  language: string,
+  relevance: number,
+  value: string,
+  top?: HighlightJs$Continuation,
+  second_best?: HighlightJs$HighlightResult
+};
+
+type HighlightJs$LanguageContains = {
+  begin?: string|RegExp,
+  beginKeywords?: string|RegExp,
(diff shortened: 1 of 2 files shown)


marudor:
For the use case the HighlightJs$Continuation should be fine.

definitions/highlight.js_v8.x.x/flow_>=v0.13.x/highlight.js_v8.x.x.js
```diff
@@ -0,0 +1,51 @@
...
+
+type HighlightResult = {
+  language: string,
+  relevance: number,
+  value: string,
+  top?: HighlightJs$Continuation,
+  second_best?: HighlightResult
+};
+
+type LanguageContains = {
```
marudor: Prefix please

definitions/highlight.js_v8.x.x/flow_>=v0.13.x/highlight.js_v8.x.x.js
```diff
@@ -0,0 +1,51 @@
...
+  begin?: string|RegExp,
+  beginKeywords?: string|RegExp,
+  className?: string,
+  contains?: Array<LanguageContains>,
+  end?: string|RegExp,
+  keywords?: string,
+  relevance?: number
+};
+
+type Language = {
```
marudor: Prefix please

PR #75, opened 2016-04-30
mocha definition

vaukalak:
Definitions for https://mochajs.org/

### DIFF
--- a/definitions/mocha_v2.4.x/flow_>=v0.22.x/mocha_v2.4.x.js
+++ b/definitions/mocha_v2.4.x/flow_>=v0.22.x/mocha_v2.4.x.js
@@ -1,17 +1,18 @@
-type TestFunction = (() => void | Promise) | ((done : Function) => void);
+type TestFunction = (() => void | Promise<mixed>) | ((done : () => void) => void);
 
-declare class describe {
-    static (name:string, spec:() => void): void;
-    static only(description:string, spec:() => void): void;
-    static skip(description:string, spec:() => void): void;
-    static timeout(ms:number): void;
-}
-declare class it {
-    static (name:string, spec:TestFunction): void;
-    static only(description:string, spec:TestFunction): void;
-    static skip(description:string, spec:TestFunction): void;
-    static timeout(ms:number): void;
-}
+declare var describe : {
+    (name:string, spec:() => void): void;
+    only(description:string, spec:() => void): void;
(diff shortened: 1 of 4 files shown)


marudor:
Please make a module for mocha. Otherwise it will always be global.
If someone needs them global (thats the usual case for mocha) he can create a global libdef that imports the methods he's using.

definitions/mocha_v2.4.x/flow_>=v0.22.x/mocha_v2.4.x.js
```diff
@@ -0,0 +1,19 @@
+type TestFunction = (() => void | Promise) | ((done : Function) => void);
+
+declare class describe {
```
jeffmo: Since `describe` is not meant to be `new`-ed, this would be better described as a callable object:

```javascript
declare var describe: {
  (name: string, spec: () => void): void;
  ...
};
```

definitions/mocha_v2.4.x/flow_>=v0.22.x/mocha_v2.4.x.js
```diff
@@ -0,0 +1,19 @@
+type TestFunction = (() => void | Promise) | ((done : Function) => void);
```
jeffmo: In the next version of Flow we're going to start requiring type parameters on parameterized types like `Promise`. Can you specify `Promise<mixed>` here to pre-empt that?

jeffmo: Also, can we type the `done` callback as `() => void` instead of `Function`? `Function` is an `any` type that is best to avoid when possible

jeffmo:
Woohoo! Thanks so much for this!

@marudor: IIUC, mocha actually defines global functions though, right?

marudor:
@jeffmo yep, it does define globals.
Doe we isolate the libdefs enough to just declare the stuff global?

jeffmo:
Should be ok. If the declaration file is installed, the expectation is that it's declared globals are present

vaukalak:
@jeffmo @marudor thanks for your feedback. So the things to do with it are:

1 - `Promise` should be `Promise<mixed>`.
2 - change `done : Function` to `done : () => void`.
3 - change `declare` and `it` to be variables rather than classes.

make sense?

jeffmo:
@vaukalak looks good!

PR #77, opened 2016-05-01
draft for any check

marudor:
To address https://github.com/flowtype/flow-typed/issues/49
very basic check. It uses flow dump-types to check all types and searches for the "any" string.
Probably error prune. But the dump-type command does not dump each type. Arrow functions are one type, even with paramter.
So a arrow function with any type would not get catched if I do an exact match for any.
Anyone got an idea how to get better accuracy?

### DIFF
--- a/npm/.flowconfig
+++ b/npm/.flowconfig
@@ -11,4 +11,4 @@ flow-typed
 experimental.const_params=true
 
 [version]
-0.22.1
+>=0.22.1
--- a/npm/src/cli.js
+++ b/npm/src/cli.js
@@ -7,6 +7,7 @@ import * as yargs from "yargs";
 import * as RunTests from "./commands/runTests.js";
 import * as ValidateDefs from "./commands/validateDefs.js";
 import * as Search from "./commands/search.js";
+import * as CheckForAny from "./commands/checkForAny.js";
 
 type CommandModule = {
   name: string,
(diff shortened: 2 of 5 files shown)


jeffmo:
Woohoo! Thanks for tackling this

So on using `dump-types`: There's a good chance this could be a bit noisy since there are places where `any` can creep in even when the user didn't explicitly state them (i.e. returning a type from something in flow's built-in libs that uses one of the `any`-types).

Instead, it would be a little more direct to just use `flow ast` to look for references to annotations of `any`, `Object`, `Function`, etc. `flow ast` should go all the way back to 0.13 -- so should be available in [all versions of Flow that we support here](https://github.com/flowtype/flow-typed/blob/master/npm/src/commands/runTests.js#L93-L94).

marudor:
Now obsolete - gonna do it with a eslint plugin.

jeffmo:
Heya -- so just to color some thoughts before you spend too much time I wanted to toss out the consideration that an eslint plugin might be a little noisy?

Specifically: we probably want to tell the user once somewhere that anys aren't ideal and then let them decide if they want to ignore the note -- since sometimes any is unavoidable. Do you have something in mind with lint that would make this doable?

marudor:
yes - count any in general, not per line base.
Use of /* eslint flow-typed/any: [2, x] */ to set the number of allowed any

jeffmo:
Cool, glad to hear you're already on top of it then :)
Looking forward to what you come up with!

PR #78, opened 2016-05-05
`install` comand

jeffmo:
Thanks to both @splodingsocks and @ryyppy, this is a working version of `flow-typed install`!

It works using the same machinery as `flow-typed search`, except it will actually fetch and save the libdef to the `flow-typed/` directory. There's some extra noise in here from the addition of eslint, but hopefully not too hard to skim/scroll past...

Note: While the `install` command here works, it has certainly uncovered some more work for us. I'll make issues for each of these once this PR lands:

1. As @ryyppy pointed out somewhere already, this will hit the github API rate-limiter fairly quickly. We really need to solve this before we can "launch" the CLI. We can discuss options for this in the issue I'll create for it, but my current thoughts are that we should do something similar to homebrew and just store a cached checkout of the git repo somewhere on the user's machine (say, in `~/.flow-typed-cache/`). From there, all `search` and `install` operations can happen locally.

1. I think it's [going to be pretty important](https://github.com/flowtype/flow-typed/blob/install-command2/npm/src/commands/install.js#L147-L159) for us to track when/where an installed libdef came from (i.e. the last-change commit hash). Similarly I'll open an issue to track this, but I think we should probably come back to this as well before we "launch" (mostly to preserve option value post-launch).

1. This can only install a single libdef at a time -- but we should probably add support for specifying multiple arguments at once. Probably not launch-blocking.

1. We should build support for some kind of `--fromPkgJson` flag (that reads a `package.json` file and tries to install libdefs for as many dependencies as it can). I don't see this as needing to be launch-blocking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,5 +31,5 @@ node_modules
 
 # CLI dist directory
 npm/dist/*
-npm/test-dir/*
-npm/flow-bins/*
+npm/.test-dir/*
+npm/.flow-bins-cache/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ addons:
 script: ./runtests.sh
 cache:
   directories:
-  - npm/flow-bins
+  - npm/.flow-bins-cache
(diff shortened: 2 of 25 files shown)


splodingsocks:
WOOOOOOOOOOOOOOOOOO! I'll take a look!

splodingsocks:
Nice! 👍 Looks good to me, let's merge!

PR #79, opened 2016-05-06
`mv npm cli`

jeffmo:
Title says it all.
I think I'm to blame for naming it `npm` originally, but this has been mildly bugging me...


splodingsocks:
I blame nobody

👍

PR #81, opened 2016-05-09
Add redux-form declaration

bassettsj:
- Add a basic redux-form declaration to the library with a basic test

### DIFF
--- a/definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
+++ b/definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
@@ -1,0 +1,66 @@
+// @flow
+import React from 'react';
+
+declare module 'redux-form' {
+  declare type InputProps = {
+    name: string,
+    value: string | number | boolean,
+    valid: boolean,
+    invalid: boolean,
+    dirty: boolean,
+    pristine: boolean,
+    active: boolean,
+    touched: boolean,
+    viisted: boolean,
+    autofiller: boolean
+  };
+  declare type FormProps = {
+    active: string,
+    asyncValidate: Function,
+    asyncValidating: string | boolean,
+    destroyForm: Function,
+    dirty: boolean,
+    error: String,
+    fields: { [fieldName: string]: InputProps },
+    handleSubmit: Function,
+    initializeForm: (data:Object) => any,
+    invalid: boolean,
+    pristine: boolean,
+    resetForm: Function,
+    formKey: string,
+    submitting: boolean,
+    submitFailed: boolean,
+    touch: (...fields: Array<string>) => void,
+    touchAll: () => void,
+    untouch: (...fields: Array<string>) => void,
+    untouchAll: () => void,
+    valid: boolean,
+    values: Object
+  };
+  declare type FormConfig = {
+    fields: Array<string>,
+    form: string,
+    alwaysAsyncValidate?: boolean,
+    asyncBlurFields?: Array<string>,
+    asyncValidate?: (values: Object, dispatch: Function, props: Object) => Promise<void>,
+    destroyOnUnmount?: boolean,
+    formKey?: string,
+    getFormState?: Function,
+    initialValues?: Object,
+    onSubmit?: Function,
+    onSubmitFail?: Function,
+    onSubmitSuccess?: Function,
+    verwriteOnInitialValuesChange?: boolean,
+    propNamespace?: string,
+    readonly?: boolean,
+    reduxMountPoint?: String,
+    returnRejectedSubmitPromise?: boolean,
+    touchOnBlur?: boolean,
+    touchOnChange?: boolean,
+    validate?: (values:Object, props:Object) => Object
+  };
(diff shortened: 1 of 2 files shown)


definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
```diff
@@ -0,0 +1,66 @@
...
+    viisted: boolean,
+    autofiller: boolean
+  };
+  declare type FormProps = {
+    active: string,
+    asyncValidate: Function,
+    asyncValidating: string | boolean,
+    destroyForm: Function,
+    dirty: boolean,
+    error: String,
```
jeffmo: lower-case-`s`?

definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
```diff
@@ -0,0 +1,66 @@
...
+    dirty: boolean,
+    pristine: boolean,
+    active: boolean,
+    touched: boolean,
+    viisted: boolean,
+    autofiller: boolean
+  };
+  declare type FormProps = {
+    active: string,
+    asyncValidate: Function,
```
jeffmo: Per the docs, it looks like this type is:

`(values: Object, dispatch: Function, props: Object) => Promise<void>`. 

Maybe we can be more specific?

definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
```diff
@@ -0,0 +1,66 @@
...
+    asyncValidating: string | boolean,
+    destroyForm: Function,
+    dirty: boolean,
+    error: String,
+    fields: { [fieldName: string]: InputProps },
+    handleSubmit: Function,
```
jeffmo: `handleSubmit` takes no params and returns `void`, am I reading that right? If so, could we be specific and type this as `() => void`?

definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
```diff
@@ -0,0 +1,66 @@
...
+  };
+  declare type FormConfig = {
+    fields: Array<string>,
+    form: string,
+    alwaysAsyncValidate?: boolean,
+    asyncBlurFields?: Array<string>,
+    asyncValidate?: (values: Object, dispatch: Function, props: Object) => Promise<void>,
+    destroyOnUnmount?: boolean,
+    formKey?: string,
+    getFormState?: Function,
```
jeffmo: This function always takes a state and a mountPoint, right?

definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
```diff
@@ -0,0 +1,66 @@
...
+    initialValues?: Object,
```
jeffmo: Docs suggest this is an object with only string values. Could we type this as `{[field: string]: string}`?

jeffmo:
Woohoo! This is awesome

I commented on a few of lines, but in general we try to discourage use of `Function` and `Object` when it's possible to be more specific. The reason is because `Object` can unexpectedly introduce `any`s (i.e. if you access a property on a thing typed as `Object`, the property will have type `any` -- which can flow through the consumer's program and lose type errors). Similarly with `Function`: If you call it, the return type is `any` -- which can flow (allowing type errors) as well.

Anyway, if there are places you want to be more specific that would be sweet. If not, no worries -- we can always merge and iterate later :)

definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
bassettsj: Updated, thank you.

definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
bassettsj: I think that would be a good start at least.

definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
bassettsj: I think it does take params, basically the field names and then their values, then you can return the a promise or void, I believe: http://redux-form.com/5.2.3/#/api/props?_k=omut2m

definitions/redux-form_v5.x.x/flow_>=v0.22.1/redux-form_v5.x.x.js
bassettsj: Updated: thanks

bassettsj:
@jeffmo Thanks for the review, I tried updating the definitions to the best of my ability. It should be a bit better.

jeffmo:
thank *you*! Looks great

PR #82, opened 2016-05-09
Cache repo in `~/.flow-typed/repo` rather than querying GitHub API

jeffmo:
(Apologies in advance for the size of this diff -- it's worth it though, I promise!)

Previously we were using the GitHub API for the `search` and `install` commands. This worked pretty well, except that it would hit the GitHub API rate limit super fast for users.

Per https://github.com/flowtype/flow-typed/issues/80, this PR switches the backend to use a local (persistent) clone of the `flow-typed` repo (stored in `~/.flow-typed/repo`). Not only does this make general use of the CLI a little faster, but it also forgoes the need for CLI users to ever hit the GitHub API. The only place we still use the GitHub API is for the test runner to fetch the list of flow binaries. This is more ok because most of those runs happen in travis -- and we've worked around the rate-limit there to some degree.

Coincidentally, this was also an opportunity to clean up the old `libDefs.js` file (which had to manage partial bits of info from GitHub due to the need to request bits of information in phases and pages...yuck). It simplified all that down into the exact same code that's used to run tests -- so yay, less complexity!


marudor:
Doing a quick review - looks great!
I've got nothing against merging.

PR #84, opened 2016-05-10
Sign libdefs during `install` command

jeffmo:
This signs libdefs with a version (their "last updated" commit sha) so that we can detect if they've been locally modified or if there is a new version available.

We don't make use of the version or signature info just yet, but we'll likely want to in the near future:

* Tell the user if any of their installed libdefs have been updated since they last installed
* Detect local modifications and encourage users to contribute their libdef improvements upstream

### DIFF
--- a/cli/.eslintrc
+++ b/cli/.eslintrc
@@ -15,7 +15,7 @@
     "semi": 2,
     "no-unused-vars": ["error", {
       "varsIgnorePattern": "^_.?.?$",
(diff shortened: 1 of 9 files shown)


jeffmo:
(Note that this sits on top of https://github.com/flowtype/flow-typed/pull/82)

ryyppy:
This looks really cool, I like it!

PR #90, opened 2016-05-11
Express v4 definition file

bassettsj:
I wanted to get started writing the definition express. I am having a hard time structuring the module defition to be be accurate without declaring the types outside of the module block. Also, I am not sure if there is a better way to set the `Router` and `static` properties on the exported `express` function. I will try to work on this a bit more soon.


jeffmo:
> I am having a hard time structuring the module defition to be be accurate without declaring the types outside of the module block

Indeed, this is a known problem ([something I'd like to fix soon](https://github.com/flowtype/flow-typed/issues/16#issuecomment-197985601)). For now there are two options:

1. Declare the types *in* the module and use `var exports: ` (as you currently are). Using `var exports:` will override all other declares in the module -- effectively making them not an export.

1. Declare the types outside the module as you're doing here, but just namespace them with something very unique like `$npm$Express4$Response` (etc).

Another thing I noticed here: It looks like you're writing actual JS code in these files, but lib-defs should only include [`declare` statements](http://flowtype.org/docs/declarations.html).

bassettsj:
> Declare the types in the module and use var exports: (as you currently are). Using var exports: will override all other declares in the module -- effectively making them not an export.

Great, I will use this option. Now I guess the real problem is being able to declare a function type that creates the application `const app = express()`, but also able to set two properties to this function `static` and `Router`. I am stuck there now. That is why I was trying to write JS to get around this, but I am sure there must be a way to do this. 🙏 Thanks in advance.

jeffmo:
>  Now I guess the real problem is being able to declare a function type that creates the application `const app = express()`, but also able to set two properties to this function `static` and `Router`.

You can declare a "callable object" like this:

```js
declare module 'express' {
  declare class Application { ... }
  declare class Router { ... }
  declare function serveStatic(): middleware;
  declare module.exports: {
    () => Application, // If you try to call like a function, it will use this signature
    static: serveStatic, // `static` property on the function
    Router: typeof Router, // `Router` property on the function
  };
};
```

bassettsj:
@jeffmo awesome thank you. I will work on that.

definitions/npm/express_v4.x.x/test_express_v4.x.x.js
```diff
@@ -0,0 +1,17 @@
...
+
+// $ExpectError property `foo` Property not found in Application:
+app.foo();
+
+const myRouter = new express.Router();
+
+myRouter.get('/dang', (req, res) => {
+    res.json({ payload: 'stuff' });
+});
+// This should work but doesn't for now... :-(
```
bassettsj: Not sure why this is throwing an error. I am a bit confused on how to declare this type.

definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
```diff
@@ -0,0 +1,101 @@
...
+        download(path: string, filename?: string, callback?: (err?: ?Error) => void): void;
+        format(typesObject: {[type: string]: Function}): void;
+        json(body?: any): void;
+        jsonp(body?: any): void;
+        links(links: {[name: string]: string}): void;
+        location(path: string): void;
+        redirect(status?: number, path: location): void;
+        render(view: string, locals?: any, callback: (err?: ?Error) => any): void;
+        send(body?: any): void;
+        sendFile(path: string, options?: Object, callback?: (err?: ?Error) => any): void;
```
bassettsj: - [ ] add sendfile options.

bassettsj:
@jeffmo I was wondering about how I am trying to define this pattern `app.use([path,] function [, function...])`, currently I am trying to overload the method definitions like this in my Router class: 

```js
use(middleware: Middleware): void;
use(...middelware: Array<Middleware>): void;
use(path: string, ...middelware: Array<Middleware>): void;
```

It is now not being able to guess the types of the params in the function:

image showing the example middleware function

How should I better define that?

alexeygolev:
@bassettsj What exactly do you mean by 'not being able to guess the types of the params'? If by that you mean the fact that atom doesn't show you a tooltip then it might be not be a problem with your declarations. Will flow throw an error if you pass a string to res.status? (If you declare the type for `res` obviously)

bassettsj:
@alexeygolev It does throw and error if I annotate the prams. Flow is unable to guess the params without the annotations though. I guess that is all that I really wanted.

bassettsj:
@jeffmo What else do I need to add to get this in?

definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
```diff
@@ -0,0 +1,128 @@
...
+      mergeParams?: boolean,
+      strict?: boolean
+    };
+    declare class RequestResponseBase {
+        app: Application;
+        get(field: string): string | void;
+    }
+    declare class Request extends http$IncomingMessage mixins RequestResponseBase {
+      baseUrl: string;
+      body: any;
```
jeffmo: Should probably type this as `mixed` as it seems like a pretty significant source of types for most code

definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
```diff
@@ -0,0 +1,128 @@
...
+    declare class Response extends http$ClientRequest mixins RequestResponseBase {
+        headersSent: boolean;
+        locals: any;
+        append(field: string, value?: string): void;
+        attachment(filename?: string): void;
+        cookie(name: string, value: string, options?: CookieOptions): void;
+        clearCookie(name: string, options?: CookieOptions): void;
+        download(path: string, filename?: string, callback?: (err?: ?Error) => void): void;
+        format(typesObject: {[type: string]: Function}): void;
+        json(body?: any): void;
```
jeffmo: Let's [switch these params to `mixed`](https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md#avoid-any-when-possible)

jeffmo:
Oh sorry, I noticed there was discussion but didn't realize things were in a stable state.
Looks good! I commented about a couple of `any`-nits, then I think it makes sense to merge

ctrlplusb:
What about all the `Application` methods? i.e. `listen`, `disable`, etc

ctrlplusb:
Additionally, the `Response` and `Request` methods should return instances of themselves to allow for the fluent interface right?

bassettsj:
@jeffmo Thanks for the feedback, sorry I didn't add mixed before. The contrib docs are much better now.

@ctrlplusb I added the missing methods, thanks for pointing that out.

jeffmo:
Looks like some of the tests are failing (fyi)

jeffmo:
Sweet! Let's get this in.

We can always fix forward if anyone notices ways to improve/extend/tweak further.

bassettsj:
@jeffmo Thanks for all your help and agreed on improving.

neurovertex:
Is there a technical reason for Route/Router/App not to have [an all(...) function](https://expressjs.com/en/4x/api.html#router.all) ? I'm confused not to see anyone here mention it.
I added 
```js
all: RouteMethodType<this>;
```
in the Route class declaration and so far it's not causing any trouble

bassettsj:
@neurovertex Go ahead and open a PR with the changes, that would be great if it works, just make sure to add tests !

jtlapp:
Was I supposed to post all issues with the Express definition file in this thread? I posted some concerns about the error type at #454.

PR #93, opened 2016-05-12
CONTRIBUTING.md

jeffmo:
Explains the structure of the repository and the steps needed to write one before sending a pull request.

(cc @nmn -- and of course anyone else -- for feedback)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,71 @@
+## Contributing to the definitions repository
+
+All definitions are contained in the [definitions/](https://github.com/flowtype/flow-typed/tree/master/definitions) 
+directory and are named and structured very explicitly.
+
+Let's say you've created a definition for an npm package called `left-pad` at 
+version 4.1.0. **[You've followed all of the 
+best practices for writing high quality definitions,](https://github.com/flowtype/flow-typed/issues/13#issuecomment-214892914)**
+and now you'd like to contribute it:
+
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,71 @@
...
+
+You can write as many test files as you like as long as their names start with
+`test_`. 
+
+All test files that sit in the package version directory will be run by
+the test-runner for *all* versions of Flow the package version supports. If you 
+ever need to write a test for a particular version of Flow, you can put the 
+`test_` file in the appropriate flow version directory.
+
+#### 5) Run your tests with `./run_def_tests.sh left-pad`
```
bsr203: is this correct as I cannot locate any  run_def_tests.sh

If `runtests.sh` was the correct one, running like `runtests.sh left-pad`, ended up running all the packages, not just `left-pad`. Please verify.

the way it worked for moe was `cd cli && node dist/cli.js run-tests left-pad`

jeffmo: run_def_tests.sh is included in this diff :)

(Should be a little faster than the current runtests.sh)

bsr203: oops.. didn't notice it :-) thanks

nmn: Looks good!
A ASCII art file tree will probably be helpful.

jeffmo:
An ASCII file structure outline is a good idea! I'll see if I can come up with something on Monday

PR #94, opened 2016-05-13
Move npm libs under a `definitions/npm/` namespace

jeffmo:
We've talked on the Flow team about potentially moving even internal libdefs into `flow-typed` one day.
Additionally there are some JS libraries out there that just aren't on npm.

In order to preserve future option value for these circumstances, I figured we should just move npm-package libdefs into their own directory namespace to leave room for non-npm libdefs in the future.


jeffmo:
The only downside to waiting and doing this later is that people will have `npm install -g`-ed the old version of the CLI -- which won't understand the new directory structure.

In that case we *could* just bump the `minimumCLIVersion` number (which would prevent those users from rebasing until they upgraded) -- but I figured we could save everyone the thrash if we just do it now.

ryyppy:
So what is blocking the merge of this repo?

jeffmo:
Nothing really, just wanted to give other collabs a chance to see it before I merged

PR #95, opened 2016-05-13
Adds ramda definitions

alexeygolev:
It's still lacking quite a bit of definitions but is already quite useful. I use mocha to test that what I'm using to type check actually make sense as code. I removed all calls to `expect` and provided fake `it`s and `describe`s so it can be used without mocha.
I planning on finishing the rest of definitions/test within a week

### DIFF
--- a/definitions/ramda_v0.21.x/flow_>=v0.23.x/ramda_v0.21.x.js
+++ b/definitions/ramda_v0.21.x/flow_>=v0.23.x/ramda_v0.21.x.js
@@ -1,0 +1,577 @@
+type UnaryFn<A,R> = (a: A) => R
+type BinaryFn<A,B,R> = ((a: A, b: B) => R) & ((a:A) => (b: B) => R)
+
+type UnarySameTypeFn<T> = UnaryFn<T,T>
+type BinarySameTypeFn<T> = BinaryFn<T,T,T>
+
+interface ObjPredicate {
+  (value: any, key: string): boolean;
+}
+type UnaryPredicateFn<T> = (x:T) => boolean
+type BinaryPredicateFn<T> = (x:T, y:T) => boolean
+
+declare class Composition {
+  pipe<A,B,C,D,E,F,G>(ab: UnaryFn<A,B>, bc: UnaryFn<B,C>, cd: UnaryFn<C,D>, de: UnaryFn<D,E>, ef: UnaryFn<E,F>, fg: UnaryFn<F,G>, ...rest: Array<void>): UnaryFn<A,G>;
+  pipe<A,B,C,D,E,F>(ab: UnaryFn<A,B>, bc: UnaryFn<B,C>, cd: UnaryFn<C,D>, de: UnaryFn<D,E>, ef: UnaryFn<E,F>, ...rest: Array<void>): UnaryFn<A,F>;
+  pipe<A,B,C,D,E>(ab: UnaryFn<A,B>, bc: UnaryFn<B,C>, cd: UnaryFn<C,D>, de: UnaryFn<D,E>, ...rest: Array<void>): UnaryFn<A,E>;
+  pipe<A,B,C,D>(ab: UnaryFn<A,B>, bc: UnaryFn<B,C>, cd: UnaryFn<C,D>, ...rest: Array<void>): UnaryFn<A,D>;
+  pipe<A,B,C>(ab: UnaryFn<A,B>, bc: UnaryFn<B,C>, ...rest: Array<void>): UnaryFn<A,C>;
+  pipe<A,B>(ab: UnaryFn<A,B>, ...rest: Array<void>): UnaryFn<A,B>;
+  compose<A,B,C,D,E,F,G>(fg: UnaryFn<F,G>, ef: UnaryFn<E,F>, de: UnaryFn<D,E>, cd: UnaryFn<C,D>, bc: UnaryFn<B,C>, ab: UnaryFn<A,B>, ...rest: Array<void>): UnaryFn<A,G>;
+  compose<A,B,C,D,E,F>(ef: UnaryFn<E,F>, de: UnaryFn<D,E>, cd: UnaryFn<C,D>, bc: UnaryFn<B,C>, ab: UnaryFn<A,B>, ...rest: Array<void>): UnaryFn<A,F>;
+  compose<A,B,C,D,E>(de: UnaryFn<D,E>, cd: UnaryFn<C,D>, bc: UnaryFn<B,C>, ab: UnaryFn<A,B>, ...rest: Array<void>): UnaryFn<A,E>;
+  compose<A,B,C,D>(cd: UnaryFn<C,D>, bc: UnaryFn<B,C>, ab: UnaryFn<A,B>, ...rest: Array<void>): UnaryFn<A,D>;
+  compose<A,B,C>(bc: UnaryFn<B,C>, ab: UnaryFn<A,B>, ...rest: Array<void>): UnaryFn<A,C>;
+  compose<A,B>(ab: UnaryFn<A,B>, ...rest: Array<void>): UnaryFn<A,B>;
+}
+
+interface CurriedFunction2<T1, T2, R> {
+  (t1: T1): (t2: T2) => R;
+  (t1: T1, t2: T2): R;
+}
+
+interface CurriedFunction3<T1, T2, T3, R> {
+  (t1: T1): CurriedFunction2<T2, T3, R>;
+  (t1: T1, t2: T2): (t3: T3) => R;
+  (t1: T1, t2: T2, t3: T3): R;
+}
+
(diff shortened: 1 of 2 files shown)


definitions/ramda_v0.21.x/flow_>=v0.23.x/ramda_v0.21.x.js
```diff
@@ -0,0 +1,603 @@
+/* @flow */
+/* eslint-disable no-unused-vars, no-redeclare */
+
+type UnaryFn<A,R> = (a: A) => R
```
jeffmo: Things outside a `declare module` get declared in the global scope -- so we should try to namespace these appropriately. Alternatively, if it's not really useful to reference these types outside of the module definition itself, you can move them inside the `declare module` and be sure the `declare class`es will be hidden from the global scope (since you are using `declare var exports: ...`).

jeffmo:
This is awesome!!
Lets move the global `declare`s into the `declare module` and then I think we can merge

alexeygolev:
Thanks! Will fix tonight

rjmk:
This is *so* timely! I have just been hamfistedly adding solely unary versions of Ramda functions to my project.

Could we maybe define some types for semigroup, etc, to capture some more of the dispatches that Ramda does?

```js
declare interface Semigroup {
  concat<s: Semigroup>(x: s): s
}
```

I dunno if there's a way to make sure `s` is the same type as the type of the implementer of Semigroup?

rjmk:
Also, I think there might be a little bit of an issue with some of the curried functions.

Like,

```js
  propOr<T>(v: T, key: string, o: Object): ?any|T;
  propOr<T>(v: T, key: string): (o: Object) => ?any|T;
```

doesn't include
```js
  propOr<T>(v: T): (key: string, o: Object) => any|T;
  propOr<T>(v: T): (key: string) => (o: Object) => ?any|T;

**Comment by jeffmo** [comment]
@rmjk: I'm very unfamiliar with the specific details of Ramda, but if I understand what you're saying correctly then I think what you wrote for `Semigroup` should work just fine.

Maybe you can test it out and confirm?

**Comment by alexeygolev** [comment]
@rjmk @jeffmo Although I do use fantasy land types myself I would like to finish the declarations for the general types first (as you pointed out there are still some missing). Fantasy land is a bit more complex as well as lenses. But when it's merged I guess we can all sort additional things out? I'm going to fix the global scope issue in a couple of hours

**Comment by rjmk** [comment]
Sure thing, @jeffmo.

To be clear, my concern is that something like the possible would be allowable as a Semigroup instance

```js
function notSemi () {
 this.concat = s => s
}
```

But then we could `(new Semi).concat([])` and not be in the type.

In pseudo-syntax, we would be looking at

```js
declare interface Semigroup {
  concat(x: typeof this): typeof this
}
```

Or perhaps I'm not reasoning about the type inference correctly?

alexeygolev:
@rjmk  https://github.com/hallettj/safety-lens will possibly give you an idea of possible fantasy land declarations. I haven't tested them but will do after finishing the "standard" types for ramda.

rjmk:
@alexeygolev OK, awesome. Thank you for the tips.

Another question: is there a reason you're not defining the Ramda functions using your excellent `CurriedFunction` interfaces?

alexeygolev:
@rjmk no reason, to be honest. As it is very much in progress I'm trying to get things right by writing the declarations manually before jumping into abstracting everything away. At the moment I'm fighting with tranducers (into, transduce etc) and boy what a wonderful fight it is:) As soon as we battle test the declarations (and I must say I really enjoy using it in one of my work projects) we can clean everything up. But then again — it's not my project, it's for all of us. So you might want to test things for yourself and submit a PR that will use abstractions sooner than I'll get to it. There are some fundamental underlying ideas that are discarded at the moment (like fantasy land) but because of the way ramda dispatches some methods it's quite tricky to get them right without some research and deep diving into ramda's code. I used some of the typescript interfaces as a references but there were too many 'any' types to my taste so it's already more specific. However it's still very much work in progress. Ramda has a lot going on behind the scenes.
I'm hoping to push transducer related code in a couple of days. As well as some other List methods that we lack. I'm dreading to start on lenses:)

rjmk:
@alexeygolev I've been hacking on this version for use with a project. I've had no problems with

```js
type Lens<s, a> = (f: (a: a) => Functor<a>) => (s: s) => Functor<s>;

lens<s, a>(f: (s: s) => ?a): (g: ((a: a, s: s) => s)) => Lens<s, a>;
lensPath<s, a>(path: string[]): Lens<s, a>;
lensProp<s, a>(prop: string): Lens<s, a>;

over<s,a>(lens: Lens<s,a>): (f: (a: a) => a) => (s: s) => s;
set<s, a>(lens: Lens<s, a>): (a: a) => (s: s) => s;
view<s, a>(lens: Lens<s, a>): (s: s) => a;
```

and I've just started rewriting several of the functions declarations with `CurriedFunction` very happily. I have had to declare some type variables on the classes, which I'm rather wary of. My wariness or my approach may well be a symptom of the fact that I just started playing seriously with flow today so have a lot to learn.

definitions/ramda_v0.21.x/test_ramda_v0.21.x.js
```diff
@@ -0,0 +1,225 @@
...
+const _ = require('ramda')
+function describe(tag, fn) {
+  fn()
+}
+
+function it(tag, fn) {
+  fn()
+}
+
+describe('Functions', () => {
```
jeffmo: whoa, this is a cool idea!

I wonder if we could generalize the notion of using a runtime test runner + types for a better way of ensuring that our libdefs match up with their corresponding libraries...

cc @marudor @splodingsocks @ryyppy

alexeygolev:
@rjmk The problem with your lens declarations is that they follow ramda's ersatz Haskell declarations. Yet I can easily do this
```js
var headLens = R.lensIndex(0);

R.over(headLens, () => 1, ['foo', 'bar', 'baz'])
```
breaking the (f: (a: a) => a) type for the second argument of over. 
In fact it's wrong even for Haskell because there is only one function that has an `a -> a` signature. And clearly we're not going to use identity only.
I love Haskell and I love it's type system but as much as I want to have Lenses defined through fantasy land interfaces I think we really need to think it through in the context of flow type system (not the same but close — [scala](http://scalabound.org/?page_id=201#localTypeInference).

jeffmo:
@alexeygolev: Thanks for moving the globals around.
I'll go ahead and merge and, as you say, you guys can continue to improve this libdef over time

alexeygolev:
@jeffmo Thank you!
Ramda has enough concepts thrown in there to play around with flow type system. I would love to keep discussion open because we might find some 'good practices' for general flowtype use in different scenarios.

alexeygolev:
@jeffmo oops... I cloned the repo before you moved everything into npm.

rjmk:
@alexeygolev That link looks awesome. I take your point that it's not quite right as (at least as I understand it) we may change the type of our record.

Also, I don't think it's quite right that there's only one function `a -> a` in this context, just as `(a -> a) -> a -> a` does not need to have `id` provided as its first argument

jeffmo:
@alexeygolev no prob -- fixed

alexeygolev:
@jeffmo Thanks a lot
@rjmk I propose to create an issue where we can discuss possible solutions if need be so we move the conversation out of the closed PR if you're OK with it :)
I'm planning on finishing the transducers stuff (at least in a very basic form) and then move on to Lenses

rjmk:
@alexeygolev Sounds good. Ping me if you open it up

rpominov:
@alexbooker 

Functor interface looks wrong. Should it be like this:

```diff
   declare interface Functor<A> {
-    map<A,B>(fn:(a: A) => B): B;
+    map<A,B>(fn:(a: A) => B): Functor<B>;
   }
```

?

alexeygolev:
@rpominov You're absolutely right. As you can see from the code I haven't yet have time to actually generalize map to use functors. But your signature is the correct one

rpominov:
Actually it can be a bit more correct I think:

```diff
   declare interface Functor<A> {
-    map<A,B>(fn:(a: A) => B): Functor<B>;
+    map<B>(fn:(a: A) => B): Functor<B>;
   }
```

But anyway, it looses type info about return value, which becomes any Functor, instead of a specific type like Maybe etc. This is probably a well known limitation. But I wonder if there any plans to make it possible to define a proper Functor interface somehow in Flow? @jeffmo Maybe you could answer?

alexeygolev:
@rpominov Check [this issue](https://github.com/facebook/flow/issues/30)

Risto-Stevcev:
This is awesome, but I don't know how far you could get until you hit a wall with this because flow can't express higher kinded types. How would you write the definition for `sequence`?

```
sequence :: forall a m. Applicative m => t (m a) -> m (t a)
```

PR #112, opened 2016-05-20
rxjs: Properly type Observables that never produce values

hansonw:
`empty`, `never`, and `ignoreElements` should be typed as "an observable of some type" rather than Observables of any. This prevents possible type degradation when they're combined with Observables of well-defined types.


jeffmo:
Can you add a couple of regression tests to `definitions/npm/rxjs_v5.0.x/test_rxjs.js` that verify this behavior as well?

hansonw:
Ah, cool! will do

hansonw:
Added two test cases that fail with the existing definitions.

jeffmo:
woo!

PR #119, opened 2016-05-25
Type rxjs_v5.0.x Observable.cache() method

shushz:
Typing the cache operator from: https://github.com/ReactiveX/rxjs/blob/5.0.0-beta.8/src/operator/cache.ts#L14


jeffmo:
Looks like this got merged over in https://github.com/flowtype/flow-typed/pull/120/files, but that PR doesn't have a 3rd `scheduler` argument... Should it have?

cc @matthewwithanm

shushz:
It should theoretically, but Schedulers are not yet typed at all in any of the other methods. I chose to put it as `mixed` as a placeholder for the time being.
I guess this can be closed then.

PR #123, opened 2016-05-27
Add underscore.string flow definitions

torifat:
http://epeli.github.io/underscore.string/

`chain` is implemented with `any` for now. I tried to make it work but couldn't. Can you please give me some pointers?


definitions/npm/big.js_v3.x.x/test_big.js_v3.x.x.js
```diff
@@ -1,6 +1,5 @@
 // @flow
-
-import Big from "big.js";
+import Big from 'big.js';
```
jeffmo: Guessing this wasn't intentional?

(Since libdefs are signed/versions once they are installed with the CLI, it's good not to change the contents unless there's an update/improvement/fix/etc)

definitions/npm/underscore.string_v3.x.x/flow_all/underscore.string_v3.x.x.js
```diff
@@ -0,0 +1,87 @@
...
+    seperator?: string,
+    trailingSpaces?: boolean,
+    cut?: boolean,
+    preserveSpaces?: boolean,
+  };
+  declare type $npm$pad$type = 'left' | 'right' | 'both';
+
+  declare class S {
+    // TODO: fix it
+    static chain(string: string): any;
```
jeffmo: I'm looking at the API docs, and I don't see that `chain()` is actually a static method?

It looks like instead the `exports` object is a function with lots of static methods on it? When the function is called, it returns an object with a version of those same methods (except they have a slightly different signature:

`toBoolean: (str: string) => boolean` vs `chain.toBoolean: () => boolean`
(note that one has an input param and the other doesn't)

In this case, you're probably better off defining both a class for `chain` objects, and a "callable object" for the `exports` object:

```js
declare module "underscore.string" {
  ...

  declare class Chain {
    toLowerCase(str: string): string;
    ...
  }

  declare module.exports: {
  };
}
```

jeffmo: (of course theres a chance I'm mis-reading the docs? If so, let me know and I'll see if I can help further)

definitions/npm/big.js_v3.x.x/test_big.js_v3.x.x.js
torifat: I changed it to keep things consistent but I get the point. Thanks for explaining :)

definitions/npm/underscore.string_v3.x.x/flow_all/underscore.string_v3.x.x.js
torifat: No, you are right. Like I mentioned before, I couldn't come up with a solution that worked properly. I will try to implement it the way you suggested. Thanks 👍

torifat:
@jeffmo I have made all the necessary changes.

definitions/npm/underscore.string_v3.x.x/flow_all/underscore.string_v3.x.x.js
```diff
@@ -0,0 +1,167 @@
...
+    substring(indexStart: number, indexStart?: number): Chain,
+    substr(start: number, length?: number): Chain,
+    concat(...params: Array<string>): Chain,
+
+    // Chain specific methods
+    tap(fn: (string: string) => string): Chain;
+    value(): string;
+  }
+
+  declare var exports: {
```
thejameskyle: `declare var exports` is now deprecated in favor of `declare module.exports:`

mind updating that here?

torifat: @thejameskyle initially I used `declare module.exports` but it doesn't work for flow version prior to 0.25 😞.

Ok, moving `flow_all` to `flow_>=v0.25.x` then.

torifat: @thejameskyle done.

torifat:
@jeffmo I have made all the necessary changes. Can you have a look? I know you are busy. But, posted this just as a reminder. Hope it's ok with you. Thanks.

thejameskyle:
This looks great, thank you so much @torifat!

PR #124, opened 2016-05-27
Add types for 'bull'

LegNeato:


definitions/npm/bull_v1.x.x/flow_>=v0.22.x/bull_v1.x.x.js
```diff
@@ -0,0 +1,56 @@
+declare type $npm$bull$jobType =
```
ryyppy: Okay, that's new... why prefix your classes with `$npm$bull` instead of just `bull$`? :-)

LegNeato: I saw most others in the repo doing so (https://github.com/flowtype/flow-typed/blob/master/definitions/npm/big.js_v3.x.x/flow_%3E%3Dv0.17.x/big.js_v3.x.x.js#L3) and it was suggested in a different PR I happened to read to make sure it is a unique name. I assume the fact that there is an `npm` subdir in this project that it means potentially there is some other package manager/universe expected. I can imagine a world where `bull` isn't a unique name, so decided to be "safe" as suggested.

ryyppy: Usually, we just skip the `$npm` an would just write `{packagename}${type|...}`
I opened an issue (#126) for further discussion!

jeffmo:
Woo, thanks! Sorry for the delay

PR #127, opened 2016-05-27
Add tape v4.5.x libdef

ryyppy:
Finally got this working... I am not 100% happy with how flow infers the test-context function `(t) => { ... }`, when provided as first or second argument,.... but yeah, it's basically working with all of my (extensive) tape tests on our current projects.


marudor:
Any Reason you made all the functions variables?
As in ```fail: () => void``` instead of ```fail(): void```

Other than that looks good.

ryyppy:
Hm, I am probably just not used to write classes or maybe I started out with a `Object` instead of an `interface`... I will change that 👍

marudor:
Okay, gonna merge once changed.

PR #129, opened 2016-05-28
add fuzzy-install to filterLibDefs's types

joe-re:
fizzy filter should filter on partial name for search command.
But I think, in case of install command, it should filter using correct package name.

How do you think?

# before
```
$ flow-typed install -o -f 0.26.0 redux
 * found 3 matching libdefs.
'redux-actions_v0.9.x.js' installed at /Users/noguchimasato/.ghq/github.com/joe-re/pomodo-desuka/flow-typed/npm/redux-actions_v0.9.x.js.
```

oops...installed redux-action.

# after
```
$ flow-typed install -o -f 0.26.0 redux
 * found 1 matching libdefs.
'redux_v3.x.x.js' installed at /Users/noguchimasato/.ghq/github.com/joe-re/flow-typed/cli/flow-typed/npm/redux_v3.x.x.js.
```

### DIFF
--- a/cli/src/commands/install.js
+++ b/cli/src/commands/install.js
@@ -98,7 +98,7 @@ export async function run(args: Args): Promise<number> {
     };
   } else {
     filter = {
-      type: 'fuzzy',
+      type: 'fuzzy-install',
       term: defName,
       flowVersion,
     };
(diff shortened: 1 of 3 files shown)


jeffmo:
Ahh, this is much better!

PR #132, opened 2016-06-08
Expand mocha coverage to include body-less it()

glenjamin:
As noted in http://mochajs.org/#pending-tests


jeffmo:
Nice, thanks!

Issue #111, opened 2016-05-20
[CLI] Need a `flow-typed update-cache` command


jeffmo:
Currently the cache updates itself if it's been more than 5 days since it was last used.
We should also include an `update-cache` command so users can quickly update the cache themselves.

(Also, while we're at it, might as well add a `--update-cache` flag to the `install` command too)

ryyppy:
Maybe I will give it  a shot if nobody is working on that!

jeffmo:
Go for it!

ryyppy:
This issue should be closed ... I cannot do it sadly :-(

PR #135, opened 2016-06-08
Add update-cache command

ryyppy:
This commit also adds another parameter to the ensureCacheRepo function,
so we can reuse the logic for updating the cache-repo without a 5-day
limit (which is still the limit by default).

Issue: #111 

@jeffmo

### DIFF
--- a/cli/src/lib/libDefs.js
+++ b/cli/src/lib/libDefs.js
@@ -66,7 +66,7 @@ async function rebaseCacheRepo(verbose?: VerboseOutput) {
 /**
  * Utility wrapper for ensureCacheRepo with an update expiry of 0 hours.
  */
-async function updateCacheRepo(verbose?: VerboseOutput): Promise<boolean>{
+async function updateCacheRepo(verbose?: VerboseOutput) {
   return await ensureCacheRepo(verbose, 0);
 }
 
--- a/cli/src/lib/libDefs.js
+++ b/cli/src/lib/libDefs.js
@@ -64,11 +64,23 @@ async function rebaseCacheRepo(verbose?: VerboseOutput) {
 }
 
 /**
+ * Utility wrapper for ensureCacheRepo with an update expiry of 0 hours.
+ */
+async function updateCacheRepo(verbose?: VerboseOutput): Promise<boolean> {
+  try {
+    await ensureCacheRepo(verbose, 0);
+    return true;
+  } catch (e) {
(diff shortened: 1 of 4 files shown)


cli/src/lib/libDefs.js
```diff
@@ -64,11 +64,23 @@ async function rebaseCacheRepo(verbose?: VerboseOutput) {
 }
 
 /**
+ * Utility wrapper for ensureCacheRepo with an update expiry of 0 hours.
+ */
+async function updateCacheRepo(verbose?: VerboseOutput): Promise<boolean> {
+  try {
```
jeffmo: Probably better to just throw and give detailed errors.
That way if there are issues, people can paste helpful debug info in the bug report

jeffmo: I see above that you remind the user to re-run the command for more detailed info...

That's fair, though I can't imagine when you'd ever *not* want the detailed info. Did you have a circumstance in mind?

jeffmo: Oh, but that detailed info doesn't actually contain the error message anyway... Yea, I still think we should just throw up to the toplevel.

ryyppy: Actually, I am not 100% happy with the granularity of the different functions... initially I was thinking about refactoring stuff around so it would be more modular, but then I saw the jest tests and whatsoever... :-)

It would also be possible to just arrow-function the updateCacheRepo and try / catch the error in the command itself (instead of returning a boolean)

jeffmo: I think it'd be fine to expose `updateCacheRepo` as a public API (I only didn't before because it didn't need to be -- but clearly does here)

What's the downside to just printing the exception on the command line, though? Exceptions should be exceptional, and when they do happen they give a lot of useful info for reporting the bug in the GitHub issue.

ryyppy: Added the changes... not sure if I should always output the complete error object (I currently hide this behind the debug flag)... if no debug flag is provided, it will print the `e.message` only...

I personally get confused when a stack-trace is printed whenever I use a commandline-tool (usually tastes like a broken tool)... what do you think?

jeffmo: Yea, I think this is reasonable. If someone comes with an error without context, we can always ask them to run the command again with `-d` I suppose

PR #148, opened 2016-06-17
Add a definition for magic-string.

eventualbuddha:
This doesn't pass tests, but I can't figure out why. Any ideas?

@Rich-Harris, do these definitions look right to you?

### DIFF
--- a/definitions/npm/magic-string_v0.15.x/flow_>=v0.27.x/magic-string_v0.15.x.js
+++ b/definitions/npm/magic-string_v0.15.x/flow_>=v0.27.x/magic-string_v0.15.x.js
@@ -1,0 +1,78 @@
+/* @flow */
+
+declare module "magic-string" {
+  declare type SourceMapOptions = {
+    file?: string;
+    source?: string;
+    includeContent?: boolean;
+    hires?: boolean;
+  };
+
+  declare class SourceMap {
+    version: number;
+    file: ?string;
+    sources: Array<?string>;
+    sourcesContent: Array<?string>;
+    names: Array<string>;
+    mappings: string;
+  }
+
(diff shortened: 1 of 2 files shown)


jeffmo:
Looks like you need to wrap all of the declarations in a `declare module "magic-string" { }` container

jeffmo:
As a random example, here's a definition for the `"yargs"` module:

https://github.com/flowtype/flow-typed/blob/master/definitions/npm/yargs_v4.x.x/flow_%3E%3Dv0.23.x/yargs_v4.x.x.js

eventualbuddha:
Okay, I added the `declare module` wrapper, but now I can't figure out how to model a default export constructor with a named export constructor. I should be able to import like so:

```js
import MagicString, { Bundle } from 'magic-string';
```

But I don't know how to do that using the `declare var exports: …;` syntax. Any ideas?

jeffmo:
The bad news: It's not possible to simultaneously declare both named and default parameters in a libdef in <=0.27 :(

The good news: Support for doing this just landed in master and will be available in 0.28! https://github.com/facebook/flow/commit/93f0614d4d69454b4d60d7cda9fb23d7e617dff8 https://github.com/facebook/flow/commit/ada27dbd1b154777d67133af0233da8e310f781a

jeffmo:
In the meantime the best thing you can do is declare statics on the `MagicClass ` class and import/use with `import MagicClass from "magic-class"; MagicClass.Bundle`

jeffmo:
FYI: Flow v0.28 shipped -- so it should be possible to use `declare export ...` syntax in this libdef now (if you bump the Flow version to `>=0.28.0`)

marudor:

PR #149, opened 2016-06-17
Bunyan definition

bassettsj:
This should add lib def for bunyan.


marudor:
Looks similar to the one I started for work.
Although there is quite some any it should be correct - mixed wouldn't help there.

PR #154, opened 2016-06-20
Add RxJS `groupBy` operator

### LINKED ISSUE #153: `groupBy` missing from rxjs types
The [GroupBy operator](http://reactivex.io/documentation/operators/groupby.html) is missing from the RxJS type definitions.

ssorallen:
> The [GroupBy operator][1] divides an Observable that emits items into an
Observable that emits Observables, each one of which emits some subset
of the items from the original source Observable.

Fixes #153

[1]: http://reactivex.io/documentation/operators/groupby.html


ssorallen:
cc @matthewwithanm

PR #156, opened 2016-06-22
support Enzyme 2.3 additions

turadg:
https://github.com/airbnb/enzyme/blob/master/CHANGELOG.md

Is this the right way to have a 2.x.x and 2.3.x matcher co-exist? (Actually, is that supported yet? The run-tests command seems to assume there will be only one because when added this dir it no longer picked up 2.x.x)


jeffmo:
Yea, this is the right way to have both. There's a weird bug in the `run-test` command when you use a filter that matches multiple libs. The bug is fixed over in https://github.com/flowtype/flow-typed/pull/143 (did you know that `RegExp.prototype.test()` is stateful??? Ugh.) -- I just need to finish that up and land it.

Luckily it looks like the Travis test did run tests for both and everything is passing here

PR #158, opened 2016-06-29
Add listen method and Router class definitions

yamadayuki:
ref. http://expressjs.com/en/starter/hello-world.html

> ```js
app.listen(3000, function () {
  console.log('Example app listening on port 3000!');
});
```

`listen` can take function parameter on the second parameter.
And add Router class declaration `$Router`.

### DIFF
--- a/definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
@@ -206,6 +206,9 @@ declare module 'express' {
     constructor(): void;
     locals: {[name: string]: mixed};
     mountpath: string;
+    listen(port: number, ...params: Array<mixed>): Server;
+    listen(port: number, hostname?: string, ...params: Array<mixed>): Server;
+    listen(port: number, hostname?: string, backlog?: number, ...params: Array<mixed>): Server;
     listen(port: number, hostname?: string, backlog?: number, callback?: (err?: ?Error) => mixed): Server;
     listen(path: string, callback?: (err?: ?Error) => mixed): Server;
     listen(handle: Object, callback?: (err?: ?Error) => mixed): Server;
--- a/definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
@@ -207,9 +207,9 @@ declare module 'express' {
     constructor(): void;
     locals: {[name: string]: mixed};
     mountpath: string;
-    listen(port: number, hostname ? : string, backlog ? : number, callback ? : (err ? : ? Error) => mixed): Server;
-    listen(port: number, hostname ? : string, callback ? : (err ? : ? Error) => mixed): Server;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jeffmo** [comment]
cc @bassettsj (who wrote the original) for general review

**Comment by jeffmo** [comment]
Looks pretty good to me, and the tests pass -- so will give @bassettsj a chance to 👍 as well before merging

**Comment by bassettsj** [comment]
:+1: Looks great. Thanks for adding that.

**Comment by jeffmo** [comment]
Good call -- mind adding 1 or 2 `// $ExpectError` tests?
I'm concerned we'll forget if we don't just do it now :)

**Comment by yamadayuki** [comment]
@jeffmo Sorry to slow response :bow:
Added 1 small negative test case. Please review once again :pray:

**Note**
Add `use(path: string, router: Router): this` declaration.

WHY
```js
const foo: $Router = new Router();

app.use('/foo', foo);
          |      |
          |     Router
        string
```

Use case in test.
https://github.com/yamadayuki/flow-typed/blob/fbdae18f4a27cccd1804244050303b9ab8f2798d/definitions/npm/express_v4.x.x/test_express_v4.x.x.js#L67-L74

definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
```diff
@@ -196,9 +196,11 @@ declare module 'express' {
     use(middleware: Middleware): this;
     use(...middelware: Array<Middleware>): this;
     use(path: string, ...middelware: Array<Middleware>): this;
+    use(path: string, router: Router): this;
 
     route(path: string): Route;
   }
+  declare type $Router = Router;
```
jeffmo: Because `Router` is a class, it's already possible to use `import type` with it -- so no need to define a special type export here.

(Same goes for `Application`/`$Application` now that I'm looking. I didn't notice this in the original libdef)

jeffmo:
Looks like there's some kind of merge conflict on this branch, too?

definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
yamadayuki: Ah, I misunderstood it. I thought that I needed to define a special type `$Router` to use `import type { $Router }`, but it was unnecessary. Thank you! I'll remove it.

jeffmo: Yea, so just to clarify: You can use `import type` with any thing that is a "type". Classes are a special kind of thing that is simultaneously a type *and* a value -- so it's ok to use `import type` on classes.

jeffmo:
Thanks! Any clue why GitHub thinks there are merge conflicts?

yamadayuki:
Thank you, too! But, I investigate what conflicts... Sorry, I don't know how to resolve it...

jeffmo:
squash-merged in https://github.com/flowtype/flow-typed/commit/d438a9780d1317a2f4056c61f5a6e6327db795a5

yamadayuki:
Thanks! I appreciate your help.

PR #159, opened 2016-06-30
Add uuid.

LegNeato:
Note this is for `uuid`, not `node-uuid` or `uuid-v4`.

This one:

https://github.com/defunctzombie/node-uuid


jeffmo:
Looks like some expected errors aren't firing in <=0.26.x.

Should we just set the minimum version of this libdef to `>=0.27.x`?

Kriegslustig:
Hi, I've since implemented it and it was merged. Sorry, I didn't know that there was an existing, open PR at the time.

@jeffmo you may want to close this.

PR #164, opened 2016-07-05
Support pad 1.x

thejameskyle:


definitions/npm/pad_v1.x.x/flow_>=v0.27.x/pad_v1.x.x.js
```diff
@@ -0,0 +1,13 @@
+type padOptions = {
```
ryyppy: Everything which would end up in the global scope has to be namespaced...
In this case it would be `$npm$pad$padOptions` or maybe just `$npm$pad$Options` :-)

thejameskyle:
Updated

marudor:
Looks nice for me - the failed tests are due to the .28 release - already on it.
Nothing keeps us from merging this.

PR #168, opened 2016-07-05
Update classnames for Flow v0.28.0

nmote:
This removes the `{[className: string]: boolean }` option for Flow v0.28.0. The union/intersection fixes force annotations on all object literals passed to `classnames`, to disambiguate between `{[className: string]: boolean }` and `{[className: string]: ?boolean }`. So, while this will no longer accept objects explicitly typed as mapping to `boolean`, that seems preferable to requiring annotations everywhere. In practice it works pretty well in the Nuclide codebase.

If Flow stops requiring annotations in obvious cases like this, or if we get readonly props, we will be better off... in the meantime, this seems like the right tradeoff.


jeffmo:
Let's copy this rather than rename it. That way people using older versions of Flow still get the older/less-naggy level of precision.

nmote:
Not sure what you mean -- this change does only apply to 0.28 forward. I just renamed the folder for the old definitions to only apply to Flow versions 0.23-0.27.

jeffmo:
My bad, looking closer I see what you mean. Looks good

PR #172, opened 2016-07-06
Improve enzyme declarations

strax:
Fixes chaining to have correct types instead of `any`, adds missing API methods and declares named exports correctly for Flow >= v0.28


jeffmo:
cc @turadg (who wrote the original) for general review

turadg:
LGTM, thanks!

definitions/npm/enzyme_v2.3.x/flow_>=v0.28.x/enzyme_v2.3.x.js
```diff
@@ -0,0 +1,74 @@
...
+    map<T>(fn: (node: this) => T): Array<T>;
+    reduce<T>(fn: (value: T, node: this, index: number) => T, initialValue?: T): Array<T>;
+    reduceRight<T>(fn: (value: T, node: this, index: number) => T, initialValue?: T): Array<T>;
+    some(selector: EnzymeSelector): boolean;
+    someWhere(predicate: PredicateFunction<this>): boolean;
+    every(selector: EnzymeSelector): boolean;
+    everyWhere(predicate: PredicateFunction<this>): boolean;
+  }
+
+  declare export class ReactWrapper extends Wrapper {
```
jeffmo: chyessss :)

turadg:
I should have tested this before replying, but I just tried this out with Flow 0.28 and I get a bunch of this error,
```
property `length`. Property not found in ShallowWrapper
```

My temp work-around is to add the prop to Wrapper,
```
    length: number;
```

Any better solution?

strax:
@turadg right, that's not documented in itself but seems to be in the code and the first example [here](https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md#shallow-rendering-api) implies it's available, so I think it's a good idea to get that merged in as well :smile:

PR #174, opened 2016-07-06
Support commander 2.x

thejameskyle:


definitions/npm/commander_v2.x.x/test_commander_v2.x.x.js
```diff
@@ -0,0 +1,13 @@
+import program from 'commander';
```
jeffmo: Let's add some [`$ExpectError`s](https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md#4-write-a-test-file-whose-name-starts-with-test_-in-the-package-version-directory) in these tests too, pls

definitions/npm/commander_v2.x.x/flow_>=0.28.x/commander_v2.x.x.js
```diff
@@ -0,0 +1,276 @@
...
+  description: string;
+}
+
+type $npm$commander$exports = $npm$commander$Command & {
+  Command: $npm$commander$Command,
+  Option: $npm$commander$Option
+};
+
+declare module "commander" {
+  declare var exports: $npm$commander$exports;
```
jeffmo: `declare var exports` is now deprecated in favor of `declare module.exports: `

I'm waiting a few more versions before I start emitting a deprecation notice here (I want to give the syntax some time to work its way into history a bit), but since this libdef is designated `>=0.28.0` it's good to start using the new syntax.

definitions/npm/commander_v2.x.x/flow_>=0.28.x/commander_v2.x.x.js
```diff
@@ -0,0 +1,276 @@
...
+   *
+   *      program.parse(process.argv);
+   *
+   * @param {String} name
+   * @param {String} [desc]
+   * @param {Mixed} [opts]
+   * @return {Command} the new command
+   * @api public
+   */
+  command(name: string, desc?: string, opts?: any): $npm$commander$Command;
```
jeffmo: At minimum, `opts` should be an optional `Object` probably, right?

(Ideally there'd be a full-on type for it though... If you don't want to write one right now, mind adding a `TODO` for it?)

definitions/npm/commander_v2.x.x/flow_>=0.28.x/commander_v2.x.x.js
```diff
@@ -0,0 +1,276 @@
+declare class $npm$commander$Command extends events$EventEmitter {
```
jeffmo: Can we just move this inside the `declare module` body? Best to avoid globals unless they're strictly needed.

(It's always possible to move it *out* of the declare module body later if some other libdef needs to reference the type before we've gotten around to implementing imports for `declare module`s)

definitions/npm/commander_v2.x.x/flow_>=0.28.x/commander_v2.x.x.js
```diff
@@ -0,0 +1,276 @@
...
+   *        .description('display verbose help')
+   *        .action(function(){
+   *           // output help here
+   *        });
+   *
+   * @param {Function} fn
+   * @return {Command} for chaining
+   * @api public
+   */
+  action(fn: (...args: any[]) => void): $npm$commander$Command;
```
jeffmo: Probably best to specify the return type of the callback as `mixed`.
This lets people use general-purpose functions (that might happen to return something) as callbacks.

definitions/npm/commander_v2.x.x/flow_>=0.28.x/commander_v2.x.x.js
```diff
@@ -0,0 +1,276 @@
...
+   *     program.option('-c, --cheese [type]', 'add cheese [marble]');
+   *
+   * @param {String} flags
+   * @param {String} description
+   * @param {Function|Mixed} fn or default
+   * @param {Mixed} defaultValue
+   * @return {Command} for chaining
+   * @api public
+   */
+  option(flags: string, description?: string, fn?: ((arg1: any, arg2: any) => void) | RegExp, defaultValue?: any): $npm$commander$Command;
```
jeffmo: Let's [use `mixed` instead of `any`](https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md#avoid-any-when-possible) for the `defaultValue` parameter

definitions/npm/commander_v2.x.x/flow_>=0.28.x/commander_v2.x.x.js
```diff
@@ -0,0 +1,276 @@
...
+   */
+  arguments(desc: string): $npm$commander$Command;
+
+  /**
+   * Return an object containing options as key-value pairs
+   *
+   * @return {Object}
+   * @api public
+   */
+  opts(): any;
```
jeffmo: `Object` would be better

thejameskyle:
Updated the types, still need to add more tests

definitions/npm/commander_v2.x.x/flow_>=v0.28.x/commander_v2.x.x.js
```diff
@@ -0,0 +1,278 @@
...
+    flags: string;
+    required: boolean;
+    optional: boolean;
+    bool: boolean;
+    short?: string;
+    long: string;
+    description: string;
+  }
+
+  declare var exports: Command & {
```
jeffmo: `declare module.exports`

jeffmo: (...probably time to get that deprecation notice into the next release...)

jeffmo:
Running through PRs and following up: Looks like this one is just waiting on some `$ExpectError`s

thejameskyle:
Should be good to go

PR #175, opened 2016-07-06
Fix rxjs combineLatest

nmote:
Workaround for https://github.com/facebook/flow/issues/2041

### DIFF
--- a/definitions/npm/rxjs_v5.0.x/flow_>=v0.25.0/rxjs_v5.0.x.js
+++ b/definitions/npm/rxjs_v5.0.x/flow_>=v0.25.0/rxjs_v5.0.x.js
@@ -23,7 +23,14 @@ type TeardownLogic = rx$ISubscription | () => void;
 declare module 'rxjs' {
   declare class Observable<+T> {
     // This is actually variadic, but we only support one or two other observables.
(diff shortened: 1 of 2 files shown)


definitions/npm/rxjs_v5.0.x/flow_>=v0.25.0/rxjs_v5.0.x.js
```diff
@@ -23,7 +23,11 @@ type TeardownLogic = rx$ISubscription | () => void;
 declare module 'rxjs' {
   declare class Observable<+T> {
     // This is actually variadic, but we only support one or two other observables.
-    static combineLatest<T, U>(t: Observable<T>, u: Observable<U>): Observable<[T, U]>;
+    static combineLatest<T, U>(
+      t: Observable<T>,
+      u: Observable<U>,
+      _: void, // Prevents Flow from picking this case when it should pick the next one
```
jeffmo: Can you add a comment that this is specifically for `0.28` so that we don't lose that context in the future if this becomes unnecessary?

nmote: Will do.

nmote:
Updated

PR #183, opened 2016-07-08
Add initial definitions for chai

glenjamin:
There's probably still some work to do for these to be widely usable, but I figured I'd share what I have so far.

Some annoyances:

```js
expect(1).to.eql("1");
```
This should really be a type error, but for some reason flow infers the type to be `ExpectChain<number | string>` rather than sticking with what `expect()` says. Unsure how to fix this.

*****

The plugin system

Chai's plugin system allows plugins to add extra methods and properties to the ExpectChain, the only way I can think of to manage this in flow is to add them all to the core def. This PR currently includes some methods from a couple of plugins that I'm using in the project this is sourced from.

To complicate matters more, chai has a slightly awkward API where getters have side effects that won't be changed for back-compat reasons. [dirty-chai](https://www.npmjs.com/package/dirty-chai) is the recommended way to fix the old API, which is a plugin that changes the definition of core methods.

### DIFF
--- a/definitions/npm/chai_v3.5.x/flow_>=v0.24.0/chai_v3.5.x.js
+++ b/definitions/npm/chai_v3.5.x/flow_>=v0.24.0/chai_v3.5.x.js
@@ -1,0 +1,105 @@
+/* @flow */
+declare module "chai" {
+
+    declare type ExpectChain<T> = {
+        and: ExpectChain<T>,
+        at: ExpectChain<T>,
+        be: ExpectChain<T>,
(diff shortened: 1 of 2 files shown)


marudor:
@jeffmo whats your opinion on how to handle this?
My suggestion was to put the plugins into the core definition (just like @glenjamin did)
And have a extra definition for dirty-chai as it changes core methods.

Seemed like the best workaround for now.

marudor:
As We got no further input I think we can merge.
As we discussed before I favor this way (for now)

PR #185, opened 2016-07-08
add eslint to check for at least 1 negative test

marudor:
simple eslint setup + writte plugin that checks for negative tests.
Curently set to throw if we have no negative tests.
We could increase it to a specific amount of negative tests.


definitions/.eslintrc
```diff
@@ -0,0 +1,9 @@
+{
+  "parser": "babel-eslint",
+  "plugins": [
+    "flow-typed"
+  ],
+  "rules": {
+    "flow-typed/negative-tests": [2, 1]
```
thejameskyle: Where is this eslint rule defined?

marudor: https://github.com/marudor/eslint-plugin-flow-typed

ryyppy:
This is great, but doesn't the `cli/.flowconfig` also require a `suppress_comment` options for `ExpectError`?

marudor:
Nope,
this PR and rule is just to test the tests of the definitions.
All of them are in definitions/
Has nothing to do with the cli

marudor:
@jeffmo whats you're opinion? would merge if you're okay to get it running asap.
new Rules for stricter checking would be major semver updates. That way I don't accidently break flow-typed builds with changes to linting and tests.

jeffmo:
Seems good, but I'm a little surprised eslint can parse everything right now considering https://github.com/flowtype/flow-typed/issues/161... Any idea why that's actually working?

Also, would it make sense to put the eslint rule directly in this repo so that there's less work for Travis to do at CI time (i.e. download another dependency)? It's a small dependency it seems, but npm is pretty slow...so these all start to add up.

marudor:
I'm not parsing the flow types. Only tests.

Would keep the plugin seperate for now. It really is a small package - extra time is nearly 0 compared to actual testing.

jeffmo:
> I'm not parsing the flow types. Only tests.

Ahhh, that explains it!

PR #187, opened 2016-07-08
Init async 1.x definitions

thejameskyle:
I'm working on this right now and I'm opening up a PR because I have some questions about how to handle different things

### DIFF
--- a/definitions/npm/async_v1.x.x/flow_>=v0.28.x/async_v1.x.x.js
+++ b/definitions/npm/async_v1.x.x/flow_>=v0.28.x/async_v1.x.x.js
@@ -1,0 +1,77 @@
+// @flow
+
+declare module 'async' {
+  declare type Collection<TItem> = Iterable<TItem> | { [key: string]: TItem };
+
+  declare type IterateeCallback<TError, TResult> = (err?: TError, result?: TResult) => void;
+  declare type Callback<TError, TResult> = (err?: TError, result?: TResult) => mixed;
+
+  // Collections
+  declare function each<TItem, TError>(
+    collection: Collection<TItem>,
+    iteratee: (item: TItem, callback: IterateeCallback<TError, void>) => mixed,
+    callback?: Callback<TError, void>
+  ): void;
+  declare function forEachOf<TItem, TError>(
+    collection: Collection<TItem>,
+    iteratee: (value: TItem, key: string | number, callback: IterateeCallback<TError, void>) => mixed,
+    callback?: Callback<TError, void>
+  ): void;
+  declare function map<TItem, TResult, TError>(
+    collection: Collection<TItem>,
+    iteratee: (value: TItem, callback: IterateeCallback<TError, TResult>) => mixed,
+    callback?: Callback<TError, Array<TResult>>
+  ): void;
+  declare function filter(): void;
+  declare function reject(): void;
+  declare function reduce(): void;
+  declare function reduceRight(): void;
+  declare function transform(): void;
+  declare function detect(): void;
+  declare function sortBy(): void;
+  declare function some(): void;
+  declare function every(): void;
+  declare function concat(): void;
+  // Control Flow
+  declare function series(): void;
+  declare function parallel(): void;
+  declare function whilst(): void;
+  declare function doWhilst(): void;
+  declare function until(): void;
+  declare function doUntil(): void;
+  declare function during(): void;
+  declare function doDuring(): void;
+  declare function forever(): void;
+  declare function waterfall(): void;
+  declare function compose(): void;
+  declare function seq(): void;
+  declare function applyEach(): void;
+  declare function applyEachSeries(): void;
+  declare function queue(): void;
+  declare function priorityQueue(): void;
+  declare function cargo(): void;
+  declare function auto(): void;
+  declare function autoInject(): void;
+  declare function retry(): void;
+  declare function retryable(): void;
+  declare function iterator(): void;
+  declare function times(): void;
+  declare function timesSeries(): void;
+  declare function timesLimit(): void;
+  declare function race(): void;
+  // Utils
+  declare function apply(): void;
+  declare function nextTick(): void;
+  declare function memoize(): void;
+  declare function unmemoize(): void;
+  declare function ensureAsync(): void;
+  declare function constant(): void;
+  declare function asyncify(): void;
+  declare function wrapSync(): void;
+  declare function log(): void;
+  declare function dir(): void;
+  declare function noConflict(): void;
+  declare function timeout(): void;
+  declare function reflect(): void;
+  declare function reflectAll(): void;
+}


definitions/npm/async_v1.x.x/flow_>=v0.28.x/async_v1.x.x.js
```diff
@@ -0,0 +1,77 @@
...
+
+  // Collections
+  declare function each<TItem, TError>(
+    collection: Collection<TItem>,
+    iteratee: (item: TItem, callback: IterateeCallback<TError, void>) => mixed,
+    callback?: Callback<TError, void>
+  ): void;
+  declare function forEachOf<TItem, TError>(
+    collection: Collection<TItem>,
+    iteratee: (value: TItem, key: string | number, callback: IterateeCallback<TError, void>) => mixed,
```
thejameskyle: Here you can see for `key` I am using `string | number`, but I would prefer to use a polymorphic type that pulls from the abstract `Collection` type (which accepts anything that implements an iterator, as well as objects).

Here is a table of the mapping of different values of `collection` to `value` and `key`:


| collection | value | key |
|:--- |:--- |:--- |
| `Array [a]` | `a` | `0` |
| `Object { a: b }` | `b` | `a` |
| `Map { a => b }` | `[a, b]` | `0` |
| `Set { a }` | `a` | `0` |

Note that `Iterable<[TKey, TValue]>` is not what I am looking for... that doesn't work for sets and doesn't match the way maps work as noted in the table above.

definitions/npm/async_v1.x.x/flow_>=v0.28.x/async_v1.x.x.js
```diff
@@ -0,0 +1,77 @@
...
+    callback?: Callback<TError, void>
+  ): void;
+  declare function map<TItem, TResult, TError>(
+    collection: Collection<TItem>,
+    iteratee: (value: TItem, callback: IterateeCallback<TError, TResult>) => mixed,
+    callback?: Callback<TError, Array<TResult>>
```
thejameskyle: In async.js the `error` in the final callback behaves a bit weird.

If you pass a **truthy** value to the iteratee callback, it will be passed to the final callback.
If you pass a **falsy** value to the iteratee callback, `null` will be passed to the final callback.

This is really hard to write a type annotation for. I tried declaring two separate methods, but then flow can't decide which to use.

I've created a simplified version of this: https://gist.github.com/thejameskyle/17e80cc7f2a6ff45d351c00185dd2db4

jeffmo: > I tried declaring two separate methods, but then flow can't decide which to use.

cc @avikchaudhuri -- This seems like something we should try to improve with the latest intersections changes.

I think the right way to type this is with overloads like you tried, but Flow is a bit infantile with this right now...so it's not great for libdef consumers. For now I'd say just define the `error` param on the callback as `null | TError` and maybe write a TODO to revisit this once Flow has made some progress with deciding how to select overrides (hopefully in a near-future version?).

This will require that people always check for `null` in their callback, but at least that's still safe.

definitions/npm/async_v1.x.x/flow_>=v0.28.x/async_v1.x.x.js
jeffmo: Could you maybe overload this function using each of the two members of the union and writing out their specific `Iterable<T>` types?

(There's a good chance I've skimmed to aggressively here -- so feel free to tell me I'm stupid and point out the context I probably missed)

thejameskyle: It's almost not part of the Iterable type though. For non-objects, they key is an index that is incremented separately.

So if I write the `Collection` type like this:

```js
declare type Collection<TKey, TValue> = Iterable<TValue> | { [key: TKey]: TValue };
```

And the method like this:

```js
declare function forEachOf<TKey, TValue, TError>(
  collection: Collection<TKey, TValue>,
  iteratee: (value: TValue, key: TKey | number, ...) => mixed,
  ...
```

What I expect is that it will use the type of `TKey` if it is an object, and it will use `number` if it is an iterable.

However, [this is what happens when I use it that way](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDCcMMeAxhgJZwB2APACoDSeWANGIwGoCGMAVzwA+MAF4wASQx4ATrwBGZJn0EiwAHzABvMAG0A1mwBcXVlgC6pnvyFgAvgG502fFJnyZeQvxgLe5AZMAKKysnCynIwASngAzgIwGKISABRysgD81qHhkWCy8YkY2VyxCUkAlOKiAG5wVAAmzq4EPqT+gSFhEVHlxSlg6WGljLm9BUVJo-1VNWAAtlQAHnjN6FACtJQ0tGBQEcEBABYA8lBM5lGqQlHjssKpqGBg5CRkO3SmxKQU1HSXNjXWwidjPMBUDy8LymVK1EHWG54ThGLDWcyaMC0AQLBRyTjkXydAymaRyaF4bxEgJBMY9fL1JrCapiURLVaNMEvQkdGmldp+GndPKcRmNYSoSqmMU6ewbQ4nc6pXS8UwAcgAgmqHJw4SCUSYsTi8fkeYLAizRNp7JVnAdZEdyGcoMqwKqwJrtfZdfC1Aa0WA4hhZFRaABzAnUi3za221BAA).

thejameskyle: An alternative might be to write it like this:

```js
type $Iterable<TKey: number, TValue> = Iterable<TValue>;
type Collection<TKey, TValue> = $Iterable<TKey, TValue> | { [key: TKey]: TValue };
```

[But that doesn't work either](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgAkAkhngE4CGARjHgDwAqA0nlgFxgB2ArgLbUKAGjCMAapRg88APjABeMKQo06TCVNkBuTLgIBhODDoBjDAEs4XJqywjxk6XMUkyVWgxZt7Gp2AA+YADeYADaANZsnF5YALrRvgQAvjq6+EpulGR4+pIw1JQm4UwAouTkcOT2AEp4AM48MBjOYAAUFOQA-NFlFVVg5PWNGN2itQ1NAJQKcgBucOYAJjrY6bnGBUWl5ZU1Q00t7eWjjL27A-sj0ePD0-JyfOYAHnjL6FA8XGaWXGBQlSVCgALADyUBs3lEiXsZ3IMlaqDAYBMRlMFisnEMxjw3ysELsUMcsiEiLA5ky2U4rVmRISRJEkQ4olsIhMeU24U4yio2XW+UKxVOO368yWMjuD2erxJSLZGwFoz5HO2fREosWMlQk046uCSXeAOBYNaIUonAA5ABBc1gJIian0sCMzi8ATCZHsgUSvWTHT-ciAkygqAmsBmsBWm12to0zQMqJgOoYcjmLgAc1ZnqK3qCSV9qCAA).

jeffmo: What I meant by "overload this function using each of the two members of the union and writing out their specific" was something like

```js
declare function forEachOf<TValue, TError>(
  collection: Array<TValue>,
  iteratee: (value: TValue, key: number, ...) => mixed,
  ...
);

declare function forEachOf<TValue, TError>(
  collection: {[key: string]: TValue},
  iteratee: (value: TValue, key: string, ...) => mixed,
  ...
);

declare function forEachOf<TValue, TError>(
  collection: Iterator<TValue>,
  iteratee: (value: TValue, key: string | number, ...) => mixed,
  ...
);
```

I'm relatively sure Flow (v0.28) won't know how to choose these either, unfortunately (cc @avikchaudhuri here as well).

In the meantime, I think what you have here already is probably the best amount of precision you can get away with.

definitions/npm/async_v1.x.x/flow_>=v0.28.x/async_v1.x.x.js
jeffmo: Just checking in -- were you still planning on updating to `null | TError` here?

thejameskyle: I believe it's fixed locally, this definition is taking a really long time to get right and I haven't been able to focus on it.

marudor:
Whats the Status on this?

marudor:

Issue #16, opened 2016-03-16
Dependencies between library definitions?


samwgoldman:
I started writing this out in the facebook/flow repo, but I think it actually might belong here. At least I think the discussion is more appropriate here, and might lead to tasks for Flow.

In a world where flow-typed manages library definitions for things like React, how should people write definitions for their libraries that depend on React types? Currently we allow library definitions to export global types, but I'm only interested in modular proposals.

I think it's worth separating libraries, like React, from environments, like node or browsers. It's possible that the same thing might solve both cases, but maybe not.

Library definitions currently describe compatibility with version(s) of the library itself and version(s) of Flow, but what about dependencies to other library definitions? What about node libraries vs. dom libraries?

samwgoldman:
Ref facebook/flow#1543

marudor:
I think flow needs a setting for the env (node/browser)
Based on that flow itself will provide basic libdefs for the globals.
(Like Number, Promise, Math and so on)
For the node env also fs, path and everything that comes with node.

Everything else should be modular. So the current react defs in flow should be in here.
We would also need to be able to use import in defs. Currently it only works partially as the referenced Issue shows.
In my time writing libdefs I often noticed I need to import stuff. So far I've worked around it by declaring a global type prefixed with the library name just like the flow defs. But this will probably run into issues and collisions sooner or later.
It will probably make it easier and faster to update definitions as well. My PR about updated libdefs in the flow repo is still open (since mid December).
Hopefully by seperating libdefs and flow code it will be easier to keep updated.

jeffmo:
#### On libdef dependencies:

I hadn't really thought about this, but it makes sense. So to just start by circling the simplest solution: How about a `dependencies.json` metadata file?

**dependencies.json**
```json
{
  "react": ">=0.3"
}
```

#### On importing things within a `declare module`:

I've been wanting to tackle this for a little while now -- I've just been waiting on someone to really need it! What I had in mind was to take `declare module {}` and add support for toplevel `export` statements within it (i.e. `export function foo(bar: BarType): RetType;`). TypeScript came to the same conclusion already and I think it makes a lot of sense.

[...]

PR #189, opened 2016-07-12
Add shelljs (v0.7.0) declarations

freebroccolo:
This PR adds flow declarations for `shelljs`. It covers the complete API but in some cases I have restricted the interfaces for certain functions to the part that can be reasonably typed. However, this should not limit functionality.

I've added tests following some of the others as examples but I had some trouble getting the `flow-typed` CLI tool to run them.

### DIFF
--- a/definitions/npm/shelljs_v0.7.x/flow_>=v0.28.x/shelljs_v0.7.x.js
+++ b/definitions/npm/shelljs_v0.7.x/flow_>=v0.28.x/shelljs_v0.7.x.js
@@ -1,0 +1,213 @@
+declare class $npm$shelljs$Array<T> extends Array<T> mixins $npm$shelljs$Result, String {}
+declare type  $npm$shelljs$Async = Class<child_process$ChildProcess>;
+declare type  $npm$shelljs$Pattern = RegExp | String | string;
+declare class $npm$shelljs$String mixins $npm$shelljs$Result, String {
+  constructor(stdout: string | string[], stderr?: string, code?: number): this;
+}
+
+declare interface $npm$shelljs$Config {
+  fatal: boolean;
+  globOpts: {
+    nodir: boolean;
+  };
+  silent: boolean;
+  verbose: boolean;
+}
+declare interface $npm$shelljs$Env {
+  [key: string]: string;
+};
+
+declare type $npm$shelljs$OptionsPoly<Flags: string> = {
+  [keys: Flags]: boolean;
+};
+declare interface $npm$shelljs$ExecThen {
+  (code: number, stdout: string, stderr: string): void;
+};
+declare type $npm$shelljs$ExecOptionsPoly<T: Object> = T & { async?: boolean; silent?: boolean };
+declare type $npm$shelljs$ExecOpts = $npm$shelljs$ExecOptionsPoly<child_process$execOpts>;
+declare type $npm$shelljs$ExecOptsSync = $npm$shelljs$ExecOptionsPoly<child_process$execSyncOpts>;
+declare type $npm$shelljs$GrepOpts = $npm$shelljs$OptionsPoly<'-l' | '-v'>;
+declare type $npm$shelljs$SedOpts = $npm$shelljs$OptionsPoly<'-i'>;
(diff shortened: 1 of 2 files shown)


definitions/npm/shelljs_v0.7.x/flow_>=v0.28.x/shelljs_v0.7.x.js
```diff
@@ -0,0 +1,381 @@
...
+  & ((options: $npm$shelljs$OptSort, rest: void) => $npm$shelljs$String)
+  & ((rest: void) => $npm$shelljs$String)
+  ;
+
+declare type $npm$shelljs$PipeTail =
+  & ((num: number, rest: void) => $npm$shelljs$String)
+  & ((rest: void) => $npm$shelljs$String)
+  ;
+
+declare module 'shelljs' {
```
jeffmo: Since it looks like this module only defined named exports, I would recommend 2 tweaks here:

1. You can switch all the `declare`s inside your `declare module` body here to be `declare export` 
2. You can then move all of the global `$npm$shelljs$*` stuff inside the `declare module` body and, because you're not using `declare export`, they will be hidden from the exports of other modules.

There was a practice of using namespaced globals in other libdefs before because there wasn't a good option of hiding `declare`s inside a `declare module` before -- but that was fixed for ES module `declare module`s in v0.28!

jeffmo:
I love how complete and extensively tested this libdef is! Awesome :D

definitions/npm/shelljs_v0.7.x/flow_>=v0.28.x/shelljs_v0.7.x.js
freebroccolo: Interesting. I just experimented with `declare export` and I think I see what you mean. I should be able to clean some of this up that way.

Regarding namespaced globals, the main reason I have been using them in my libdefs so far is so that I can refer to types defined in libdefs for some module `m0` in libdefs for another module `m1`.

For instance, I have another small library built on `shelljs` that provides a promise based API for some parts like `exec`. I use flow for the code and babel + rollup to transform it to normal javascript on prepublish. Along with the package, I provide a libdef that refers to the types defined in this one.  In order to make the libdef for my library usable, it seemed like I needed to use namespaced globals for the `shelljs` libdef.

Is there another way to preserve this type information across module/package boundaries? I've been keeping an eye on https://github.com/flowtype/flow-typed/issues/16 since it seems relevant.

Just to be clear, here's a simplified example of the scenario I'm describing:

```javascript
// .flowconfig
[libs]
libdef-m0.js
libdef-m1.js

// libdef-m0.js
declare module "m0" {
  declare interface I0 {
    p0: T0;
  };
}

// libdef-m1.js
declare module "m1" {
  declare interface I1 {
    // How do we provide the type for p0? We can `import type { … }` in this
    // file (libdef-m1.js) but flow doesn't seem to preserve that information
    // when we use this as a lib from another file (see below).
    p0: m0.I0;
    p1: string;
  };
}

// index.js
// @flow
import type { I1 } from "m1";

declare var foo: I1;
// foo.p0 <- flow reports no type
// foo.p1 <- flow reports type string
```

jeffmo: > the main reason I have been using them in my libdefs so far is so that I can refer to types defined in libdefs for some module m0 in libdefs for another module m1.

Ah ok, yea that's definitely useful. If that's your purpose then fine by me until #16 is resolved

freebroccolo: Okay cool. I'll be happy to update this to remove the namespace stuff once there's a resolution.

Aside from that, there are a couple things I'm cleaning up a bit in this PR but once that's done it should be good to go from my side.

freebroccolo:
Okay, I cleaned up some of the exports and moved most of the stuff that's probably not needed in other libdefs into the module declaration. Think it's good to go on my end.

jeffmo:
woo!

PR #192, opened 2016-07-13
Add an editor config for easier colaboration

bassettsj:
- Add a simple editor editorconfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,5 +15,6 @@ charset = utf-8
 tab_size = 2
 tab_style = space
 
+[*.{js,json}]
 # trim any trailing whitespace
 trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*]
+tab_size = 2
+tab_style = space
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+
+# Set default charset
+charset = utf-8
+
+# 2 space indentation
+tab_size = 2
+tab_style = space
+
+# trim any trailing whitespace
+trim_trailing_whitespace = true
```
jeffmo: Is there anyway to whitelist specific files for this? For example, I think this has bitten me on `.md` files in the past?

.editorconfig
```diff
@@ -0,0 +1,19 @@
...
+# Unix-style newlines with a newline ending every file
+end_of_line = lf
+insert_final_newline = true
+
+# Set default charset
+charset = utf-8
+
+# 2 space indentation
+tab_size = 2
+tab_style = space
```
jeffmo: tabs_v_spaces

bassettsj: 😆  LOL that is correct for this project right?

.editorconfig
bassettsj: Yeah, we could just add that to js / json files let me change that.

bassettsj: Changed :-)

.editorconfig
jeffmo: Hehe, yea :p

PR #193, opened 2016-07-13
[libdef] platform v1.x

bassettsj:
- Add libdef for [platform](https://github.com/bestiejs/platform.js/blob/master/doc/README.md#readme)
- Add tests for it.

### DIFF
--- a/definitions/npm/platform_v1.x.x/flow_>=v0.21.x/platform_v1.x.x.js
+++ b/definitions/npm/platform_v1.x.x/flow_>=v0.21.x/platform_v1.x.x.js
@@ -3,7 +3,7 @@ declare module 'platform' {
     toString(): string;
     parse(userAgent: string): Platform;
     description: ?string;
-    layoyt: ?string;
+    layout: ?string;
     manufacturer: ?string;
     name: ?string;
     prerelease: ?string;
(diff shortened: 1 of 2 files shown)


definitions/npm/platform_v1.x.x/flow_>=v0.21.x/platform_v1.x.x.js
```diff
@@ -0,0 +1,22 @@
+declare module 'platform' {
+  declare class Platform extends Object {
```
bassettsj: It is supposed to be an object, but defining the class makes more sense for the parse method which returns a new instance.

definitions/npm/platform_v1.x.x/test_platform_v1.x.x.js
```diff
@@ -0,0 +1,49 @@
+// @flow
+import platform from 'platform';
+import type { $Platform } from 'platform';
```
marudor: Why the extra type import/export?
you can just use platform as type.

bassettsj: Yup, seems good. `typeof platform` works just fine.

marudor:
The travis error is rx for flow 0.29 - merging.

PR #203, opened 2016-07-19
Allow use of process.stdout as one of bunyan streams

dolphin278:
While `process.stdout` is not only a `stream$Writable`, but also a `tty$WriteStream` (https://github.com/facebook/flow/blob/master/lib/node.js#L1483), passing it as a `bunyan` stream causes errors.

Test was corrected to include originally failing case.


bassettsj:
Seems like a good addition @dolphin278 Thanks!

PR #204, opened 2016-07-20
Add node-hid definition

runn1ng:
Adding definitions for https://github.com/node-hid/node-hid


runn1ng:
It's a rather simple library, so there should be no issue.

PR #210, opened 2016-07-21
Add ecurve

runn1ng:
I am adding support for ecurve - a library for elliptic curves.

https://github.com/cryptocoinjs/ecurve

The bad thing is that it is using `bigi` module classes. I don't know how to solve this ( https://github.com/flowtype/flow-typed/issues/16 ).

So I just copy-pasted the definitions, so they are both in the same file. It's not a nice solution, but I don't know how to solve it correctly. Now it works.


marudor:
I think this is the best solution we have currently.

PR #211, opened 2016-07-21
[WIP] Add part of bitcoinjs-lib

runn1ng:
I have added SOME definitions for `bitcoinjs-lib`.

Bitcoin.js is a rather large library, I have made only a part of all the classes. The classes I have defined should be complete though.

Again this library depends on `bigi` and `ecurve`, and because I didn't know how to do that, I just copy-pasted the definitions. It seems to work, even when it's icky.


marudor:
I'm not sure how we want to handle unfinished.
Would like to keep this PR unmerged until the definitions  are finished.

runn1ng:
Now there shouldn't be anything missing, I went through all the exported modules.

jeffmo:
> I'm not sure how we want to handle unfinished.

I think the best way to judge this is to consider whether Flow will give errors for usage of any part of a libdef that is still TODO. We want to be able to trust that libdefs from flow-typed will not give false-negatives as much as possible, so we don't want libdefs that might tell us there are no errors when in fact we do have errors.

So for example: If there's an omission of a few complicated exports in the definition, that's ok (because trying to import something that's not declared is a clear error that the user will ultimately have to come declare it -- and hopefully contribute their addition).

But, as another example, we wouldn't want to take things like libdefs with methods that return `any` or `Object` simply because those return signatures are still a TODO.

PR #214, opened 2016-07-23
cache-manager

marudor:


definitions/npm/cache-manager_v2.x.x/flow_>=v0.28.x/cache-manager_v2.x.x.js
```diff
@@ -0,0 +1,24 @@
+declare module 'cache-manager' {
+  declare type _CacheOptions = {
+    ttl?: number,
```
thejameskyle: What about `max`, `store`, `isCacheableValue`?

Looking at the source there is also `ignoreCacheErrors` and `promiseDependency`

thejameskyle: Oh sorry this was meant for `caching(opts)`

marudor: Thanks, This was copied straight from out private work libdefs.
Wasn't surprised something was missing.

Added options & wrap function.

definitions/npm/cache-manager_v2.x.x/flow_>=v0.28.x/cache-manager_v2.x.x.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+  declare type CacheOptions = $Shape<_CacheOptions> | Object
```
jeffmo: `$Shape<_CacheOptions> | Object` is actually just as permissive/restrictive as just `Object`: "Either an object with this shape, or any object" is the same as "any object".

We don't really have a good way of saying what I think you're trying to say here.

marudor: Nope there is not.
All extra Settings get passed to the specific Store implementation. So we can't restrict it to _CacheOptions.

PR #227, opened 2016-08-03
backbone@v1.x.x Library Definition

bassettsj:
We are still using backbone unfortunately, so I have been struggling to get the typings correct for the backbone library and extended classes using the .extend function. I need the most guidance around dealing the the .extend functionality in each of the base classes.


definitions/npm/backbone_v1.3.x/flow_>=v0.25.x/backbone_v1.3.x.js.flow
```diff
@@ -0,0 +1,68 @@
+// @flow
+import typeof Underscore from 'underscore';
+
+declare module 'backbone' {
+   declare var $: any;
```
bassettsj: Since we still don't have a jQuery definition or a way to deal with dependencies between library definitions, I wasn't sure the best way to `Backone.$` or `Backbone._` What should I do now for these, leave `any`?

thejameskyle: Using `any` is likely the best way for now since the import you made above won't work.

If you leave a comment `// $FlowFixMe` above the `any` we will return to it.

definitions/npm/backbone_v1.3.x/flow_>=v0.25.x/backbone_v1.3.x.js.flow
```diff
@@ -0,0 +1,68 @@
...
+   declare var _: Underscore;
+   declare var version: string;
+   declare var emulateHTTP: boolean;
+   declare var emulateJSON: boolean;
+
+   declare type eventCallback = (event: Event) => void;
```
thejameskyle: Should use `mixed` instead of `void` here. Backbone doesn't care about it, but it will allow reused functions with returns to work with Flow.

definitions/npm/backbone_v1.x.x/flow_>=v0.25.x/backbone_v1.x.x.js
```diff
@@ -0,0 +1,80 @@
...
+    id: string | number;
+    attributes: Attrs;
+    cid: string;
+    cidPrefix: string;
+    chagned: ?Object,
+    validationError: ?Object;
+    initialize(): void;
+    toJSON(): Attrs;
+    sync: sync;
+    fetch(options: Object):  XMLHttpRequest;
```
thejameskyle: `fetch()` actually returns a jQuery XHR. Also you have it defined a second time returning `any`

thejameskyle:
This is looking pretty good minus the `fetch()` thing I mentioned. Would you mind just adding some more tests?

bassettsj:
@thejameskyle Sorry for the very long delay. I have added a few more tests and completed all the public APIs for the backbone classes. I think it should be okay for an initial draft and we can improve it from there. Let me know if I need to add more.

bassettsj:
@thejameskyle / @jeffmo 
Do you think this is good?

jeffmo:
Nice

bassettsj:
@jeffmo Thanks for the review and merge 👍

definitions/npm/backbone_v1.x.x/flow_v0.25.x-/backbone_v1.x.x.js
```diff
@@ -0,0 +1,248 @@
...
+    attributes: Attrs;
+    cid: string;
+    cidPrefix: string;
+    chagned: ?Object,
+    validationError: ?Object;
+    initialize(): void;
+    toJSON(): Attrs;
+    sync: sync;
+    set(attrs: Attrs, options?: Object): this;
+    set(attr: string, value: mixed, options?: Object): this;
```
philippotto: Could it be the `Model` is missing a method for `get`? Or am I missing something?

Issue #222, opened 2016-07-29
Moment definition for moment().set({...})


ggravarr:
I think the definition for set on the moment object should be something more like `set(options: { [key:string]:void|number }): this;` it's currently `set(options: { unit: string, value: number }): this;`. [This](http://momentjs.com/docs/#/get-set/set/) is the relevant moment documentation section.  The void union is there in my example because of some code formatting in my project I didn't want to change that uses object shorthand;  I suppose it doesn't really match the documented function.

ryyppy:
Would you mind sending a PR with the proposed changes?

thejameskyle:
Fixed by #234

PR #234, opened 2016-08-07
Fixes to Moment

danharper:
relevant documentation:

set: http://momentjs.com/docs/#/get-set/set/ (addresses #222)
local: http://momentjs.com/docs/#/manipulating/local/
utc: http://momentjs.com/docs/#/manipulating/utc/
utcOffset: http://momentjs.com/docs/#/manipulating/utc-offset/


marudor:
You are correct, thanks.

PR #237, opened 2016-08-08
[prismic.io] - Adds convenient type aliases for classes and fixes a few minor bugs.

ctrlplusb:
Having used the definition for a while now the following changes were helpful.


definitions/npm/prismic.io_v3.x.x/flow_>=v0.28.x/prismic.io_v3.x.x.js
```diff
@@ -1,5 +1,5 @@
 declare module 'prismic.io' {
-  declare type linkResolver = (...args: any) => any;
+  declare type LinkResolver = (...args: any) => any;
```
thejameskyle: @jeffmo Technically this is a breaking change because it changes the name of the type which could be imported. Do we want to allow these breaking changes? Do we want some way of versioning them?

ctrlplusb: James is right about this, and in almost every case I would agree, howeveri would be surprised if anyone but myself was using this definition file (I originally submitted it). Without the types that I expose in this PR the definition isn't very useful. There was also a bug in one of the major API endpoints of this definition. 
I have learnt that I need to give my defs more usage time before submitting.

thejameskyle: No worries, I'm not that concerned about this case. I'm just wondering about it in general

jeffmo: Probably ok here.

More generally: flow-typed libdefs are meant to be checked in, and upgrades are meant to be an actionable event by the developer (something like `flow-typed update`) and is never implicit. We should probably set expectations around updating libdefs that doing so may require some changes (i.e. a libdef can be improved to find new kinds of things).

jeffmo: Also: Should the resolver have a return type of `mixed`?

ctrlplusb: @jeffmo I've improved the `LinkResolver` definition by digging into the source and fishing out the actual implementation details. I've annotated the type with the expected inputs/outputs.

@thejameskyle good to go? :)

PR #238, opened 2016-08-08
Correcting bitcoinjs definitions

runn1ng:
I found some errors in the bitcoinjs-lib definitions. 

The inputs to `ECPair` is not buffer, but either biginteger or ecpoint, both imported from different packages.

### DIFF
--- a/definitions/npm/bitcoinjs-lib_v2.x.x/flow_>=v0.17.x/bitcoinjs-lib_v2.x.x.js
+++ b/definitions/npm/bitcoinjs-lib_v2.x.x/flow_>=v0.17.x/bitcoinjs-lib_v2.x.x.js
@@ -36,68 +36,70 @@ declare module 'bigi' {
   declare var exports: typeof $npm$bigi$BigInteger;
 }
 
-declare module 'ecurve' {
+declare class $npm$ecurve$Curve {
+  p: $npm$bigi$BigInteger;
+  a: $npm$bigi$BigInteger;
+  b: $npm$bigi$BigInteger;
+  G: $npm$ecurve$Point;
+  n: $npm$bigi$BigInteger;
+  h: $npm$bigi$BigInteger;
 
-  declare class Curve {
-    p: $npm$bigi$BigInteger;
-    a: $npm$bigi$BigInteger;
-    b: $npm$bigi$BigInteger;
-    G: Point;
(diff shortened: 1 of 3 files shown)


definitions/npm/bitcoinjs-lib_v2.x.x/flow_>=v0.17.x/bitcoinjs-lib_v2.x.x.js
```diff
@@ -97,6 +97,8 @@ declare module 'ecurve' {
 }
 // ---------- copypasta end ---- 
 
+import {Point as ECPoint} from 'ecurve';
```
thejameskyle: You cannot use imports within lib defs like this, which is why there is TODO at the top of this file.

runn1ng: Hmmm... I wrote the TODO :) since I wanted to explain that I copy-paste the definitions instead of importing them :)

Anyway, bitcoin.js ECPair constructor requires Point from ecurve as a parameter (see other changes in the commit; the current version in the master is wrong - I made a mistake when writing it). How to do that, if not like that? I cannot write `Q: ?ecurve.Point` or anything similar. `ecurve` definition is copy-pasted as a whole in the start of the definition.

Also, I think the "test" runs successfully, so it seems to me that it typechecks like it should...

runn1ng: I have added more tests for the ECPair constructor, to test if it really checks the parameters with the `ecurve.Point` type, let's see what happens....

runn1ng: Hm. Seems you are right and it's not possible to do this, since it doesn't typecheck correctly.

So I will have to move `ecurve.Point` outside of `ecurve` and refactor it out into `$npm$ecurve$Point` or something similar. Well, I am not looking forward to that. :)

runn1ng:
I refactored the ecurve definition (both in the copy-pasted one and in the original definition) to allow the constructor to use it directly, now it seems to work and the test works also.

PR #241, opened 2016-08-09
Slightly improve types of map-obj

nmn:
Not sure how much it'll actually help in practice, but I read an article by @thejameskyle and this kept bugging me in my head.


thejameskyle:
Ha! This is the right level of obsessiveness.

marudor:
Looks like Travis fails due to some NPM Problem (ECONNRESET)
Gonna restart Tests once thats resolved. If they pass its a merge.

PR #246, opened 2016-08-10
initial file-type

marudor:
https://www.npmjs.com/package/file-type

### DIFF
--- a/definitions/npm/file-type_v3.x.x/flow_>=v0.28.x/file-type_v3.x.x.js
+++ b/definitions/npm/file-type_v3.x.x/flow_>=v0.28.x/file-type_v3.x.x.js
@@ -1,0 +1,3 @@
+declare module 'file-type' {
+  declare module.exports: (buffer: Buffer | Uint8Array) => ?{ ext: string, mime: string };
+}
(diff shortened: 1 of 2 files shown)


definitions/npm/file-type_v3.x.x/flow_>=v0.28.x/file-type_v3.x.x.js
```diff
@@ -0,0 +1,3 @@
+declare module 'file-type' {
+  declare module.exports: (buffer: Buffer | Uint8Array) => ?{ ext: string, mime: string };
```
thejameskyle: Should this be `{ ext: string, mime: string } | null` instead?

marudor: What exactly is the difference?
Isn't ?Type exaclty that? A nullable type.

thejameskyle: Well technically there's a difference because `?` says it can be `null` or `undefined`.

That doesn't have a real impact so just ignore me

ryyppy: Wait, ... `fn(): ?Object` means it may return `null|undefined`? Also thought this would be null only... ¯\_(ツ)_/¯

PR #248, opened 2016-08-11
Add multi-typeof

dustinspecker:
I think I followed the contributing guidelines correctly. (Thanks for the thorough walkthrough!)

Please let me know if I can do anything else.

### DIFF
--- a/definitions/npm/multi-typeof_v1.x.x/flow_>=v0.25.x/multi-typeof_v1.x.x.js
+++ b/definitions/npm/multi-typeof_v1.x.x/flow_>=v0.25.x/multi-typeof_v1.x.x.js
@@ -1,0 +1,6 @@
+declare module 'multi-typeof' {
+  declare module.exports: (
+    input: mixed,
+    types: string|Array<string>
+  ) => boolean;
+}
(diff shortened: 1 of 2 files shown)


thejameskyle:
@dustinspecker awesome thanks

@panagosg7 question... can/how do predicate types handle this?

panagosg7:
@thejameskyle: I'm afraid Flow predicates do not support examples like `multi-typeof`. There's a couple of reasons for this:

* The function implies a *relational* constraint between the two inputs of `multi_typeof`. To make   this more clear, suppose the function's signature currently looks like this:
  ```javascript
  declare function multi_typeof(input: mixed,  types: string[]): boolean;
  ```
  This function establishes a predicate, say `typeof_belongs_to`, that expects *two* arguments (`input` and `types`). Flow currently only supports single argument predicates (`typeof` actually desugars to a single argument predicate as well).

* The use of an array as the second argument implies the notion of a set. Reasoning about sets (in full generality) is hard.

This is probably a bit off-topic here, but the support we hope to deliver soon is something like the following:
```javascript
declare function multi_typeof_pred<P:$Pred>(v: mixed, f: P): boolean checks(P);
```
Read this as: `multi_typeof_pred` takes an input `v` and a function `f` that will be treated as a condition check (this is denoted by the constraint `: $Pred`), and returns `true` if `v` fulfills `f`'s condition and `false` otherwise.

Now `multi_typeof_pred` can accept *arbitrary* predicate functions that express the desired check. For example, one such case that corresponds to `['string', 'number']` from the `multi-typeof` example before is the  predicate function:
```javascript
declare function str_or_num(x): boolean checks(typeof x === 'string' || typeof x === 'number');
```
This last signature is currently supported by Flow, i.e. one can use `str_or_num(x)` in a conditional to narrow down a type to the union `string | number`, as if the relevant check was inlined.

PR #249, opened 2016-08-12
Added travis ci support for node LTS versions

amilajack:


jeffmo:
@amilajack: It seems that Flow is hanging forever for some reason when checking the CLI code against node@4...

I'm not sure why, but I just updated the CLI to use the latest version of Flow (which I believe includes a few various fixes for hanging issues) -- would you mind rebasing and see if that makes the Travis tests happy?

marudor:
Considering this is open since August and Node 6 will be LTS this month I'm gonna close this.

amilajack:
@jeffmo Maybe we should try rerunning the tests. I dont have permission to do this so maybe you could look into this?

PR #258, opened 2016-08-17
Improve type information for react-dnd

cdchambers:
Updated the react-dnd definitions based on existing TypeScript
definitions from DefinitelyTyped. Type definitions are now complete
based on react-dnd documents.

Added unit tests exercising this functionality.

Also fixed incorrect signatures for DragLayer.

### DIFF
--- a/definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
+++ b/definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
@@ -1,6 +1,202 @@
+// Shared
+// ----------------------------------------------------------------------
+// todo: add symbol once flow supports it
+type Identifier = string;
+
+type ClientOffset = {
+  x: number,
+  y: number
+}
+
+type DndOptions<P> = {
+  arePropsEqual?: (props: P, otherProps: P) => boolean
+};
+
+// Decorated Components
+// ----------------------------------------------------------------------
+declare class DndComponent<P> extends React$Component<any, P, any> {
+  static DecoratedComponent: Class<this>;
+  getDecoratedComponentInstance(): this;
+  getHandlerId(): Identifier;
+}
+
+declare class ContextComponent<P> extends React$Component<any, P, any> {
+  static DecoratedComponent: Class<this>;
+  getDecoratedComponentInstance(): this;
+  // getManager is not yet documented in ReactDnd Docs
+  getManager(): any;
+}
+
+// Drag Source
+// ----------------------------------------------------------------------
+type DragSourceType<P> =
+  Identifier |
+  (props: P) => Identifier;
+
+type DragSourceSpec<P> = {
+  beginDrag: (
+    props: P,
+    monitor?: DragSourceMonitor,
+    component?: React$Component<any, P, any>
+  ) => Object,
+
+  endDrag?: (
+    props: P,
+    monitor?: DragSourceMonitor,
+    component?: ?React$Component<any, P, any>
+  ) => void,
+
+  canDrag?: (
+    props: P,
+    monitor?: DragSourceMonitor
+  ) => boolean,
+
+  isDragging?: (
+    props: P,
+    monitor?: DragSourceMonitor
+  ) => boolean
+};
+
+type DragSourceMonitor = {
+  canDrag: () => boolean,
+  isDragging: () => boolean,
+  getItemType: () => Identifier,
+  getItem: () => Object,
+  getDropResult: () => Object,
+  didDrop: () => boolean,
+  getInitialClientOffset: () => ClientOffset,
+  getInitialSourceClientOffset: () => ClientOffset,
+  getClientOffset: () => ClientOffset,
+  getDifferenceFromInitialOffset: () => ClientOffset,
+  getSourceClientOffset: () => ClientOffset
+}
+
+type DragSourceConnector = {
+  dragSource: () => ConnectDragSource,
+  dragPreview: () => ConnectDragPreview
+}
+
+type DragSourceOptions = {
+  dropEffect?: string
+}
+
+type DragPreviewOptions = {
+  captureDraggingState?: boolean,
+  anchorX?: number,
+  anchorY?: number
+}
+
+type ConnectDragSource = (
+  elementOrNode: React$Element<any>
+  options?: DragSourceOptions
+) => React$Element<any>;
+
+type ConnectDragPreview = (
+  elementOrNode: React$Element<any>
+  options?: DragPreviewOptions
+) => React$Element<any>;
+
+type DragSourceCollector = (
+  connect: DragSourceConnector,
+  monitor: DragSourceMonitor
+) => Object;
+
+type DragSource = <P, C: React$Component<any, P, any>>(
+  type: DragSourceType<P>,
+  spec: DragSourceSpec<P>,
+  collect: DragSourceCollector,
+  options?: DndOptions<P>
+) => (component: Class<C>) => Class<DndComponent<P>>;
+
+// Drop Target
+// ----------------------------------------------------------------------
+type DropTargetTypes<P> =
+  Identifier |
+  Array<Identifier> |
+  (props: P) => Identifier | Array<Identifier>;
+
+type DropTargetSpec<P> = {
+  drop?: (
+    props: P,
+    monitor?: DropTargetMonitor,
+    component?: React$Component<any, P, any>
+  ) => ?Object,
+
+  hover?: (
+    props: P,
+    monitor?: DropTargetMonitor,
+    component?: React$Component<any, P, any>
+  ) => void,
+
+  canDrop?: (
+    props: P,
+    monitor?: DropTargetMonitor
+  ) => boolean
+};
+
+type DropTargetMonitor = {
+  canDrop: () => boolean,
+  isOver: (options?: { shallow: boolean }) => boolean,
+  getItemType: () => Identifier,
+  getItem: () => Object,
+  getDropResult: () => Object,
+  didDrop: () => boolean,
+  getInitialClientOffset: () => ClientOffset,
+  getInitialSourceClientOffset: () => ClientOffset,
+  getClientOffset: () => ClientOffset,
+  getDifferenceFromInitialOffset: () => ClientOffset,
+  getSourceClientOffset: () => ClientOffset
+}
+
+type DropTargetConnector = {
+  dropTarget: () => ConnectDropTarget
+}
+
+type ConnectDropTarget = (
+  elementOrNode: React$Element<any>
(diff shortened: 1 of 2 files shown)


thejameskyle:
cc @gaearon

gaearon:
Nice!

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
```diff
@@ -1,6 +1,202 @@
...
+}
+
+type DragPreviewOptions = {
+  captureDraggingState?: boolean,
+  anchorX?: number,
+  anchorY?: number
+}
+
+type ConnectDragSource = (
+  elementOrNode: React$Element<any>
```
gaearon: This could be a DOM node too.

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
```diff
@@ -1,6 +1,202 @@
...
+  options?: DragSourceOptions
+) => React$Element<any>;
+
+type ConnectDragPreview = (
+  elementOrNode: React$Element<any>
```
gaearon: And this could be a DOM node too.

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
```diff
@@ -1,6 +1,202 @@
...
+  getDifferenceFromInitialOffset: () => ClientOffset,
+  getSourceClientOffset: () => ClientOffset
+}
+
+type DropTargetConnector = {
+  dropTarget: () => ConnectDropTarget
+}
+
+type ConnectDropTarget = (
+  elementOrNode: React$Element<any>
```
gaearon: This could be a DOM node too.

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
```diff
@@ -1,6 +1,202 @@
...
+  connect: DragSourceConnector,
+  monitor: DragSourceMonitor
+) => Object;
+
+type DragSource = <P, C: React$Component<any, P, any>>(
+  type: DragSourceType<P>,
+  spec: DragSourceSpec<P>,
+  collect: DragSourceCollector,
+  options?: DndOptions<P>
+) => (component: Class<C>) => Class<DndComponent<P>>;
```
gaearon: I'm not sure what `Class` means in Flow but any of these could be a functional React component as well.

definitions/npm/react-dnd_v2.x.x/test_react-dnd-v2.js
```diff
@@ -7,7 +7,7 @@ import { DragSource, DropTarget, DragLayer, DragDropContext } from 'react-dnd';
 // ----------------------------------------------------------------------
 type KnightProps = {
   connectDragSource: (e: React$Element<*>) => React$Element<*>,
-  connectDragPreview: (e: React$Element<*>) => React$Element<*>,
+  connectDragPreview: (e: Image) => Image,
```
gaearon: I think this can be element, image, or node, not just an image.

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
cdchambers: Fixed. Made it `React$Element<any> | HTMLElement`

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
cdchambers: Fixed. Made it `React$Element<any> | HTMLElement`

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
cdchambers: Fixed. Made it `React$Element<any> | HTMLElement`

definitions/npm/react-dnd_v2.x.x/test_react-dnd-v2.js
gaearon: Ah never mind, I thought this was lib typings.

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
```diff
@@ -12,6 +12,8 @@ type DndOptions<P> = {
   arePropsEqual?: (props: P, otherProps: P) => boolean
 };
 
+type ElementOrNode = React$Element<any> | HTMLElement;
```
gaearon: I think `?HTMLElement` is more correct. When passing element, we should be able to `null` it out later.

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
```diff
@@ -87,14 +89,14 @@ type DragPreviewOptions = {
 }
 
 type ConnectDragSource = (
-  elementOrNode: React$Element<any>
+  elementOrNode: ElementOrNode
   options?: DragSourceOptions
-) => React$Element<any>;
+) => ElementOrNode;
```
gaearon: Is there any way to express that if you pass element, you get an element, but if you pass a node, you get a node?

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
cdchambers: I'm not completely sure about this one either. I added a test in which DragSource is applied to a functional React components and things seem to still type check.

definitions/npm/react-dnd_v2.x.x/flow_>=v0.23.x/react-dnd_v2.x.x.js
cdchambers: Yeah, something like this? We can `null` out both HTMLElement and ReactElement this way. Thoughts?
```javascript
type ElementOrNode = React$Element<any> | HTMLElement;
```
```javascript
type ConnectDragSource = <T : ElementOrNode>(
  elementOrNode: T
  options?: DragSourceOptions
) => ?T;
```

gaearon: Looks reasonable

cdchambers:
@thejameskyle Two questions:
* What's the right way to define a higher-order component in flow? Is the following the right way to type a higher-order component?

The DragSource is a higher-order component that takes three arguments and it transforms a component C into a draggable componet (typed as `DndComponent`).

```javascript
type DragSource = <P, C: React$Component<any, P, any>>(
  type: DragSourceType<P>,
  spec: DragSourceSpec<P>,
  collect: DragSourceCollector,
  options?: DndOptions<P>
) => (component: Class<C>) => Class<DndComponent<P>>;
```
```javascript
declare class DndComponent<P> extends React$Component<any, P, any> {
  static DecoratedComponent: Class<this>;
  getDecoratedComponentInstance(): this;
  getHandlerId(): Identifier;
}
```
* How to test the types of the two instance methods on `DndComponent`? As an example, `dk` below is an instance of the `DndKnight` component. Flow doesn't like `dk.getHandlerId()` because the type of `dk` is `React$Element<{}>`. 

```javascript
const DndKnight = DragSource('knight', knightSource, knightCollect)(Knight);
const dk = <DndKnight />;
```

cdchambers:
@thejameskyle @gaearon 

I've been trying to get the `DecoratedComponent` (named `DndComponent` below) in react-dnd to type check. React-dnd extends React's Component class with methods such as `getDecoratedComponentInstance` and `getHandlerId`. These methods are then used on a ReactElement. See [here](https://github.com/gaearon/react-dnd/blob/8d1260e42d48def5b9d30f9b0881589082aeabd8/examples/01%20Dustbin/Single%20Target/__tests__/Box-test.js) for an example.

I started with something like this:
```javascript
declare class DndComponent<D, P, S> extends React$Component<D, P, S> {
  static DecoratedComponent: Class<this>;
  getDecoratedComponentInstance(): this;
  getHandlerId(): Identifier;
}
```
But this is not getting me anywhere, because as far as I can tell, the React and Flow definitions do not allow changing the result type of `React.createElement` beyond passing in a `Config`, which only affects the props.

I think to support this, it would be necessary to add a parameter to ReactComponent for the "extension" methods. It may also fit on the `Config`. Something like:

```javascript
// Extensions to React$Element enabled by react-dnd:
type DndExt<C> = {
 getDecoratedComponentInstance : () => C
 getHandlerId: () => Identifier 
};

// Additional parameter to React$Component specifies these extensions:
declare class DndComponent<D, P, S> extends React$Component<D, P, S, DndExt<this>> { ... }

// Type of React.createElement changes
// from React.createElement : ReactClass<C> -> ReactElement<C>
// to React.createElement : ReactClass<C, X> -> X<ReactElement<C>>
// Where X is a higher-order type that adds additional methods to the ReactElement.
```

Since the DecoratedComponents are only used in tests, I've dropped them from the type definitions for now.

gaearon:
>React-dnd extends React's Component class with methods such as getDecoratedComponentInstance and getHandlerId. These methods are then used on a ReactElement. 

Hmm, I don’t think this is correct. These methods are defined on the component *instance*, not *element*. So it shouldn’t actually involve `ReactElement`.

gaearon:
In other words, in your example, those methods should be on `class DndComponent`, not the element it uses.

cdchambers:
Gotcha @gaearon. Fixed in the last check-in. Added (back) the DndComponent and ContextComponent classes and parameterized them over types of default props, props and state. DndComponent for DragSource, DropTarget and DropLayer. ContextComponent for ContextComponent.

cdchambers:
@thejameskyle @gaearon does the PR look good to merge?

gaearon:
Looks good to me, and we can always improve this in the future!

gaearon:
Great work by the way, thank you so much for taking the time to do that (even for undocumented methods). 💯

cdchambers:
Ping @thejameskyle  +  @marudor, the original author of the react-dnd flow definition. 
Is this PR ready to go in?

PR #260, opened 2016-08-17
rxjs: add some combineLatest and withLatestFrom definitions

justinwoo:
I removed code marked as not necessary for >= 0.29 and added definitions that I needed in my own projects.

I'm guessing there's not another way to do this, but I'd be happy to found out if there were.

### DIFF
--- a/definitions/npm/rxjs_v5.0.x/flow_>=v0.29.0/rxjs_v5.0.x.js
+++ b/definitions/npm/rxjs_v5.0.x/flow_>=v0.29.0/rxjs_v5.0.x.js
@@ -22,33 +22,6 @@ type TeardownLogic = rx$ISubscription | () => void;
 
 declare module 'rxjs' {
   declare class Observable<+T> {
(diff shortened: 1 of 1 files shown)


definitions/npm/rxjs_v5.0.x/flow_>=v0.29.0/rxjs_v5.0.x.js
```diff
@@ -260,6 +207,238 @@ declare module 'rxjs' {
...
+      resultSelector: (a: A, b: B, c: C, d: D, e: E, f: F, g: G) => H,
+    ): Observable<H>;
+
+    withLatestFrom<A>(
+      a: Observable<A>
+    ): Observable<[T, A]>;
+
+    withLatestFrom<A, B>(
+      a: Observable<A>,
+      resultSelector: (a: A) => B,
```
jbaudanza: Is there some reason that `t: T` was removed from the resultSelector? I needed to add it back in order to get my flow checks to pass.

PR #263, opened 2016-08-18
Let tape$TestCb return a promise

cdlewis:
Let tape$TestCb return a promise
    
If you want to use async/await with Tape you will need to prefix your test function with async, e.g.

```
test('Promise resolves true', async (assert: tape$Context) => {
  const result = await promiseFunction();
  assert.true(result);
  assert.end();
}
```
    
But this will cause type checking to fail as the callback no longer returns void.

### DIFF
--- a/definitions/npm/tape_v4.5.x/flow_>=v0.25.x/tape_v4.5.x.js
+++ b/definitions/npm/tape_v4.5.x/flow_>=v0.25.x/tape_v4.5.x.js
@@ -11,7 +11,7 @@ declare type tape$TestOpts = {
 };
 
 
-declare type tape$TestCb = (t: tape$Context) => void;
+declare type tape$TestCb = (t: tape$Context) => void | Promise<any>;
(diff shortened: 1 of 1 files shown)


definitions/npm/tape_v4.5.x/flow_>=v0.25.x/tape_v4.5.x.js
```diff
@@ -11,7 +11,7 @@ declare type tape$TestOpts = {
 };
 
 
-declare type tape$TestCb = (t: tape$Context) => void;
+declare type tape$TestCb = (t: tape$Context) => void | Promise<mixed>;
```
thejameskyle: It'd be better just to write `mixed` instead and just accept anything since it doesn't make a difference to tape (AFAIK)

ryyppy:
Ah well, I just started using async functions, in that time I forgot to include them in the definitions...

Does our test setup support async / await syntax?
Would like to see that test case in the tape test file

jeffmo:
> Does our test setup support async / await syntax?
Would like to see that test case in the tape test file

Yes, anything that Flow can read is valid flow-typed test code

PR #265, opened 2016-08-19
Jest updates

elliottsj:
Three distinct changes in this PR; see each commit message for details.

Do these changes make sense? Can split into multiple PRs if necessary. Any additional tests I should add?

Also wondering if it makes sense to remove the other Jasmine methods from the Jest type definition:
* `describe`, `it`, etc.
* `expect`, et al

I think Jest users should be installing both the Jest and Jasmine type definitions; a couple reasons described in https://github.com/flowtype/flow-typed/pull/265/commits/21334b864b475ab5e0574fd243e09c229fff5e13


thejameskyle:
cc @cpojer

cpojer:
Can you add all jasmine definitions to Jest? I commented on the issue you created.

elliottsj:
Can do! Will update later today

elliottsj:
@cpojer Updated

definitions/npm/jest_v14.x.x/flow_>=v0.23.x/jest_v14.x.x.js
```diff
@@ -65,8 +97,16 @@ declare var jest: {
   setMock(moduleName: string, moduleExports: any): void;
   unmock(moduleName: string): void;
   useFakeTimers(): void;
+  useRealTimers(): void;
 }
 
 declare var jasmine: {
   DEFAULT_TIMEOUT_INTERVAL: number;
+  any(val: mixed): void;
```
cpojer: is the return value of jasmine.any really void?

cpojer:
yay this is way better! :)

definitions/npm/jest_v14.x.x/flow_>=v0.23.x/jest_v14.x.x.js
elliottsj: Nope; looks like it actually returns an "Any" object with properties `asymmetricMatch` and `jasmineToString`:

https://github.com/jasmine/jasmine/blob/f196faf0bbec55b2b912644f8b513214d6000980/src/core/asymmetric_equality/Any.js

I copied the definition over from jasmine_v2.4.x.js; do you think we should a definition for [Jasmine's "asymmetricMatch" objects](http://jasmine.github.io/2.4/introduction.html#section-Custom_asymmetric_equality_tester)?

cpojer: yeah that would be nice :)

elliottsj: Cool, adding now 😄

elliottsj: I've added a `JestAsymmetricEqualityType` type, omitting the `jasmineToString` method, since it's not documented on http://jasmine.github.io/2.4/introduction.html

PR #271, opened 2016-08-21
include html in Express render callback

turadg:
http://expressjs.com/en/api.html#res.render

### DIFF
--- a/definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
@@ -91,12 +91,12 @@ declare module 'express' {
   declare interface RouteMethodType<T> {
     (middleware: Middleware): T;
     (...middleware: Array<Middleware>): T;
-    (path: string|RegExp, ...middleware: Array<Middleware>): T;
+    (path: string|RegExp|string[], ...middleware: Array<Middleware>): T;
   }
   declare interface RouterMethodType<T> {
     (middleware: Middleware): T;
     (...middleware: Array<Middleware>): T;
-    (path: string|RegExp, ...middleware: Array<Middleware>): T;
+    (path: string|RegExp|string[], ...middleware: Array<Middleware>): T;
     (path: string, router: Router): T;
   }
   declare class Route {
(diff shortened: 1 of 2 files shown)


turadg:
`./travish.sh` works for me locally.

@marudor looks like something's up with the Travis config https://travis-ci.org/flowtype/flow-typed/builds/154836301

definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
```diff
@@ -73,7 +73,7 @@ declare module 'express' {
     location(path: string): this;
     redirect(url: string, ...args: Array<void>): this;
     redirect(status: number, url: string, ...args: Array<void>): this;
-    render(view: string, locals?: mixed, callback: (err?: ?Error) => mixed): this;
+    render(view: string, locals?: mixed, callback: (err?: ?Error, html: string) => mixed): this;
```
gabelevi: Nit: This type is a little odd. Flow probably should be changed to ban non-optional parameters from following optional parameters.

If the callback method is always passed two arguments

1. `err`, which is either an instance of `Error` or `null`
2. `html`, which is a string

then the most correct type would be `(err: Error | null, html: string) => mixed)`.

However, as written, this is basically `(err: Error | null | void, html: string) => mixed)`, which is only a little bit more strict than it needs to be.

So I'm happy to merge without this being fixed. Just let me know what you want to do!

gabelevi:
Not sure what's going on with travis, but this looks good to me! I have one nit, which I mentioned inline, but happy to merge even without it being fixed.

definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
turadg: Good catch. Looks like `html` is omitted if there's an error: https://github.com/expressjs/express/blob/master/lib/application.js#L581

And it's always called with at least the first arg. So I'll change the signature.

The Response render just calls to app render so I'll share the type too.

gabelevi:
Good catch on the callback being called sometimes with only one argument!

I think you might need to update the test too, though

gabelevi:
Looks good! Thanks for bearing with me and thanks for your contribution!

turadg:
btw, what's your current philosophy on aligning with definitelytyped? Seems a waste to write all these types twice.

I've heard Facebook wants a flow2typescript rather than typescript2flow because Flow can express more. Seems like it would be easier to make a subset work in a superset than reduce that extra expressivity, but I don't know much of the details here.

Before either one, could the two repos be made to support the same test files? That's a lot of the work right there, figuring out what the spec is for these JS libs.

PR #272, opened 2016-08-21
Debug v2.x.x: relax first param of Debugger declaration

gpittarelli:
As currently declared, Debugger functions can only accept a string as their first argument.
However, Debugger functions are specifically implemented to also accept an Error as their first argument (instead of a formatter)--or no formatter/error at all.

See:
https://github.com/visionmedia/debug/blob/master/debug.js#L88-L93
and 'coerce' at: https://github.com/visionmedia/debug/blob/master/debug.js#L194-L197

Note that these changes allow a debugger function to be called with any pattern of arguments--there may be value in restricting that, but I couldn't find any definitive guidance on the intended API of those functions in the project's docs/code.

### DIFF
--- a/definitions/npm/debug_v2.x.x/flow_>=v0.28.x/debug_v2.x.x.js
+++ b/definitions/npm/debug_v2.x.x/flow_>=v0.28.x/debug_v2.x.x.js
@@ -1,6 +1,7 @@
 declare module 'debug' {
   declare type Debugger = {
     (formatter: string, ...args: Array<any>): void,
+    (err: Error, ...args: Array<any>): void,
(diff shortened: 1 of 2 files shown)


definitions/npm/debug_v2.x.x/flow_>=v0.28.x/debug_v2.x.x.js
```diff
@@ -1,6 +1,8 @@
 declare module 'debug' {
   declare type Debugger = {
+    (...args: Array<any>): void,
```
marudor: Please used mixed at least.

PR #276, opened 2016-08-23
add fail(reason: string) to jasmine v2.4.x definitions

quentin-:


gabelevi:
Looks like you can call `fail()` with 0 arguments. Mind updating the PR to make the argument optional?

At least I see a test with `fail()` in it here: https://github.com/jasmine/jasmine/blob/b1344d5c73c5e01a07e1ea435be3ed980f6db9de/spec/core/integration/EnvSpec.js#L135

quentin-:
Oops, right. It also can take an Error.

https://github.com/jasmine/jasmine/blob/master/spec/core/integration/EnvSpec.js#L198
https://github.com/jasmine/jasmine/blob/master/src/core/Env.js#L494-L509

gabelevi:
Ooh, you're right! Nice catch.

Thanks a ton for the contribution and for the quick update!

PR #287, opened 2016-08-30
Fix problems with redux interop

ryyppy:
There were some minor problems with redux libdefs e.g.
applyMiddleware. Also fixed the @@iterator problem with `takeLatest` /
`takeEvery`.

Credit goes to @aaronjensen


ryyppy:
@aaronjensen I slightly adapted your solution for fixing  interoperability with redux' libdefs ... does this work for you?

aaronjensen:
@ryyppy yep, this still checks on my project, thanks!

Out of curiosity, why define the generator manually rather than using `Generator`?

ryyppy:
@aaronjensen Because it's in fact not a generator! it also has attributes which are non standardized like `name` : https://github.com/yelouafi/redux-saga/blob/master/src/internal/sagaHelpers.js#L28-L31

ryyppy:
Maybe I am completely wrong, but the lack of ES7 features obfuscates the real interface :-/

PR #288, opened 2016-09-01
React test utils return Elements from get-by-class and get-by-tag and can return null

DanielHeath:
Per PR title.

Looking up rendered nodes by tag name or className always return `Element`, not `React$Component`.


DanielHeath:
Also in this change: `findRenderedDOMComponentWithTag` and similar can return null.

jeffmo:
41f45a7d8cfa462b87418b2e7d42f14942f476ac

Issue #252, opened 2016-08-16
[CLI] `flow-typed install`: When no flags are present, should install from package.json


jeffmo:
Currently it is necessary to specify the name of a library and a flow version when calling `flow-typed install`.

Before we can launch out of beta, we want the flag-less version of the command to do most of the work for you: It should determine your dependencies from the `package.json` file, scan the flow-typed repo for the expected libdef versions (when present), and install libdefs that match.

PR #289, opened 2016-09-01
Install libdefs from package.json

palange:
Find and install libdefs from package.json
After installing flow-typed, you should be able to fetch exisiting libdefs like this:
```
cd path/to/my/project
flow-typed install
```


ryyppy:
That's huge! I know there were some plans doing this, but @jeffmo was talking about some issues with the package.json approach... will try it out hopefully soon 🎉👍

cli/src/commands/install.js
```diff
@@ -63,19 +62,114 @@ export async function run(args: Args): Promise<number> {
...
+  if (!Object.keys(depsMap).length === 0) {
     return failWithMessage(
-      "Please specify a package name of the format `PackageFoo` or " +
-      "PackageFoo@0.2.2"
+      "No package dependencies were found in package.json or specified on the command line."
     );
   }
 
+  // Get a list of defs to install
+  const defs: [LibDef] = [];
```
jeffmo: I think you mean `Array<LibDef>` here

cli/src/commands/install.js
```diff
@@ -63,19 +62,114 @@ export async function run(args: Args): Promise<number> {
...
+  for(let dep in depsMap) {
+    console.log(`Found npm dependency: ${dep} v${depsMap[dep]}`);
+    let def = await findLibDef(
```
jeffmo: You're probably not done here, but when you are: Best to group these `await`s into an array and use `Promise.all()` to dispatch them all together (rather than serially)

cli/src/commands/install.js
```diff
@@ -63,19 +62,114 @@ export async function run(args: Args): Promise<number> {
...
+      dep, 
+      depsMap[dep],
```
jeffmo: This is going to be where things get tricky: `package.json` holds version ranges which are a super set of version strings themselves...and we don't really handle them yet.

Moreover, flow-typed versions libdefs with ranges as well -- so this means we're going to have to compare the range specified for a dependency in `package.json` with all of the possible ranges available for a given package in flow-typed and choose the "best" one.

Making the choice basically amounts to picking the version of the package that is both the lowest-compatible version of the range in package.json as well as the lowest-compatible version in flow-typed.

The easiest way to do this would be to have a list of all the versions for a given package -- but the caveat there is that would require calling out to npm to query the list of versions. A simpler way might be to find the lowest-possible version for each range (package.json range, and all the flow-typed ranges) and choose the lowest matching range in flow-typed.

cli/src/commands/install.js
palange: Tried that.
It failed with an error accessing the file index.lock while rebasing the flow-typed cache. 
I suppose we could/should redesign ensureCacheRepo to serialize requests to rebase the cache, instead of forcing serialization up to the callers.

palange:
This PR addresses issue #252

PR #290, opened 2016-09-02
validate disjointness of flow versions per package

avikchaudhuri:
Ensure that the flow versions for a given package are mutually disjoint.

Fixes https://github.com/flowtype/flow-typed/issues/30


avikchaudhuri:
Looks like I wasn't running flow! Will fix and resend.

thejameskyle:
This file as a whole seems to be duplicating some functionality of https://github.com/npm/node-semver which is already a dependency.

cli/src/lib/semver.js
```diff
@@ -8,7 +8,7 @@ export type Version = {
   major: number | "x",
```
thejameskyle: This file as a whole seems to be duplicating some functionality of https://github.com/npm/node-semver which is already a dependency.

jeffmo: @thejameskyle: It wraps that dependency, not duplicates it. The semver forms that we accept in flow-typed are a limited form of the full-blown types in `semver` -- so we use this module to help ensure we stick to those limited forms.

Additionally, we now have a need to compare 2 ranges (whereas `node-semver` will only compare exact versions to a range)

cli/src/lib/semver.js
```diff
@@ -48,6 +48,75 @@ export function sortVersions(a: Version, b: Version): number {
...
+}
+
+/**
+ * Given an array of versions, returns whether they are mutually disjoint.
+ */
+function _disjointVersionsAll(vers, i) {
+  const n = vers.length;
+  if (i+1 === n) return true;
+  for (var j = i+1; j < n; j++) {
+    if (!disjointVersions(vers[i], vers[j])) return false;
```
jeffmo: Doesn't this only verify that two sequential versions are mutually disjoint? What about something like:

`disjointVersionsAll(['v1.1.x', 'v0.1.x', 'v1.x.x'])`

jeffmo:
Overall looks good, but I think we'll either need to sort the ranges in `disjointVersionsAll` or do something to ensure non-sequential disjointedness.

The travis errors appear unrelated to this diff, so I'll follow up on those after lunch and you can just rebase.

jeffmo:
You should be able to rebase now to kill the Travis errors

cli/src/lib/semver.js
avikchaudhuri: No, I think it's correct. We check `vers[0]` with `vers[1], vers[2], ..., vers[n-1]`; then `vers[1]` with `vers[2], ..., vers[n-1]`, and so on. I'll add your test and rebase.

Issue #225, opened 2016-08-02
Windows issues


Phoenixmatrix:
Now that Flow[ supports Windows](https://flowtype.org/blog/2016/08/01/Windows-Support.html) (yay!), this probably should too.

2 issues I'm seeing so far: 

First, the obvious one [here](https://github.com/flowtype/flow-typed/blob/master/cli/src/commands/runTests.js#L20), probably just needs the binary path added.

I tried to fix that locally, but I encountered another bigger issues: the file names. Git for Windows really didn't seem to enjoy paths like `flow_>=v0.28.x`.

Don't know what a good solution would be.

thejameskyle:
Looked at this with @gabelevi and it appears that `>` and `<` are reserved characters on windows paths.

@jeffmo We should find another way to communicate the version ranges

jeffmo:
It's not possible to escape `<` and `>`?

jeffmo:
What happens when you check out this repo on a windows machine? Does it just mangle the directory names?

Phoenixmatrix:
It just ignores them and give an error that checkout failed (but everything else gets checked out fine), so `git status` then shows those directories as deleted.

gabelevi:
Nope, no way to escape as far as I can tell. And yeah, I think git has the blob with the info for the file, it just cannot create the file on the filesystem.

jeffmo:
whoa, that's pretty disappointing :(

So having a clear and structured naming format is pretty important for automated tooling here. Any objects to just switching `>=` and `<=` to `gte` and `lte`?

`flow_>=v0.30` -> `flow_gte0.30`
`flow_<=v0.30` -> `flow_lte0.30`
`flow_v0.30` -> `flow_0.30`
`flow_>=0.30_<=0.35` -> "flow_gte0.30_lte0.35"

gabelevi:
What about

`flow_>=v0.30` -> `flow_v0.30-`
`flow_<=v0.30` -> `flow_-v0.30`
`flow_v0.30` -> `flow_v0.30`
`flow_>=v0.30_<=v0.35` -> `flow_v0.30-v0.35.0`

or

`flow_>=v0.30` -> `flow_v0.30...`
`flow_<=v0.30` -> `flow_...v0.30`
`flow_v0.30` -> `flow_v0.30`
`flow_>=v0.30_<=v0.35` -> `flow_v0.30...v0.35.0`

[...]

PR #296, opened 2016-09-08
check version sat

avikchaudhuri:
flow versions of packages should be satisfiable, in the sense that we should not allow bogus ranges whose lower bound and upper bound point in the wrong directions


jeffmo:
This will also be moot given #225, right?

PR #297, opened 2016-09-08
rename upperBound to otherBound

avikchaudhuri:
version.upperBound is not necessarily an upper bound, because we don’t enforce that it begins with <=. (E.g., we might have <=v.1.1.x_>=v.1.0.x, where version.upperBound is actually a lower bound.) Thus renaming to otherBound.

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -249,17 +249,17 @@ async function runTestGroup(
         return true;
       }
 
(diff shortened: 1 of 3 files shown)


avikchaudhuri:
Given the plans around #225, looks like we can keep the name `upperBound` and enforce that it is indeed an upperBound instead of this PR. In other words, don't accept `<=v.1.1.x_>=v.1.0.x`, instead insist on `v1.0.x-v1.1.x`.

PR #298, opened 2016-09-08
fix: Response.render callback is optional

gcedo:
The render callback is not mandatory http://expressjs.com/en/api.html#res.render


thejameskyle:
Can you fix the tests? https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md#5-run-your-tests-with-run_def_testssh-left-pad

jeffmo:
d2849ed4c42566f1f0662ee6f68ee7d0450a75c5

PR #300, opened 2016-09-08
Fix redux-saga for flow 0.32

### LINKED ISSUE #292: redux-saga takeLatest does not typecheck when called declaratively
Issue discovered here: https://github.com/yelouafi/redux-saga/issues/358#issuecomment-244812498

From @ryyppy
```
Hmm, this is because the takeLatest requires a Generator type, but gets my own backed simulation of a Generator...

I did this to my definitions and the tests run through:

  declare interface TakeXRet extends Generator<*,*,*> {
    name: string,
  }
Not sure if I should patch this... probably yes :-)
```

aaronjensen:
Also fixes #292


aaronjensen:
@ryyppy would you mind taking a look over my changes?

definitions/npm/redux-saga_v0.11.x/flow_>=v0.28.x/redux-saga_v0.11.x.js
```diff
@@ -87,38 +87,35 @@ declare module 'redux-saga' {
   declare type Saga5<Y: IOEffect, R, N, T1, T2, T3, T4, T5> = (t1: T1, t2: T2, t3: T3, t4: T4, t5: T5) => Generator<Y, R, N>;
   declare type Saga6<Y: IOEffect, R, N, T1, T2, T3, T4, T5, T6> = (t1: T1, t2: T2, t3: T3, t4: T4, t5: T5, t6: T6) => Generator<Y, R, N>;
 
-  declare type TakeXRet = {
-    @@iterator(): $Iterator<*,*,*>,
+  declare interface TakeXRet extends Generator<*,*,*> {
```
aaronjensen: 💭 I wasn't sure if this was idiomatic or if `Generator<*,*,*> & { name: string }` was more-so. They seem to be the same, yeah?

ryyppy: If the tests go through, it should be fine :-)
Looking good!

jeffmo:
Looks like this doesn't update the Flow version upper bound of "<=0.31" in the path name (thus tests won't run against 0.32 and the CLI won't see this libdef if someone requests redux-saga for Flow >0.31)

jeffmo:
Doh mobile github is hard to read -- disregard my previous comment

PR #302, opened 2016-09-14
Updated Readme + Contributing docs

jeffmo:
This just tweaks the README.md and CONTRIBUTING.md docs a bit + updates them with post-launch details (This is part of https://github.com/flowtype/flow-typed/issues/254).

Note that assumes the [`flow-typed install` command (no flags)](https://github.com/flowtype/flow-typed/pull/289) feature has landed, so it shouldn't land until after that does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,17 +15,33 @@ You can grab definitions directly from this GitHub repo, or you can use the CLI 
 $ npm install -g flow-typed
 
 $ cd /path/to/my/project
-$ flow-typed install -f 0.30 rxjs@5.0.0 # `-f 0.30` specifies the Flow version we're using for this project
+$ npm install
+$ flow-typed install
 'rxjs_v5.0.x.js' installed at /path/to/my/project/flow-typed/npm/rxjs_v5.0.xjs
 ```
 
+**We recommend checking installed libdefs in to your project's version control.**
+
+Libdefs in flow-typed are tagged at both Flow and library version when they are
+installed, but libdefs themselves can improve over time. For example, they may
+have a bug or there may be an improvement to their accuracy or completeness.
+
+When a libdef is improved or updated in flow-*typed*, there's some chance that
+the change could introduce new Flow errors into your project. As good as it is
+to find new issues, **we also want to make sure that Flow errors in your project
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,32 +119,29 @@ working with this repository:
...
+structured. It should be run *after* running 
+[`flow-typed validate-defs`](#flow-typed-validate-defs).
 
 ##### `flow-typed update-cache [--debug]`
 
 By default flow-typed retrieves all available libdefs from its related upstream
-repository. To make this process more efficient, those libdefs will be cached to
-your local filesystem. Usually, the cache will automatically be updated after a
-certain grace period during a libdef installation, but sometimes it is useful to
+repository. To make this process more efficient, those libdefs arce cached once
```
thejameskyle: s/arce/are

jeffmo: "Requested changes" -- so awesome!

jeffmo: Ok, so I updated this -- but GitHub makes it seem like I didn't...
(Not that we can merge this yet, just noting)

jeffmo:
Aw, it looks like @GitHub missed a critical aspect of accept/request changes: That when an update comes back after an RC, it should reset :(

migueloller:
Hey @jeffmo, quick q. Do you recommend to check in lib typedefs (i.e., flowtyped dir) into version control? Why or why not?

jeffmo:
@migueloller: Yes, we recommend checking in libdefs.

Libdefs are tagged at flow/library version when installed, but libdefs themselves -- even for the same flow/library version -- can improve over time. There may be a bug or an improvement to accuracy, etc.

So in the scenario where a libdef was improved, it's possible that that improvement may cause new Flow errors that weren't previously present in your repo (similar to what happens when you upgrade Flow: Flow gets smarter with each release, so it may find new issues it wasn't able to before). And thus, we want to make sure your flow errors are *consistent* over time for each checkout of the library.

If/when you want to upgrade a libdef, simply update it and check that update (plus any potential fixes) in for all of your library's contributors to benefit from as well.

---

Thanks for asking that question, I'll add the above to the README here :)

PR #280, opened 2016-08-25
Add react-storybook module definition

ajhyndman:
Hey.  I put this together for my own project, and thought it might be useful for others.

I tried to follow the definition and test style from other packages.  Am I missing anything?

PR #303, opened 2016-09-15
support for scoped lib dirs

avikchaudhuri:
Build support for libs corresponding to scoped npm packages, e.g., https://github.com/flowtype/flow-typed/pull/280


avikchaudhuri:
I don't really know how to fully test this feature. I made a test dir following https://github.com/flowtype/flow-typed/pull/280, and ran validate-defs and run-tests on it. Those work. What else?

In particular, the `pkgName` field of a scoped lib def, e.g., `@foo/bar` is still the base name `bar`, and I feel like it should be `@foo/bar` but then we use the `pkgName` in a bunch of places to create paths, and the path to a scoped lib def is of the form `definitions/npm/@foo/bar_v.../flow_v.../bar_v...`, which means that `bar` isn't always preceded by `@foo`. Should I add an optional `scope` field to the `LibDef` type?

cc @jeffmo @gabelevi

avikchaudhuri:
Alright, need to make search and install work. Question: for testing, I'm faking the online repo by adding my test dir to the cache and bumping the expiry to `Number.POSITIVE_INFINITY`...is there a better way?

jeffmo:
> Should I add an optional `scope` field to the LibDef type?

Yea, that would make sense to me. Then handle appropriately where necessary.

> I don't really know how to fully test this feature. I made a test dir following #280, and ran validate-defs and run-tests on it. Those work. What else?

That's a good start. There's a test file for `libDefs.js` here: https://github.com/flowtype/flow-typed/blob/master/cli/src/lib/__tests__/libDefs-test.js

It mocks out the FS API and Git APIs as needed for each test -- so you can probably mimick some of that.

> for testing, I'm faking the online repo by adding my test dir to the cache and bumping the expiry to Number.POSITIVE_INFINITY...is there a better way?

Over in that test file, I made some "test-only" exports with the convention of using a `_` prefix on the name of the export to make it clear somethings up. You could probably add a test-only API that sets `CACHE_REPO_EXPIRY` to `Number.POSITIVE_INFINITY` (or just add support for a special `-1` value or something).

Alternatively you could thread a `noCache` arg through all the various library functions, but that probably gets annoying fast...so I'd probably just opt for the former.

avikchaudhuri:
@jeffmo Thanks for the review. Haven't changed LibDef yet, but I did test that all commands (install, search, run-tests, and validate-defs) works, sometimes with temporary changes to get around (e.g., I also had to to tweak the code to ignore histories when creating a git-versioned install path).

So current status is, this works, but testing is mostly manual. How should we proceed? Do you want me to make this more robust / testable in this PR? Or merge this and I follow-up?

jeffmo:
If you'd prefer to do a follow-up PR with tests, I'm ok with that as long as merging doesn't break current operation of `run-tests` and `validate-defs` (the pending travis test will confirm this).

The master version of the CLI runs in travis for each libdef PR for validation, so we just need to keep that working and robust. Since we don't have any namespaced libdefs merged right now that basically just means: As long as nothing regresses we can merge.

The GitHub review stuff is pretty new (to everyone I guess), so also want to make sure you saw my comments about tweaking the comment block, etc before merging.

avikchaudhuri:
Sure, ok will address minor comments here and then follow up later.

cli/src/commands/runTests.js
```diff
@@ -43,7 +43,7 @@ async function getTestGroups(repoDirPath): Promise<Array<TestGroup>> {
   const libDefs = await getLibDefs(repoDirPath);
   return libDefs.map(libDef => {
     const groupID =
-      `${libDef.pkgName}_${libDef.pkgVersionStr}--flow_` +
+      `${libDef.pkgName}_${libDef.pkgVersionStr}${path.sep}flow_` +
```
avikchaudhuri: @jeffmo See line 194 below. The reason for the change is that `libDef.pkgName` can now also contain `/`, so I'm just escaping in one place instead of two.

jeffmo:
https://github.com/flowtype/flow-typed/commit/6916065655492e1196c5d5ebbb0e9cacc9e3bd54

PR #304, opened 2016-09-15
Add non-legacy method names to kefir.Emitter

benhartley:
After doing this I see that Kefir actually ships with [up-to-date types](https://github.com/rpominov/kefir/blob/master/kefir.js.flow), however, many people may not notice this and look to flow-typed for the definitions instead.

In [3.3.0](https://github.com/rpominov/kefir/blob/master/changelog.md#330-16072016), two methods were renamed. The old names work for now so I've just added new definitions and kept the old ones.


jeffmo:
8e74738876de96ff810ef18ab01a96e38d3803b3

jeffmo:
For future ref: Since this libdef is `3.x.x`, that means it must also be compatible with `3.2.x` as well -- but `3.2.x` doesn't have these new methods.

So instead of adding these methods here, I went ahead and just created a `3.3.x` version of the libdef over in 8e74738876de96ff810ef18ab01a96e38d3803b3 and reverted the changes in the `3.x.x` libdef.

PR #307, opened 2016-09-16
Express: Added all(...) method for Route

neurovertex:
As I commented on #90, the express def file is missing the all() method for Route, so here it is.
I was also advised to update the test files accordingly, but upon inspection, said file does not test for any other method of Route with the exception of get() so I wasn't sure what else to add (also all said methods share the same signature so I'm guessing testing one should be enough).


jeffmo:
https://github.com/flowtype/flow-typed/commit/08386ddd61f7cc0a06f243e8bd642959d534daff

PR #308, opened 2016-09-16
Change filename convention to support local libdefs on Windows

### LINKED ISSUE #225: Windows issues

samwgoldman:
The range syntax used in `/definitions/npm` uses `<=` and `>=`, which does not work on Windows. This PR changes the file names to use a different range syntax.

To keep this small and avoid mid-flight collisions, I only changed the code in `lib/libDef.js`, which deals with filesystem paths.

The "codemod" to move the files:

```
find definitions/npm -type d -name flow_* | gawk -f move.awk
```

Where `move.awk` is:

```
{ r = gensub(/>=(v[0-9x.]+)_<=(v[0-1x.]+)/, "\\1-\\2", 1)
  r = gensub(/>=(v[0-9x.]+)/, "\\1-", 1, r)
  r = gensub(/<=(v[0-9x.]+)/, "-\\1", 1, r)
  system("git mv '" $0 "' '" r "'") }
```

Fixes #225


samwgoldman:
Not sure what's up with Travis, but tests pass on my machine.

jeffmo:
Travis error looks like an npm thing (seems to happen sometimes). I'll restart it

Sinewyk:
Shouldn't you also edit [theses lines](https://github.com/flowtype/flow-typed/blob/07eca72148e4134634b7ee30e5422a6dfec32b13/cli/src/commands/runTests.js#L15-L24 ?) for windows support ?

samwgoldman:
Will circle back and look at overall Windows support in a separate PR, but this part of the work is done and ready for review.

(Note that the renames are likely to cause merge conflicts with all the other open PRs. I'm happy to rebase on my end at any point to minimize conflicts.)

PR #309, opened 2016-09-17
Add definition for aphrodite.

wbinnssmith:
This adds a definition for [the aphrodite styles-in-js](https://github.com/Khan/aphrodite) library. It functions all the way back to flow 28 and I targeted v0.5 in my definition. 

A couple places annotate simply `Object` as I ran into some issues with more specific types conforming to the expected shape. Let me know if something more specific is preferred and I can try my best to make that happen :)


jeffmo:
https://github.com/flowtype/flow-typed/commit/7fc28637a1ec29b5e4f04ee054ca38a81ac05df5

PR #312, opened 2016-09-18
Update express_v4.x.x.js

vaukalak:
when working with express you usually do 

```
const foo = req.body.foo;
```

when `body` is `mixed` - this is impossible.

### DIFF
--- a/definitions/npm/express_v4.x.x/flow_v0.25.x-/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_v0.25.x-/express_v4.x.x.js
@@ -12,7 +12,7 @@ declare module 'express' {
   }
   declare class Request extends http$IncomingMessage mixins RequestResponseBase {
     baseUrl: string;
-    body: mixed;
+    body: any;
     cookies: {[cookie: string]: string};
     fresh: boolean;
     hostname: boolean;
@@ -138,7 +138,7 @@ declare module 'express' {
     use: RouterMethodType<this>;
 
     route(path: string): Route;
-    
+
     static (): Router;
   }
 


definitions/npm/express_v4.x.x/flow_>=v0.25.x/express_v4.x.x.js
```diff
@@ -12,7 +12,7 @@ declare module 'express' {
   }
   declare class Request extends http$IncomingMessage mixins RequestResponseBase {
     baseUrl: string;
-    body: mixed;
+    body: any;
```
bassettsj: Yeah, this seems to be a problem, but I am wordering if we can add something like a type paramater 

```
declare class Request<B> extends http$IncomingMessage mixins RequestResponseBase {
  baseUrl: string;
  body: B;

}

**Comment by jeffmo** [comment]
After updating this PR, would you mind rebasing on top of #308?

We knew this would be a little painful, but luckily the merge is easy: Just rename the flow version directory from "flow_>=v0.25.x" to "flow_v0.25.x-".

**Comment by vaukalak** [comment]
@bassettsj yes, but what will be the usage of this? match, if we can append a certain handler to the route? in this case will have to rewrite almost all the declarations...

**Comment by bassettsj** [comment]
@vaukalak This would allow you to do something like this:


```
// @flow

import type { $Response, $Request } from 'express';

app.use(function(req: $Request<?{id: string, name: string}>, res: $Response) {
  if (req.body == null) {
    res.code(400).send({ message: 'error expected body to be sent wth id and name' });
  }
  // can help find issues like missspelling an attribute name on the body.
  User.find(req.body._id);
});

```

File: definitions/npm/express_v4.x.x/flow_v0.25.x-/express_v4.x.x.js [hunk 2]
```diff
@@ -12,7 +12,7 @@ declare module 'express' {
   }
   declare class Request extends http$IncomingMessage mixins RequestResponseBase {
     baseUrl: string;
-    body: mixed;
+    body: any;
```
jeffmo: At minimum this should probably be `Object`, right?

vaukalak: @jeffmo shouldn't everything in javascript be an `Object` ? what is the point of `any` in this case?

jeffmo: `Object` guarantees, (as just 1 example), that you don't pass this value into something that's a `number`:

```js
function sum(a: number, b: number) { return a + b; }
sum(req.body); // oops!
```

jeffmo: @vaukalak: TBH I would probably never advocate for mutating an internal thing in that way as a side-effect-based contract, but if it's a common pattern then we should probably handle it.

All this said, it occurs to me that this is controversial because different developers/use-cases demand/expect certain levels of strictness/looseness. What if we made the `body` type a type parameter and asked developers to opt-in to their stated body type?

```js
declare class Request<BodyT> {
  baseUrl: string;
  body: BodyT;
  cookies: {[cookie: string]: string};
  ...etc...
}
```

Then to use:

```js
app.post('/', (req, res) => {
  console.log(`req.body: ${JSON.stringify(req.body)}`);
  console.log(`typeof req.body: ${typeof req.body}`); // string
  res.send('kuku');
});
```

^^ works leaving the `BodyT` type as just the "`empty`" type since Flow doesn't know what it is (`empty` can flow into anything, but nothing can flow into it..a bit of a strange type, but it models Flow's understanding of the type correctly).

And a user who wishes to be specific can simply annotate the `res` parameter:

```js
type MyResponseT = /* ... */;
app.post('/', (req: Response<MyResponseT>, res) => {
  console.log(`req.body: ${JSON.stringify(req.body)}`);
  console.log(`typeof req.body: ${typeof req.body}`); // string
  res.send('kuku');
});
```

What do you think? (cc also: @bassettsj )

vaukalak: @jeffmo In this case I think we should list fields that should be generic. For example:
***Request***
_body_
***Response***
_body_ in the _`send`_ method.

Also. I did some [tests](https://flowtype.org/try/#0MYGwhgzhAEBKCmBHArvCAXAPAFWgLmgHkAjAK3mHQD5oBvAKAEhiB7AEwE8DsBuegXz710HAA7xoACTAA7NiHgAnHDQC80ABSKkBBCjRZsVAJTRVNAG4sAlmyGhIMWC2ToJDaNFEsMKjaLB0AAsCDEVrGQBzABpoEKlZeSUVYwIrWzp6T08s6H4BenoZeAB3OBc3DWMAOm8MDQByAHoG2K1dJFRfAEFFRTAOTDCIyKoTMxpgFhkIFgVqkBZIrWrWTmNjHiA), and find out that you can't assign an `Array` to an `Object` field o__O
```
let a : Object = []; // empty array literal. This type is incompatible with object type.
```

not sure if this is a flowtype bug, but I think that if a request is a json, it can be an array.. BTW, if you use json middleware for array, you can receive all values, that are valid json.. So I don't think we have to constrain it to object.

jeffmo: > BTW, if you use json middleware for array, you can receive all values, that are valid json.. So I don't think we have to constrain it to `Object`. So just `Request<T>` without `Request<T:Object>`.

Ok, that's pretty compelling -- I'm onboard with leaving bounds off the typeparam then.
As long as @bassettsj is also onboard here, sounds like we're all in agreement!

jeffmo: > you can't assign an `Array` to an `Object`

This was an intentional design decision that trades off between utility and exactly modeling JS. It's easy to make an argument in both directions really, so we went with the more conservative expectation (that an "Array" is disjoint from the concept of an "Object": Objects have named properties, arrays are usually treated more vector-y).

Pedantically: Anything that you can do with an `Object` could also be done with an `Array`, but practically this is probably not often the intention.

marudor:

PR #313, opened 2016-09-18
Implement libdefs for Marked

Kriegslustig:
This PR add libdefs for [Marked](https://www.npmjs.com/package/marked) NPM package.

PR #315, opened 2016-09-19
Implement libdefs for marked-man v0.1.x

Kriegslustig:
This PR implements libdefs for [_marked-man_](https://www.npmjs.com/package/marked-man). This PR depends on #313, since _marked_ is a dependency of _marked-man_. As soon as the _marked_ libdefs have been merge, this can be rebased onto `master`.

I'm not sure though if my declarations are ok. Since Flow doesn't support intersections with object types yet, I had to make some cuts to avoid code duplication between this and _marked_. But since these definitions covers all "normal" use cases (which is to extend _marked_ and only use _marked-man_ as described in the documentation), I thought it was ok to do it like this. Any thoughts on this would be welcome.


Kriegslustig:
I updated it to the state of master. So it can be merged now. Thanks!

definitions/npm/marked-man_v0.1.x/flow_v0.28.x-/marked-man_v0.1.x.js
```diff
@@ -0,0 +1,64 @@
+import marked from 'marked'
```
jeffmo: Unfortunately this doesn't work. We haven't yet implemented support for dependencies between libdefs (see also https://github.com/flowtype/flow-typed/issues/16). (As-is, this will just turn out as effectively an `any` type)

For now your best bet is going to be to copy/paste from `marked` into this libdef until we add property support for `declare import` inside `declare module` bodies.

definitions/npm/marked-man_v0.1.x/flow_v0.28.x-/marked-man_v0.1.x.js
```diff
@@ -0,0 +1,64 @@
...
+
+type NodeCallback<T> = (e: ?Error, d: ?T) => void
```
jeffmo: Can you move these into the body of the `declare module`? As-is, these create global types

definitions/npm/marked-man_v0.1.x/flow_v0.28.x-/marked-man_v0.1.x.js
Kriegslustig: Wouldn't it be better to use the `$` convention? So exporting `marked$Marked` from the marked libdef?

jeffmo: Yea, that works too if you want to update the `marked` libdef to have global + namespaced types.

Kriegslustig:
I copied over the global marked definitions (`marked$*`) and used these to implement the libdefs. The new libdefs are a lot more restrictive. IMO they cover the "only real" use case of the module.

PR #318, opened 2016-09-21
Add redux and react-redux definitions

gcanti:
This PR implements libdefs for redux and react-redux
These libdefs have been already merged in the redux repo
Here's the related PR https://github.com/reactjs/redux/pull/1887

A definition file for both redux and react-redux already exist, I'd love to involve their respective authors @marudor and @nmn. Is this ok for you? Do you have any objections / suggestions?


definitions/npm/react-redux_v4.x.x/test_react-redux_v4.x.x.js
```diff
@@ -1,65 +0,0 @@
...
-    dispatch
-  })
-)(Test);
-
-// Works without dispatch OR timeOfDay OR updateTime. Injected implicitly
-const d1 = <ConnectedTestD name='hjh' />;
-// $ExpectError
-const d2 = <ConnectedTestD />; // missing prop: name
-// $ExpectError
-const d3 = <ConnectedTestD name='345' timeOfDay={23443} updateTime={() => undefined} />; // timeOfDay: string
```
nmn: I don't see any new test.

Looking at the types it looks very loose and I think it will no longer have a strict type for the connected components.

I see a lot of `void` and `any` being used.

I don't see this working in a practical setting.

gcanti: @nmn thanks for reviewing this PR. Could you please elaborate a bit more? Running the new test suite against the old definition file raises 12 errors, which means there are (at least) 12 new tests and the new definition is (theorically) safer. Why do you think types are "very loose"? More details would be helpful, either to improve this definition or dismiss the changes. 

The raised errors:

```
src/test_connect.js:49
 49: // $ExpectError

src/test_connect.js:51
 51: // $ExpectError

src/test_connect.js:53
 53: // $ExpectError

src/test_connect.js:70
 70: // $ExpectError

src/test_connect.js:73
 73: // $ExpectError

src/test_connect.js:77
 77: // $ExpectError

src/test_connect.js:91
 91: // $ExpectError

src/test_connect.js:93
 93: // $ExpectError

src/test_connect.js:102
102: // $ExpectError

src/test_connect.js:104
104: // $ExpectError

src/test_connect.js:133
133: const CC8 = connector8(C3);
                 ^^^^^^^^^^^^^^ function call
 37: class C3 extends React.Component<void, Props2, void> {
                                            ^^^^^^ property `b`. Property not found in
134: <CC8 a={1} />;
     ^^^^^^^^^^^^^ props of React element `CC8`

src/test_connect.js:170
170: (CC1.WrappedComponent: C1);
          ^^^^^^^^^^^^^^^^ property `WrappedComponent`. Property not found in
170: (CC1.WrappedComponent: C1);
      ^^^ statics of React$Component
```

and the tests

```js
// @flow
import React from 'react'
import { connect } from 'react-redux'
// import type { Connector } from 'react-redux'

// copy & paste from redux libedef :(
type ReduxDispatch<A: { type: $Subtype<string> }> = (action: A) => A;

type Action = { type: 'A' } | { type: 'B' };
type Dispatch = ReduxDispatch<Action>

type State = {
  c: number,
  d: string
};

type Props1 = {
  a: number,
  b: string
};

const C1 = (props: Props1) => <div>{props.a} {props.b}</div>

type Props2 = {
  a: number,
  b: string,
  dispatch: Dispatch
};

class C2 extends React.Component<void, Props2, void> {
  render () {
    // $ExpectError
    return <div onClick={this.props.dispatch()}>{this.props.a} {this.props.b}</div>
  }
}

class C3 extends React.Component<void, Props2, void> {
  render () {
    return <div onClick={this.props.dispatch({ type: 'A' })}>{this.props.a} {this.props.b}</div>
  }
}

//
// connect()
//

const CC1 = connect()(C1);
<CC1 a={1} b="s" />;
// $ExpectError
<CC1 />; // missing a, b
// $ExpectError
<CC1 a={1} />; // missing b
// $ExpectError
<CC1 a="s" b="s" />; // wrong a type

const CC3 = connect()(C3);
<CC3 a={1} b="s" />;
// $ExpectError
<CC3 />; // missing a, b
// $ExpectError
<CC3 a={1} />; // missing b
// $ExpectError
<CC3 a="s" b="s" />; // wrong a type

//
// connect(null, null, nul, options)
//

connect(null, null, null, { pure: true })(C1);
// $ExpectError
connect(null, null, null, { pure: 1 })(C1); // wrong type
connect(null, null, null, { withRef: true })(C1);
// $ExpectError
connect(null, null, null, { withRef: 1 })(C1); // wrong type
const CC4 = connect(null, null, null, {})(C1);
<CC4 a={1} b="s" />;
// $ExpectError
<CC4 />; // missing a, b

//
// connect(mapStateToProps)
//

type OwnProps1 = { b: string };
// without ownProps
const connector5 = connect(
  (state: State) => ({ a: state.c })
);
const CC5 = connector5(C1);
<CC5 b="s" />;
// $ExpectError
<CC5 />; // missing b
// $ExpectError
<CC5 b={1} />; // wrong b type

// with ownProps
const connector6 = connect(
  (state: State, ownProps: OwnProps1) => ({ a: state.c, b: ownProps.b + 's' })
);
const CC6 = connector6(C1);
<CC6 b="s" />;
// $ExpectError
<CC6 />; // missing b
// $ExpectError
<CC6 b={1} />; // wrong b type

//
// connect(null, mapDispatchToProps)
//

type OwnProps2 = {
  a: number,
  b: string
};
// without ownProps
const connector7 = connect(
  null,
  (dispatch: Dispatch) => ({ dispatch })
);
const CC7 = connector7(C3);
<CC7 a={1} b="s" />;
// $ExpectError
<CC7 a={1} />; // missing b
// $ExpectError
<CC7 a="s" b="s" />; // wrong b type

type OwnProps3 = { a: number };
// with ownProps
const connector8 = connect(
  null,
  (dispatch: Dispatch, ownProps: OwnProps3) => ({ dispatch, b: ownProps.a + 's' })
);
const CC8 = connector8(C3);
<CC8 a={1} />;
// $ExpectError
<CC8 />; // missing a
// $ExpectError
<CC8 a="s" />; // wrong b type

//
// connect(mapStateToProps, mapDispatchToProps)
//

const connector9 = connect(
  (state: State) => ({ a: state.c }),
  (dispatch: Dispatch) => ({ dispatch })
);
const CC9 = connector9(C3);
<CC9 b="s" />;
// $ExpectError
<CC9 />; // missing b
// $ExpectError
<CC9 b={1} />; // wrong b type

//
// connect(mapStateToProps, mapDispatchToProps, MergeProps)
//

const connector10 = connect(
  (state: State) => ({ a: state.c }),
  (dispatch: Dispatch) => ({ dispatch }),
  (stateProps, dispatchProps) => Object.assign({}, stateProps, dispatchProps, { b: 's' })
);
const CC10 = connector10(C3);
<CC10 />;

//
// ConnectedComponent
//
(CC1.WrappedComponent: C1);
```

nmn: I think I missed your new tests file when I reviewed this last.
Give me a few days. I think I need to do a more thorough testing of both the type definitions.
When I tried your type definitions last, there were errors that Flow didn't pick up.
But now that Flow has been updated things may have changed.

Q: The errors here are errors thrown with your test file but the old types, right?

definitions/npm/react-redux_v4.x.x/flow_v0.30.x-/react-redux_v4.x.x.js
```diff
@@ -1,78 +1,86 @@
...
+
+    S = State
+    A = Action
+    OP = OwnProps
+    SP = StateProps
+    DP = DispatchProps
+
+  */
+
+  declare type MapStateToProps<S, OP: Object, SP: Object> = (state: S, ownProps: OP) => SP | MapStateToProps<S, OP, SP>;
```
nmn: Doesn't the `mapStateToProps` argument always return an object?

definitions/npm/react-redux_v4.x.x/flow_v0.30.x-/react-redux_v4.x.x.js
```diff
@@ -1,78 +1,86 @@
...
+
+  declare type MapDispatchToProps<A, OP: Object, DP: Object> = ((dispatch: Dispatch<A>, ownProps: OP) => DP) | DP;
```
nmn: This case doesn't work correctly in practice. When this argument is not a function, the types don't match up. (DP here can be just an object of actions, but the type of the same props in the component are all functions.

definitions/npm/react-redux_v4.x.x/flow_v0.30.x-/test_connect.js
```diff
@@ -0,0 +1,170 @@
...
+const CC4 = connect(null, null, null, {})(C1);
+<CC4 a={1} b="s" />;
+// $ExpectError
+<CC4 />; // missing a, b
+
+//
+// connect(mapStateToProps)
+//
+
+type OwnProps1 = { b: string };
```
nmn: This is the place where I think the new types are loose.
You have to pretty much manually write the props for the wrapped components.

My main effort with the types was to avoid having to write the types for wrappedComponent Props. This is why I think these types are loose. They don't infer as much.

If you're applying the same `connector` to many different props, you'll have to manually type the props for each component twice. Once before it's wrapped, and once after.

nmn:
There also seems to something that has changed with Flow 0.33

This simple example failed to type-check correctly:

```typescript
type P1 = {
  a: number,
  b: string,
  dispatch: Function
}
class C1 extends React.Component<void, P1, void> {
  render () {

  }
}

<C1 a={122} b='sfgb' dispatch={a => {}} />;
// $FlowFixMe - incorrect prop type
<C1 a='1234' b='sfgb' dispatch={a => {}} />;
// $FlowFixMe - missing props
<C1 />;
// $FlowFixMe - missing props
<C1 a={123} />;

const CC1 = connect()(C1);

// This line should give me an error. but it doesn't.
<CC1 />;
```

nmn:
Yeah, looks like my type-defs don't work with functional components with Flow 0.33

This is a problem that keeps changing with different versions of flow with my types.
Can you try my types with only `class` components and see if there are any other serious issues?

gcanti:
I put up a repo which will help us go through this process. It contains my tests and yours. https://github.com/gcanti/react-redux-libdef-tests

> The errors here are errors thrown with your test file but the old types, right?

Right

> in your type definitions, when you using the mapStateToProps or mapDispatchToProps you have to manually type the Props of the resulting component

Yes and no. Type inference is nice but I value type safety over anything else. Being able to express explicitly the contract of the connected component and make sure Flow enforces that contract is valuable. 

Anyway you can also rely on type inference, the same tests are green even if you remove the `Connector` annotations (see `test_connect_without_connector.js` in the repo above)

> Doesn't the mapStateToProps argument always return an object?

See https://github.com/reactjs/react-redux/blob/master/docs/api.md#arguments

"Note: in advanced scenarios where you need more control over the rendering performance, mapStateToProps() can also return a function. In this case, that function will be used as mapStateToProps() for a particular component instance"

> This case doesn't work correctly in practice. When this argument is not a function, the types don't match up. (DP here can be just an object of actions, but the type of the same props in the component are all functions.

Not sure I'm following, could you please elaborate or show an example? In your definition file this case seems not handled

> This simple example failed to type-check correctly, with your type definitions.

Maybe we have a different setup because I can see the type checker complaining (see `new_test.js` in the repo above)

nmn:
Thanks for all the work @gcanti . The repo looks very helpful, and I'll do some more tests.

First let me answer some of your questions:

> See https://github.com/reactjs/react-redux/blob/master/docs/api.md#arguments

I didn't know about this. I'm sure it can be addressed though.

> >This case doesn't work correctly in practice. When this argument is not a function, the types don't match up. (DP here can be just an object of actions, but the type of the same props in the component are all functions.

> Not sure I'm following, could you please elaborate or show an example? In your definition file this case seems not handled

I'll double check if this case actually affects your type definitions. But the gist of what I was trying to say was this. If the `dispatchToProps` argument is not a function, it's type look something like this:
`{[key: string]: Action}`. But the actual props injected into the component look like this:
`{[key: string]: () => dispatch(Action)}`

This case affected my type definitions as I couldn't just subtract this object from the prop object for the Component. The reason I left this case out was because I agree with you and value type-safety over the other stuff.

---

On removing the `Connector` annotations. I tried that locally, and I got no type-checking as a result.
I'll try again with your repo.

If however, you do have to type the `Connector` type annotation every time, then I feel the type definitions don't really help. This is because this is possible even without type-defs for `connect`:

```
type Props
class X extends React.Component<void, Props, void> {}

export default (connect(a, b)(X): ReactClass<PropsNeededOutside>)
```

I don't mind writing type manually. I just don't feel like we should have to repeat ourselves for type-safety.

---

Some of the things that are clearly broken in the current type-defs:

1. I use `React$Component` itself as the return type. You defined a `ConnectedComponent` class. This is something I missed. I missed the `getWrappedInstance` use-case. This needs to be fixed.

2. My type-defs were written for Flow 0.28 or 0.29. At that point, type definitions for Functional Components didn't work. As a result, I used some hacks to make it all work correctly. In fact, at that time, wrapping a function in `connect` would make it start working. This is no longer the case. My type-defs are currently broken for all functional components. I'm trying to find a way to fix it.

3. My type-defs don't account for the `mapDispatchToProps` being an object case. I had done this as the extra overload was breaking all the other type definitions. I will revisit and see if it works now.

4. My type-defs for `Provider` were very weak. Your idea of using type defs from `redux` for `Store` and `Dispatch` are a great idea.

----

I'm having a busy few days right now, but I would love to perhaps pair on this sometime and figure this out.

aaronjensen:
> I'll double check if this case actually affects your type definitions. But the gist of what I was trying to say was this. If the dispatchToProps argument is not a function, it's type look something like this:
> {[key: string]: Action}. But the actual props injected into the component look like this:
> {[key: string]: () => dispatch(Action)}

When you say `{[key: string]: Action}` do you actually mean `{[key: string]: ActionCreator}`? ["If an object is passed, each function inside it will be assumed to be a Redux action creator"](https://github.com/reactjs/react-redux/blob/master/docs/api.md#arguments)

If so...

`dispatch` returns the action: https://github.com/reactjs/redux/blob/master/src/createStore.js#L181

and bindActionCreator returns what dispatch returns:
https://github.com/reactjs/redux/blob/master/src/bindActionCreators.js#L2

so the type is actually exactly the action creator function as far as I can tell, so `dispatch` could be reduced out of a type: `() => dispatch(Action)` === `() => Action`

gcanti:
> On removing the Connector annotations. I tried that locally, and I got no type-checking as a result.
I'll try again with your repo.

Yes please. If we come up with an agreement we could merge these libdefs right away, no more work to do for both of us. Tests are green and functional components are handled. Then we could improve the libdef (if ever needed).

I don't mind being the author of this libdefs, all I want is to put an end to this very long journey (56 days and counting) and produce something useful and actionable for the users. Otherwise just let me know, and please feel free to take any parts of them and improve yours.

(Actually I'm not even sure that having a huge repository for definition files is a good idea. Maybe libdefs should compete on npm like libraries do)

jeffmo:
@gcanti/@nmn/@aaronjensen: Please ping me when you're in agreement on this and I'll merge it. You three are more familiar with the library than I, so I'll defer to your judgement on when it's ready.

> Actually I'm not even sure that having a huge repository for definition files is a good idea.

We decided to keep them centrally stored (but community-managed) so that there are canonical, high-quality, tested+versioned libdefs for each library that everyone can coalesce on.

Additionally, one problem we hit while exploring the idea of publishing each libdef to npm is that npm packages are immutable. This would be ok if libdefs were to be versioned separately from the package they are describing, but in reality it should be possible to iterate on a libdef for the same `pkg@version` to improve it over time. If libdefs are versioned separately from the package they describe, though, it's hard to tell which version of the library it corresponds to. Moreover, npm offers no way to version against both the library *and* Flow's version itself.

Anyway...hopefully that makes sense!

(Happy to discuss the npm further if you like, but I guess we should open a separate issue for that if so)

gcanti:
@jeffmo thanks for explaining, both TypeScript and Flow share the same problem with definition files, I'm sure we will find a good solution based on feedback from both communities.

If I'm not wrong you are about to launch the flow-typed cli, so I think it's important to provide high-quality libdefs for such popular libraries like redux and react-redux, it might end up with better adoption if users can acknowledge the benefits that Flow brings to development. I want to stress that these libdefs have been reviewed, improved and tested on real world codebases by the redux community for more than 2 months now, hopefully we can come up with an agreement before the cli launch. @nmn please ping me if there are some issues, I'm available for any clarifications. In the meanwhile I rebased with master so now everything is green

jamesisaac:
Hi @gcanti, I can't seem to get your Flow v0.28-v0.32 `combineReducers` def to work with reducers any more complex than those provided in the tests.  Maybe I'm using the defs wrong though.  Simple example of failing case (took your combineReducers test and turned `state.a` into (a) a full reducer definition, (b) managing an object rather than a number):

``` js
// @flow
import type { Reducer } from 'redux'
import { combineReducers } from 'redux'

type AState = { foo: number }
const initialAState: AState = { foo: 1 }
const reducerA: Reducer<AState, Action> = (state = initialAState, action) => {
  return state
}

type State = {
  a: AState,
  b: string,
}
type Action = { type: 'A' } | { type: 'B' }

// $FlowExpectError
combineReducers() // wrong reducers argument
// $FlowExpectError
combineReducers([]) // wrong reducers argument
const r1: Reducer<State, Action> = combineReducers({
  a: reducerA,
  b: () => 's'
})
// $FlowExpectError
const r2: Reducer<State, Action> = combineReducers({
  a: reducerA,
  b: () => 's',
  c: () => true // wrong key
})
```

Flow 0.32 output:

```
js/reducers/__tests__/typedefs-test.js:21
 21: const r1: Reducer<State, Action> = combineReducers({
                                        ^ function call
  7: const reducerA: Reducer<AState, Action> = (state = initialAState, action) => {
                             ^^^^^^ object type. This type is incompatible with
 52:   declare function combineReducers<S: Object, A>(reducers: {[key: $Keys<S>]: Reducer<*, A>}): Reducer<S, A>;
                                                                                          ^ string. See lib: flow/redux.js:52
```

In the mean time I've had to alter the combineReducers def to `... Reducer<any, A> ...`.

gcanti:
Hi @jamesisaac, during the porting from redux https://github.com/reactjs/redux/blob/master/flow-typed/redux.js#L50 I wrongly replaced `any` with `*`, thanks for reporting

jamesisaac:
Great, thanks for quick follow up.

Which repo am I better off checking if I want to keep up to date with the very latest version of these libdefs?  Redux or flow-typed?  I presumed flow-typed would be more definitive as it keeps track of libdefs targeting multiple Flow versions, but maybe that was misguided if these ones are just being ported over from redux?

gcanti:
> Which repo am I better off checking

Definitely flow-typed. In the redux repo there are (old) copies of these libdefs because of the todos-flow example (development started [there](https://github.com/reactjs/redux/pull/1887))

maraujop:
Hi @gcanti ,

I've been some weeks following this thread and redux one too. I'm a little confused and I would like to ask a couple things if you don't mind. I'm quite a new comer to flow myself, but I'm fast becoming a fan of it.

It's my understanding that the purpose of typing redux is that types don't get lost when connecting redux to the components and also that redux is correctly used and called. I've looked at your code examples modified <a href="https://github.com/reactjs/redux/pull/1887/files">here</a>. In one of my components I'm currently doing:

```
export default connect(
  state => ({
    nextServicePk: state.Services.nextServicePk,
  }),
  dispatch => ({
    addService: data => dispatch(servicesActions.addService(data)),
  })
)(AddService)
```

And from those examples it's my understanding that I would have to change this into:

```
const connector: Connector<{}, Props> = connect(
   state:State => ({
     nextServicePk: state.Services.nextServicePk,
   }),
  dispatch: Dispatch => ({
    addService: data => dispatch(servicesActions.addService(data)),
  })
) 

export default connector(AddService)
```

1. I'm not sure if this is correct, and why flow is not able to infer types from my connect call without modifying it.
2. Without redux type defs I can already type my state actions (except dispatch), what extra gains do I get?

I know my question is not very clear, but I'm quite confused. I expected redux to be able to use my type definition when I create the store and use it everywhere in my connect calls without having to add it everywhere I use it.

Thanks a lot for your effort and your open source work.

Cheers,
Miguel

gcanti:
> why flow is not able to infer types from my connect call without modifying it

@maraujop Flow can't know that your `state` in mapStateToProps and `dispatch` in mapDispatchToProps are precisely those which comes from redux, you have to help Flow.

> what extra gains do I get?

It's a trade off, in order to get strict type checking you have to write a little more

aaronjensen:
@gcanti would it be possible/worth documenting a way to export your own `connect` function that is typed to your actual state and actions?

jeffmo:
Ok, let's move forward and we can continue to iterate here if necessary with follow-ups.

gcanti:
@aaronjensen mmmh not sure I understand what you mean

aaronjensen:
@gcanti if I understand correctly, `connect` cannot infer the type of `State` or `Action` (and therefore `Dispatch` unless it is provided some additional information, like from your implementation of `mapStateToProps`. We'll often declare our selectors/`mapStateToProps` in other files and use the `State` type we have defined elsewhere, which is fine, but if I'm going to build an ad-hoc selector, it's nice to not have to type it and to be able to just write `connect(s => s.foo)` or whatever.

I can't do this with a polymorphic `connect`, but if I import that `connect` and put it in a variable that is monomorphic and then export that and use it wherever I would have otherwise used `connect`, wouldn't I get the type benefits consistently and w/o any additional annotation?

Maybe I'm off the rails here 😄 , please correct me if I'm missing something.

gcanti:
Ah got it, a kind of "curryng" at the type level.

Modified libdef

```js
import type { Dispatch, Store } from 'redux'

declare module 'react-redux' {

  /*

    S = State
    A = Action
    OP = OwnProps
    SP = StateProps
    DP = DispatchProps

  */

  declare type MapStateToProps<S, OP: Object, SP: Object> = (state: S, ownProps: OP) => SP;

  declare type MapDispatchToProps<A, OP: Object, DP: Object> = (dispatch: Dispatch<A>, ownProps: OP) => DP;

  declare type MergeProps<SP, DP: Object, OP: Object, P: Object> = (stateProps: SP, dispatchProps: DP, ownProps: OP) => P;

  declare type StatelessComponent<P> = (props: P) => ?React$Element<any>;

  declare class ConnectedComponent<OP, P, Def, St> extends React$Component<void, OP, void> {
    static WrappedComponent: Class<React$Component<Def, P, St>>;
    getWrappedInstance(): React$Component<Def, P, St>;
    static defaultProps: void;
    props: OP;
    state: void;
  }

  declare type ConnectedComponentClass<OP, P, Def, St> = Class<ConnectedComponent<OP, P, Def, St>>;

  declare type Connector<OP, P> = {
    (component: StatelessComponent<P>): ConnectedComponentClass<OP, P, void, void>;
    <Def, St>(component: Class<React$Component<Def, P, St>>): ConnectedComponentClass<OP, P, Def, St>;
  };

  declare class Provider<S, A> extends React$Component<void, { store: Store<S, A>, children?: any }, void> { }

  declare type ConnectOptions = {
    pure?: boolean,
    withRef?: boolean
  };

  declare type Null = null | void;

  declare type Connect1<A> = <OP>(...rest: Array<void>) => Connector<OP, $Supertype<{ dispatch: Dispatch<A> } & OP>>;

  declare type Connect2<A> = <OP>(
    mapStateToProps: Null,
    mapDispatchToProps: Null,
    mergeProps: Null,
    options: ConnectOptions
  ) => Connector<OP, $Supertype<{ dispatch: Dispatch<A> } & OP>>;

  declare type Connect3<S, A> = <OP, SP>(
    mapStateToProps: MapStateToProps<S, OP, SP>,
    mapDispatchToProps: Null,
    mergeProps: Null,
    options?: ConnectOptions
  ) => Connector<OP, $Supertype<SP & { dispatch: Dispatch<A> } & OP>>;

  declare type Connect4<A> = <OP, DP>(
    mapStateToProps: Null,
    mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
    mergeProps: Null,
    options?: ConnectOptions
  ) => Connector<OP, $Supertype<DP & OP>>;

  declare type Connect5<S, A> = <OP, SP, DP>(
    mapStateToProps: MapStateToProps<S, OP, SP>,
    mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
    mergeProps: Null,
    options?: ConnectOptions
  ) => Connector<OP, $Supertype<SP & DP & OP>>;

  declare type Connect6<S, A> = <OP, SP, DP, P>(
    mapStateToProps: MapStateToProps<S, OP, SP>,
    mapDispatchToProps: MapDispatchToProps<A, OP, DP>,
    mergeProps: MergeProps<SP, DP, OP, P>,
    options?: ConnectOptions
  ) => Connector<OP, P>;

  declare var connect: any;

}
```

Usage

```js
// @flow

import type { Connect5 } from 'react-redux'
import { connect as c } from 'react-redux'

type State = {
  ...
};

export type Action = ...;

type BoundConnect = Connect5<State, Action>;

export var connect: BoundConnect = c
```

Not sure how to type `declare var connect` without duplicating the declarations

```js
declare var connect: {
  <A, OP>(...rest: Array<void>): Connector<OP, $Supertype<{ dispatch: Dispatch<A> } & OP>>;
  <A, OP>(
    mapStateToProps: Null,
    mapDispatchToProps: Null,
    mergeProps: Null,
    options: ConnectOptions
  ): Connector<OP, $Supertype<{ dispatch: Dispatch<A> } & OP>>;
  ...etc...
};
```

aaronjensen:
@gcanti 

Could you just do:

```js
  declare type Connect<S, A> =
        Connect1<A> &
        Connect2<A> &
        Connect3<S, A> &
        Connect4<A> &
        Connect5<S, A> &
        Connect6<S, A>
```

and use `Connect<S, A>` where you need it? If so you can probably inline the `ConnectN` definitions as well and not have to define them explicitly.

gcanti:
yes but `Connect<S, A>` would be just a type, `connect` must be exported somehow, as a `var` or as a `function`, otherwise you get an error: "Named import from module `react-redux`. This module has no named export called `connect`" when you try to `import { connect as c } from 'react-redux'`

aaronjensen:
I see, because you can't declare a polymorphic var and `declare var connect: Connect<*, *>` would be too permissive? @jeffmo do you have a recommendation here?

maraujop:
Ok, my previous sample code:

```
export default connect(
  state => ({
    nextServicePk: state.Services.nextServicePk,
  }),
  dispatch => ({
    addService: data => dispatch(servicesActions.addService(data)),
  })
)(AddService)
```

without being modified raises:

```
function call
Function cannot be called on any member of intersection type
intersection: App/services/AddService.js:106
Member 1:
polymorphic type: function type: flow-typed/npm/react-redux_v4.x.x.js:50
Error:
arrow function: App/services/AddService.js:107
This type is incompatible with
undefined: flow-typed/npm/react-redux_v4.x.x.js:51
Member 2:
polymorphic type: function type: flow-typed/npm/react-redux_v4.x.x.js:54
Error:
arrow function: App/services/AddService.js:107
This type is incompatible with
union: null | undefined: flow-typed/npm/react-redux_v4.x.x.js:55
Member 3:
polymorphic type: function type: flow-typed/npm/react-redux_v4.x.x.js:61
Error:
arrow function: App/services/AddService.js:110
This type is incompatible with
union: null | undefined: flow-typed/npm/react-redux_v4.x.x.js:63
Member 4:
polymorphic type: function type: flow-typed/npm/react-redux_v4.x.x.js:68
Error:
arrow function: App/services/AddService.js:107
This type is incompatible with
union: null | undefined: flow-typed/npm/react-redux_v4.x.x.js:69
Member 5:
polymorphic type: function type: flow-typed/npm/react-redux_v4.x.x.js:75
Error:
arrow function: App/services/AddService.js:110
Could not decide which case to select
union type: flow-typed/npm/react-redux_v4.x.x.js:77
Member 6:
polymorphic type: function type: flow-typed/npm/react-redux_v4.x.x.js:82
Error:
arrow function: App/services/AddService.js:110
Could not decide which case to select
union type: flow-typed/npm/react-redux_v4.x.x.js:84
```

I'm using the new redux and react-redux flow type libdefs and it's my understanding this code should not raise any flow errors as it is.

Thanks, cheers
Miguel

gcanti:
> it's my understanding this code should not raise any flow errors as it is

@maraujop we are just discussing possible future enhancements, with the current libdefs you must still add type annotations to `state` and `dispatch`

maraujop:
Ok, thanks @gcanti again for your fast and clear answer :)

definitions/npm/redux_v3.x.x/flow_v0.28.x-v0.32.x/redux_v3.x.x.js
```diff
@@ -0,0 +1,54 @@
+declare module 'redux' {
+
+  /*
+
+    S = State
+    A = Action
+
+  */
+
+  declare type Dispatch<A: { type: $Subtype<string> }> = (action: A) => A;
```
pcardune: Looking at this, I'm guessing that making this work with [redux-thunk](https://github.com/gaearon/redux-thunk) which allows action creators to return functions is on the todo list? Or is there a special trick to get that to work?

aaronjensen: @pcardune I asked [the same question](https://github.com/reactjs/redux/pull/1887#issuecomment-241446828) and was pointed [here](https://github.com/agentcooper/react-redux-flow-example/blob/master/src/types/index.js#L20) for the solution.

kristian-puccio: I'd love to see an example of this in action if possible. Thanks

PR #323, opened 2016-09-21
flow-typed create-stub

gabelevi:
This introduces a `create-stub` command. `create-stub` takes one or more package names.

Examples:

```Bash
flow-typed create-stub foo
flow-typed create-stub foo@1.2.3
flow-typed create-stub foo@^1.2.3
flow-typed create-stub foo@^1.2.3 bar@~12.3.2
```

The stub for `foo` goes in `flow-typed/foo.js`

A generic stub just has

```
declare module "foo" {
  declare module.exports: any;
}
```

A specialized stub looks like

```
declare module "foo" {
  declare module.exports: any;
}

declare module "foo/fileA" {
  declare module.exports: any;
}

...

declare module "foo/fileZ" {
  declare module.exports: any;
}
```

When generating a stub for `foo`, we try to resolve `foo` in the node_modules directory to see what files are in `foo`. If we can find it, we generate a specialized stub. If we can't find it, we generate a generic stub.

If the version is not provided, we use the currently installed version. If it isn't currently installed, we look in `package.json`'s `dependencies` and then `devDependencies`.

Stubs are signed, which lets us detect if they've been modified. Unmodified stubs are overwritten. Modified stubs are not overwritten, unless you set the `--overwrite` flag.

Closes https://github.com/flowtype/flow-typed/issues/255


cli/src/lib/createStub.js
```diff
@@ -0,0 +1,178 @@
...
+    }
+  }
+
+  await fs.writeFile(filename, signCode(output, version));
+  return filename;
+}
+
+/**
+ * createStub("/path/to/root", "foo") will create a file
+ * /path/to/root/flow-typed/foo.js that contains a stub for the module foo.
```
jeffmo: This makes me think:

This will clearly work (any .js filename under `/flow-typed/` will technically do), but `flow-typed install foo@1.2.3` will create a file called `<PROJECT_ROOT>/flow-typed/npm/foo_v1.2.3.js` (conceptually: A copy of the libdef and its filename from the repo).

Maybe the stub that we create should have the same name as what would be created if the libdef were installed from the repo so that if an "official" libdef ever does come along, a hypothetical `flow-typed update` comand (which would basically be sugar for `flow-typed install --overwrite`) would notice that the stub could be replaced with something more substantial.

Thoughts?

cli/src/lib/createStub.js
```diff
@@ -0,0 +1,178 @@
...
+    // Look at the package.json for the installed module
+    if (pathToPackage != null) {
+      try {
+        version = (require: any)(path.join(pathToPackage, 'package.json')).version;
+      } catch (e) {}
+    }
+
+    // If that failed, try looking for a package.json in the root
+    if (version == null) {
+      (require: any)(path.join(projectRoot, 'package.json')).version;
```
jeffmo: I don't understand what this is for...can you explain?

It seems to be looking at the version of the outer project, but I'm not sure why that version is useful and moreover this doesn't actually seem to do anything with that version?

cli/src/lib/createStub.js
```diff
@@ -0,0 +1,178 @@
...
+    }
+  }
+
+  if (version == null) {
+    try {
+      const package_json = (require: any)(path.join(projectRoot, 'package.json'));
+      const deps = package_json.dependencies || {};
+      const devDeps = package_json.devDependencies || {};
+      version = deps[packageName] || devDeps[packageName] || null;
```
jeffmo: `flow-typed install` [will also check `peerDependencies` and `bundledDependencies`](https://github.com/flowtype/flow-typed/pull/289#discussion_r79907126). I suppose this should also check those too.

cli/src/lib/createStub.js
```diff
@@ -0,0 +1,178 @@
...
+      projectRoot,
+      packageName,
+      version,
+      overwrite,
+      files,
+    );
+    console.log(colors.green("✓\t%s@%s -> %s"), packageName, version, filename);
+    if (resolutionError) {
+      console.log(
+        colors.yellow("\t  Only created generic stub for '%s' (%s)"),
```
jeffmo: Can we add a little more context to this for the user? Not sure I'd understand what "generic stub" means exactly if I saw it in the output.

Maybe something like:

`\t  Unable to stub all files in '%s', so only created a stub for the main index.js (%s)`

cli/src/lib/findFlowRoot.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import {searchUpDirPath} from "./fileUtils";
+import {fs, path} from './node';
+
+// Find the project root
+export default async function (start: string): Promise<?string> {
```
jeffmo: So #289 just landed, which adds this to one of the commands: https://github.com/flowtype/flow-typed/blob/master/cli/src/commands/install.js#L247

If you rebase, would you mind updating that to just use this utility module?

cli/src/lib/createStub.js
gabelevi: My original though was to put something in `<PROJECT_ROOT>/flow-typed/stubs`. But I scrapped that idea based on the reasoning:

1. If the user never touches the stub, then it doesn't matter where it is.
2. If the user improves the stub, they'd probably want to move it to wherever they keep their custom libdefs. Which is probably not a stubs dir.

But I think your "if an official libdef shows up" argument is pretty convincing. I'll change dirs and add the version to the file name.

cli/src/lib/createStub.js
gabelevi: Err...this looks like dead copy-pasta. Reading the version and throwing it away doesn't do very much :P

cli/src/lib/createStub.js
gabelevi: Sure!

cli/src/lib/createStub.js
gabelevi: Sure. It's a hard thing to explain concisely!

cli/src/lib/createStub.js
jeffmo: Thinking more: If a libdef comes in that replaces a stub, there's a reasonable chance we'll have a de-duping problem even if we tried to name the stub correctly:

stub: `<PROJECT_ROOT>/flow-typed/npm/foo_v1.2.3.js`
libdef: `<PROJECT_ROOT>/flow-typed/npm/foo_v1.2.x.js`

So maybe a `/stubs/` directory actually would be better (just to help make it clear where all your stubs are to get a sense for what's stubbed and what's actually libdef-ed).

I could go either way here, so I'll leave it to you to decide

Issue #322, opened 2016-09-21
Remove the definition for enzyme@2.x.x? Or a way to specify which libdef to use when there are multiple?


cdchambers:
There are two enzyme definitions right now, 2.x.x and 2.3.x. The latter is a superset of the former. 

When installing through the flow-typed cli, any version that matches 2.3.x will also match 2.x.x and the cli installs 2.x.x by default.

Is there a way for me to force flow-typed to install 2.3.x? The version number approach doesn't appear to cut it...

wbeard:
I ran into this too. It looks like `filterLibDefs`'s sorting in `lib/libDefs` may not be correct? It ends up returning a list of libdefs by version in ascending order, rather than by descending. If it were descending, it'd indicate a more exact match.

Maybe one of the maintainers could help us understand why it works this way?

jeffmo:
@palange is knee-deep in this code right now over in https://github.com/flowtype/flow-typed/pull/289

palange:
#289 is merged, but didn't address this issue.

jeffmo:
FWYI: I just chatted with @palange about this and here's the updated plan to address this:

Have `flow-typed install` find the *lowest* satisfying version of the range specified on the CLI/package.json, then find the *highest* satisfying libdef version in flow-typed to install (and assume an implicit `^` in front of all libdef versions in flow-typed).

So to prove we're sane here, lets run through a few examples to test this procedure:
Assume we have 2 libdefs: One for 2.x.x and one for 2.3.x:

**If a project states that it uses `2.1.x`**

1. `2.1.0` is the lowest-satisfying version of `2.1.x`
1. Between `2.x.x` and `2.3.x`, `^2.x.x` is the *highest* satisfying libdef -- and so it will be installed

**If a project states that it uses `2.3.x`**

1. `2.3.0` is the lowest-satisfying version of `2.3.x`
1. Between `2.x.x` and `2.3.x`, `^2.3.x` is the *highest* satisfying libdef -- and so it will be installed

**If a project states that it uses `2.7.x`**

1. `2.7.0` is the lowest-satisfying version of `2.7.x`
1. Between `2.x.x` and `2.3.x`, `^2.3.x` is the *highest* satisfying libdef (it does satisfy since `3` is a minor version and we don't have anything closer to `7` to install) -- and so it will be installed.

---

[...]

PR #326, opened 2016-09-23
Improve libdef versioning in the cli

palange:
This PR addresses #322 

Versioning now supports the use cases @jeffmo describes. 
libdef versions will be treated as ^x.x.x semver ranges by default.

If multiple versions match, they will now be returned in the descending order.

### DIFF
--- a/cli/src/commands/install.js
+++ b/cli/src/commands/install.js
@@ -6,7 +6,7 @@ import {getCacheLibDefVersion, getCacheLibDefs, filterLibDefs} from "../lib/libD
 import type {LibDef} from "../lib/libDefs.js";
 import {fs, path} from '../lib/node.js';
 import semver from 'semver';
-import {emptyVersion, stringToVersion, versionToString} from "../lib/semver.js";
+import {emptyVersion, stringToVersion, versionToString, getLowerBound} from "../lib/semver.js";
 import type {Version} from "../lib/semver.js";
 import {getPackageDependencies, getPackageFlowBinSemver} from "../lib/npmHelper.js";
 import type {DepsMap} from "../lib/npmHelper.js";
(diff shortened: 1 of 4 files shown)


cli/src/lib/libDefs.js
```diff
@@ -570,18 +570,26 @@ function packageNameMatch(a: string, b: string): boolean {
 }
 
 function satisfiesSemver(pkgSemver: string, defVersion: string): boolean {
+  // the libdef version should treated as a semver prefixed
```
jeffmo: "should **be** treated"

cli/src/lib/libDefs.js
```diff
@@ -570,18 +570,26 @@ function packageNameMatch(a: string, b: string): boolean {
 }
 
 function satisfiesSemver(pkgSemver: string, defVersion: string): boolean {
```
jeffmo: I'm concerned about this function being used in contexts where the assumed-`^` isn't intended.
Maybe we should update the name of the function to indicate that it's a non-standard semver comparison?

Maybe something like `pkgVerSatisfiesLibDefVer`? (not tied to that name necessary)

cli/src/commands/install.js
```diff
@@ -118,23 +118,24 @@ export async function run(args: Args): Promise<number> {
...
+    const depVersion = depsMap[dep]; 
     let def = await findLibDef(
       dep,
-      depsMap[dep],
+      depVersion,
       flowVersion);
     if(def) {
       defs.push(def);
+      if(libdefNeedsUpdate(def.pkgVersionStr, depVersion)) {
+        console.log(`${def.pkgName}_${def.pkgVersionStr} may need an update to fully support ${dep} at ${depVersion}.`);
```
jeffmo: Suggestion:

`Installed ${def.pkgName}_${def.pkgVersionStr}, but you may want to create an updated libdef in flow-typed for your version (${depVersion}).`

jeffmo:
Agh, I did a bunch of cleanup work thinking this was merged but apparently not.
I'll resolve the merge conflicts and land this

jeffmo:
https://github.com/flowtype/flow-typed/commit/b0eee29fbee8dfbf03de2c27d17ea90d487d2ee2

PR #327, opened 2016-09-23
Adding RxJS operators let, letBind, and switch

inlinestyle:
Heyo, I've been using flow with RxJS for a few days, and noticed that a few method definitions are missing. This adds a couple that I've been wanting to use without ignoring in flow.

It's unfortunate that the typing for `.switch()` can't assert that `T` be an `Observable` (at least I assume this is the case, backed up by what I see with `.mergeAll()`), but I'd be happy to take suggestions to make that better.


inlinestyle:
@marudor (and other maintainers) sorry to bother you, do think this is mergeable?

marudor:
Sorry, the other maintainer are quite busy currently. So have I - going through them slowsly.

inlinestyle:
@marudor no problem! Thank you for spending time on this!

PR #330, opened 2016-09-24
make global typings in marked local

Kriegslustig:
I initially thought typings defined outside of the `module` block were local. Which they aren't. This PR moves all these typings into the module scope.


Kriegslustig:
Closing this, in order to make them global, to be able to extend the module definition without copying the module's typings.

PR #335, opened 2016-09-27
Add flowtype/no-weak-types lint rule in definitions

### LINKED ISSUE #49: [Test CI] Automatically comment on PRs that add `any` types
We want the libdefs in this repository to remain high quality. As such, we want to minimize `any` types as much as is reasonably possible. In some cases an `any` type of some kind (including things like `Object`, `Function`, etc) is unavoidable -- so we can't simply ban them altogether.

To minimize the number of unnecessary `any` types, it would be good to have Travis check for these and comment on any pull requests that add them -- pointing out that they should be avoided if at all possible.

Perhaps something like a `--strict` flag could be added to the `validate-defs` command that emits warnings on these kinds of things?

relekang:
Fixes #49

### DIFF
--- a/definitions/.eslintrc
+++ b/definitions/.eslintrc
@@ -1,9 +1,11 @@
 {
   "parser": "babel-eslint",
   "plugins": [
+    "flowtype",
     "flow-typed"
   ],
   "rules": {
+    "flowtype/no-weak-types": 1,
     "flow-typed/negative-tests": [2, 1]
   }
 }
--- a/definitions/package.json
+++ b/definitions/package.json
@@ -8,7 +8,8 @@
     "eslint": "^3.0.1"
   },
   "devDependencies": {
-    "eslint-plugin-flow-typed": "^1.0.0"
+    "eslint-plugin-flow-typed": "^1.0.0",
+    "eslint-plugin-flowtype": "^2.19.0"
   },
   "scripts": {
     "test": "eslint npm/**/test*"


relekang:
Should I add some kind of tool that only checks files that have been changed? (see comment in issue)

Are there any other rules that should be added while I am at it? `no-dupe-keys` looks like a good one to have.

jeffmo:
Does eslint have a notion of warning vs error? If so, maybe we could make this a warning and only block travis on errors...

`any`/`Object`/`Function` are certainly to be avoided whenever possible, but sometimes it's legitimately unreasonable to do so.

For example, some functions take an array of "anything" and return a concrete type. Using `Array<mixed>` in this case might seem preferable, but since arrays are mutable their type params are invariant -- and that means [it'd be an error](https://flowtype.org/try/#0GYVwdgxgLglg9mABMAFABwFyIIICdcCGAngDwC2MAHgKYAmAfAJSIDeAvgFABuBuiBWPIVJgQZAEbVc9RAF5EAbQCMAGgBMKgMwBdANwdUBRrqA) to pass anything but an array of exactly `mixed`-typed items in.

aaron-harvey:
FYI https://github.com/gajus/eslint-plugin-flowtype

relekang:
@jeffmo Thank you for the input. I updated the PR now so that it only raises warnings for weak types.

jeffmo:
This is awesome! It would be really cool if there were some way to surface these in the diff-view of PRs too...

PR #349, opened 2016-09-30
Fix `flow-typed install` for packages without a 'main'/index.js

### LINKED ISSUE #338: Support: flow-typed install resolve error
When running the `flow-typed install -o` it will run into an error from the resovle module:
`UNCAUGHT ERROR: Error: Cannot find module 'babel-runtime' from '/path/to/root-project-dir'`

What might be the best way to debug this issue? Is it coming from the cli or my project?

jeffmo:
Previously we'd use `resolve()` to find the path to an `npm install`ed
dependency. This usually works, but in some cases a package doesn't have
either an `index.js` file or a `main` entry in its package.json. `react-scripts`
is one such example.

This makes us resilient to this case when looking up the directory path for the installed package.

Fixes #338

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -11,7 +11,7 @@
     "lint": "eslint .",
     "prepublish": "mkdir -p dist; npm run test",
     "test": "npm run clean; npm run build; npm run test-quick",
-    "test-quick": "npm run lint && jest",
+    "test-quick": "jest && npm run lint",
     "watch": "mkdir -p dist; babel --source-maps --watch=./src --out-dir=./dist"
   },
   "repository": {
--- a/cli/src/lib/stubUtils.js
+++ b/cli/src/lib/stubUtils.js
@@ -10,7 +10,7 @@ import {getPackageJsonData} from './npmProjectUtils';
 import {getPackageJsonDependencies} from './npmProjectUtils';
 import {mkdirp} from './fileUtils';
 import {path} from './node';
-import resolveAsync from 'resolve';
+import {resolve} from "./npmProjectUtils";
 import {signCode} from './codeSign';
 import {verifySignedCode} from './codeSign';
 import {versionToString} from './semver';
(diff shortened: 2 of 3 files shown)


cli/package.json
```diff
@@ -11,7 +11,7 @@
     "lint": "eslint .",
     "prepublish": "mkdir -p dist; npm run test",
     "test": "npm run clean; npm run build; npm run test-quick",
-    "test-quick": "npm run lint && jest",
+    "test-quick": "npm run lint; jest",
```
gabelevi: Was this on purpose? Won't this pass even if lint fails?

cli/src/lib/stubUtils.js
```diff
@@ -27,16 +27,52 @@ export function glob(pattern: string, options: Object): Promise<Array<string>> {
...
+  let opts = {
+    ...options,
+    packageFilter: function(pkgJson, pkgFilePath) {
+      if (!pkgJson.main) {
+        pkgFiles.push(path.join(path.dirname(pkgFilePath), 'index.js'));
+        pkgJson.main = "index.js";
+      }
+      return pkgJson;
+    },
+    isFile: function(filePath, cb) {
```
gabelevi: Nit: Given that the return type of `isFile` doesn't matter, could you write it as an async function? Then you could just do

```JavaScript
try {
  const stat = await fs.stat(filePath);
  cb(null, stat.isFile() || stat.isFIFO());
} catch (e) {
  if (e.code === 'ENOENT') {
    cb(null, false);
  } else {
    cb(e);
  }
}
```

The less I see of the Promise API the happier I am :)

cli/package.json
jeffmo: Ah didn't mean to check this in, thanks.

It's annoying that `test-quick` won't run tests if there are lint errors, but the `&&` is necessary to make travis fail on lint errors (barring more complex bashisms)

PR #353, opened 2016-10-02
make libdefs for marked global under marked$*

Kriegslustig:
This  PR makes all libdefs for marked global under `marked$[...]`.

### DIFF
--- a/definitions/npm/marked_v0.3.x/flow_v0.28.x-/marked_v0.3.x.js
+++ b/definitions/npm/marked_v0.3.x/flow_v0.28.x-/marked_v0.3.x.js
@@ -1,10 +1,10 @@
-type AlignFlag = 'left' | 'right' | 'center'
+type marked$AlignFlag = 'left' | 'right' | 'center'
 
-type NodeCallback<T> = (e: ?Error, d: ?T) => void
+type marked$NodeCallback<T> = (e: ?Error, d: ?T) => void
 
-class Renderer {
-  constructor: (o?: MarkedOptions) => Renderer
-  options: MarkedOptions;
+class marked$Renderer {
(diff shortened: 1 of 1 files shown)


ryyppy:
I don't know `marked`, but is it really necessary to put everything in the global scope?
Why is it not in the `declare module` scope?

Kriegslustig:
You're probably right. It's not necessary in most use-cases. `marked` has a pretty flexible system for extension. For example `marked-man` is a package that just extends `marked` and adds the ability to format the output as `roff`. So for the `marked-man` libdef it would be necessary to access the `marked` libdef.

But since we can't define dependencies between libdefs, I had to copy over the `marked` libdef to the `marked-man` libdef. In `marked-man`'s libdefs I declared the `marked` libdefs in global-scope as well. Redeclaring the `marked` libdefs inside the `marked-man` module scope would technically be incorrect. Since users could then for example import `marked`'s `Lexer` directly from `marked-man`. Which would be incorrect.

Also; if a user wanted to write their own extension of `marked` though it might be useful to have them available globally. **I thought there was probably no harm to declaring them globally with the `$`-convention.**

Do you think we should leave them inside the module scope and have users copy them over?

ryyppy:
Yeah, for extensions it is reasonable... I wonder if there will be a possibility to do `import type from 'somepkg'` in declaration files soon? Then it would be a no-brainer to just declare reusable types in the `declare module` scope... anyways, other than that I don't see any problems with this

Kriegslustig:
For that to work, libdefs would need to be able to reference their dependencies (including the required version). You probably know this, but a solution is beeing worked on [here](https://github.com/flowtype/flow-typed/issues/16).

PR #360, opened 2016-10-05
RxJS: Remove committed signature

matthewwithanm:
Looks like this snuck through in 46017ce8a27c489ee96b7f69cf3670fb73fed97c.


jeffmo:
Oops!

PR #366, opened 2016-10-09
Jest 16 definitions

AgentME:
This adds definitions for Jest 16. I copied Jest 14's definitions and added in some missing things. (I think most of the definitions were based on Jasmine's, but Jest appears to have more functions than Jasmine did.) A bunch of missing expect() object methods were added, and I made some types more specific so that `{test,it}.{only,skip}` are typed now. (These APIs may or may not exist in older versions of Jest, dunno if anyone feels strongly about updating the older definitions too.) I re-ordered a few things not very far to match the order on [Jest's API page](https://facebook.github.io/jest/docs/api.html).

The first commit copies Jest 14's definitions as-is to new files for Jest 16, and then the second commit modifies them.

### DIFF
--- a/definitions/npm/jest_v16.x.x/flow_v0.33.x-/jest_v16.x.x.js
+++ b/definitions/npm/jest_v16.x.x/flow_v0.33.x-/jest_v16.x.x.js
@@ -1,0 +1,116 @@
+// flow-typed signature: e2130120dcdc34bf09ff82449b0d508c
+// flow-typed version: 230d7577ce/jest_v12.0.x/flow_>=v0.23.x
+
+type JestMockFn = {
+  (...args: Array<any>): any;
+  mock: {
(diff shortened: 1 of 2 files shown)


definitions/npm/jest_v16.x.x/flow_v0.33.x-/jest_v16.x.x.js
```diff
@@ -0,0 +1,121 @@
+// flow-typed signature: e2130120dcdc34bf09ff82449b0d508c
+// flow-typed version: 230d7577ce/jest_v12.0.x/flow_>=v0.23.x
```
AgentME: Do I need to change anything right here?

AgentME: Dropped it.

PR #373, opened 2016-10-12
Add definitions for pluralize 2.0 and 3.0

danez:
The definitions for 1.x, 2.0 and 3.0 are completely identical as the public API hasn't changed, but as far as I could see there is no way right now to inherit or have version ranges, right? 
Maybe this would be a nice addition, although not sure how often this case really happens.


marudor:
Problem about duplication tracked here:
https://github.com/flowtype/flow-typed/issues/374

PR #380, opened 2016-10-13
Add graceful-fs definition

palange:
graceful-fs is the node fs api, with the addition of the method `graceful-fs.gracefulify()`.


definitions/npm/graceful-fs_v4.1.x/flow_v0.28.x-/graceful-fs_v4.1.x.js
```diff
@@ -0,0 +1,197 @@
+interface ErrnoError extends Error {
```
marudor: Please move this into the module declaration.
Otherwise you're creating globals

PR #383, opened 2016-10-16
Add more underscore definitions

jdelStrother:
This adds a bunch of definitions which our app was relying on.  Also, I wasn't sure that the old definition for `map` was correct:

```diff
-  declare function map<T, U>(a: T[], iteratee: (val: T, n?: number)=>U): U[];
+  declare function map<T, U>(a: T[], iteratee: (val: T, n: number)=>U): U[];
```

The old version means you can't do `_.map(['a', 'b', 'c'], function(e, i) { return i < 2 })` because flow thinks `i` may be undefined.   I've changed it to allow that, which doesn't appear to any adverse effects on use of map where you don't include the index argument (eg `_.map(['a', 'b', 'c'], function(e) { return e != 'b' })`

### DIFF
--- a/definitions/npm/underscore_v1.x.x/flow_v0.13.x-/underscore_v1.x.x.js
+++ b/definitions/npm/underscore_v1.x.x/flow_v0.13.x-/underscore_v1.x.x.js
@@ -5,6 +5,10 @@ declare module "underscore" {
   declare function findWhere<T>(list: Array<T>, properties: {[key:string]: any}): ?T;
   declare function clone<T>(obj: T): T;
 
+  declare function findIndex<T>(list: T[], predicate: (val: T)=>boolean): number;
+  declare function indexOf<T>(list: T[], val: T): number;
+  declare function contains<T>(list: T[], val: T, fromIndex?: number): boolean;
+
   declare function isEqual(a: any, b: any): boolean;
   declare function range(a: number, b: number): Array<number>;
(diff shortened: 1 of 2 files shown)


definitions/npm/underscore_v1.x.x/flow_v0.13.x-/underscore_v1.x.x.js
```diff
@@ -13,39 +17,82 @@ declare module "underscore" {
...
+  declare function some<T>(a: Array<T>, pred: (val: T)=>boolean): boolean;
+  declare function any<T>(a: Array<T>, pred: (val: T)=>boolean): boolean;
+
+  declare function intersection<T>(...arrays: Array<Array<T>>): Array<T>;
+  declare function difference<T>(array: Array<T>, ...others: Array<Array<T>>): Array<T>;
 
   declare function initial<T>(a: Array<T>, n?: number): Array<T>;
   declare function rest<T>(a: Array<T>, index?: number): Array<T>;
 
+  declare function first<T>(a: Array<T>): T;
```
marudor: Swap the declare order. The one with more parameters first.
You can include the array check afterwards.

jdelStrother: Oh, cool.   I've updated with those changes.

PR #385, opened 2016-10-17
Add left-pad definition file

gabro:
As per https://twitter.com/lbljeffmo/status/788055941073731584


jeffmo:
Excellent work.

I guess we should now ship this then delete it for a day in protest, right?

gabro:
That's the plan: make flow a keystone of the global infrastructure and then threaten the world with a leftPad yielding any 😱

PR #388, opened 2016-10-18
[libdef] slug v 0.9.x

alexeygolev:


ckknight:
`SlugOptions` should be used as `$Shape<SlugOptions>` or similar because `slug('some string', { mode: 'rfc3986' });` should be completely valid, despite not having all properties of `SlugOptions`

jeffmo:
@ckknight: Mind sending a pull request?

ckknight:
Yeah, I can when I get home - `flow-typed run-tests` won't work behind my corporate network, apparently.

jeffmo:
Ah, yea `run-tests` needs to be able to download the flow binaries from github to run the tests.
(Maybe you need to set HTTP_PROXY? ....or just do it at home :) )

PR #391, opened 2016-10-18
fix ramda zipObj type

mgmcdermott:
Currently, the R.zipObj type is incorrect. If you look at the docs,
http://ramdajs.com/docs/#zipObj, you'll see that zipObj returns an
object given two arrays (hence the name zipObj). I have confirmed this
error in my own projects and thought I'd make a PR to fix it here.


jeffmo:
Nice catch!

To be sure, I've confirmed this to be true specifically in the v0.21 source here: https://github.com/ramda/ramda/blob/v0.21.0/src/zipObj.js#L16-L30

cc @alexeygolev as an FYI

jeffmo:
Can you add a test to verify this as well? Not sure why it was mistaken originally, but it would be good to be sure this fix doesn't regress.

mgmcdermott:
Sure thing. I updated the tests but I'm guessing you're thinking something along the lines of keeping one of the old tests and adding an `//$ExpectError`? Not familiar with testing flow libraries, I'm sure theres a more robust solution

jeffmo:
Oh...I think I somehow missed that you already updated/added tests.
Sorry about that, looks good. Thanks!

alexeygolev:
@mgmcdermott thank you!

PR #405, opened 2016-10-21
fix: ramda 0.21.x more cases of Maybe

mattgstevens:
Came across these issues when introducing flow-type into an existing project with ramda. Did not look to see if there are more instances where a Maybe is possible in the ramda docs.


mattgstevens:
Saw that the previous test run was failing with a 503 from npmjs.com

Rebased with the latest master and if the tests complete is there something else to do for the PR @marudor?

PR #408, opened 2016-10-23
Add `all` method to Express Application

LegNeato:
I'm not sure why, but this was needed to not throw an error in my code.


LegNeato:
Actually, this is now working.

PR #411, opened 2016-10-24
[WIP] Added _.create test, improved _.clone test for lodash

Leeds-eBooks:
Firstly, the `_.clone` test was not failing consistently, for two reasons: `$ExpectError` appears to cover the following block, not just the following line. If you have 2 lines that throw flow errors, and put `// $ExpectError` before the first line but not the second, both lines are allowed to pass the test. Secondly, the object passed to `_.clone()` was not typed strongly enough.

The main reason I have opened this PR is that `_.create` is not typed correctly. However, I can't find a way to do it correctly, so I'm opening this in the hope someone can help out.

``` javascript
var createdObj = _.create(createProto, {b: 2});
createdObj.b == 2;
```

This should pass, but flow errors out because the return type of `create<T>(prototype: T, properties?: Object)` is currently `$Supertype<T>`, making no mention of `properties` in the returned object.

Two points I can make with my limited understanding of OOP: Shouldn't the returned object be a _subtype_ of `T`, not a _supertype_? And how can we, in flow syntax, express the relationship of the returned object both to the prototype `T` and the properties `P`?

### DIFF
--- a/definitions/npm/lodash_v4.x.x/test_lodash-v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/test_lodash-v4.x.x.js
@@ -19,11 +19,15 @@ _.find({x: 1, y: 2}, { x: 3 });
 /**
  * _.clone
  */
-_.clone({a: 1}).a == 1;
+ var cloned: {a: 1} = _.clone({a: 1})
+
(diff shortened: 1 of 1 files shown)


marudor:

PR #424, opened 2016-10-29
 Improve Bluebird's interoperability with native promises

jdelStrother:
Bluebird accepts native promises pretty much everywhere, as far as I can tell.  For example:

``` js
Bluebird.resolve(Promise.resolve('hello')).then(function(a: string) { })
```

ought to behave ok, but flow-typed currently expects 'a' to be `Promise<string>`, rather than a simple `string`.

I've attempted to fix that in this PR - WDYT?


jdelStrother:
... one other thought - ideally, I suspect all Bluebird functions would accept native promise and implicitly unwrap them - ie:

```js
Bluebird.all([1, Bluebird.resolve(1), Promise.resolve(1)]).then(function(r: Array<number>) {
  
})
```

ought to typecheck ok, but I couldn't persuade that to work.  Any suggestions?

PR #426, opened 2016-10-29
Use node v6.1 and v7.0 in travis

douglasduteil:
Node 7 has arrived!
https://nodejs.org/en/blog/release/v7.0.0/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: false
 language: node_js
 node_js:
-  - '5.11'
   - '6.1'
+  - '7.0'
 compiler: clang-3.6
 env:
   - CXX=clang-3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 sudo: false
 language: node_js
 node_js:
-  - '6.1'
-  - '7.0'
+  - '6'
+  - '7'
 compiler: clang-3.6
 env:
   - CXX=clang-3.6
--- a/cli/src/lib/fileUtils.js
+++ b/cli/src/lib/fileUtils.js
@@ -50,7 +50,15 @@ export async function recursiveRmdir(dirPath: string): Promise<void> {
       await fs.unlink(itemPath);
     } else {
       await recursiveRmdir(itemPath);
-      fs.rmdir(itemPath);
+      await fs.rmdir(itemPath).catch((err) => {
+        if(err.code === 'ENOENT') {
+          // Ignore ENOENT error
+          // it's okay if the files are already removed
+          return;
+        }
+
+        throw err;
+      });
     }
   }));
   return fs.rmdir(dirPath);


douglasduteil:
Might be better to always test over the latest LTS version (6.9.1)

jeffmo:
I'm not really opposed to this, but does going from 6 -> 7 do anything for the test runner?

douglasduteil:
> Does going from 6 -> 7 do anything specific for the test runner?

Yes. On Node 7.0.0, running the `./run_def_tests.sh` script is emitting `UnhandledPromiseRejectionWarning` and `DeprecationWarning` while on the last line `node dist/cli.js run-tests "$@"`

Extract:
```
(node:4618) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: ENOENT: no such file or directory, rmdir '/home/travis/build/flowtype/flow-typed/cli/.flow-bins-cache/TMP-flow-v0.33.0/flow'
(node:4618) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
[Reproduced on Travis](https://travis-ci.org/flowtype/flow-typed/jobs/171656130#L1190 )

Following [7.0.0 node changelog]
(https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V7.md#7.0.0) I guess that there is just an error emitted during the clean up process in [fileUtils.js#L46-L56)](https://github.com/flowtype/flow-typed/blob/3ada18a151b4a07e5b501ff1a6dcc1f727ccec2a/cli/src/lib/fileUtils.js#L46-L56)

In the diff below, I kind of fix it but I'm not convince by this solution.... :-1: 

```diff
export async function recursiveRmdir(dirPath: string): Promise<void> {
  // [...]
  await P.all(dirItems.map(async (itemName, idx) => {
    const itemStat = dirItemStats[idx];
    const itemPath = path.join(dirPath, itemName);
    if (itemStat.isFile()) {
      await fs.unlink(itemPath);
    } else {
      await recursiveRmdir(itemPath);
-      fs.rmdir(itemPath);
+      await fs.rmdir(itemPath).catch(() => {
+        // ignore already removed itemPath errors
+      });
    }
  }));
  return fs.rmdir(dirPath);
}
```

jeffmo:
Ohhh, nice catch! Let's fix this before we bump to node7, then.

On your diff, is it possible to discriminate `ENOENT` errors? Seems we'd only want to catch and ignore those (if I'm reading your fix right?).

douglasduteil:
hum... K.

douglasduteil:
I'm pushing something here. Wait for it :)

douglasduteil:
done @jeffmo. What you think ?

douglasduteil:
Isn't it strange that the file we want to remove is already removed at that moment BTW ?

jeffmo:
> Isn't it strange that the file we want to remove is already removed at that moment BTW ?

That's a great point -- and in fact that second `rmdir` appears to be unnecessary!
It's redundant with the `rmdir` on the last line of `recursiveRmdir`.

marudor:
If we merge this I would like to use LTS instead of 6.1
Basically we should always test against LTS and newest.
That would be 6.9.1 and 7.0 for now.

douglasduteil:
:+1:
Sorry guys I didn't took the time to it

PR #446, opened 2016-11-15
Update rxjs 5 definitions

hansonw:
In Flow 0.34, Object keys are now invariant by default. This makes `rxjs$Observer` incompatible with `rxjs$PartialObserver`, which is pretty annoying when subscribing observers.

The fix is to make the properties covariant; however we now have to make the defs Flow 0.34+ only.

This also adds `distinctUntilKeyChanged` and amends `using` to allow subtypes of `rxjs$ISubscription`.

cc @matthewwithanm


marudor:
I support the breaking change to 0.34+
But I'm unsure how we should handle this.
If we copy it to have a 0.29 and a 0.34+ version we wouldn't break backwards compat. But this would introduce a lot of old, probably unmaintained, versions.

cc @jeffmo

nmote:
I pinged @jeffmo and he said to copy the old ones out. For tests that pass in the new versions and fail in the old versions, we can put the test files inside of the `flow_vwhatever` folders.

hansonw:
@nmote split the defs and added the covariance-dependent test in the `flow_v0.34.x-` folder - does this look OK to you?

nmote:
Yep, LGTM

PR #457, opened 2016-11-17
Ramda: Evolve should create an object with the same type as its input

ajhyndman:
http://ramdajs.com/docs/#evolve

According to the Ramda documentation, evolve should accept an object of functions which return the same type as their input.  This means that we actually have stronger guarantees about the shape of the return value.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.28.x-v0.30.x/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.28.x-v0.30.x/ramda_v0.x.x.js
@@ -552,8 +552,8 @@ declare module ramda {
 
   // TODO over
 
-  declare function path<V,A:NestedObject<V>>(p: Array<string>, ...rest: Array<void>): (o: A) => ?V;
-  declare function path<V,A:NestedObject<V>>(p: Array<string>, o: A): ?V;
+  declare function path<V,A:?NestedObject<V>>(p: Array<string>, ...rest: Array<void>): (o: A) => ?V;
+  declare function path<V,A:?NestedObject<V>>(p: Array<string>, o: A): ?V;
 
   declare function pathOr<T,V,A:NestedObject<V>>(or: T, ...rest: Array<void>):
   ((p: Array<string>, ...rest: Array<void>) => (o: ?A) => V|T)
@@ -720,7 +720,7 @@ declare module ramda {
     f2: (...args: Array<any>) => C
   ): (...args: Array<A>) => B|C;
 
-  declare function isEmpty(x:Array<any>|Object|string): boolean;
+  declare function isEmpty(x:?Array<any>|Object|string): boolean;
 
   declare function not(x:boolean): boolean;
 
--- a/definitions/npm/ramda_v0.x.x/flow_v0.28.x-v0.30.x/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.28.x-v0.30.x/ramda_v0.x.x.js
@@ -499,8 +499,8 @@ declare module ramda {
   declare function dissocPath<T>(key: Array<string>, val:T, ...args: Array<void>): (src: {[k:string]:T}) => {[k:string]:T};
   declare function dissocPath<T>(key: Array<string>, val:T, src: {[k:string]:T}): {[k:string]:T};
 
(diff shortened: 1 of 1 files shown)


mwalkerwells:
@ajhyndman Did you get any follow-up on this?

ajhyndman:
I haven't heard anything.  I'd still like to see this PR go through!

stevenyap:
@ajhyndman I did not see from http://ramdajs.com/docs/#evolve that the type should be enforced. It looks like we are free to apply transformation to change the type (I do use it to coerce types eg, from a submitted form string into integers)

definitions/npm/ramda_v0.21.x/flow_v0.28.x-v0.30.x/ramda_v0.21.x.js
```diff
@@ -500,8 +500,8 @@ declare module ramda {
   declare function dissocPath<T>(key: Array<string>, val:T, ...args: Array<void>): (src: {[k:string]:T}) => {[k:string]:T};
   declare function dissocPath<T>(key: Array<string>, val:T, src: {[k:string]:T}): {[k:string]:T};
 
-  declare function evolve<V,R>(fn: NestedObject<(x:any) => R>, ...rest: Array<void>): (src: NestedObject<any>) => NestedObject<R>;
-  declare function evolve<V,R>(fn: NestedObject<(x:any) => R>, src: NestedObject<any>): NestedObject<R>;
+  declare function evolve<R: NestedObject<any>>(fn: NestedObject<UnarySameTypeFn<any>>, ...rest: Array<void>): (src: R) => R;
```
mwalkerwells: These two lines are not accurate. In Flow, polymorphic types represent exact values, not exact types. This means that ```src``` & the return type would have to be referentially transparent, which they are not, since they are being modified by ```fn``` argument.

mwalkerwells:
My understanding is that the ```evolve ``` fn does this opposite of what your title indicates: "Evolve should create an object with the same type as its input." This declares the transformation of an object, thus should result in a different type.

definitions/npm/ramda_v0.21.x/flow_v0.28.x-v0.30.x/ramda_v0.21.x.js
```diff
@@ -553,8 +553,8 @@ declare module ramda {
 
   // TODO over
 
-  declare function path<V,A:?NestedObject<V>>(p: Array<string>, ...rest: Array<void>): (o: A) => ?V;
```
mwalkerwells: Can these functions not handle nullable values?

ajhyndman:
@mwalkerwells You are correct in pointing out that nothing in Ramda's code prevents passing an argument to the evolve function that breaks the assumption that the evolved object will have the same type as the input.

I guess I feel that adopting the stronger assumption makes evolve's type signature more useful than the weaker.  If we adopt the strong assumption, in some cases, it will be accurate and precise, and in some cases we will need to abuse casting to any, or `// $Flowissue` comments.  If we adopt the weak assumption, then we will lose type information across the boundary of the `evolve` invocation, and need to use casting or `// $FlowIssue` comments at every invocation to restore the type guaranteed.

I have found the former to be more useful in practice.

definitions/npm/ramda_v0.21.x/flow_v0.28.x-v0.30.x/ramda_v0.21.x.js
ajhyndman: That's absolutely not true.

I've used this definition in my own code.  This definition will pass a typecheck even if the input and result are not referentially transparent.

There is one test already that demonstrates exactly this use-case, and which you can see passed CI.

https://github.com/ajhyndman/flow-typed/blob/ramda-evolve-definition/definitions/npm/ramda_v0.21.x/flow_v0.28.x-v0.30.x/test_ramda_v0.21.x_object.js#L30

definitions/npm/ramda_v0.21.x/flow_v0.28.x-v0.30.x/ramda_v0.21.x.js
ajhyndman: Hmm.  They definitely can handle nullable values.  It's possible that I just accidentally copied these lines from an older version of the typedef.  Sorry about that!

mwalkerwells:
@ajhyndman Can you provide an example?

GAntoine:
Closing as the current implementation of `evolve` needs to be fixed before it can be changed.

PR #460, opened 2016-11-18
Remove recursive definition

allenylzhou:
Wrapped react component this.props was unable to infer types for properties defined in SP


cjhoward92:
Is there a reason this was not removed in v5.x.x?

PR #462, opened 2016-11-18
Update `table` version range.

### LINKED ISSUE #456: `table` version.
Apologizes for the question, but why is `table` [locked to a specific version](https://github.com/flowtype/flow-typed/blob/master/cli/package.json#L61) while most of the other dependencies are not?  It is not a big deal, but it is preventing a clean dedupe with `eslint`.

wtgtybhertgeghgtwtg:
Fixes #456


jeffmo:
Thanks!

wtgtybhertgeghgtwtg:
Thank you for your time.

PR #464, opened 2016-11-19
Adding namedspaced globals for express with backward compatibility

mwalkerwells:
Imports are not currently supported in declaration files. This is helpful for typing other classes that extend Express like the Webpack Development Server.

**Context:** https://github.com/facebook/flow/issues/2847


mwalkerwells:
Actually, do these need to be namespaced with npm$express$ ?

marudor:
npm$express$ and express$ should both be fine.
As long as  its unique. And express$ should be as unique as npm$express$

PR #466, opened 2016-11-20
knex

marudor:
This one was sitting in my work repo for quite a while.
It's missing some tests - but is tested on our live code. It should be pretty good.

knexjs.org


dwillie:
I've just been using this I've noticed there doesn't seem to be any support for schema changes, is that correct? (As in http://knexjs.org/#Schema)

marudor:
@dwillie correct.
We're not using it at work so I probably just forgot about it.
Feel free to send a PR.

dwillie:
Thanks 👍 

I'm pretty new to Flow, but I'll have a look when I get a chance

PR #467, opened 2016-11-20
Create react-scripts annotation from jest

rricard:
react-scripts don't expose anything to the end user except the jest api.
That's why we're just copying the jest annotations.

Coming from here: https://github.com/facebookincubator/create-react-app/issues/72


gaearon:
For context, it is part of Create React App default experience and ideally should work by default.

thejameskyle:
We could probably improve this using libdefs imports once they are supported

PR #473, opened 2016-11-22
Fix (non-static) Observable.concat type

hansonw:
Concatting an `Observable<T>` with an `Observable<U>` should result in an `Observable<T | U>`.
Currently this can expand the type of the first observable if it's not fixed, causing spurious errors down the line.


hansonw:
@nmote I tested this in the Nuclide codebase and this caused no additional Flow errors (although it would have prevented the one seen in tonight's temporary trunk breakage)

PR #475, opened 2016-11-22
define CombinedReducer type for redux_v3.x.x

nickarora:
Updates to redux flow based on conversation [here](https://github.com/reactjs/redux/pull/2115)


steida:
I just spend two hours trying https://flow.org/en/docs/frameworks/redux/#toc-typing-redux-state-immutability but "Covariant property `xyz` incompatible with invariant use in object type" ruined my day.

This PR fixes it. I hope it will be merged soon.

PR #480, opened 2016-11-23
Add ESDoc support to Jest

orta:
Adds comments from the documentation inline into the Flow annotations.

A rough overview of using it:

As this doesn't change any of the flow structure, I don't think there's anything to change WRT the tests.

/cc @cpojer @dmitriiabramov


cpojer:
wow this is so great.

orta:
ah yeah, looks like its not green - let's fix this up, was just missing these autocompletions in another project

orta:
Alright, all green 👍

PR #484, opened 2016-11-24
Chai's assert style api

petemill:
This is to add definitons for `chai.assert` and associated functions (http://chaijs.com/api/assert/)


ryyppy:
Nice, thanks for the contribution! (will merge this, since it doesn't interfere with existing typings)

PR #485, opened 2016-11-24
express: support req.body

petemill:
`req.body` can be an object with attributes or undefined.

Previously, this was marked as `mixed`, resulting in flow throwing an error when attempting to access a property, e.g. `req.body.Username`.

According to [the express docs](http://expressjs.com/en/api.html#req.body), the `body` property can be undefined, or an object.

With passing tests.

### DIFF
--- a/definitions/npm/express_v4.x.x/flow_v0.25.x-/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_v0.25.x-/express_v4.x.x.js
@@ -14,7 +14,7 @@ declare class express$RequestResponseBase {
 
 declare class express$Request extends http$IncomingMessage mixins express$RequestResponseBase {
   baseUrl: string;
-  body?: {[key: string]: string};
+  body?: string | Buffer | {[key: string]: mixed};
   cookies: {[cookie: string]: string};
   fresh: boolean;
   hostname: boolean;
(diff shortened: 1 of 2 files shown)


petemill:
@kamek-pf fair enough, I misunderstood the *mixed* type and how to access it when it's an object. According to the docs, `body` can be either an *object*, a *string*, a *Buffer*, or non-existant. I've now changed the declaration to account for that:

`body?: string | Buffer | {[key: string]: mixed}`

If anything, I've added req.body tests, whether we accept this stricter body definition or not. Those should, at least, be useful.

The one issue I see is when developers know they are calling [`bodyParser.json()`](https://github.com/expressjs/body-parser/blob/master/README.md#bodyparserjsonoptions) higher up in the middleware cycle, and therefore don't need to type check req.body as an object each time. In those cases, perhaps developers should be encouraged to extend `express$Request` and define `body` as `{[key: string]: mixed}`. 
We could perhaps provide that helper type here for developers to use when defining the types of params for their middleware functions.

petemill:
Thinking about this more, I realize `express$Request.body` should not exist and should actually be implemented by a *body-parser* module definition. Do you agree @kamek-pf ? 

I've tried to implement this, but some issues I think are due to the flow-typed test environment are described in #488

kamek-pf:
Yup, I guess if you want to do things by the book, `body` should be undefined unless the `body-parser` middleware is injected.
Should probably ping some of the original authors of the express libdef and see what they think about it.

definitions/npm/express_v4.x.x/flow_v0.25.x-/express_v4.x.x.js
```diff
@@ -14,7 +14,7 @@ declare class express$RequestResponseBase {
 
 declare class express$Request extends http$IncomingMessage mixins express$RequestResponseBase {
```
bassettsj: Why not make it parameterized instead?

petemill: how do you mean?

marudor:

PR #486, opened 2016-11-25
Fix missing types in Lodash

ggregoire:
* Fix `_.matchesProperty` iteratee shorthand (e.g. `_.find(users, ['active', false])`)

* Fix `_.property` iteratee shorthand (e.g. `_.map(users, 'age')`)


marudor:
Could you add some more tests with your fix?

ggregoire:
@marudor Done. :)

PR #487, opened 2016-11-25
Add cpy@5.x.x

thejameskyle:


ryyppy:
Hmm, this test requires a negative test... can we still just merge this without breaking other things?

ryyppy:
Oh snap, the file name format has changed since then... there shouldn't be any >= in it for windows compat

ryyppy:
Okay, I will take it on from here, thanks a bunch :-)

PR #489, opened 2016-11-28
rxjs: add bindCallback funtion

mitrandir77:
as documented in
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-bindCallback


nmote:
Could you add a test for this?

superherointj:
Hello. Consider:

  static bindCallback(
      f: (...args: any) => void, // Change to this part.
      selector?: () => T,
      scheduler?: rxjs$SchedulerClass
      ) : (...args: any) => rxjs$Observable<T>;

When converting function to Observable. Setting return to void is desirable because the return won't be returning and will be lost. Makes sense?

marudor:

PR #502, opened 2016-12-01
Definition for bser and node-int64

danez:
This is the definition for the package `bser` (https://github.com/facebook/watchman/tree/master/node/bser)

I have two questions:

* Is it valid that one of the objects extends `EventEmitter`, because flowtype.org/try says `"Could not resolve name EventEmitter"`?
* How would I use types from other definitions? Is that already possible? Because `bser` also accepts and returns instances of `Int64` (npm.im/node-int64), so assuming a type definition would exists for it how would I reference it?

Tests are still missing, but wanted to ask the questions first.

### DIFF
--- a/definitions/npm/bser_v1.x.x/flow_v0.28.x-/bser_v1.x.x.js
+++ b/definitions/npm/bser_v1.x.x/flow_v0.28.x-/bser_v1.x.x.js
@@ -1,0 +1,59 @@
+declare module 'bser' {
+  /*
+  TODO: Support importing of other libdefs
+  import typeof Int64 from 'node-int64'; 
+  */
+  declare type BserValidTypes =
+    | number
+    | string
+    | boolean
+    | null
+    /* |Int64 */
+    | Array<BserValidTypes>
+    | { [key: string]: BserValidTypes };
+
+
+  declare export class Accumulator {
+    buf: Buffer,
+    readOffset: number,
+    writeOffset: number,
+    constructor(initsize?: number): Accumulator,
+    writeAvail(): number,
+    readAvail(): number,
+    reserve(size: number): void,
+    append(buf: string|Buffer): void,
+    assertReadableSize(size: number): void,
+    peekString(size: number): string,
+    readString(size: number): string,
+    peekInt(size: number): number/* |Int64 */,
+    readInt(size: number): number/* |Int64 */,
+    peekDouble(): number,
+    readDouble(): number,
+    readAdvance(size: number): void,
+    writeByte(value: number): void,
+    writeInt(value: number, size: number): void,
+    writeDouble(value: number): void,
+    
+  }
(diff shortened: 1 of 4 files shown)


ryyppy:
Hey. Awesome stuff!

To your questions:

* Try to use `events$EventEmitter`, since this is exposed by [flow's node libdefs](https://github.com/facebook/flow/blob/master/lib/node.js#L640)
* For the second question, there will be support for `import type {...} from ...` syntax inside a `declare module` scope, but there is no `flow-typed` support for resolving these dependencies yet :-(

danez:
I updated the PR to also include tests and the definition+tests for `node-int64`.

I for now left the connection between those two libdefs out, as the `import type` cannot be used yet.

definitions/npm/bser_v1.x.x/flow_v0.22.x-/bser_v1.x.x.js
```diff
@@ -0,0 +1,58 @@
+declare type $npm$bser$BserTypes =
```
ryyppy: Awesome stuff! Just one question: Why does this type needs to be global? I can't see any other file / code part except the `bser` module using it...?

danez: Oh, it doesn't need to be I just copied this pattern from another def.
So I just move it inside the `declare module` as is? Or should I rename it to declare type `BserType`? But wouldn't any `declare type` inside a `declare module` define a named export?
I think that is why I initially put it on the top scope, as I thought prefixing it with $npm means it is private and does not count towards the named exports.

ryyppy: As soon as you declare at least one `declare` statement as `declare export ...`, the module will be treated as an ES6 module and will not automatically export everything... so if you put a `declare type BserTypes = ...` inside the `declare module` scope and export types etc. as `declare export class Acumulator` etc., it should work.

Example:
```
declare module 'bser' {
   declare type BserTypes = {...};

   declare export class Accumulator { ... };
   declare export function loadFromBuffer(......);
}
```
And so on... the `declare export` syntax is only supported by more modern versions of flow (I guess >= 0.30 or so... you need to try that).

Hope this helps

ryyppy: Also, maybe have a look at my slides for writing libdefs: 
https://github.com/ryyppy/flow-libdef-talk/blob/master/dist/flow-libdef-talk.pdf

danez: Thanks, the slides help a lot.

I changed it.

ryyppy:
Awesome job! Thanks :-)

PR #505, opened 2016-12-01
Add libdef for react-i18next

echenley:
Adds libdef for react-i18next. Should be fairly robust, though there are some holes (not checking the i18n object). Special attention has been given to the `translate` HOC, which was modeled after `connect` from react-redux. I am new to flow, so please let me know if there is anything I can improve.

Usage:

```js
// @flow
import { translate } from 'react-i18next'
import type { Translator, TFunction } from 'react-i18next'

type OwnProps = { s: string }
type Props = OwnProps & { t: TFunction }

const Comp = ({ t }: Props) => <div s={}>{ t('whatever') }</div>

const translator: Translator<OwnProps, Props> = translate('whatever')

const WrappedComp = translator(Comp)

const OtherComp = () => <Comp s={ 2 } /> // TypeError
```

### DIFF
--- a/definitions/npm/react-i18next_v1.10.x/flow_v0.36.x-/react-i18next_v1.10.x.js
+++ b/definitions/npm/react-i18next_v1.10.x/flow_v0.36.x-/react-i18next_v1.10.x.js
@@ -1,0 +1,32 @@
+import type { TFunction } from 'i18next'
+
+declare module 'react-i18next' {
+  declare type TFunction = TFunction;
+  declare type Locales = string | Array<string>;
+
+  declare type StatelessComponent<P> = (props: P) => ?React$Element<any>;
+
+  declare type Comp<P> = StatelessComponent<P> | Class<React$Component<void, P, void>>;
+  declare type TranslatedComponent<OP> = Class<React$Component<void, $Diff<OP, { t: TFunction }>, void>>;
+  declare type Translator<OP, P> = (component: Comp<P>) => TranslatedComponent<OP>;
+
+  declare function translate<OP, P>(locales: Locales): Translator<OP, P>;
+
+  declare type NamespacesProps = {
+    components: Array<Comp<*>>,
+    i18n: { loadNamespaces: Function },
+  };
+
+  declare function loadNamespaces(props: NamespacesProps): Promise<void>;
+
+  declare type ProviderProps = { i18n: Object, children: React$Element<any> };
+
+  declare var I18nextProvider: Class<React$Component<void, ProviderProps, void>>;
+
+  declare type InterpolateProps = {
+    children?: React$Element<any>,
+    className?: string,
+  };
+
+  declare var Interpolate: Class<React$Component<void, InterpolateProps, void>>;
+}
--- a/definitions/npm/react-i18next_v1.10.x/flow_v0.36.x-/react-i18next_v1.10.x.js
+++ b/definitions/npm/react-i18next_v1.10.x/flow_v0.36.x-/react-i18next_v1.10.x.js
@@ -1,7 +1,5 @@
-import type { TFunction } from 'i18next'
-
 declare module 'react-i18next' {
-  declare type TFunction = TFunction;
+  declare type TFunction = (key?: ?string, data?: ?Object) => string;
   declare type Locales = string | Array<string>;
 
   declare type StatelessComponent<P> = (props: P) => ?React$Element<any>;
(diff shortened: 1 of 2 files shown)


definitions/npm/react-i18next_v1.10.x/flow_v0.36.x-/react-i18next_v1.10.x.js
```diff
@@ -0,0 +1,35 @@
...
+  declare type TFunction = TFunction;
+  declare type Locales = string | Array<string>;
+
+  declare type StatelessComponent<P> = (props: P) => ?React$Element<any>;
+
+  declare type Comp<P> = StatelessComponent<P> | Class<React$Component<*, P, *>>;
+
+  declare type Translator<OP, P> = {
+    (component: StatelessComponent<P>): Class<React$Component<void, OP, void>>;
+    <Def, St>(component: Class<React$Component<Def, P, St>>): Class<React$Component<Def, OP, St>>;
```
echenley: I admittedly don't understand this syntax. It appears to define a function (callable object) that can be called in two different ways, the first taking a stateless component, the other a class component. The `<Def, St>` part is particularly strange. I borrowed it from https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-redux_v4.x.x/flow_v0.30.x-/react-redux_v4.x.x.js#L35, but couldn't find any other examples in the docs or elsewhere. If anyone could offer an explanation, I would greatly appreciate it.

ryyppy: Are you confused by the `<Def, St>` syntax? These are generic placeholders... just imagine those as type variables, which will be substituted in specific calls... you don't really care about the concrete types, you just want them to be the same type consistently for this function call....

e.g.
```

// Generic N -> we don't care about the type,
// we just want the `test` parameter and the return type
// to be the same type
function foo<N>(test: N): N {
   return test;
}

// test => Number... return type => Number... t => Number... works
const t: number = foo(1);

// N => Number... return type => Number.... t => String ... doesn't work
const t: string = foo(1);
```

does this help?

echenley: No, not the generics syntax in general, just this particular instance where `<Def, St>` comes before the function parameters.

I guess there's two questions:

1. Is this:

```js
type Foo<A> = <B>(bar: Bar<A, B>) => Bar<A, B>
```

Equivalent to this?

```js
type Foo<A> = {
  <B>(bar: Bar<A, B>): Bar<A, B>
}
```

2. In the above, is it true that `<B>` correspond to the type of the `bar` parameter?

```js
const foo: Foo<A> = blah
const bar: Bar<A, B> = blah
const baz = foo(bar) // baz has type Bar<A, B>
```

So my understanding would be that this syntax is useful when the type of the return value of a function depends on the type of the input value?

ryyppy: By intuition, I think that (1) they are not equivalent... by putting the generic definitions in front of the function paranthesis, it will always redo the check for each function call, so you cannot lock down the type to a specific type in forefront (unless you do some subtyping stuff).

For (2), yes <B> would be of the same type as bar (Bar<X, Y> -> A, B are only in the context of Bar, so I call them X, Y)... so <Bar<X, Y>>(bar: Bar<A, Bar<X, Y>>): Bar<A, Bar<X,Y>>

It's hard to have discussions on generics without sounding very confusing :-)

definitions/npm/react-i18next_v1.10.x/flow_v0.36.x-/react-i18next_v1.10.x.js
```diff
@@ -0,0 +1,35 @@
+import type { TFunction } from 'i18next'
```
echenley: Good question. I believe it does work, but I need to add a test for it.

definitions/npm/react-i18next_v1.10.x/flow_v0.36.x-/react-i18next_v1.10.x.js
echenley: Haha, for sure. I think that makes sense, thanks! Flow can get pretty "mathy" and abstract for someone without a CS background (at least at first), so I appreciate the help.

definitions/npm/react-i18next_v1.10.x/flow_v0.36.x-/react-i18next_v1.10.x.js
ryyppy: if tests are in place, I think we would be ready to merge... I saw a lot of imports in other libdefs now, so maybe I am just wrong here :-)

If you still need any help, just ping me

echenley: @ryyppy Updated to remove the type import from i18next. Apparently this was silently not working. Added a couple tests for it. Should be good to go unless you have any other suggestions.

ryyppy:
Nice, ready as you are? Everything done here?

echenley:
Ready to go!

PR #507, opened 2016-12-02
Remove common eslint warnings from jest.js

altano:
* There are many common eslint warnings that trigger on jest.js, this diff removes them.
* There are a few errors in which methods allow chaining of the jest object and this diff addresses those errors by changing the return type from void to JestObjectType.

### DIFF
--- a/definitions/npm/jest_v17.x.x/flow_v0.33.x-/jest_v17.x.x.js
+++ b/definitions/npm/jest_v17.x.x/flow_v0.33.x-/jest_v17.x.x.js
@@ -30,7 +30,7 @@ type JestMockFn = {
   /**
    * Accepts a function that should be used as the implementation of the mock.
    * The mock itself will still record all calls that go into and instances
-   * that come from itself – the only difference is that the implementation
+   * that come from itself -- the only difference is that the implementation
    * will also be executed when the mock is called.
    */
   mockImplementation(fn: Function): JestMockFn,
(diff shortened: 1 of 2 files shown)


ryyppy:
Hey! Thanks for the PR... I don't really understand what you mean by "common eslint warnings"... you mean in this repository? Since we don't enforce any strict eslint rules here?

altano:
Right, not this repository (since there is no eslint configured) but the one I'm importing it into. These are pretty straightforward warnings and fixing them makes things much more readable, so I figured no one would mind.


________________________________

ryyppy:
The thing is, we don't want to force any eslint rules onto the user... what might work for you, not necessarily will work for others... maybe you can put your `flow-typed/npm` directory in your `.eslintignore`? That would effectively fix your workflow I think

ryyppy:
I would be happy to merge the jest bug tho!

altano:
I agree, trying to conform to certain eslint rules in this repository is a little silly, but if you look at the diff I think you'll agree the JS file is much more readable now, as it previously had inconsistent semicolons, really long lines, etc. And if you agree, why not accept the changes as a general readability improvement? :)

If you don't agree I'd be happy to move the method chaining fix into a separate pull request. Just let me know!

ryyppy:
I just want to discourage people committing style changes mixed with feature changes, since the diff noise is too high and I cannot see which feature is actually being changed :-) .... will need to revisit this again on my desktop

ryyppy:
Could you add some type tests for jest showing the usage of your changes? (chaining and such)? :-)

altano:
> I just want to discourage people committing style changes mixed with feature changes, since the diff noise is too high and I cannot see which feature is actually being changed :-)

The pull request is split into multiple commits, all of which do very specific things. If you'd prefer I'd be happy to split each commit up into a separate pull request, just let me know.

> Could you add some type tests for jest showing the usage of your changes? (chaining and such)? :-)

Absolutely. Added https://github.com/flowtype/flow-typed/pull/507/commits/afd945bf220740415becd7fb5b2401d66ee4147f with some tests that failed before my changes but now succeed.

ryyppy:
Thanks for your help :-)
@cpojer did the review here, so also thanks for that!

PR #508, opened 2016-12-02
Express: Make use method covariant to allow class extensions

mwalkerwells:
> Class methods are covariant but fields are invariant by default. — @jeffmo 
https://github.com/facebook/flow/issues/2770#issuecomment-258955097

This allows me to extend the class & override the use method:
```javascript
type SessionExpressMiddleware =
  ((request: expressSession$Request, response: express$Response, next: express$NextFunction) => mixed) |
  ((error: ?Error, request: expressSession$Request, response: express$Response, next: express$NextFunction) => mixed);

type sessionConstructor = expressSession$Options => express$Middleware

export function SessionExpress(
  sessionOptions: expressSession$Options,
  sessionConstructor: sessionConstructor,
  expressApplication: () => express$Application
) {
  const express = expressApplication()
  express.use(sessionConstructor(sessionOptions))
  return express
}

declare class SessionExpress extends express$Application {
    constructor(
        sessionOptions: expressSession$Options,
        sessionConstructor: sessionConstructor,
        expressApplication: () => express$Application
    ): void;
    use(middleware: SessionExpressMiddleware): this;
    use(...middleware: Array<SessionExpressMiddleware>): this;
    use(path: string|RegExp|string[], ...middleware: Array<SessionExpressMiddleware>): this;
    use(path: string, router: express$Router): this;
}

```

Without this, you get the following error:
screen shot 2016-12-02 at 3 40 05 pm


ryyppy:
LGTM! Could you also add a test case for overriding `use`? Would be happy to merge then  :-)

mwalkerwells:
@ryyppy A bit more complicated than I hoped, but should work! There were some flow compatibility issues with the test, so I needed to separate by version.

mwalkerwells:
@ryyppy Do I need to address anything else before merge?

ryyppy:
No, all done, thank you so much for doing the extra work :-)

PR #511, opened 2016-12-05
Chai tweaks

scottnonnenberg:
A couple changes were required to get rid of all Flow warnings on a 6000-line project:

1. Chai allows you to check length with a function call: `.length(4)`
2. Chai's configuration option `truncateThreshold` is a number, not a boolean

### DIFF
--- a/definitions/npm/chai_v3.5.x/flow_v0.24.0-/chai_v3.5.x.js
+++ b/definitions/npm/chai_v3.5.x/flow_v0.24.0-/chai_v3.5.x.js
@@ -183,6 +183,6 @@ declare module "chai" {
     declare var config: {
         includeStack: boolean,
         showDiff: boolean,
-        truncateThreshold: boolean
+        truncateThreshold: number
     };
 }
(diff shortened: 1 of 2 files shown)


ryyppy:
Alright, in the beginning I was confused by the official `lodash#map` documentation, since it didn't state anything about `string` as iteratee... but then I saw the examples which explicitly states 

```
var users = [
  { 'user': 'barney' },
  { 'user': 'fred' }
];
 
// The `_.property` iteratee shorthand.
_.map(users, 'user');
// => ['barney', 'fred']
```

Rest LGTM as well... I think we can merge this? @marudor

ryyppy:
Dang, this looks quite familiar to #486 as well :(

ggregoire:
Indeed, #486 fixes this case of `_.map`. I'm going to add some tests as asked by @marudor.

I will probably just use the examples from the Lodash docs.

scottnonnenberg:
Happy to remove the `lodash` aspects of this PR, just leaving the `chai` elements. Seems like y'all would like some tests as well? Do you have any other feedback I could address which might help this get approved?

ryyppy:
@scottnonnenberg Yeah, please remove the `lodash` stuff and add some tests for your chai changes, then I am more than happy to merge...  for lodash we will merge #486 to fix the issues :-)

@ggregoire Thanks for chiming as well! 👍

ryyppy:
Sorry for the delay and thanks for all the work 👍

PR #512, opened 2016-12-05
fix: express$Response mixins with http$ServerResponse

mattgstevens:
using the builtin http$ServerResponse class since the express.response class has these methods (in my case noticed `writeHead` missing)
https://github.com/facebook/flow/blob/master/lib/node.js#L879


marudor:

PR #515, opened 2016-12-06
Update moment queries (isAfter...) to be called without params

AugustinLF:
`isAfter|isBefore`etc. can also be called without any parameters. From the doc: 

> If nothing is passed to `moment#isBefore`, it will default to the current time.


masylum:
Let's get this merged please 🙏

masylum:
thanks!

PR #516, opened 2016-12-06
Add definition for jsuri

davidswinegar:
Add flowtypes and tests for jsuri npm package.


ryyppy:
Hmm, I looked at the API and some examples would not work with these definitions, I assume:

```
// query() should return an { params: Array<string|Array<string>> } I guess
new Uri('?a=b&c=d').query().params   // [ ['a', 'b'], ['c', 'd']]
```

Also, why does `protocol: (val: ?string) => string;` and similar methods take a parameter?

davidswinegar:
@ryyppy hmm, the documentation doesn't seem to match up with the code: https://github.com/derek-watson/jsUri/blob/master/Uri.js#L177 

Those methods take an optional parameter to set the value, and then return the value: https://github.com/derek-watson/jsUri/blob/master/Uri.js#L141

ryyppy:
Oh... wow... well :-)

PR #517, opened 2016-12-07
redux-form: Fix react component type

JelteF:
I got this error: Component. Application of polymorphic type needs <list of 3 arguments>. (Can use `*` for inferrable ones)


CodyReichert:
@JelteF - are you using flow v0.22.1? I believe that `React.Component` is valid in that version of flow and so this PR should actually add a new file under the correct version of flow. Eg,

```
redux-form_v5.x.x/flow_v0.36.x-/redux-form_v5.x.x.js
```

JelteF:
That might definitely be true. I was using flow v0.36.0.

CodyReichert:
@JelteF It looks like you're correct, this would be a bug in [flow v0.22](https://github.com/facebook/flow/blob/1f5fad20e546a5e64a3581c365f6957538d737c1/lib/react.js#L16) also, so you can ignore my comment. FWIW I hit this also, is why I cam looking for it :+1:

PR #518, opened 2016-12-07
Update node-slug options configuration

AugustinLF:
All the keys of the option don't have to be set, we can for instance see in the package's tests such an example: `slug("foo bar baz", {charmap})`.


AugustinLF:
Any update on this one? @marudor perhaps, since you merged my last PR in this repo? :)

PR #519, opened 2016-12-07
Update moment_v2.x.x.js

tiodan81:
utcOffset() should only return number. allowing string as a possible return type breaks compatibility with the add() method, since add() doesn't accept string

### DIFF
--- a/definitions/npm/moment_v2.x.x/flow_v0.28.x-/moment_v2.x.x.js
+++ b/definitions/npm/moment_v2.x.x/flow_v0.28.x-/moment_v2.x.x.js
@@ -176,7 +176,7 @@ declare class moment$Moment {
   local(): this;
   utc(): this;
   utcOffset(offset: number|string): void;
-  utcOffset(): number|string;
+  utcOffset(): number;
   format(format?: string): string;
   fromNow(removeSuffix?: bool): string;
   from(value: moment$Moment|string|number|Date|Array<number>, removePrefix?: bool): string;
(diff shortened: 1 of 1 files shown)


ryyppy:
I just tried `moment().utcOffset(9)` and `moment().utcOffset()` and got this:

image

Shouldn't it actually be `number|this`?

tiodan81:
At first I thought you were right, but looking more closely at the existing definitions it seems only the setter methods have `this` as a return type. For example, 

```
quarter(number: number): this;
  quarter(): number;
  year(number: number): this;
  years(number: number): this;
  year(): number;
  years(): number;
  weekYear(number: number): this;
  weekYear(): number;
  isoWeekYear(number: number): this;
  isoWeekYear(): number;
```

ryyppy:
Yeah exactly... it's hard to express the return type based on an existing parameter without doing some kind of overloading... maybe try something like this:

```javascript
utcOffset(): number,
utcOffset(v: number|string): this,
```

Also, could you add some tests for verification? :-)

nvie:
I've added this requested change in #722 — @ryyppy looks good?

ryyppy:
Hmm, there seem to be some conflicts... I am currently really stressed right now, if you can fix this, I will merge the PR asap :-)

nvie:
@ryyppy You can merge #722 — it's based on the latest master ;)

marudor:
As https://github.com/flowtype/flow-typed/pull/722 is merged this one is obsolete.

PR #521, opened 2016-12-08
Adding constructor & the expected misuse error

mwalkerwells:
Bug Fix: Adding correct constructor definition in test


mwalkerwells:
@ryyppy Quick follow-up to https://github.com/flowtype/flow-typed/pull/508

PR #522, opened 2016-12-09
Makes `scripts` Windows-compatible.

wtgtybhertgeghgtwtg:
Bumps `jest`, adds [`mkdirp`](https://github.com/substack/node-mkdirp) and [`rimraf`](https://github.com/isaacs/rimraf).

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -46,7 +46,6 @@
     "eslint-plugin-flow-vars": "^0.4.0",
     "flow-bin": "^0.32",
     "jest-cli": "^17.0.3",
-    "mkdirp": "^0.5.1",
     "rimraf": "^2.5.4"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


cli/package.json
```diff
@@ -46,7 +46,6 @@
     "eslint-plugin-flow-vars": "^0.4.0",
     "flow-bin": "^0.32",
     "jest-cli": "^17.0.3",
-    "mkdirp": "^0.5.1",
```
ryyppy: Why is it removed?

wtgtybhertgeghgtwtg: I noticed it was already in `dependencies`.

ryyppy: Ah, true... I don't like the diffing preview sometimes :-)

ryyppy:
Okay I just tested it on OSX as well... looking good, thanks! 👍  ... I will need to check back with other flow-typed maintainers if we want to go the `src for testing` way.. I would actually prefer it

PR #523, opened 2016-12-09
mersenne-twister types

AgentME:
Adds types for https://github.com/boo1ean/mersenne-twister.

### DIFF
--- a/definitions/npm/mersenne-twister_v1.x.x/flow_v0.33.x-/mersenne-twister_v1.x.x.js
+++ b/definitions/npm/mersenne-twister_v1.x.x/flow_v0.33.x-/mersenne-twister_v1.x.x.js
@@ -1,0 +1,18 @@
+declare module 'mersenne-twister' {
+  declare class $npm$MersenneTwister {
+    constructor(seed?: number|Array<number>): $npm$MersenneTwister;
+    init_seed(seed: number): void;
+    init_by_array(init_key: Array<number>, key_length: number): void;
+
+    random(): number;
+    random_int(): number;
+    random_excl(): number;
+    random_long(): number;
+    random_int31(): number;
+
+    mt: Array<number>;
+    mti: number;
+  }
+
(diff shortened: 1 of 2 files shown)


definitions/npm/mersenne-twister_v1.x.x/flow_v0.33.x-/mersenne-twister_v1.x.x.js
```diff
@@ -0,0 +1,18 @@
+declare module 'mersenne-twister' {
+  declare class $npm$MersenneTwister {
```
ryyppy: You don't need to namespace your stuff inside the `declare module` body... it's already concealed and will not bleed to global scope

AgentME: Won't Flow think it was something exported from the module? I didn't want Flow to allow `import {MersenneTwister} from 'mersenne-twister'`.

ryyppy: You can try doing that... if you have `declare var exports: ...` it will assume it's a commonjs module and will not automatically export all inner declares (I am pretty sure... did not test it tho)

ryyppy: whatever, I will just merge this for now... it doesn't break things :-)

AgentME: Oh, I don't know about `declare var exports`, that's worth a try in the future. I've seen a few other modules prefix things (though leave them in the global scope) instead of leave them regular-named inside the module, so I don't think this is too inconsistent with what else is here. I think Flow is sorely lacking a real way to mark things private in a definition.

PR #524, opened 2016-12-09
Test on `src` instead of `dist`.

wtgtybhertgeghgtwtg:
Builds off of #522.
This has a few benefits.
- `__tests__` can be moved out of `src` so that they will not be transpiled and packaged.
- The `test` script can be rewritten so that it does not run `build`.
- "Uncovered Lines" in `jest --coverage` will point to source code instead of transpiled code.

### DIFF
--- a/cli/.babelrc
+++ b/cli/.babelrc
@@ -1,8 +1,4 @@
 {
-  "ignore": [
-    "__mocks__",
-    "__tests__"
-  ],
   "plugins": [
     "transform-flow-strip-types",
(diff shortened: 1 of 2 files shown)


ryyppy:
Hmm, `Test on src instead of dist` is a little too much... @jeffmo wants to test the compiled code for a reason (I also wanted to do that in the first place, but he didn't think it was a good idea)

@jeffmo would this be an option?

marudor:

PR #533, opened 2016-12-11
Update express

philikon:
express's response object clearly inherits from node's http$ServerResponse. However, this type is only available in Flow 0.32+.


definitions/npm/express_v4.x.x/flow_v0.32.x-/express_v4.x.x.js
```diff
@@ -0,0 +1,179 @@
...
+
+declare type express$SendFileOptions = {
+  maxAge?: number,
+  root?: string,
+  lastModified?: boolean,
+  headers?: {[name: string]: string},
+  dotfiles?: 'allow' | 'deny' | 'ignore'
+};
+
+declare class express$Response extends http$ServerResponse mixins express$RequestResponseBase {
```
philikon: This file is the same as the Flow v0.28.x - v0.31.x version, except for this line, where we make `express$Response` inherit from `http$ServerResponse`.

See also https://github.com/expressjs/express/blob/master/lib/response.js#L41

philikon:
Sigh, now hitting https://github.com/facebook/flow/issues/3084. That's annoying. Even if that gets fixed we'll have to account for that forever.

philikon:
Re-pushed with a workaround for https://github.com/facebook/flow/issues/3084.

philikon:
Can haz merge?

PR #537, opened 2016-12-13
Update express_v4.x.x.js

asci:
it is possible to use a router without root path [Expressjs docs](http://expressjs.com/en/4x/api.html#router)


asci:
opened another PR https://github.com/flowtype/flow-typed/pull/561

PR #540, opened 2016-12-13
RxJS: Add repeat() operator

matthewwithanm:


GAntoine:
Closing since the `repeat` method was added by #821.

PR #541, opened 2016-12-14
Add moment's updateLocale method

AgentME:
Add a missing method type.


ryyppy:
Merged, since @jeffmo already fixed the libs causing the CI to fail

PR #543, opened 2016-12-14
react-dnd definition improvements

AgentME:
This pull request makes three main improvements to react-dnd's definitions. Each improvement is in its own commit.

The definitions of react-dnd prohibits legal code like this because `monitor` in the callback type is incorrectly typed as optional:

```js
import {DragSource} from 'react-dnd';

const ManageItemDragSource = {
  beginDrag(props: Props) {
    return {
      item: props.item,
      index: props.index
    };
  },

  endDrag(props, monitor){
    if(!monitor.didDrop()){
      props.onSetIndex(monitor.getItem().index, monitor.getItem().item);
    }
  },

  isDragging(props, monitor){
    return props.item.guid === monitor.getItem().item.guid;
  }
}


export default DragSource(..., ManageItemDragSource, ...);
```

The user of react-dnd has to pass some callback functions into the library. react-dnd calls these callbacks with a bunch of parameters and the user might not care about all of the parameters. The current react-dnd definitions incorrectly type some of the parameters as optional. The parameters are always passed. I believe they were typed as optional to allow the user's callbacks to not need to deal with them, but this is unnecessary (Flow allows a function to be called with extra parameters) and means that user callbacks that do want to deal with these extra parameters has to null-check values that will never be null. This pull request fixes it.

---

Another problem is that the `getDecoratedComponentInstance()` method had the wrong type. It was typed as returning the wrapper component, when it's supposed to return the type of the component being wrapped (so that methods could be called on it, etc). `DndComponent` has another type parameter referring to the wrapped class so that this just works now. Test included.

---

When you wrap a component with DragSource, DropTarget, or DragLayer, you pass a "collect" function which returns an object that provides more props to the component. The returned wrapper component does not need this props passed to it. However, the old definitions had the wrapper component have the same type (and therefore the same required props) as the wrapped component, so Flow still believed you needed to pass props that were provided by the collect function. This was fixed by setting the return value of the collect function into a type parameter, and then using `$Diff` on the wrapped component's prop type and the collect function's return type. This solution is consistent with the [documentation's recommendations on React higher order components](https://flowtype.org/docs/react.html).

### DIFF
--- a/definitions/npm/react-dnd_v2.x.x/flow_v0.23.x-/react-dnd_v2.x.x.js
+++ b/definitions/npm/react-dnd_v2.x.x/flow_v0.23.x-/react-dnd_v2.x.x.js
@@ -46,24 +46,24 @@ type DragSourceType<P> =
 type DragSourceSpec<D, P, S> = {
   beginDrag: (
     props: P,
-    monitor?: DragSourceMonitor,
-    component?: React$Component<D, P, S>
+    monitor: DragSourceMonitor,
+    component: React$Component<D, P, S>
   ) => Object,
 
   endDrag?: (
     props: P,
-    monitor?: DragSourceMonitor,
-    component?: ?React$Component<D, P, S>
+    monitor: DragSourceMonitor,
+    component: ?React$Component<D, P, S>
   ) => void,
 
   canDrag?: (
     props: P,
-    monitor?: DragSourceMonitor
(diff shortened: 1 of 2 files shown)


Nasafato:
How can I contribute to help make sure Travis CI passes? The changes you're proposing to making monitor required instead of optional are great and necessary.

PR #545, opened 2016-12-17
Avoid `import type` in declaration files

mkscrg:
This is a workaround for facebook/flow#3040: when used in a declaration file, `import type { T } from 'm'` breaks other files' usage of `m.T`.

`flow-typed` only has two definitions that use `import type`:

- `express_v4.x.x` includes `import type { Server } from 'http'`. With this definition installed, the user gets a Flow error anywhere they use the `http.Server` type. Unfortunately this error is raised for any dependencies that are themselves using the `express` definition, since the definition includes a `// @flow` pragma and Flow type checks all files with pragmas. 😭 

- `react-redux_v4.x.x` includes `import type { Dispatch, Store } from 'redux'`. This definition doesn't cause problems due to an odd quirk of facebook/flow#3040: it's only reproducible when the declaration with `import type { T } ...` is listed (by the filesystem or explicitly in `.flowconfig`) _after_ the declaration that exports `T`. So this one slips by because lexicographically, `'react-redux' < 'redux'`. 

This PR updates both to use the syntax `import m from 'm'` and `m.T`. I could have left `react-redux_v4.x.x` alone, but it felt right to get away from that strange order-dependent behavior.

### DIFF
--- a/definitions/npm/express_v4.x.x/flow_v0.25.x-v0.27.x/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_v0.25.x-v0.27.x/express_v4.x.x.js
@@ -1,4 +1,4 @@
-import type { Server } from 'http';
+import http from 'http';
 
 declare type express$RouterOptions = {
   caseSensitive?: boolean,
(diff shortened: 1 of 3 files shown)


ryyppy:
Hey! Just wanted to chime in and say that I am aware of this PR, but have no idea how to handle it for now... maybe @marudor & @bassettsj can also have a look at that?

mkscrg:
@ryyppy do you mean how we can avoid the problem going forward? We could try to catch these in `validate-defs`, though if facebook/flow#3040 is fixed soon that might not be worth the effort.

migueloller:
There's also another side effect that comes from importing types like this. When using `react-redux` the types `Dispatch` and `Store` are available globally because they are imported outside the module declaration.

The solution to this would be to change the type definition to one of these:

```js
import Redux from 'redux'

declare module 'react-redux' {
  declare type Dispatch = Redux.Dispatch
  declare type Store = Redux.Store
```
```js
declare module 'react-redux' {
  import type { Dispatch, Store } from 'redux'
```

The latter looks like the correct one. It would seem that importing Flow types at the top level in a library declaration yields bad results.

sibelius:
I think we can use the later solution

marudor:
Yes, we should use the later one.
If someone could make a PR for this that would be great. 

Closing this in favor of a PR that includes on Module level.

PR #550, opened 2016-12-18
Add definition for react-native-audio-toolkit

keplersj:
This is my first attempt at writing a Flow binding, so please tell me where to improve.


definitions/npm/react-native-audio-toolkit_v1.x.x/flow_v0.32.x-/react-native-audio-toolkit_v1.x.x.js
```diff
@@ -0,0 +1,58 @@
+type $npm$react_native_audio_toolkit$PlaybackOptions = {
```
ryyppy: Why are these types declared in the global and not in the `declare module` scope?

keplersj: That is an excellent question. Fixing the scope now.

ryyppy:
Looking good, thanks 👍

PR #552, opened 2016-12-18
Add bluebird spread

SimenB:
http://bluebirdjs.com/docs/api/spread.html

### DIFF
--- a/definitions/npm/bluebird_v3.x.x/flow_v0.32.x-/bluebird_v3.x.x.js
+++ b/definitions/npm/bluebird_v3.x.x/flow_v0.32.x-/bluebird_v3.x.x.js
@@ -142,6 +142,7 @@ declare class Bluebird$Promise<R> {
   each<T, U>(iterator: (item: T, index: number, arrayLength: number) => Bluebird$Promisable<U>): Bluebird$Promise<Array<T>>;
   asCallback<T>(callback: (error: ?any, value?: T) => any, options?: Bluebird$SpreadOption): void;
   return<T>(value: T): Bluebird$Promise<T>;
+  spread<T>(...args: Array<T>): Bluebird$Promise<*>;
 
   reflect(): Bluebird$Promise<Bluebird$PromiseInspection<*>>;
 
--- a/definitions/npm/bluebird_v3.x.x/test_bluebird-v3.x.x.js
+++ b/definitions/npm/bluebird_v3.x.x/test_bluebird-v3.x.x.js
@@ -53,3 +53,7 @@ fooPromise(1, 'b').catch(function(e: Error) {
 fooPromise('a', 1)
 // $ExpectError
 fooPromise()
+
+Bluebird.resolve(['arr', { some: 'value' }, 42])
+  .spread((someString: string, map: Object, answer: number) => answer)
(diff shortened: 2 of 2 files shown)


definitions/npm/bluebird_v3.x.x/flow_v0.32.x-/bluebird_v3.x.x.js
```diff
@@ -142,6 +142,7 @@ declare class Bluebird$Promise<R> {
   each<T, U>(iterator: (item: T, index: number, arrayLength: number) => Bluebird$Promisable<U>): Bluebird$Promise<Array<T>>;
   asCallback<T>(callback: (error: ?any, value?: T) => any, options?: Bluebird$SpreadOption): void;
   return<T>(value: T): Bluebird$Promise<T>;
+  spread(...args: Array<any>): Bluebird$Promise<any>;
```
ryyppy: Why `any` and not a generic `T`? ...e.g  `spread<T>(...args: Array<T>): Bluebird$Promise<T>;`?

SimenB: Mostly because the promise can resolve with many values of different types, e.g. a string and an object. Would a generic still be correct? (I'm not too well versed in flow generics, but that'd be wrong in Java)

ryyppy: Hmm, not sure how `flow` will treat different promise results in one array as a generic... do you mind adding a simple test in your test file :-)? If this works it would be better to use generics than `any`

SimenB: Updated with a test of multiple types, and flow doesn't complain!

definitions/npm/bluebird_v3.x.x/flow_v0.32.x-/bluebird_v3.x.x.js
```diff
@@ -142,6 +142,7 @@ declare class Bluebird$Promise<R> {
   each<T, U>(iterator: (item: T, index: number, arrayLength: number) => Bluebird$Promisable<U>): Bluebird$Promise<Array<T>>;
   asCallback<T>(callback: (error: ?any, value?: T) => any, options?: Bluebird$SpreadOption): void;
   return<T>(value: T): Bluebird$Promise<T>;
+  spread<T>(...args: Array<T>): Bluebird$Promise<T>;
```
ryyppy: Okay I double checked it myself... I think the better return type of this function would be `Bluebird$Promise<*>` ... otherwise we would force the user to return the same type `T` as the input type... which is not what bluebird wants, right? `*` will take whatever type is returned in the `then` body... this is definitely better than `any` :-)

SimenB: What's the difference between `any` and `*`?

Pushed the change though 😄  And expanded the test so it actually failed like you thought it would before using `*`

ryyppy:
Thanks for the effort 👍

PR #555, opened 2016-12-20
Added parse definition

mmazzarolo:
Added definition for [Parse JS SDK](https://github.com/ParsePlatform/Parse-SDK-JS). 
Some less used modules are still not defined (and marked as TODO).

### DIFF
--- a/definitions/npm/parse_v1.9.x/flow_v0.37.x-/parse_v1.9.x.js
+++ b/definitions/npm/parse_v1.9.x/flow_v0.37.x-/parse_v1.9.x.js
@@ -1,16 +1,16 @@
 declare module 'parse/node' {
-  declare module.exports: Parse
+  declare module.exports: $npm$parse$Parse
 }
 
 declare module 'parse/react-native' {
-  declare module.exports: Parse
+  declare module.exports: $npm$parse$Parse
 }
 
 declare module 'parse/browser' {
-  declare module.exports: Parse
+  declare module.exports: $npm$parse$Parse
 }
 
-type Parse = {
+type $npm$parse$Parse = {
   initialize(applicationId: string, javaScriptKey: string): void,
   Analytics: any, // TODO
   ACL: Class<ParseACL>,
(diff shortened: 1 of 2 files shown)


definitions/npm/parse_v1.9.x/flow_v0.37.x-/parse_v1.9.x.js
```diff
@@ -0,0 +1,326 @@
...
+
+declare module 'parse/react-native' {
+  declare module.exports: Parse
+}
+
+declare module 'parse/browser' {
+  declare module.exports: Parse
+}
+
+type Parse = {
```
ryyppy: Needs namespacing... like this: `$npm$Parse$Parse`

mmazzarolo:
@ryyppy Thank you for checking the code. Can you point me to the documentation about the namespacing? Is namespacing the exported module (Parse) enough? Thanks!

ryyppy:
Hey @mmazzarolo no worries, namespaces are a `flow-typed` specific thing, because we don't want to have collisions with global userspace types (which will eventually happen without this weird `$npm$[library]` prefix.

For more information please check our [CONTRIBUTING.md](https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md)

ryyppy:
I just realized that you declared almost all type declarations etc. in the toplevel scope of the declaration file.. that would mean you would be required to namespace every single type (which is kinda hard to read and also tedious to change)... could you try this:

```
declare module '@@parse' {
  // declare all the stuff you need
  declare type Parse = {...};
  declare module.exports: Parse;
}

declare module 'parse/node' {
  import Parse from '@@parse';
  declare module.exports: typeof Parse;
}

// same for 'parse/react-native' etc.
```

I didn't test it like this, but I think this should work to prevent global scope pollution

mmazzarolo:
Tried it but it seems it doesn't work. 
I even tried playing with the import itself but with no results:  
```javascript
declare module 'parse/browser' {
  import type Parse from '@@parse'
  declare module.exports: Parse
}
```

mmazzarolo:
@ryyppy I just namespaced everything :)

ryyppy:
Oh snap, probably the `import` inside `declare module` is not supported yet... maybe in the next feature version?

Anyways, something seems to be broken during the renaming process :-(

ryyppy:
Thank you for your contribution :-)

mmazzarolo:
@ryyppy no worries, thank you for the super-support. 
I'll try to keep it updated, we use Parse everyday here.

PR #560, opened 2016-12-22
axios@v0.15.x fix for flow@0.37.x

bassettsj:
Address the error that was introduced after flow@0.37.x was released.

https://github.com/flowtype/flow-typed/issues/558


bassettsj:
Hmmm.. This is frustrating, I think it could be an issue with flow `v0.37.x` maybe?

marudor:
You're running into the same Problems I did when I tried to fix it...
I think we need the help of someone from the flow team here.
@jeffmo @samwgoldman @gabelevi

inv8der:
This appears to be the same issue we're experiencing here: https://github.com/facebook/flow/issues/3084

marudor:
fixed with d28912a0365e24da7e7a99d7b6a787c1e551a2f1

PR #427, opened 2016-10-30
Imlement most missing rxjs libdefs

Kriegslustig:
This PR extends the rxjs libdefs, so that they deflect the official documentation. I did find a few cases where they defined a type signature loosesly because of restrictions in TypeScript. In those cases I tried to be as restrictive as possible.

PR #562, opened 2016-12-22
Add rxjs5 bindNodeCallback static method

ooHmartY:
Hi all,

rxjs 5 supports `Observable.bindNodeCallback` here's the flow definition for it.

From:
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html
https://github.com/ReactiveX/rxjs/blob/master/MIGRATION.md

### DIFF
--- a/definitions/npm/rxjs_v5.0.x/flow_v0.34.x-/rxjs_v5.0.x.js
+++ b/definitions/npm/rxjs_v5.0.x/flow_v0.34.x-/rxjs_v5.0.x.js
@@ -37,6 +37,13 @@ type rxjs$EventListenerOptions = {
 } | boolean;
 
 declare class rxjs$Observable<+T> {
+
+  static bindNodeCallback(
+      func: (e: Error, ...args: any) => void,
+      selector?: () => T,
+      scheduler?: rxjs$SchedulerClass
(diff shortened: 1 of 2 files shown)


superherointj:
The version you posted didn't work for me. I'm really unexperienced and might be saying something really stupid. So, please bear with me. I just want it to work with my current code. That works without Flow.

Considering:

import fs from 'fs'
import Rx from 'rxjs'
const bindNodeCallback = Rx.Observable.bindNodeCallback
var fsReaddir = bindNodeCallback(fs.readdir)
fsReaddir('../').subscribe(n => console.log('n', n), err => console.log('err', err), () => console.log('Completed.') )

// From Node.JS docs:
// fs.readdir(path[, options], callback)


----

inconsistent use of library definitions:
763:   declare function readdir(path: string, callback?: (err: ?ErrnoError, files: Array<string>) => void): void;
                                      ^^^^^^ string. This type is incompatible with an argument type of. See lib: 

 46:     func: (e: Error, ...args: any) => void,
                   ^^^^^ Error. See lib: flow-typed/npm/rxjs_v5.0.x.js:46

Found 1 error

---

ooHmartY:
looks like #427 covers this. closing.

PR #566, opened 2016-12-23
Add Bluebird.promisifyAll array form

bchociej:
`Bluebird.promisifyAll` also accepts an array form as per http://bluebirdjs.com/docs/api/promise.promisifyall.html#promisifying-multiple-classes-in-one-go


ryyppy:
Thanks 👍

PR #568, opened 2016-12-24
add type definitions for partial and compose

andreiglingeanu:
This pull request add type definitions for [`_.partial`](http://underscorejs.org/#partial) and [`_.compose`](http://underscorejs.org/#compose).

The type definition for `compose` is very simple and I really don't know how do I describe the fact that the first function's return value is the return value of the function that is returned from `compose`:

```javascript
declare function compose(...functions: Function): Function;

// USAGE:
(

_.compose(
  function (name:string):string { return name + ', hello!'; },
  function (user:Object):string { return user.name; }
): (user: Object) => string

)
```

I'd be happy to provide a more stricter type definition but I really don't know how to approach that. I'd be happy to read any feedback. Thanks.

### DIFF
--- a/definitions/npm/underscore_v1.x.x/flow_v0.13.x-/underscore_v1.x.x.js
+++ b/definitions/npm/underscore_v1.x.x/flow_v0.13.x-/underscore_v1.x.x.js
@@ -155,7 +155,7 @@ declare module "underscore" {
 
   // TODO: negate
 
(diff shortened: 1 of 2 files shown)


andreiglingeanu:
Tests passing!

andreiglingeanu:
Is there something I can do in order to get this merged?

andreiglingeanu:
Thanks!

PR #570, opened 2016-12-26
fix chalk_v1.x.x failing in react-native project

ifiokjr:
chalk_v1.x.x was failing in react-native project due to missing method/property (dim).


ryyppy:
Thanks! :-)

PR #572, opened 2016-12-27
Definition for xorshift

dbrockman:
https://github.com/AndreasMadsen/xorshift

### DIFF
--- a/definitions/npm/xorshift_v1.x.x/flow_v0.32.0-/xorshift_v1.x.x.js
+++ b/definitions/npm/xorshift_v1.x.x/flow_v0.32.0-/xorshift_v1.x.x.js
@@ -1,0 +1,18 @@
+declare module 'xorshift' {
+  declare class XorShift {
+    /**
+     * Create a new random generator, with a specific seed.
+     */
+    constructor(seed: [number, number, number, number]): XorShift;
+    /**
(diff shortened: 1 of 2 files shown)


definitions/npm/xorshift_v1.x.x/flow_v0.32.0-/xorshift_v1.x.x.js
```diff
@@ -0,0 +1,18 @@
+declare module 'xorshift' {
+  declare class $npm$xorshift {
```
ryyppy: If you use `declare module.exports`, then this class will not be automatically exported.. you can theoretically leave out the `$npm$` part

dbrockman: Oh, I was wondering about that. Is there any documentation on what becomes exported/visible?

ryyppy: There is a major refactoring for the official flow doc going on, where I want to add this specific section... for now, you can see most syntax features for module definition in my libdef-talk: https://github.com/ryyppy/flow-libdef-talk/blob/master/dist/flow-libdef-talk.pdf

PR #573, opened 2016-12-28
Add koa@2.x definition.

iamchenxin:
Currently, there is an issue [flow/issues/3009](https://github.com/facebook/flow/issues/3009).
The overload functions (`accepts`, `acceptsCharsets`, `acceptsEncodings`, `acceptsLanguages`) in `Request` type need additional annotations to disambiguate cases.
ex: [in test file L274](https://github.com/flowtype/flow-typed/pull/573/files#diff-ef19f871389577e2759a1decdff9a2c2R274)
```
  const typeOrFalse3: string|false = request.accepts((['json', 'text']:Array<string>));
```
Maybe, this will be fixed later.


ryyppy:
Looking good :-)

PR #581, opened 2017-01-01
lodash: defs for flatmap to work on objects

ngzhian:
Existing defs for flatMap only works for array, but it actually works for objects too ([lodash docs](https://lodash.com/docs/4.17.4#flatMap))

This pr:
- adds OFlatMapIteratee for flatMapping on objects
- adds appropriate defs for flatMap, flatMapDeep, flatMapDepth on objects


marudor:
Could you add tests please?

ngzhian:
@marudor added!

PR #585, opened 2017-01-01
Added stack to express$Router definition

jagreehal:
Without this I can't execute tests to ensure routes have been added to an Express router e,g,

`  assert.ok(router.stack.length);`

### DIFF
--- a/definitions/npm/express_v4.x.x/flow_v0.28.x-/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_v0.28.x-/express_v4.x.x.js
@@ -135,6 +135,7 @@ declare class express$Router extends express$Route {
   use(...middleware: Array<express$Middleware>): this;
   use(path: string|RegExp|string[], ...middleware: Array<express$Middleware>): this;
   use(path: string, router: express$Router): this;
+  stack: mixed[];
 }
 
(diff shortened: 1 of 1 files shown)


ryyppy:
is this kind of private api?

jagreehal:
That is an interesting point.

I'm converting over from a project over from TypeScript to Flow.

It worked in TS as stack is included in the definition for a Router https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/express-serve-static-core/index.d.ts#L121

Is the aim for Flow-Typed to only provide definitions for 'public' functions and properties?

If so what happens when users want to convert TS definitions to Flow e.g. https://github.com/flowtype/flow-typed/pull/590?

ryyppy:
I guess we can include it, but the stack would consist of `Layer` objects, which is not `mixed` no?

jagreehal:
@ryyppy no lets not include it. I'd rather wait to see what happens to #590 rather than creating lots of small PRs and open a can of worms.

For now (and while https://github.com/facebook/flow/issues/396 is unresolved) I'll just use a work around 

`assert.ok((router: any).stack.length);`

Thanks for your help.

ryyppy:
good decision 👍 😊

PR #587, opened 2017-01-02
Allow a dependency to be listed in multiple package.json dep fields a…

### LINKED ISSUE #379: Recommended approach for using flow-typed with peerDependencies

``` bash
❯ flow-typed version
flow-typed@2.0.0-beta.11

```

ahutchings:
…s long as the versions are equivalent. Refs #379 

According to the npm docs, [`bundledDependencies` is an array of dependency names](https://docs.npmjs.com/files/package.json#bundleddependencies). Also, a `bundledDependency` [must be in dependencies, devDependencies, optionalDependencies, or node_modules](https://github.com/npm/npm/issues/6435#issuecomment-58599735) so we don't need to look for version ranges in that field - either a dependency's version range will be defined in another field, or the dependency will be in node_modules, potentially with local modifications such that we cannot install libdefs for it.

The primary change relaxes validation to allow for a dependency to be in multiple package.json dependency fields, as long as the version ranges are the same. This should fix #379.

### DIFF
--- a/cli/src/lib/__tests__/npmProjectUtils-test.js
+++ b/cli/src/lib/__tests__/npmProjectUtils-test.js
@@ -1,0 +1,48 @@
+// @flow
+import {getPackageJsonDependencies} from '../npmProjectUtils';
+
+describe('npmProjectUtils', () => {
+  describe('getPackageJsonDependencies', () => {
+    it('throws when a dep of differing versions is found in multiple dep fields', () => {
+      const pkgJson = {
+        pathStr: '',
+        content: {
+          name: '',
+          version: '',
+          devDependencies: {
+            'flow-bin': '^0.38.0'
+          },
+          peerDependencies: {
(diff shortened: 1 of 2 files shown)


ahutchings:
This does not address the use case raised in https://github.com/flowtype/flow-typed/issues/379#issuecomment-264651811. There are a few related cases to consider:

1. The same version/version range exists in peerDeps and devDeps. This one seems simple -resolve to that version/version range.
2. peerDeps and devDeps have differing versions. Should peerDeps take precedence?
3. There's no reason for a dependency to be in both deps and any other category (devDeps or peerDeps). In these scenarios we have at least 3 options - throw, let deps take precedence, or only throw if the version ranges are not the same.

jedwards1211:
@ahutchings devDeps should take precedence, because it's what's actually installed.

GAntoine:
Closed in favour of #1467

PR #588, opened 2017-01-02
Update parse_v1.9.x.js

robarnold:
Updates the parse definitions to match the documentation at https://parseplatform.github.io/Parse-SDK-JS/api/index.html along with my reading of the code where it is ambiguous or unclear. This is all according to my current use, so I've probably missed some things.

### DIFF
--- a/definitions/npm/parse_v1.9.x/flow_v0.37.x-/parse_v1.9.x.js
+++ b/definitions/npm/parse_v1.9.x/flow_v0.37.x-/parse_v1.9.x.js
@@ -11,7 +11,7 @@ declare module 'parse/browser' {
 }
 
 declare type $npm$parse$Parse = {
-  initialize(applicationId: string, javaScriptKey: string): void,
+  initialize(applicationId: string, javaScriptKey: ?string): void,
   Analytics: any, // TODO
   ACL: Class<$npm$parse$ParseACL>,
   Cloud: $npm$parse$ParseCloud,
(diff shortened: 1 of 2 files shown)


definitions/npm/parse_v1.9.x/flow_v0.37.x-/parse_v1.9.x.js
```diff
@@ -32,7 +32,9 @@ declare type $npm$parse$Parse = {
   Session: any, // TODO
   Storage: any, // TODO
   Query: Class<$npm$parse$ParseQuery>,
-  User: Class<$npm$parse$ParseUser>
+  User: Class<$npm$parse$ParseUser>,
+  masterKey: ?string,
```
ryyppy: is `masterKey` etc. required as a property? Otherwise use `masterKey?: string`

ryyppy:
Cool! Could you also add some matching tests for your changes to prevent regressions in later updates? That would be neat :-)

definitions/npm/parse_v1.9.x/flow_v0.37.x-/parse_v1.9.x.js
robarnold: You don't need to set it, but it's always defined via a getter/setter and appears to default to null

ryyppy:
Thanks for the effort 👍

PR #592, opened 2017-01-03
Underscore.js: Add secondary reduce that works on objects, not arrays.

wbecker:


GAntoine:
@wbecker How keen are you on making this change? Seems versions with `flow 38+` have a secondary reduce for objects.

PR #596, opened 2017-01-06
Parse.Query should be parameterized by object type

robarnold:
In the constructor you tell it the class you expect to be returned. We can have Flow infer that and properly type the query results.

Reference: https://parseplatform.github.io/Parse-SDK-JS/api/classes/Parse.Query.html


robarnold:
Anything else needed for this PR to be accepted?

mmazzarolo:
Hey @robarnold, nice job.  
I created the first typedef of the Parse lib and I can confirm that the PR seems valid.  
Just a question though: why did you pure a `?`on the type here?  
```javascript
  get(objectId: string, options?: $npm$parse$FullOptions): Promise<?T>,
```

PR #600, opened 2017-01-07
[rxjs] Add dummy void arguments to overloaded methods

nmote:
I don't have an isolated test case, but these help Flow determine which option to choose. Without these, it might choose e.g. a one-argument version even though the callsite has three arguments, and it will just ignore the last two arguments, leading to type errors.

The changes to `combineLatest` are necessary to typecheck Nuclide's code with Flow master.


nmote:
I'm going to merge this myself unless anybody objects in the next few days

PR #601, opened 2017-01-07
declare Bluebird as subtype of Native Promise, add tests

orlandoc01:
The Bluebird API effectively provides the same interface as the Native Promise API (despite having a different under-the-hood implementation) with some additional functionality as well. As a result, it can be thought of as a subtype of the Promise type. By declaring Bluebird as a subclass of Native Promise (since the Native Promise is a Class), most existing code which has been constructed to operate with Promise type instances can now also operate with Bluebird type instances without having to be rewritten and made aware of the Bluebird library. The test cases I added to the Bluebird library best illustrate the benefits that come from making this sub-typing relationship.

However, in attempting this, there was an issue with the `Promise.all` declaration from Flow version v0.32 to v0.33. As a result, I let the v0.32 flow version type library infer the type signature of Bluebird's Promise.all from it's "superclass" (the Native Promise) and this allowed it to pass existing tests. I then fixed the implementation of `Promise.all` in flow version v0.33 and up (when that method was modified in `core.js`) so that it can function as the Native Promise.all method does as well as support additional Promise manipulation functionality. (See here for more info http://bluebirdjs.com/docs/api/promise.all.html). Thus, the only difference between `v0.32` and `v0.33-` is the addition of the type signature for the `all` static method.

Lastly, I also noticed a bug and submitted a previous PR regarding the static `reduce` method for Bluebird. However, that PR did not include new unit tests for that so I've also added tests for that as well and combined it here with this PR. (See here for info about the type signature for `reduce` http://bluebirdjs.com/docs/api/promise.reduce.html)

### DIFF
--- a/definitions/npm/bluebird_v3.x.x/flow_v0.32/bluebird_v3.x.x.js
+++ b/definitions/npm/bluebird_v3.x.x/flow_v0.32/bluebird_v3.x.x.js
@@ -41,62 +41,59 @@ declare type Bluebird$PromisifyAllOptions = {
   promisifier?: (originalMethod: Function) => () => Bluebird$Promise<any> ;
 };
 
-declare type Bluebird$Promisable<T> = Bluebird$Promise<T> | Promise<T> | T;
+declare type $Promisable<T> = Promise<T> | T;
 
-declare class Bluebird$Promise<R> {
+declare class Bluebird$Promise<+R> extends Promise<R> {
   static Defer: Class<Bluebird$Defer>;
   static PromiseInspection: Class<Bluebird$PromiseInspection<*>>;
 
-  static all<T, Elem: Bluebird$Promisable<T>>(Promises: Array<Elem>): Bluebird$Promise<Array<T>>;
-  static props(input: Bluebird$Promisable<Object|Map<*,*>>): Bluebird$Promise<*>;
-  static any<T, Elem: Bluebird$Promisable<T>>(Promises: Array<Elem>): Bluebird$Promise<T>;
-  static race<T, Elem: Bluebird$Promisable<T>>(Promises: Array<Elem>): Bluebird$Promise<T>;
+  static props(input: $Promisable<Object|Map<*,*>>): Bluebird$Promise<*>;
+  static any<T, Elem: $Promisable<T>>(Promises: Array<Elem>): Bluebird$Promise<T>;
+  static race<T, Elem: $Promisable<T>>(Promises: Array<Elem>): Bluebird$Promise<T>;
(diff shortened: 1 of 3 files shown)


StreetStrider:
@orlandoc01, 
> As a result, it can be thought of as a subtype of the Promise type

This may be a good idea.

> Lastly, I also noticed a bug and submitted a previous PR regarding the static reduce method for Bluebird

I presume, you do the same that I did in #583. Sadly, my PRs are hanging for a while. I do not see any real attention from maintainers here.

orlandoc01:
I think it would be a good idea, as you can see in the tests I added it allows typed functions which expect to manipulate native Promises to also function correctly with Bluebird Promises (which fully support the Native Promise API) without having to explicitly import and be aware of the Bluebird library types.

As for  [#583](https://github.com/flowtype/flow-typed/pull/583), it seems like you picked up a different error than the one I found haha. I rebased and incorporated your fix in this branch but the error I was fixing referred to the static reducer function's return value [here](https://github.com/flowtype/flow-typed/pull/601/files#diff-284a8d6c72c820c624103907f48e1dd5R68), which is already correct actually on the instance reduce method

StreetStrider:
@orlandoc01, damn right, reducer retval should be not `U`, but `$Promisable<U>`, which allows reducer to be sync or async. Bluebird is awesomly flexible.

Thanks for  the rebasing, by the way. We had already conflict at this point. I got other two old patches, #594 #559, I'm expecting they will also conflict in `Bluebird$Promisable<U>` → `$Promisable<U>`.

orlandoc01:
> Bluebird is awesomly flexible
Totally agree, love that library haha 👍 

I'll watch out for when those patches merge and rebase this accordingly

orlandoc01:
Rebased and incorporated patches from #594 and #559

danharper:
@marudor @orlandoc01 I think this PR broke the CLI, see #610.

> UNCAUGHT ERROR: Error: npm/bluebird_v3.x.x/flow_v0.32: Malformed flow-version directory name! Expected the name to be formatted as 'flow_all' or 'flow_-?v<VERSION>' or 'flow_v<VERSION>-?' or 'flow_v<VERSION>-v<VERSION>' (for a range).

orlandoc01:
Looks like it :/, someone opened an issue for it #611 too. It would seem that the `validate-defs` command supports directory name formats that the `flow-typed install` command does not, which is why my tests passed but, after attempting to install the library in another repo, the installation process failed. It looks like someone already solved it also by submitting a PR which renames the folder #611

PR #616, opened 2017-01-15
add react-redux v5 (same as v4)

turadg:
Even though the major version bumped, the API didn't change
https://github.com/reactjs/react-redux/releases/tag/v5.0.0

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.30.x-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.30.x-/react-redux_v5.x.x.js
@@ -1,0 +1,86 @@
+import type { Dispatch, Store } from 'redux'
+
+declare module 'react-redux' {
+
+  /*
+
+    S = State
+    A = Action
+    OP = OwnProps
+    SP = StateProps
+    DP = DispatchProps
+
+  */
+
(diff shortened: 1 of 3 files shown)


turbometalskater:
Hello, can you explain how to properly use the connect function using flow types (maybe via code sample)?

I'm not very good with flow, but I can see that the `connect` fn is declared six times with different argument types. How do I properly annotate the connect function?

mull:
I'm having a lot of trouble with the `connect` function as well, constantly getting an error about flow not finding a matching member. Would love an example! Any chance you're sitting on one @turadg ? :)

turadg:
Here's one from our codebase at Remind,
```
function mapStateToProps(state: Object) {
  return {
    rides: state.joyrides,
  };
}

export default connect(mapStateToProps, undefined)(Joyrides);
```

turbometalskater:
@mull , I've found some great samples here https://github.com/reactjs/redux/tree/master/examples/todos-flow which is linked to a gcanti and timdorr PR: https://github.com/reactjs/redux/pull/1887

mull:
@turadg @turbometalskater thanks a lot! 

My only remaining problem now is properly mapping action creators & using bindActionCreators from redux itself. Any of you sitting on an example? T_T

PR #621, opened 2017-01-19
[express@4.x.x] Allow mixed paths

cema-sp:
In [express docs](http://expressjs.com/en/4x/api.html#path-examples) mixed paths (containing `RegExp` along with `string`) are allowed for `get()`, `use()`, etc. but current flow typings restrict them.

Here is the fix providing a new `express$Path` type and using it in `express$RouteMethodType` interface.


GAntoine:
@cema-sp Since this still seems to be the case for the express API, would you mind rebasing this PR or opening a new one? I'd be nice to get this merged.

cema-sp:
@GAntoine Sorry for the long response, I have rebased the PR, let me know if you have any questions

PR #629, opened 2017-01-25
[koa] app.callback should be a function return callback

chentsulin:
https://github.com/koajs/koa/blob/v2.x/docs/api/index.md#appcallback


GAntoine:
Closing as these changes are already in master.

PR #641, opened 2017-02-02
Express: error middleware must have error

idris:
The `error` in error handling middleware functions cannot be null.
See docs: https://expressjs.com/en/guide/error-handling.html


GAntoine:
I read through the docs you linked and couldn't find any mention that `err` cannot be null.

idris:
@GAntoine the whole point of `app.use` with a handler that takes 4 arguments is to setup an error handler.  In those docs you can see that the examples don't check for null:
```js
app.use(function (err, req, res, next) {
  console.error(err.stack)
  res.status(500).send('Something broke!')
})
```
If `err` could be null, `err.stack` would throw an error.  But it doesn't because this middleware is only called when there is an error.

Here's evidence in the Express source code showing how it's used (`fn` in this code is the middleware function passed to `app.use`):
* `handle_error` checks to make sure `fn.length === 4`: https://github.com/expressjs/express/blob/master/lib/router/layer.js#L62-L71
* `handle_request` checks to make sure `fn.length <= 3`: https://github.com/expressjs/express/blob/master/lib/router/layer.js#L89-L92

The two places where `handle_error` is called both check for a non-null error first:
* https://github.com/expressjs/express/blob/c0136d8b48dd3526c58b2ad8666fb4b12b55116c/lib/router/route.js#L134-L135
* https://github.com/expressjs/express/blob/c0136d8b48dd3526c58b2ad8666fb4b12b55116c/lib/router/index.js#L314-L315

GAntoine:
Wow, thanks for the thorough explanation! I was mostly arguing semantics, in that even if accepts 4 parameters, `err` could still theoretically be null, and that the 4 param check would still count `null` as a valid parameter.

However your 2 last links to the source code are exactly what I was looking to see, so as far as I'm concerned this is good to merge. 💯 

As for the build failing, give it 10 minutes to rest and try again.

PR #649, opened 2017-02-07
lodash: Define isString as true only if string type is passed in

smdern:


definitions/npm/lodash_v4.x.x/flow_v0.28.x-v0.37.x/test_lodash-v4.x.x.js
```diff
@@ -115,6 +115,24 @@ _.zip([{x:1}], [{x:2,y:1}])[0][2]
...
+boolTrue  = _.isString('foo');
+boolFalse = _.isString(['']);
+boolFalse = _.isString({});
+boolFalse = _.isString(5);
+boolFalse = _.isString(function(f) { return f });
+boolFalse = _.isString();
+
+// $ExpectError
+boolFalse = _.isString('');
+
```
aaronjensen: Maybe add another negative test for non-string

PR #657, opened 2017-02-11
Add dive() method to enzyme

tricoder42:


ryyppy:
As defined in https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/dive.md this should actually be something like this:

`dive(option?: { context?: Object }): this`

right?

tricoder42:
@ryyppy You're absolutely right! Sorry, I missed that in docs. Patch updated.

PR #665, opened 2017-02-17
Test only changed defs

marudor:
This is my attempt on Testing only the libDefs that got changed.
Please have a look. It is pretty simple.
@jeffmo @ryyppy

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -3,9 +3,10 @@
 import {child_process, fs, os, path} from "../lib/node.js";
 import {copyFile, recursiveRmdir} from "../lib/fileUtils.js";
 import {gitHubClient} from "../lib/github.js";
-import {getLibDefs} from "../lib/libDefs.js";
+import {getLibDefs, parseRepoDirItem} from "../lib/libDefs.js";
 import isInFlowTypedRepo from "../lib/isInFlowTypedRepo";
 import {toSemverString as flowVerToSemverString} from "../lib/flowVersion";
+import {getDiff} from '../lib/git';
 
 import request from "request";
 import * as semver from "semver";
@@ -42,8 +43,24 @@ type TestGroup = {
  * structs. Each TestGroup represents a Package/PackageVersion/FlowVersion
  * directory.
  */
-async function getTestGroups(repoDirPath): Promise<Array<TestGroup>> {
-  const libDefs = await getLibDefs(repoDirPath);
+ const basePathRegex = new RegExp(`definitions${path.sep}npm${path.sep}[^${path.sep}]*${path.sep}?`);
+async function getTestGroups(repoDirPath, onlyChanged: bool = false): Promise<Array<TestGroup>> {
+  let libDefs = await getLibDefs(repoDirPath);
+  const diff = await getDiff();
+  let changedDefs;
+  if (diff) {
+    // $FlowFixMe
+    const baseDiff: string[] = diff.map(d => {
+      const match = d.match(basePathRegex);
+      if (match) {
+        return match[0];
+      }
+    }).filter(d => d != null);
+    changedDefs = baseDiff.map(d => parseRepoDirItem(d).pkgName);
+  }
+  if (onlyChanged) {
+    libDefs = libDefs.filter(def => changedDefs.includes(def.pkgName));
+  }
   return libDefs.map(libDef => {
     const groupID =
       `${libDef.pkgName}_${libDef.pkgVersionStr}${path.sep}` +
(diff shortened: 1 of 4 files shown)


ryyppy:
I like it! So it will trigger on changed tests or it's libdefs? Changes are not intrusive as well

marudor:
Gonna leave it until Thursday. Unless someone else sees any problem I would merge it.

jeffmo:
Nice! So glad to see this going back in...
I'll take a closer look today/tomorrow and see if I can give any feedback

cli/src/lib/libDefs.js
```diff
@@ -159,10 +159,24 @@ async function addLibDefs(pkgDirPath, libDefs: Array<LibDef>, validationErrs?: V
 /**
  * Given a 'definitions/npm' dir, return a list of LibDefs that it contains.
  */
+ const basePathRegex = /definitions\/npm\/[^\/]*\/?/;
 export async function getLibDefs(
```
jeffmo: Rather than adding complexity to this general-purpose (and fairly fast) function, can we just filter the list of libdefs it returns in the `run-tests` command (and only do so during travis runs -- maybe an `--only-changed` flag or something)?

For example: `validate-defs` is very fast, so no good reason to make it not run on all libdefs (but the `run-tests` command is itself very slow, so only running relevant stuff there is a good idea).

cli/src/lib/libDefs.js
```diff
@@ -159,10 +159,24 @@ async function addLibDefs(pkgDirPath, libDefs: Array<LibDef>, validationErrs?: V
 /**
  * Given a 'definitions/npm' dir, return a list of LibDefs that it contains.
  */
+ const basePathRegex = /definitions\/npm\/[^\/]*\/?/;
```
jeffmo: This needs to consider Windows, too. Can you use the `RegExp` constructor + `path.sep`?

cli/src/lib/git.js
```diff
@@ -25,6 +25,19 @@ async function getGitPath() {
...
+export async function getDiff() {
+  const gitPath = await getGitPath();
+  try {
+    const {stdout} = await child_process.spawnP(
+      gitPath,
+      ['diff', '--name-only', 'HEAD~1'],
+    );
+    return stdout.split('\n');
+  } catch (e) {
+    return null;
```
relekang: Would it make sense to return an empty list here and then drop the `if (diff)` in `getTestGroups`?

cli/src/lib/git.js
```diff
@@ -25,6 +25,19 @@ async function getGitPath() {
   }
 };
 
+export async function getDiff() {
+  const gitPath = await getGitPath();
+  try {
+    const {stdout} = await child_process.spawnP(
+      gitPath,
+      ['diff', '--name-only', 'HEAD~1'],
```
relekang: This only checks the diff of the last commit. If we want it to check what differs from origin/master we can get it by 

```shell
git diff --name-only $(git merge-base origin/master HEAD)
```

cli/src/commands/runTests.js
```diff
@@ -42,8 +43,22 @@ type TestGroup = {
...
  */
-async function getTestGroups(repoDirPath): Promise<Array<TestGroup>> {
-  const libDefs = await getLibDefs(repoDirPath);
+ const basePathRegex = new RegExp(`definitions${path.sep}npm${path.sep}[^${path.sep}]*${path.sep}?`);
+async function getTestGroups(repoDirPath, onlyChanged: bool = false): Promise<Array<TestGroup>> {
+  let libDefs = await getLibDefs(repoDirPath);
+  if (onlyChanged) {
+    const diff = await getDiff();
+    let changedDefs;
+    // $FlowFixMe
```
jeffmo: I'm curious: What's under the FixMe here?

jeffmo: Oh, it's probably the .map().filter() 👍

cli/src/commands/runTests.js
```diff
@@ -42,8 +43,22 @@ type TestGroup = {
...
+    const baseDiff: string[] = diff.map(d => {
+      const match = d.match(basePathRegex);
+      if (match) {
+        return match[0];
+      }
+    }).filter(d => d != null);
+    changedDefs = baseDiff.map(d => parseRepoDirItem(d).pkgName);
```
jeffmo: nit: Maybe make `changedDefs` a dict or a map so the `.filter()` below isn't exponential?

jeffmo:
I think this looks good. The only new risk this presents is that it means if one libdef implicitly depends on a global type of another, we won't catch new downstream errors in the blame diff.

We need to address explicit dependency tracking between libdefs at some point anyway, and the current wait times of Travis are bad enough that it's probably worth just dealing with that every so often -- so I think this is acceptable risk.

PR #666, opened 2017-02-20
express Request.hostname is a string, not a boolean

idris:


philikon:
Bonus points if you could add a test.

idris:
@philikon I'm not sure what I would test here.  A test would just make sure that `hostname` is a string, but that's just a copy of the line I changed, so the test really isn't providing anything other than duplicate code.

PR #677, opened 2017-02-27
Introduce jest_v19.x.x

jedmao:
## Added the following new features from the [Jest 19 Changelog](https://github.com/facebook/jest/blob/master/CHANGELOG.md#jest-1900):

- [x] New `expect.addSnapshotSerializer` to add custom snapshot serializers for tests.
- [x] New `jest.spyOn`.
- [x] New expect.stringContaining asymmetric matcher.

### DIFF
--- a/definitions/npm/jest_v19.x.x/flow_v0.33.x-/jest_v19.x.x.js
+++ b/definitions/npm/jest_v19.x.x/flow_v0.33.x-/jest_v19.x.x.js
@@ -419,7 +419,7 @@ declare var expect: {
   objectContaining(value: Object): void,
   /** Matches any received string that contains the exact expected string. */
   stringContaining(value: string): void,
-  stringMatching(value: string): void,
+  stringMatching(value: string | RegExp): void,
 };
 
 // TODO handle return type
--- a/definitions/npm/jest_v19.x.x/flow_v0.33.x-/jest_v19.x.x.js
+++ b/definitions/npm/jest_v19.x.x/flow_v0.33.x-/jest_v19.x.x.js
@@ -1,0 +1,446 @@
+type JestMockFn = {
+  (...args: Array<any>): any,
+  /**
+   * An object for introspecting mock calls
+   */
+  mock: {
+    /**
+     * An array that represents all calls that have been made into this mock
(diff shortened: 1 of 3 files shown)


definitions/npm/jest_v19.x.x/flow_v0.33.x-/jest_v19.x.x.js
```diff
@@ -0,0 +1,446 @@
...
+  /** Add a module that formats application-specific data structures. */
+  addSnapshotSerializer(serializer: (input: Object) => string): void,
+  assertions(expectedAssertions: number): void,
+  any(value: mixed): JestAsymmetricEqualityType,
+  anything(): void,
+  arrayContaining(value: Array<mixed>): void,
+  objectContaining(value: Object): void,
+  /** Matches any received string that contains the exact expected string. */
+  stringContaining(value: string): void,
+  stringMatching(value: string): void,
```
Judahmeek: Shouldn't this be `stringMatching(value: string | RegExp): void,` to match https://github.com/facebook/jest/blob/master/packages/jest-matchers/src/asymmetric-matchers.js#L216 (Not that either given parameter types match [Jest's documentation](https://facebook.github.io/jest/docs/expect.html#expectstringmatchingregexp) exactly)?

jedmao: Totally agree; though, this wasn't at all mentioned in the changelog, so I'd say this is pre-existing. Perhaps another PR to fix it not just in Jest 19 but also in 18 would be appropriate.

jedmao: It looks like [the original implementation](https://github.com/facebook/jest/commit/f597c07c78aa9dfc525a2e905909ff5bbb9f7afd#diff-4050ceab41127d98859329b8db36bb7aR420) should have also been `string | RegExp` since version 18.1.0.

jedmao: added in 7975993

PR #681, opened 2017-03-03
Lodash: Fix issue in type of result of _.keyBy

demands:
[`_.keyBy`](https://lodash.com/docs/4.17.4#keyBy) returns an object with a finite set of values, which means that sometimes accessing an attribute will result in `undefined` instead of the result you may have expected.

Thanks for this project, it's been fantastic for us so far. Please let me know if there's anything I can do to make this PR better! :)


philippotto:
I don't really agree with this change. From the [flow docs](https://flow.org/en/docs/types/objects/#objects-as-maps-a-classtoc-idtoc-objects-as-maps-hreftoc-objects-as-mapsa):

> When an object type has an indexer property, property accesses are assumed to have the annotated type, even if the object does not have a value in that slot at runtime. It is the programmer’s responsibility to ensure the access is safe, as with arrays.

Your change suggests that lodash might write undefined values for keys returned by the iteratee function, but clearly this does not happen. According to your reasoning, arrays can only hold optional types, since arrays are finite, too.

PR #682, opened 2017-03-06
add Context argument to StatelessComponent

faceyspacey:


faceyspacey:
any word on getting this merged? When packages have flow errors and they are ignored with `$flowIgnore` they won't be ignored in user apps unless they have the same directive. 

So that means any packages using stateless components accepting a 2nd `context` argument (whose error is ignored) will show up in user apps that dont have `$flowIgnore`.

PR #683, opened 2017-03-06
Add definition for undescore.js delay

mudetroit:
## What this does
Adds in a definition for the underscore `delay` function. Mirrored off of the existing `defer` and `throttle` definitions.

## Notes/caveats
- I attempted to add an $ExpectedError line to the underscore test with the following call:
```
_.delay(() => {}); // No delay specified
```
But I was getting an 'Unused suppression error. Not certain what I was doing wrong there.

- I would prefer to not use the `Array<any>` function here, and figure out a way to match the arguments to the function passed in, but that requires a little more thought.


samMeow:
looks like `wait: number` is optional
```js
let greet = () => console.log("hello");
_.delay(greet) //it will be treat as setTimeout(greet, 0)
```

samMeow:
_.delay also return `number` EventId
can tighten the rules from `void` to `number`

PR #698, opened 2017-03-13
Remove trailing spaces and new lines from git commit

jkimbo:
I noticed a bug when running `run-tests --onlyChanged` for https://github.com/flowtype/flow-typed/pull/697 where it doesn't seem to pick up any changed files. Traced the issue to the fact that the command that gets the current master commit doesn't trim the trailing new line and so the diff command fails (it also fails silently which I'm not sure is desired but I didn't want to change it).

Anyway this PR trims the new lines so that the command works as expected.

### DIFF
--- a/cli/src/lib/git.js
+++ b/cli/src/lib/git.js
@@ -28,10 +28,13 @@ async function getGitPath() {
 export async function getDiff() {
   const gitPath = await getGitPath();
   try {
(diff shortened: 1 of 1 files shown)


jkimbo:
Looks like this has already been fixed (https://github.com/flowtype/flow-typed/commit/aa2965c7916d27039ce8b818031b04a3d3e54d50)

PR #703, opened 2017-03-14
Add gl-matrix

Platane:


Platane:
As the lib methods are quite numerous but yet simple, I used a codemon to read the type from the well formatted jsdoc block.

I guess project to do so already exists, but I wrote my own: https://gist.github.com/Platane/5b2fc3877ea64d8d9f65d810424e830b

( In case you may be interested )

PR #704, opened 2017-03-15
Add assert.throws to chai definitions

kalman:
See http://chaijs.com/api/assert/#method_throws.


kalman:
Also have a look at https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js#L1626

kalman:
Come to think of it, this should match `ExpectChain.throw`, but the definition for `throw` is wrong. Give me a moment and I'll fix both.

jonfreedman:
What about doesNotThrow as well?

kalman:
Right - it's documented there, but I actually don't see it in the chai source code - https://github.com/chaijs/chai/blob/master/lib/chai/core/assertions.js. Am I missing something?

kalman:
And never mind, I found it in a different file. https://github.com/chaijs/chai/blob/master/lib/chai/interface/assert.js#L1788

Issue #341, opened 2016-09-28
Support for react-router 4


aaronjensen:
It's alpha right now so we may not want it here, but I'm opening this issue to at least share the definitions I have so far for 4.0.0-alpha.3

https://gist.github.com/aaronjensen/48fcfaa0d540805d30fa18114c224f3f

jeffmo:
Nice. `flow-typed` is pretty strict about the versioning patterns that are allowed, and I think `maj.min.ptc-alpha` versions aren't currently supported yet -- but we probably could/should support that!

jeffmo:
Linking this in to https://github.com/flowtype/flow-typed/issues/392 to track support for prerelease identifiers

aaronjensen:
Works for me, if #714 is the only API change so far then awesome. @jeffmo if you could merge #714 in asap that'd be great, it's a breaking change for the release version.

PR #714, opened 2017-03-20
Correct react-router ContextRouter definition

kouak:
React-router v4 will inject specific props in components. My understanding of these definitions is that these props were defined as `ContextRouter`.

The definition of `ContextRouter` wasn't correct. This PR replaces `ContextRouter` with a definition that follows [React-router docs](https://reacttraining.com/react-router/web/api/Route/Route-props).

Rel: #341


kouak:
Partially merged via #820

PR #722, opened 2017-03-24
Fix moment_v2.x.x.js typedef for utcOffset()

nvie:
Fixes `utcOffset()` typedef for moment_v2.x.x.js.  This was already started in #519, but this adds the correct definition and a test case for it.


ryyppy:
Cool, thanks! :+1:

PR #736, opened 2017-03-26
Fixed all remaining problems on the windows platform now allowing to run run_def_tests.cmd

doberkofler:
* only use flow-bin starting with version 0.30.0 as there was no windows support before
* use node module to extract from zip file because unzip must not be available on windows
* fixed problem with file renaming
* fixes issue https://github.com/flowtype/flow-typed/issues/735


marudor:
Nice. This one will take me a while though. I want to make sure it runs. So I can't rely on travis here.
May take me until next week. Thanks though!

doberkofler:
@marudor I basically continued to fix issues while constantly running the tests on the windows platform and am quite confident that it should work. I'm sure, that there are edge cases that I missed and it's great if you can double check. Generally speaking, it did not work at all on Windows before.

PR #738, opened 2017-03-27
add `defaults` to axios

deecewan:
The `defaults` key is not on the [axios](https://github.com/mzabriskie/axios) instance.

`defaults` is defined in the [TypeScript definition](https://github.com/mzabriskie/axios/blob/master/index.d.ts#L98), so I just used the same definition


quanganhtran:
could someone please give an update on this?

deecewan:
sorry @quanganhtran. I'll make the tests pass today, and then hopefully it will get a review.

sebastienbarre:
I tried using your patch, I'm getting this error when running flow:
```
Library type error:
flow-typed/npm/axios_v0.16.x.js:99
 99:     defaults: AxiosXHRConfig<T>
                                  ^ identifier `T`. Could not resolve name
```

deecewan:
yup.  it's not working right now, and i haven't had time to fix it.

hopefully i'll fix it soo.

sebastienbarre:
Thanks @deecewan, would be helpful.

definitions/npm/axios_v0.15.x/flow_v0.28.x-/axios_v0.15.x.js
```diff
@@ -93,6 +93,7 @@ declare module 'axios' {
       request: AxiosRequestInterceptor<mixed>,
       response: AxiosResponseInterceptor<mixed>,
     };
+    defaults: AxiosXHRConfig<T>
```
quanganhtran: Based on
https://github.com/deecewan/flow-typed/blob/master/definitions/npm/axios_v0.15.x/flow_v0.28.x-/axios_v0.15.x.js#L83
and
https://github.com/mzabriskie/axios/blob/master/lib/core/Axios.js#L16
I'm compelled to say that `defaults` is an instance of `AxiosXHRConfigBase`
I have no idea about what to put into the type parameter, but based the request and response interceptors I think it might be `mixed`

sebastienbarre: I just started with flow, so I'm afraid I might not be the most helpful here...

sebastienbarre: Maybe we can try to ask past contributors?

@marudor sorry to bother, I see you worked on previous iterations of the axios interface definition -- if you have a minute do you think you could point us in the right direction? Thank you.

marudor:
I've merged master into this PR and changed the libdef to (I hope) match the implementation.
Basically overwritten the defaults to always have the header object. The others should be fine as nullable.

sebastienbarre:
@marudor to the rescue, thanks a lot, that worked for axios!

deecewan:
Thanks a lot!

PR #745, opened 2017-03-29
Add libdefs for styled-components

relekang:


favll:
I just started using `styled-components` and `flow` yesterday, however, it seems that the current `ThemeProvider` error is easily fixed by adding `typeof` to `module.exports`:

```
declare module.exports: {
  ...
  ThemeProvider: typeof ThemeProvider,
  ...
}
```

After adding this there still remains an issue with interpolation in tagged template literals. This might be related to facebook/flow#2616. Maybe someone else knows how to deal with this issue?

relekang:
@favll Thanks, looks like it works now :)

favll:
I'm still getting a `Not covered by Flow` warning when using an expression interpolation with a tagged template literate. Flow does not seem to know what tagged template literates evaluate to. 

So while the current libdef is correct, shouldn't an alternative definition be shipped until flow fully supports expression interpolations and tagged template literates?

relekang:
Do you have an example? For this snippet I get 100% coverage

```es6
// t.js
import styled from "styled-components"

const Center = styled.div`
  text-align: center;
`

export default () => <Center>👋</Center>
```

```
$ flow coverage t.js
Covered: 100.00% (6 of 6 expressions)
```

favll:
```javascript
// @flow
//
import styled, { ThemeProvider } from 'styled-components'

const theme = {
  background: "papayawhip"
}

const Title = styled.h1`
  background-color: ${({theme}) => theme.background};
`;

const Component = () => (
  <ThemeProvider theme={theme}>
    <Title>Hello World, this is my first styled component!</Title>
  </ThemeProvider>
)
```
```
$ flow coverage sc-test.js
Covered: 80.00% (12 of 15 expressions)
```

favll:
I just looked at this again and I noticed that the problem is not with tagged template literates but rather with object destructuring. Flow does not know that the object passed to the expression interpolation has a `theme` property. And even if it would know that the object had a theme property it wouldn't know that it has the type `Theme`. So basically Flow doesn't know that ThemeProvider injects `theme` into its children and it doesn't know that the props passed to a styled component can be used in the expression interpolation.

relekang:
Ah, that makes sense. I have no idea if we can creates libdefs that fixes that?

jacobk:
Thanks for the defs!

One thing, this doesn't seem to work with stateless/function components being passed to `styled(...)`.

E.g. 

```javascript
const Wrapper = styled(({ height, ...props }) => <SomeComponent {...props} />)`
  height: ${(props) => props.height}px;
`;
```

results in
```
 33: const Wrapper = styled(({ height, ...props }) => <SomeComponent {...props} />)`
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ arrow function. This type is incompatible with the expected param type of
 31:     (baseComponent: React$Component<*, * , *>): StyledComponent,
                         ^^^^^^^^^^^^^^^^^^^^^^^^^ React$Component. See lib: flow-typed/npm/styled-components_v1.4.x.js:
```
which is a very common idiom to prevent props used for styling being passed to underlying component (think that is fixed in v2)

jacobk:
I've changed it to
```
(baseComponent: React$Component<*, *, *> | (props: any) => ?React$Element<*>): StyledComponent,
```

for now but there's probably better ways to do that.
E.g. looking at https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-redux_v5.x.x/flow_v0.30.x-/react-redux_v5.x.x.js#L33-L36 which seems like a similar concert, but I don't understand that declaration.

dnepro:
Is there anything planned for styled-components v2.0?

DarrylD:
any progress on this?

PR #750, opened 2017-03-31
_.find may return void

tempname11:
As per [the lodash docs](https://lodash.com/docs/4.17.4#find): `_.find ... Returns the matched element, else undefined.`

The type signature of the declaration didn't match that, causing potential `undefined` errors to be ignored. Same applies for the lesser-known `findLast`.

If this is merged, I guess a lot of people's code will have new errors once they get the updated libdefs. But correctness is obviously more important.


vaskevich:
There are likely a few other functions that can also return `undefined`, including `_.first`, `_.head`, and `_.last`.

PR #760, opened 2017-04-03
Add libdef for Redux Forms and Redux Forms React

oreqizer:
https://github.com/oreqizer/redux-forms


oreqizer:
should I change something?

PR #764, opened 2017-04-04
Update bluebird_v3.x.x Bluebird$BluebirdConfig

EvHaus:
According to `http://bluebirdjs.com/docs/api/promise.config.html`, `Promise.config()` doesn't just take a boolean `warning` key but instead can take an object with a `wForgottenReturn` key. This should fix that.


GAntoine:
@EvHaus Since this change is correct (and still relevant!), if you're keen to rebase it and fix any broken specs, we can probably get it merge in.

EvHaus:
@GAntoine I've rebased and all looks good. Tests are failing but it's not due to this change. The tests are failing due to this bug which is still open: https://github.com/flowtype/flow-typed/issues/1272

GAntoine:
Closing in favour of #764

EvHaus:
In favor of https://github.com/flowtype/flow-typed/pull/1457 actually. :)

PR #775, opened 2017-04-12
Fix next.js libdef issues and strengthen render typings

keplersj:
Ref facebook/flow#3685


steida:
Please, merge it. It's a bummer.

PR #787, opened 2017-04-17
Enable underscore `_.reduce` on objects

laurensclaessen:
`_.reduce` is a collection method, so both arrays and objects should be taken care of.


sibelius:
@laurensclaessen tests are failing, could you fix them?

laurensclaessen:
@sibelius, they aren't failing on my end (locally). Do you think it could be due to the latest version of Flow and missing files? Haven't had the time to investigate it yet.

Travis is showing error below, nothing test specific:
`Could not locate flowlib files`

GAntoine:
Closing, since versions of the dev for flow 38+ have support for `Object` on `reduce`. If you really want to, you could submit a new PR for flow 15-37, since that range only supports `Array`.

PR #799, opened 2017-04-24
fixed bug with invalid npm scope in install script

vadimzzz:
I tried to install libdef for axios library (0.15.3) but flow-typed could find anything to install despite existing definition in repositiory (0.15.x). My flow version (0.40.0) also matches. So I started to inspect the code and saw that the name of the package is passed as "undefinedaxios" into search library. I tried to find the source of the bug and here is it.

### DIFF
--- a/cli/src/commands/install.js
+++ b/cli/src/commands/install.js
@@ -187,7 +187,7 @@ async function installNpmLibDefs({
       }
 
       const [_, npmScope, pkgName, pkgVerStr] = termMatches;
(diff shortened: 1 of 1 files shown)


marudor:
@jeffmo a new release would be nice.

jeffmo:
Thanks for sending this over! I'll go ahead and send out a patch release later today

jeffmo:
Ok, 2.1.2 is out with this now

PR #800, opened 2017-04-24
Add 'clean' function to semver definition

callumlocke:
This function is mentioned in the [usage example](https://github.com/npm/node-semver) but not otherwise documented.

Here is the [source](https://github.com/npm/node-semver/blob/8fff3050401397e99aaf7a118015c0e33884c3f8/semver.js#L264-L268):

```js
exports.clean = clean;
function clean(version, loose) {
  var s = parse(version.trim().replace(/^[=v]+/, ''), loose);
  return s ? s.version : null;
}
```

It always returns either a version string or `null`.


GAntoine:
Closing since `clean` was added with the complete api on #1049.

PR #804, opened 2017-04-25
RxJS: Add last and mergeMap

BarryThePenguin:
mergeMap [reference](https://github.com/ReactiveX/rxjs/blob/e69d87681008a0ec1a954309c677dde812f713ef/src/operator/mergeMap.ts#L10-L11)

last [reference](https://github.com/ReactiveX/rxjs/blob/e69d87681008a0ec1a954309c677dde812f713ef/src/operator/last.ts)


marudor:
Could you add tests for the new Methods?

PR #813, opened 2017-04-29
change req.body to any

haasDev:
I know we want to avoid the any type when possible.  However using the mixed type consistently throws errors when attempting to pull values out of req.body.

### DIFF
--- a/definitions/npm/express_v4.x.x/flow_v0.32.x-/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_v0.32.x-/express_v4.x.x.js
@@ -13,7 +13,7 @@ declare class express$RequestResponseBase {
 
 declare class express$Request extends http$IncomingMessage mixins express$RequestResponseBase {
   baseUrl: string;
-  body: mixed;
+  body: any;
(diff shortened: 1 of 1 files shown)


masad-frost:
I think it should be completely removed as it's not part of express.

For express you'll always need to define your own request/response depending on the middlewares you're using.

philikon:
> I know we want to avoid the any type when possible. However using the mixed type consistently throws errors when attempting to pull values out of req.body.

`any` is a cancerous type. It disables all type checking for anything that touches that value. So yeah, of course it fixes your errors, but that's only because it wallpapers over them. You can very easily ensure you *don't* have errors, e.g.

```js
const invariant = require('assert');

/// later somewhere:
const {body} = req;
invariant(body instanceof Buffer);
// now flow knows that body is a buffer, so you can do buffer-y things to it.
```
We should avoid `any` like the plague. I don't understand how this kind of change can pass code review.

> I think it should be completely removed as it's not part of express. https://expressjs.com/en/api.html#req.body

Welllll, yes and no. Express explicitly supports it as an API, but it leaves the value up to a middleware.

PR #814, opened 2017-05-02
styled-components: Fix keyframes definition

jameskraus:
https://github.com/styled-components/styled-components/issues/686


relekang:
👍 Can you add an example to the test file?

jameskraus:
@relekang - Added an example of using keyframes

PR #686, opened 2017-03-07
Add an option to skip stubs when installing

lime:
This is an attempt to resolve #554.

This would allow you to run e.g. `flow-typed install --overwrite --skipStubs` to update existing libdefs, install new ones, and leave (potentially modified) stubs untouched.

PR #817, opened 2017-05-03
install.js: add option to skip stub generation

rufman:
When `--skip` is passed, untyped dependencies won't be stubbed. This
adds back the functionality from #686 that was lost.


marudor:
@jeffmo could you release a new Version with skip enabled again?

PR #821, opened 2017-05-05
Add definitions for other Observable methods

mcortesi:
Observable.timer:
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-timer

Observable.repeat:
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-repeat


mcortesi:
I don't know why travis is failing. 

I tried running it on my computer `npm run clean && npm run build && npm run test-quick` and everything's green!

marudor:
I've merged master into this - didn't really do anything.
I'Ve got no Idea what the problem is.
@jeffmo could you take a look?

GAntoine:
This may be a dumb comment, but does it have to do with `const numbers` being defined twice in the same scope (test_rxjs.js, line 5 and 20)?

marudor:
Yes it does have to do with that.
Guess I missed that.
Fixed and merged.

Issue #384, opened 2016-10-17
Does anyone already have flow-typed definitions for jQuery and jQueryUI ?


doberkofler:

marudor:
You can find jquery here:
https://github.com/marudor/flowInterfaces

doberkofler:
@marudor Thank you. Excellent work. Two questions:
- will you move the definitions to flow-typed?
- do you also have definitions for jQurey plugins especially for jQuery UI?

marudor:
@doberkofler Yes, the plan is to move it to flow-typed.
No, I do not have any plugins typed.

The one in my repository was a PR. Didn't type it myself.

doberkofler:
@marudor Would you know how to actually extend the existing jQuery definitions (https://github.com/marudor/flowInterfaces) without actually having to copy and paste them together when jQuery plugins must be defined? I cannot seem to find a problem solution in flowjs.

doberkofler:
@marudor Any news on when this will be available?

marudor:
Once someone writes Tests for it.
I'm busy and haven't been able to yet.
Feel free to copy paste it and get it in here. (With tests)

doberkofler:
@marudor I started working on this and am currently adding tests. Unfortunately there are quite a few issues with the existing definition and would need some help to better understand the library definitions.

I'm for example unable to create tests for the with functions:

The tests look like this:
```
//width(): number;
testNumber = $('div').width();

//width(value: number | string): JQuery;
testSelector = $('div').width(50);

//width(func: (index: number, width: number) => number | string): JQuery;
testSelector = $('div').width(function (index) {
	return index * 50;
});
```

and I get the following error message:
```
0-test_jquery.js:471
471: testSelector = $('div').width(50);
                    ^^^^^^^^^^^^^^^^^^ number. This type is incompatible with
 22: let testSelector: JQuery = $('div');
                       ^^^^^^ JQuery

0-test_jquery.js:474
474: testSelector = $('div').width(function (index) {
                    ^ number. This type is incompatible with
 22: let testSelector: JQuery = $('div');
                       ^^^^^^ JQuery
```

Could you help me to understand what's wrong here?
Is it generally allowed to "overload" function definitions?
Whom else might I contact to get some help on this?

Thank you!

GAntoine:
Closing as the PR above added a jquery definition.

PR #835, opened 2017-05-12
Added definitions for jquery. (#384)

doberkofler:


marudor:
Nice.
It looks pretty good. But could someone who is using jQuery check it on their codebase? Would like to get some feedback before merging it.

doberkofler:
@marudor I'm actually using the type definitions in a very large application. Currently i need to find a way on how to deal with the fact that I still have not found a way on how to extend the existing jquery interfaces with all the plugins that i need without having to modify the definition file itself . Any help would be appreciated.

doberkofler:
@marudor I did some quite extensive testing with the new definitions and I actually did find a few minor problems that I've already fixed.

doberkofler:
@marudor Anything else I can do, to get this merged? (I spend lot of time on this)

PR #836, opened 2017-05-12
Add support for styled-components/native

joeybaker:
The original type definition omitted support for the React Native version of styled-components. This adds support for that module.

Fixes https://github.com/styled-components/styled-components/issues/708


joeybaker:
Thanks @marudor

PR #842, opened 2017-05-15
Add redux-mock-store type definitions

lifeiscontent:
First time submitting a type-def, please feel free to give feedback as I'm not sure if I'm doing this 100% right.


lifeiscontent:
@marudor is there someone I should ping about getting this in?

PR #865, opened 2017-05-25
Backbone partial modules

turadg:
https://www.npmjs.com/package/backbone-model and https://www.npmjs.com/package/backbone-collection provide subsets of https://www.npmjs.com/package/backbone

This adds rules for those. And fixes some in the existing `backbone` one.

Is there a way to DRY this out?


GAntoine:
@turadg If you'd still like to see this merged it, you'll need to address the failing specs!

PR #868, opened 2017-05-25
Add DragDropContextProvider to react-dnd_v2.x.x

randing89:
Add missing definition of DragDropContextProvider

Reference:
https://github.com/react-dnd/react-dnd/blob/master/docs/01%20Top%20Level%20API/DragDropContextProvider.md


amccloud:
@randing89 this seems useful, why did you close this?

PR #875, opened 2017-05-26
Accept array for lodash _.get first argument (and fix _.isEqual tests for v0.47.x)

jamesshedden:
**What**
- Updating the Lodash libdef to accept an array for `_.get` first argument.
- Adding new `flow_v0.47.x-` folder to Lodash with new tests for `_.isEqual`

**Why**
Although Lodash groups its `_.get` method with other 'Object' methods (separately from 'Array' methods) & only documents its usage as pertains to object literals, it does accept arrays. There are no plans to extend the documentation to make this explicit (see https://github.com/lodash/lodash/issues/3038) due to arrays being considered ultimately to be objects, however since `Object` & `Array` are distinct types in Flow, we should extend the current definitions for `_.get` to accept an array for the first argument.

Also, it seems that Lodash tests are currently failing because they expect `_.isEqual` to accept more than two arguments, but it appears this is not supported for Flow `0.47.x` — not strictly part of what I originally intended for this PR, but had to update this to get my tests to pass 😇

**Changes**
- Updating existing Lodash libdefs to support arrays for `_.get` first argument
- Adding specs to existing Lodash test files for `_.get`
- Adding a new `flow_v0.47.x-` folder for lodash & as a result renaming the previous `flow_v0.38.x-` to `flow_v0.38.x-v0.46.x`
- Adding new `$ExpectError` comment to the `flow_v0.47.x-` `_.isEqual` tests — apart from this, the libdef & tests are identical to `flow_v0.38.x-v0.46.x`

**References**
Related issue at https://github.com/flowtype/flow-typed/issues/854.


jamesshedden:
Taking a look at failing specs 👀

definitions/npm/lodash_v4.x.x/flow_v0.47.x-/test_lodash-v4.x.x.js
```diff
@@ -0,0 +1,348 @@
...
+_.isEqual('a', 'b');
+_.isEqual({x: 1}, {y: 2});
+
+// Flow considers this compatible with isEqual(a: any, b: any).
+// Reasonable people disagree about whether this should be considered a legal call.
+// See https://github.com/splodingsocks/FlowTyped/pull/1#issuecomment-149345275
+// and https://github.com/facebook/flow/issues/956
+_.isEqual(1);
+
+// $ExpectError function type expects no more than 2 arguments
```
jamesshedden: This is only addition to this test file, otherwise is identical to the Lodash test file for `flow_v0.38.x-v0.46.x`

PR #876, opened 2017-05-26
Accept string as valid collection type in lodash’s _.size

sdg9:
Per [lodash size docs](https://lodash.com/docs/4.17.4#size) `string` is a valid type for the collection argument


GAntoine:
@sdg9 Closing as the lodash defs have change quite a lot since this opened. The current definition is still missing `string` as a valid collection type, so if you're still interested in adding it, a new PR would be mucho appreciated!

PR #886, opened 2017-05-29
Handle Function in redux-actions/handleActions

tricoder42:
`handleActions` in `redux-actions` can handle anything with `toString()` method as a action type (https://github.com/acdlite/redux-actions/blob/master/src/handleAction.js#L11). In fact, actions created using `createAction` has `toString()` defined to return action type (https://github.com/acdlite/redux-actions/blob/master/src/createAction.js#L39).

I extended type to `string | Function`, because these two are the most common (either action type or action itself).

### DIFF
--- a/definitions/npm/redux-actions_v2.x.x/flow_v0.34.x-/redux-actions_v2.x.x.js
+++ b/definitions/npm/redux-actions_v2.x.x/flow_v0.34.x-/redux-actions_v2.x.x.js
@@ -81,7 +81,7 @@ declare module 'redux-actions' {
   ): Reducer<State, Action>;
 
   declare function handleActions<State, Action>(
-    reducers: { [key: string]: Reducer<State, Action> | ReducerMap<State, Action> },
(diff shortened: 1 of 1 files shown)


hlomzik:
But Flow still throws error "Computed property cannot be assigned with function type". How did you get around this?

akkie:
I've the same "Computed property cannot be assigned with function type" issue. A workaround is to call `toString` directly on the function.

```js
const noop = createAction('INCREMENT');

const reducer = handleActions({
  [noop.toString()]: (state, action) => ({
    counter: state.counter + action.payload
  })
}, { counter: 0 });
```

ryuone:
I also still throws error `object literal. Computed property cannot be assigned with function type`
, after apply patch.
Is there any idea to resolve this?

GAntoine:
@tricoder42 I can't merge this in while the specs are failing and there's disagreement on the correctness of the fix. If you're still interested in getting this merged, please rebase and fix the broken specs.

PR #902, opened 2017-05-31
Only test against latest 10 flow versions

marudor:
Doing a full test takes ages and does not really give any benefits.
I doubt someone still uses flow 0.25 for instance.
So lets just test the last 10 Versions. This should be enough.

I've talked with @jeffmo about this - with this we also only test against the latest patch version.

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -84,7 +84,7 @@ async function getOrderedFlowBinVersions(): Promise<Array<string>> {
     _flowBinVersionPromise = (async function() {
       console.log("Fetching all Flow binaries...");
 	  const IS_WINDOWS = os.type() === "Windows_NT";
-      const FLOW_BIN_VERSION_ORDER = [];
+      let FLOW_BIN_VERSION_ORDER = [];
       const GH_CLIENT = gitHubClient();
       const QUERY_PAGE_SIZE = 100;
       const OS_ARCH_FILTER_RE = new RegExp(BIN_PLATFORM);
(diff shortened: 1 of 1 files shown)


sibelius:
this would improve a lot the work flow

@jeffmo can we merge this?

jeffmo:
Sorry to have left this PR hanging for so long -- I think this is probably reasonable (cc @calebmer who I think is thinking more about this these days than I am?)

If flow ships a new version every 2 weeks, that puts the oldest supported version at ~5 months. It might be nice to not include non-major versions in that 10-count, though

GAntoine:
This was recently aggravated by #1053, and I'm starting to see the pain of running tests with an extra ~15 flow versions to test. I really hope people are passing `--onlyChanged` or specifying which libdef(s) to test when `run-tests`ing locally, or else they might be in for a long wait...

Is there somewhere we still run a full test? I don't think build on Travis are as relevant to this issue, and they always use `--onlyChanged` either way.

calebmer:
Hmm, I do know people who are running as far back as 36 😣

Is there a smarter way to do this? For instance looking at what changed in git and then only running the versions where type definitions changed. Or only running one or two versions locally, but running everything against the repo daily? (So PRs could merge if they don’t break the last 3 or so versions, but we still have insight into older versions that are breaking.)

We could also freeze type definitions for 5 versions back. So if someone wanted to change a type definition for a `flow_v0.23.x-` they would need to raise the version. Then we could guarantee that there are no errors in older versions, while not needing to test them.

villesau:
what about making a command line argument allowing you specify the flow version manually by your self? You can test against the latest version when developing, and occasionally run the full test suite to see where things start to break. `flow-typed run-tests my-libs --flow-version 0.59.0` or `flow-typed run-tests my-libs --flow-version >0.53.0` which would run flows starting from 0.53.0

GAntoine:
@marudor Closing this as your latest changes do the same thing.

PR #906, opened 2017-05-31
Adding explicit export of using()

moberemk:
When using CommonJS modules using syntax as seen below, Flow will actually fail to recognize the export of the static `using` method, giving the error `property `using` (Property not found in CommonJS exports of "bluebird")`. This makes the library annoying to use in non-ES6 environments; by adding a copy of the function declaration to the explicit exports instead of implicitly relying on the static function, Flow will now accept the existence of the function.

```javascript
const Promise = require('bluebird');
const using = Promise.using;
```


moberemk:
Upon further experimentation to resolve a different error, I found that this actually works if you do the following:

```javascript
const Promise = require('bluebird').default;
const using = Promise.using;
```

Not ideal behaviour but until the project can be migrated to Babel/some other way of supporting ES6 I guess it's what I have to do; and it makes more sense than my proposed change because this will also encompass all static properties of Bluebird.

PR #916, opened 2017-06-02
Extend firebase v4 for other older flow versions

stevenyap:
This commit extends the typing for firebase v4 to flow version 0.34.
The versioning of the lib type def is also renamed to v4.x.x instead of
v4.0.0 for better compatibility to future minor versions of firebase v4.

This commit also fixes two small bugs with Query.remove() and
ServerValue.TIMESTAMP()


stevenyap:
@PCreations Thanks for contributing the Firebase type def!

I have added some fixes and also extended the type def for older flow versions at #916 

I am also considering renaming User to FirebaseUser as User seems a little too imposing on the global namescope... same goes for App. Would love to know your thoughts on this.

PCreations:
@stevenyap Totally agree, to be honest I said to myself the same thing just after pushing so...Your contribution is welcomed :)

definitions/npm/firebase_v4.x.x/flow_v0.46.x-/firebase_v4.x.x.js
```diff
@@ -0,0 +1,410 @@
+/* @flow */
+
+/**** firebase ****/
+
+declare interface FirebaseConfig {
+  apiKey: string;
+  authDomain?: string;
+  databaseURL: string;
+  projectId: string;
```
kennethtruong: Just wondering is projectId needed? Inside https://firebase.google.com/docs/web/setup it doesn't mention projectId

definitions/npm/firebase_v4.x.x/flow_v0.46.x-/firebase_v4.x.x.js
```diff
@@ -0,0 +1,410 @@
...
+  setCustomParameters(customOAuthParameters: Object): this;
+}
+
+declare type OAuthProvider = | FacebookAuthProvider
+                             | GithubAuthProvider
+                             | GoogleAuthProvider
+                             | TwitterAuthProvider;
+
+/****** database ******/
+declare type FirebaseValue = {} | Array<FirebaseValue> | string | number | boolean | null;
```
kennethtruong: I wonder if this should just be type `any`.

definitions/npm/firebase_v4.x.x/flow_v0.46.x-/firebase_v4.x.x.js
stevenyap: some of the example code in firebase does show projectid and I believe that is only recently introduced. anyway since firebase does not treat it as mandatory, we should not treat it as mandatory too... feel free to send a PR to make projectId optional

definitions/npm/firebase_v4.x.x/flow_v0.46.x-/firebase_v4.x.x.js
stevenyap: i don't think this should be `any` as it cannot be `undefined`

but i think the `{}` might be better typed as:
`declare type FirebaseValue = {[key: string]: FirebaseValue} | Array<FirebaseValue> | string | number | boolean | null;`

stevenyap: on further thoughts, FirebaseValue cannot be `any` because it cannot be a function or a class or an instance

kennethtruong: I was looking how typescript people also set it as. It seems they made it any https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/firebase/index.d.ts#L20

Also working with nested values coming from val() makes it really hard to work with. Especially when you pass the values to different functions and want to be a type you specified in your project.

stevenyap: the nested values of val() is captured by `{}` so there shouldn't be any issue with it... perhaps you can provide an example where the nested value is an issue?

definitions/npm/firebase_v4.x.x/flow_v0.46.x-/firebase_v4.x.x.js
kennethtruong: Ya I realize that when looking at the firebase sdk their typescript does show the projectId as optional.

definitions/npm/firebase_v4.x.x/flow_v0.46.x-/firebase_v4.x.x.js
kennethtruong: I'm on my phone right now but hopefully this makes sense. I think the main problem I have with this is if I had a variable that had the type Workbook = { connection: { label: string }. Now if I wanted to call firebase.database.ref.update flow will now say it's an error that my type Workbook does not match FirebasValue (thought it's clearly an object). Another issue is if I wanted to assign this.workbook = snapshot.val() flow wouldn't let me even if I decided to try to check if snapshot.val instance Object and then check if it had a property connection and then check if label was a string. To get it to stop giving me errors I had to cast it to type any and then cast it to the type I declared. I think it might better to let the developer use their own types. Though if you have good examples on how you use the FirebaseValue I wouldn't mind updating my code. I'm still kind of new to flow so let me know if I'm doing anything wrong.

stevenyap: >Now if I wanted to call firebase.database.ref.update flow will now say it's an error that my type Workbook does not match FirebasValue (thought it's clearly an object).

This case should not happen. Can you replicate your error at https://flow.org/try?

> Another issue is if I wanted to assign this.workbook = snapshot.val() flow wouldn't let me even if I decided to try to check if snapshot.val instance Object and then check if it had a property connection and then check if label was a string.

Yes this caught me a couple of times too and it is actually a good thing. You cannot assign snapshot.val() to workbook because snapshot.val() can be a string, array, null, object, etc... 

Flow is doing its job to tell you that the val() is not 100% a workflow object/type even though you may know yourself 100% it is definitely a workflow type (which is not true because someone can change the database to return an array/string/etc for your snapshot!)

When you check explicitly that it is an object, flow will not throw you any error. An often missed out gotcha that an Array is actually an object so you need to check for that too. Here's the code snippet that you can try:

```
// Return the param itself if it is an object else return null
const isObject = (a: any) => Array.isArray(a) && typeof a !== 'object' ? null : a
```

Hope this helps!

stevenyap:
@PCreations I have sent in the PR #955 to rename User to FirebaseUser

I have also made various other changes as below:
Breaking change:
- User is renamed to FirebaseUser
- UserCredential is renamed to FirebaseUserCredential
- AdditionalUserInfo is renamed to FirebaseAdditionalUserIno
- UserInfo is renamed to FirebaseUserInfo

Changes made:
- Formatted using Prettier
- projectId & databaseURL of FirebaseConfig is now optional
- Refactor class Query (extracted #FirebaseDatabaseCallback)
- Reference#push() returns a Promise or ThenableReference

Can you review it please?

@kennethtruong: Thanks for pointing out that projectId is optional! Please help me to review the PR too!

definitions/npm/firebase_v4.x.x/flow_v0.46.x-/firebase_v4.x.x.js
kennethtruong: Thanks for responding back to me! The code below is a simple case of what I have trouble with. Which is why I started looking into how what typescript did for their types. 

```
/* @flow */
type FirebaseValue = { [string| number]: FirebaseValue } | Array<FirebaseValue> | string | number | boolean | null;
type Workbook = { connection: { label: string } };

function update(x: { [string | number] : FirebaseValue }) {
  console.log(x);
}

const wb: Workbook = { connection: { label: 'test' }};

update(wb);
```

Another case is where I'm unsure what I need to do to be able to store my value into the workbook value

```
/* @flow */
type FirebaseValue = { [string| number]: FirebaseValue } | Array<FirebaseValue> | string | number | boolean | null;

class Test {
  workbook: { connection: { label: string } };
  snap = (snapshot: FirebaseValue) => {
    if (snapshot instanceof Object) {
      if (snapshot.connection && snapshot.connection instanceof Object) {
        this.workbook = snapshot;
      }
    }
  }
}
```

stevenyap: Your typing for FirebaseValue is different from the def. See:
- [Type def for update](https://github.com/stevenyap/flow-typed/blob/1c3af2f28cea3333ed6b192f6117b63e632918fb/definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js#L370)
- [Type def for FirebaseValue](https://github.com/stevenyap/flow-typed/blob/1c3af2f28cea3333ed6b192f6117b63e632918fb/definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js#L251)

Anyhow, I have recreated your code and found that the update throws a flow error.
After digging around, I realise the issue is with the variance typing of the object.

In short, when we write `update(wb)`, flow requires that the wb be an invariant of `{[path:string]: FirebaseValue}` and hence throws an error. The correct way is to declare update's param as a covariant which is `{+[path: string]: FirebaseValue}` (Note the extra `+`).

Please see my fix for your code:
[Flow Try](https://flow.org/try/#0PTAEAEDMBsHsHcBQJQHkBGArApgYwC6j4CeADtqAJYDOVAdgG4CGATpU3foieaAOqwWAa3SxYQ0AF5QAb1C5YdOnnyVFALlmhoTdNmibq+NnQDmoAL6XE3MhQBilFtnRNq2AGpNoAVwqTEUFAAH1kLQJDQAEEWFiZiAB5HZ1d3L19sAD4I0KMTUxzQOh8AWz0WQtFYaGwOQuLoaBsUAGVYUB9SABMmVTNQDlBYLBUiO1BqAAtYH2gu0D1QEtYhbHm3eVhmNg4uSB86AjU6Du7e7AAKZgzqTTkAagBtUl7Jw2NKMwBdTWSXN083j8lgAlJoGLBKPMZBEFHRqNVsAA6OCmK5A7DUEGIcKIOFGUDwdCaATCKoSaRyOHKI4aLQ6PQGUAAcnwmPwzMsuM6PTZFyJIKAA)

I have updated PR #955 with the fix above.

PR #918, opened 2017-06-02
Add jest-enzyme plugin to jest definition

pascalduez:
Hi,

after quite some searching I couldn’t find a "proper" way to extend the already declared jest `expect` definition. 

If I refer to facebook/flow/issues/396, and especially [how it's done for other frameworks, plugins](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/chai_v3.5.x/flow_v0.24.0-/chai_v3.5.x.js#L75-L106) this is somehow acceptable for now?

Looking for directions. Thanks.


marudor:
Sadly no "proper" way possible so far.

PR #921, opened 2017-06-05
Improve moment.duration definitions a bit

futpib:
This allows handy import like
```js
import { duration } from 'moment';
```

And adds some pretty simple duration type tests.

### DIFF
--- a/definitions/npm/moment_v2.x.x/flow_v0.34.x-/moment_v2.x.x.js
+++ b/definitions/npm/moment_v2.x.x/flow_v0.34.x-/moment_v2.x.x.js
@@ -237,5 +237,6 @@ declare class moment$Moment {
 }
 
 declare module 'moment' {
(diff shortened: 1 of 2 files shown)


futpib:
Please don't merge yet, I just realized this breaks CommonJS imports, gonna try to fix that

futpib:
No luck on trying to make this work with both es6 import and commonjs (01239e1e).

With `declare module.exports: Class<moment$Moment>;` this fails (even thought moment.duration is defined in the `moment$Moment` class, even if I change it from a static method to a static property):
```js
import { duration } from 'moment';
```

With es6 export definitions, test from 01239e1e fail.

And you can't have both declared 😕.

Any advice on this?

marudor:
@jeffmo Any Idea about the es6/commonjs issue?

GAntoine:
@futpib Is this change still relevant?

PR #933, opened 2017-06-08
Update rxjs.Observable/sampleTime with optional scheduler argument, and add the TestScheduler to typed exports

inlinestyle:
Hi there!
With the advent of Flow 0.47.0's arity-based checking, it's more important than ever to account for optional arguments. It would also be useful to type out some of rxjs's testing suite.


inlinestyle:
@marudor thanks!

PR #935, opened 2017-06-09
Update libdef for redux-actions v2.x.x

floriangosse:
Action creators now fully accept multiple arguments.


showmeyourhits:
It is really surprizing, that this bug exists.
I wonder: is it bad practice to pass multiple arguments to payloadCreator?

floriangosse:
I don't think that is bad practice because otherwise we don't need a payload create function because we could simply set the first argument (which would be an object if we have multiple arguments) as payload. But this whole process makes the payload creator obsolete for the most parts.

dignifiedquire:
I started using this  and it works very well for me 👍

floriangosse:
Maybe someone of the contributors could help me with the error.
// cc @marudor

GAntoine:
@floriangosse `redux-actions_v2.x.x` has an existing definition, where the flow version is `flow_v0.34.x-`. Change that folder's name to `flow_v0.34.x-0.38.x`, and your version will be used when flow's version is 39 or higher.

floriangosse:
@GAntoine thank you. I will change it.

PR #937, opened 2017-06-10
Add Storybook 3.x definitions

pascalduez:
From version 3.x storybook moved to a new org/scope and clarified addons.

Note:
Tests are passing locally, but the CI crashes on the package scope.

```
UNCAUGHT ERROR: Error: @storybook: '@storybook' is a malformed definitions/npm/ directory name! Expected the name to be formatted as <PKGNAME>_v<MAJOR>.<MINOR>.<PATCH>
```


GAntoine:
The failing specs are unrelated to this PR. I'll take some time to fix 'em, and we'll rebase this afterwards.

pascalduez:
I think I should improve those definitions, they don't cover the whole API range and plugins. 
But should be enough to get started with FLow+Storybook.
Maybe in another PR.

GAntoine:
@pascalduez Alright I've fixed the specs, want to rebase and we'll cross our fingers?

PR #941, opened 2017-06-11
add react-jss_v6.1.x type declaration

wizardzloy:
Add typedefs for `react-jss`. This implementation covers the most "useful" parts of the API. 

I would ask @kof to review if the typedefs are correct from the library author perspective.

### DIFF
--- a/definitions/npm/react-jss_v6.1.x/test_react-jss_v6.1.x.js
+++ b/definitions/npm/react-jss_v6.1.x/test_react-jss_v6.1.x.js
@@ -1,0 +1,122 @@
+// @flow
+import React from 'react';
+import injectSheet, { create } from 'react-jss';
+import type { Classes, Sheet } from 'react-jss';
+
+
+const styles = {
+  root: {
+    backgroundColor: 'red',
+  },
+  [`@media (min-width: ${10}px)`]: {
+    root: {
+      width: 200
+    }
+  }
+};
+
+type Styles = typeof styles;
+
+type Props = {
+  classes: Classes<Styles>,
+  sheet: Sheet<Styles>,
+  content: string,
+}
+
+const FunctionComponent = (props: Props) => {
+  if (!props.sheet.attached) {
+    return null;
+  }
+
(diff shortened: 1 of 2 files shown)


definitions/npm/react-jss_v6.1.x/test_react-jss_v6.1.x.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+};
+
+type Props = {
+  classes: Classes<typeof styles>,
+  content: string,
+}
+
+const FunctionComponent = (props: Props) => (
+  <div className={props.classes.root}>{props.content}</div>
```
kof: normally you would want to use props.children instead

definitions/npm/react-jss_v6.1.x/test_react-jss_v6.1.x.js
```diff
@@ -0,0 +1,102 @@
...
+// ===================================
+
+// $ExpectError - missing "styles" argument
+injectSheet()(FunctionComponent);
+
+// $ExpectError - wrong type of "styles" argument
+injectSheet(123)(FunctionComponent);
+
+// no errors
+injectSheet(styles)(FunctionComponent);
```
kof: you could add this `injectSheet(styles)()`  - sometimes it is useful

wizardzloy: What does it return in this case?

kof: a no-render component `render() {return null}` but the sheet is still injected, for e.g. useful if you need global styles only without any dom nodes in that particular component.

kof:
Otherwise examples look ok, didn't read exactly the type defs.

kof:
btw. you could add the `sheet` prop as well

wizardzloy:
@kof added `sheet` property and covered `injectSheet(styles)()` situation

Issue #925, opened 2017-06-06
Enzyme: Wrapper#instance() returns React$Component instead of the original component class


Sparragus:
[`Wrapper#instance()`](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/enzyme_v2.3.x/flow_v0.28.x-/enzyme_v2.3.x.js#L50) has a return type of `React$Component<any, any, any>`. That is true, but not specific enough and if forcing me to add `$FlowFixMe` almost every time I use it.

Consider this class:

```js
class Map extend React.Component {
  zoom (zoomLevel: number): void {
    // ...
  }

  render () {
    //...
  }
}
```

And the test for `Map#zoom`:

```js
describe('#zoom', () => {
  it('zooms the map', () => {
    const tree = shallow(<Map />)
    tree.instance().zoom(16) // Flow error: property `zoom` Property not found in React$Component
  })
})
```

I think that error can be fixed using generics. If `Wrapper` knows that it is *wrapping* a type `T`, then it would be possible to do something like:

```js
declare class Wrapper<T> {
  instance(): T;
}
```

Furthermore, I think we can set a type of `React$Component<any, any, any>` to the generic `T` like so:

```js
declare class Wrapper<T: React$Component<any, any, any>> {
  instance(): T;
}
```

I would be very happy to send a PR to fix this, but I need guidance. Basically, I need to know if I'm on the right direction regarding this.

Sparragus:
Fixed in https://github.com/flowtype/flow-typed/pull/950

GAntoine:
@Sparragus Is this still an issue?

Sparragus:
Not anymore, at least with `enzyme@3` and `flow-bin@0.53.0` :D

alexgurr:
I'm having this issue:

enzyme@3.7.0
flow-bin@0.89.0

ideas?

HsuTing:
enzyme@^3.8.0
flow-bin@^0.91.0

Have the same issue.

PR #950, opened 2017-06-14
[enzyme 2.3] fix ReactWrapper constructor and Wrapper#instance return…

### LINKED ISSUE #949: Enzyme 2.8.2 constructor expects no arguments
```
153:           const portalWrapper = new ReactWrapper(portal, portal);
                                                              ^^^^^^ unused function argument
         v----------------------
   13:   declare class Wrapper {
   14:     find(selector: EnzymeSelector): this;
   15:     findWhere(predicate: PredicateFunction<this>): this;
  ...:
   67:   }
         ^ default constructor expects no arguments. See lib: flow-typed/npm/enzyme_v2.3.x.js:13
```

I'll PR this and #925 shortly.

rosskevin:
Closes #949 #925

### DIFF
--- a/definitions/npm/enzyme_v2.3.x/flow_v0.28.x-/enzyme_v2.3.x.js
+++ b/definitions/npm/enzyme_v2.3.x/flow_v0.28.x-/enzyme_v2.3.x.js
@@ -47,7 +47,7 @@ declare module 'enzyme' {
     setState(state: Object, callback?: Function): this,
     setProps(props: Object): this;
     setContext(context: Object): this;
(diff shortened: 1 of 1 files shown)


Sparragus:
Oh man, thank you so much for this.

Sparragus:
I just tried this out I still can't use `Wrapper#instance()` correctly. It still causes the same problem described in https://github.com/flowtype/flow-typed/issues/925.

Setting it to `React$Component<*, *, *>` only sets the defaultProps, props, and state. See https://github.com/facebook/flow/blob/v0.46.0/lib/react.js#L16.

PR #952, opened 2017-06-14
[moment_v2.x] update based on the official included typescript file

rosskevin:
I started from scratch with the official `moment.d.ts` from `2.18.1`, used `flowgen`, then cleaned up the module namespacing.

### DIFF
--- a/definitions/npm/moment_v2.x.x/flow_v0.28.x-/moment_v2.x.x.js
+++ b/definitions/npm/moment_v2.x.x/flow_v0.28.x-/moment_v2.x.x.js
@@ -6,40 +6,40 @@ declare module 'moment' {
 
 
   declare interface Locale {
-    calendar(key?: CalendarKey, m?: Moment, now?: Moment): string,
-    longDateFormat(key: LongDateFormatKey): string,
-    invalidDate(): string,
-    ordinal(n: number): string,
-    preparse(inp: string): string,
-    postformat(inp: string): string,
+    calendar(key?: CalendarKey, m?: Moment, now?: Moment): string;
+    longDateFormat(key: LongDateFormatKey): string;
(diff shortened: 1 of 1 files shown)


rosskevin:
I will look at these CI errors and fix.

rosskevin:
I could use some help from someone here, many of the errors are completely expected, with `$ExpectError`, for example:

```js
// $ExpectError
moment().calendar(null, {
  sameElse: () => {},  // L52
});
```

```
0-test_moment-v2.js:52
    52:   sameElse: () => {},
                         ^ undefined. This type is incompatible with
    57:   declare type CalendarSpecVal = string | ((m?: MomentInput, now?: Moment) => string);
                                                                                      ^^^^^^ string. See lib: moment_v2.x.x.js:57
```
I'm not sure why the `$ExpectError` is not having any effect.

rosskevin:
@marudor any thoughts on why `$ExpectError` isn't working as expected in this case?

marudor:
Error is on Line 52, your $ExpectError on Line 50 - therefore only expected on Line 51?

rosskevin:
I tried moving it down with the same result. This is actually untouched test code that was working...so I'm confused.

marudor:
Will take a look later.

rosskevin:
I've just updated this and still have no idea why tests don't run.  There is quite a bit of work put into this, anyone else is free to take it over.  I'm using it locally successfully.

GAntoine:
How are you importing the typedef locally? The error in the spec is **Default import from `moment`. This module has no default export. Did you mean `import {Moment} from "moment"`?** This leads me to believe the test file for this def is importing incorrectly.

GAntoine:
Closing as there is a new, updated definition for moment v2.

PR #955, opened 2017-06-16
Rename FirebaseUser and various fixes for Firebase v4

stevenyap:
Breaking change:
- User is renamed to FirebaseUser
- UserCredential is renamed to FirebaseUserCredential
- AdditionalUserInfo is renamed to FirebaseAdditionalUserIno
- UserInfo is renamed to FirebaseUserInfo

Changes made:
- Formatted using Prettier
- projectId & databaseURL of FirebaseConfig is now optional
- Refactor class Query (extracted #FirebaseDatabaseCallback)
- Reference#push() returns a Promise or ThenableReference

### DIFF
--- a/definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js
+++ b/definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js
@@ -248,13 +248,7 @@ declare type OAuthProvider =
   | TwitterAuthProvider
 
 /** **** database ******/
-declare type FirebaseValue =
-  | {}
-  | Array<FirebaseValue>
-  | string
-  | number
-  | boolean
-  | null
+declare type FirebaseValue = any
 declare type OnCompleteCallback = (error: ?Object) => void
 declare type QueryEventType =
   | 'value'
--- a/definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js
+++ b/definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js
@@ -297,7 +297,7 @@ declare class OnDisconnect {
     onComplete?: OnCompleteCallback
   ): Promise<void>,
   update(
-    values: { [path: string]: FirebaseValue },
+    values: { +[path: string]: FirebaseValue },
     onComplete?: OnCompleteCallback
   ): Promise<void>
 }
--- a/definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js
+++ b/definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js
@@ -1,327 +1,385 @@
 /* @flow */
-
-/**** firebase ****/
+/** ** firebase ****/
 
 declare interface FirebaseConfig {
-  apiKey: string;
-  authDomain?: string;
-  databaseURL: string;
-  projectId: string;
-  storageBucket?: string;
-  messagingSenderId?: string;
+  apiKey: string,
+  authDomain?: string,
+  databaseURL?: string,
+  projectId?: string,
+  storageBucket?: string,
+  messagingSenderId?: string
 }
 
 declare interface AuthError {
-  code: | 'auth/app-deleted'
-        | 'auth/app-not-authorized'
-        | 'auth/argument-error'
-        | 'auth/invalid-api-key'
-        | 'auth/invalid-user-token'
-        | 'auth/network-request-failed'
-        | 'auth/operation-not-allowed'
-        | 'auth/requires-recent-login'
-        | 'auth/too-many-requests'
-        | 'auth/unauthorized-domain'
-        | 'auth/user-disabled'
-        | 'auth/user-token-expired'
-        | 'auth/web-storage-unsupported'
-        | 'auth/invalid-email'
-        | 'auth/account-exists-with-different-credential'
-        | 'auth/invalid-credential'
-        | 'auth/user-not-found'
-        | 'auth/wrong-password'
-        | 'auth/invalid-verification-code'
-        | 'auth/invalid-verification-id'
-        | 'auth/expired-action-code'
-        | 'auth/invalid-action-code'
-        | 'auth/invalid-verification-code'
-        | 'auth/missing-verification-code'
-        | 'auth/captcha-check-failed'
-        | 'auth/invalid-phone-number'
-        | 'auth/missing-phone-number'
-        | 'auth/quota-exceeded'
-        | 'auth/credential-already-in-use'
-        | 'auth/email-already-in-use'
-        | 'auth/provider-already-linked'
-        | 'auth/auth-domain-config-required'
-        | 'auth/cancelled-popup-request'
-        | 'auth/popup-blocked'
-        | 'auth/operation-not-supported-in-this-environment'
-        | 'auth/popup-closed-by-user'
-        | 'auth/unauthorized-domain'
-        | 'auth/no-such-provider';
-  message: string;
+  code:
+    | 'auth/app-deleted'
+    | 'auth/app-not-authorized'
+    | 'auth/argument-error'
+    | 'auth/invalid-api-key'
+    | 'auth/invalid-user-token'
+    | 'auth/network-request-failed'
+    | 'auth/operation-not-allowed'
+    | 'auth/requires-recent-login'
+    | 'auth/too-many-requests'
+    | 'auth/unauthorized-domain'
+    | 'auth/user-disabled'
+    | 'auth/user-token-expired'
+    | 'auth/web-storage-unsupported'
+    | 'auth/invalid-email'
+    | 'auth/account-exists-with-different-credential'
+    | 'auth/invalid-credential'
+    | 'auth/user-not-found'
+    | 'auth/wrong-password'
+    | 'auth/invalid-verification-code'
+    | 'auth/invalid-verification-id'
+    | 'auth/expired-action-code'
+    | 'auth/invalid-action-code'
+    | 'auth/invalid-verification-code'
+    | 'auth/missing-verification-code'
+    | 'auth/captcha-check-failed'
+    | 'auth/invalid-phone-number'
+    | 'auth/missing-phone-number'
+    | 'auth/quota-exceeded'
+    | 'auth/credential-already-in-use'
+    | 'auth/email-already-in-use'
+    | 'auth/provider-already-linked'
+    | 'auth/auth-domain-config-required'
+    | 'auth/cancelled-popup-request'
+    | 'auth/popup-blocked'
+    | 'auth/operation-not-supported-in-this-environment'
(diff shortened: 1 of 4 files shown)


definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js
```diff
@@ -1,344 +1,402 @@
...
-  setWithPriority(newVal: FirebaseValue, newPriority: Priority, onComplete?: OnCompleteCallback): Promise<void>;
+  key: ?string,
+  parent?: Reference,
+  root: Reference,
+  child(path: string): Reference,
+  onDisconnect(): OnDisconnect,
+  push(
+    value?: FirebaseValue,
+    onComplete?: OnCompleteCallback
+  ): ThenableReference & Promise<void>,
```
kennethtruong: You mentioned that in your comments that it should be ThenableReference or Promise. Should this be |?

kennethtruong:
Btw thanks for adding the firebase flow stuff! It's definitely nice to be able to have flow types with firebase. (: Also just a suggestion for the future but it would be easier to review if you split the prettier into a separate commit. Then add your changes on top or vice versa.

definitions/npm/firebase_v4.x.x/flow_v0.46.x-/firebase_v4.x.x.js
```diff
@@ -1,343 +1,402 @@
...
-                              | 'child_moved';
-declare type Priority = string | number | null;
+declare type OAuthProvider =
+  | FacebookAuthProvider
+  | GithubAuthProvider
+  | GoogleAuthProvider
+  | TwitterAuthProvider
+
+/** **** database ******/
+declare type FirebaseValue =
```
kennethtruong: In my opinion I still think this should be type `any`. Even though you mentioned it can't be certain values. I think the developer should know what value is coming out of their .val() and they should cast it to the right type making sure that the correct value goes in or out. I think this is why in typescript for this is `any`. Having FirebaseValue not be `any` makes it hard to call update to with a value that has a type that I declared. I could be wrong though..

definitions/npm/firebase_v4.x.x/flow_v0.34.x-v0.45.x/firebase_v4.x.x.js
stevenyap: it is a Promise and ThenableReference at the same time :)

marudor:
@kennethtruong @stevenyap 
I've got not experience with firebase-js.
I would need your help wether this should be merged or not.
Especially regarding the FirebaseValue.
Could you two decide wether any or not should be used?
Once that is decided I can merge this PR.

stevenyap:
@marudor ok we will discuss a bit more and let you know again

@kennethtruong I have replied to your issue about the `update` function... if the fix for update works, we can go ahead with FirebaseValue as `{}  | Array<FirebaseValue>  | string  | number  | boolean | null`

While I understand that `any` will resolve many issues for developers, it defeats the purpose of using flow because we won't be able to catch type issues when we cast it to `any`.

kennethtruong:
@stevenyap I really appreciate you taking the time in looking through my use cases and thanks for fixing the update method! That did solve one of my issue.

If there was a good solution for the below code I think I will be fine with your PR.

```
/* @flow */
type FirebaseValue = {} | Array<FirebaseValue> | string | number | boolean | null;

class Test {
  workbook: { connection: { label: string } };
  snap = (snapshot: FirebaseValue) => {
    if (snapshot instanceof Object) {
      if (snapshot.connection && snapshot.connection instanceof Object) {
        if (typeof snapshot.connection.label === 'string') { 
          this.workbook = snapshot;
        }
      }
    }
  }
}
```

I get this error from flow.org/try
```
10:           this.workbook = snapshot;
              ^ property `connection`. Property not found in
10:           this.workbook = snapshot;
                              ^ Object
```

stevenyap:
Two things to take into context before you read my solutions are:
- Please note that an Array is actually an object in JS so type checking or refinement should take that into consideration.
- You should know what is [type refinement](https://flow.org/en/docs/lang/refinements/)

Your code looks valid and therefore I guess that the error is with flow.
It seems that Flow object refinement is limited to simple if statements and nested if statements do not work (See https://github.com/facebook/flow/issues/3614#issuecomment-290622704)

Here's an example to rewrite your code in simple if statements that works:
```
/* @flow */
// I need to change {} to { [key:string]: FirebaseValue } else flow will complain
type FirebaseValue = { [key:string]: FirebaseValue } | Array<FirebaseValue> | string | number | boolean | null;

class Test {
  workbook: { connection: { label: string } };
  snap = (snapshot: FirebaseValue) => {
    if (snapshot == null || typeof snapshot !== 'object' || Array.isArray(snapshot)) return null
	
    const { connection } = snapshot
    if (connection && typeof connection === 'object' && !Array.isArray(connection)) {
      const { label } = connection
      
      if (typeof label === 'string') {
        return snapshot // no error
      }
    }
  }
}
```

There are other strategies to solve it too.
One way is to simply re-define the returned object:
```
// Simplest workaround
/* @flow */
type FirebaseValue = {} | Array<FirebaseValue> | string | number | boolean | null;

class Test {
  workbook: { connection: { label: string } };
  snap = (snapshot: FirebaseValue) => {
    if (snapshot instanceof Object) {
      if (snapshot.connection && snapshot.connection instanceof Object) {
        if (typeof snapshot.connection.label === 'string') { 
          // this is the only change: return a definite shape of the object
          this.workbook = { connection: { label: snapshot.connection.label }};
        }
      }
    }
  }
}
```

Finally, I was wondering why I did not hit this problem since I am using the typedefs too. I am guessing the reason is that I am using Ramda and functional programming in my app, so I rarely use type refinement but instead use functions to transform the variables. Flow can track the types easily via functions. Here's an example to demonstrate it:

```
/* @flow */
type FirebaseValue = {} | Array<FirebaseValue> | string | number | boolean | null;

/* @flow */
// Return value if it is an object else return null
const toObject = (value: any): ?Object => value && typeof value === 'object' && !Array.isArray(value) ? value : null

type Workbook = { connection: { label: string } }
class Test {
  workbook: ?Workbook;
  snap = (snapshot: FirebaseValue): void => {
    // flow won't complain as long as your snapshot is an object
    // because Workbook is an inexact type meaning the object can always be changed
    this.workbook = toObject(snapshot)
  }
}
```

DataSnapshot from the firebase is the first point into our application and it is indeed a chaotic type because it can really be anything like null or a string instead of an object, etc. I am actually very grateful to flow for pointing this out and forces us to add more checks around it to create a safer app.

kennethtruong:
Thanks for looking into it! Hmm I'm not entirely sure if those are good solutions. One thing I noticed when you talked about using functions to transform your variables, is that in `const toObject = (value: any): ?Object => value && typeof value === 'object' && !Array.isArray(value) ? value : null` It's doing a cast to any which seems to defeat the purpose of having the FirebaseValue. 

(Also first example doesn't show the type refinement since its just returning the value without setting it to the variable.)

Seeing your second example is a workaround for me but I'm not sure about the benefits with FirebaseValue here.

I still don't believe the FirebaseValue is necessary because if you have good typedefs throughout your app you shouldn't see an incorrect type. And if the user does get a value that isn't what is meant to be based on the type you're expecting.. the app is already broken which will happen either way if you did type checking or not. Flow didn't help catch it during compile time which is what I believe what flow is for. Though in the end my app would have thrown an error either way if the values didn't match. (Note: we're planning on doing is making a script that checks our database to make sure that all the values are correct. So we essentially will catch all the areas where the values won't be correct if there is going to be any.)

Another reason why I want to use any is because in the official firebase js library it shows that they're just using type any. 
https://github.com/firebase/firebase-js-sdk/blob/master/typings/firebase.d.ts#L242
also in the typescript equivalent of flow-typed is also type any
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/firebase/index.d.ts#L20
But seeing how the official firebase library uses only any in their typescript means that they probably don't see the benefits in trying to add all the values that it could be probaby?

Sorry if I'm being stubborn but I would like a second opinion if possible.

stevenyap:
Thanks for pointing out the mistake! I have removed the example for the sake of not misleading other developers.

`const toObject = (value: any): ?Object => value && typeof value === 'object' && !Array.isArray(value) ? value : null` does not cast it to `any`... in fact, it checks that ONLY object is allowed and ensure it is also not an Array. It is a kind of type refinement which is part of the reasons why we are using flow.

I don't agree with typescript definition. To me, `any` is the last resort if there are no solutions to work around. However, we do have workarounds which maintain the benefits of type safety.

> And if the user does get a value that isn't what is meant to be based on the type you're expecting.. the app is already broken which will happen either way if you did type checking or not. Flow didn't help catch it during compile time which is what I believe what flow is for. Though in the end my app would have thrown an error either way if the values didn't match.

In theory, if we define ALL the possible type of value we are expecting in all cases (like FirebaseValue), then our app will never break because of type error. However, in practice, it is not perfect because Flow is still in development (eg type refinement is not so good) and there are limitations to static analysis type checker. That's why sometimes you don't see Flow catching errors for you. But if we don't type our values as accurately as possible (eg. FirebaseValue), we are not giving Flow a better chance to catch the possible type errors for us. Don't forget that the Flow is always improving.. we may not need our workarounds in the future releases but the types never change.

So can we go ahead with `FirebaseValue = {} | Array<FirebaseValue> | string | number | boolean | null`?

PS: I can be quite stubborn too sometimes haha :)

@PCreations: Would love to hear your thoughts on this.

kennethtruong:
Ya I just can't agree after thinking about it I started to look at how Facebook gives their response type for fetch requests which is the equivalent of firebase's `.val()`. They also use type any
https://github.com/facebook/flow/blob/master/lib/bom.js#L898
The json response can't be a function and yet they kept it as `any`. I think you're making FirebaseValue is be too restrictive. The fetch response should pretty much be the equivalent to what `.val()` should return. I'm pretty sure most people are not checking if their API response is the correct response and type. 
Another example is `JSON.parse` 
https://github.com/facebook/flow/blob/fa77d91c52ecc68d17ff0c82c94613abbb10499b/lib/core.js#L536
I don't believe you can get undefined nor a function from it yet it still type `any`.

I don't think this FirebaseValue has to be strict at all and by checking this in. It makes it a lot harder to use the firebase API with flow. I actually feel like this causes me to lose a lot of benefit with flow because its adding all these extra checks when I figure out if there was errors with or without the checks. If Facebook had given the fetch response a non `any` value I would agree with you but everything I've seen with flow or typescript seems to be pointing that it should be type any.

stevenyap:
i would love to hear more opinions but i guess we would be wasting too much time discussing over it. When you gave the reference to `JSON.parse` and `.json()`, I felt there could be other reasons that I am not aware of.

So let's go with `FirebaseValue: any` for now. :)
I have updated the PR.

@marudor: Please assist to merge the PR.

kennethtruong:
Thanks @stevenyap! Ya I felt bad that we spent so much time in discussing this. But I do appreciate you taking the time to update everything!

marudor:
There we go :)
Thank you two for discussing it to come up with the best option (for now)

stevenyap:
It was a great discussion and necessary so that we can all make the best decision for the community! :)

PR #956, opened 2017-06-16
Add chai-redux-mock-store definitions

themouette:
See the [github project page](https://github.com/themouette/chai-redux-mock-store)
or [npm project page](https://www.npmjs.com/package/chai-redux-mock-store)


GAntoine:
IMO these don't belong in the definition for `chai`, and should instead be in a new one for `chai-redux-mock-store`. Thoughts?

PR #958, opened 2017-06-17
Test against node 8

amilajack:


marudor:
I've made the changes and pushed travis 8 checks manually.

PR #963, opened 2017-06-19
Add tinycolor2

lifeiscontent:
@marudor ping


marudor:
Could you add a few more tests?

lifeiscontent:
@marudor what would be considered "good tests" for libdefs? Can you give me some basic examples?

marudor:
For somthing like this just very basic tests should be enough
I made libdefs for another color library a year ago. 
https://github.com/flowtype/flow-typed/blob/master/definitions/npm/color_v0.7.x/test_color_v0.7.x.js

lifeiscontent:
@marudor can you tell me why I'm getting an error in this test?

PR #980, opened 2017-06-24
Message should not be optional.

steida:


steida:
Hmm, maybe it should.

PR #994, opened 2017-06-29
add more enzyme typings

akre54:


GAntoine:
Closing since some of the changes are already in master, the specs are failing and the typedefs have changed drastically in the last few months. @akre54 We'd fully welcome an updated PR if you're still keen on getting these changes merged!

On a related note, I can't seem to find references for `component`, `props`, `node` or `refs` in the [API docs](https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md). Could you explain their addition?

PR #1006, opened 2017-07-03
Jest matchers can have multiple "expected" arguments

gausie:


definitions/npm/jest_v20.x.x/flow_v0.33.x-/jest_v20.x.x.js
```diff
@@ -91,7 +91,7 @@ type JestMatcherResult = {
   pass: boolean
 };
 
-type JestMatcher = (actual: any, expected: any) => JestMatcherResult;
+type JestMatcher = (actual: any, ...expected: any) => JestMatcherResult;
```
pascalduez: I guess rest parameters are better defined as `Array<any>`.

GAntoine:
@gausie Are you still looking to get this merged? Looking at the docs for [`expect.extend`](https://facebook.github.io/jest/docs/en/expect.html#expectextendmatchers), it seems to me that the signature for `JestMatcher` is `(received, argument)`, where `argument` is singular.

GAntoine:
Closing as I don't think this change is necessary. Please re-open if that's not the case!

PR #1007, opened 2017-07-03
Add redux-observable type definition

Ailrun:


steida:
Is there any working example repo? Can't make it work. Thank you.

steida:
@marudor In the other words... action$ are somehow untyped since it doesn't detect typos like ffilter. Maybe a bug in my code... I don't know.

```js
export const addUser: Epic = (
  action$: ActionsObservable<Action>,
  store,
  { createUuid, getNow },
) =>
  action$.filter('ADD_USER').mergeMap(action => {
```

Ailrun:
@steida I think it's from `import` in module bug of Flow,

When a libdef imports a type definition of external (for ex, `redux`) library, Flow doesn't recognize correct type of external one.
Actually, I can type this library like what author of [this](https://github.com/flowtype/flow-typed/pull/238/files) libdef does, but it's to complex and I believe this bug will be fixed.

Ailrun:
https://github.com/flowtype/flow-typed/issues/16
Related issue, and you're already there.

steida:
So what should I do? Wait until it will be fixed, or go back to plain observable without extending which works?

steida:
As I see it, there is no point to have defs providing less type safety than Observable itself.

Ailrun:
@steida 
If you want to use it right now, you can extend this with copy of `rxjs` type definition and resend another PR to have 'correct' type for *right now*.

My work is itself just for future Flow.

steida:
https://github.com/flowtype/flow-typed/issues/1020

PR #1012, opened 2017-07-04
Add basic support normalize-styled

omninonsense:
I am not actually sure what types of tests to add for this, but I am open for suggestions if anyone has an idea.


omninonsense:
Added alternate syntax, it seems to work as far as I can tell, but ESLint is complaining about it a bit. 

Haven't looked in detail as to why the CI tests are failing, but at a glance it seems to be unrelated to this (I think it was `underscore` related, but maybe I misread), will check later.

omninonsense:
CI Jobs seem to pass now. :tada: :balloon:

marudor:
Mind adding a test?

omninonsense:
@marudor I am not sure what to test for, to be honest, short of doing something like this:

~~~javascript
import normalizeStyled from 'normalize-styled'

const styles : Array<Interpolation> = normalizeStyled
~~~

However I am not sure how useful this test is. Can you recommend something, or point me to an example?

marudor:
Indeed. Haven't really looked at this...
Its fine like this. Thanks.

PR #1031, opened 2017-07-12
koa_v2.x.x.js: make Context middleware params customizable

dunnock:
`any` type can be cast to any strict type, therefore allowing covariant customization of Context. Useful when writing custom middleware.

With this change developer can customize Context as per below:

```
type CustomContext = Context & {
  reqId: string
};

app.use(requestId);
app.use(logRequests);

async function logRequests(ctx: CustomContext, next: *) {
   log('got request', ctx.reqId, ctx.request.url);
   await next();
}
```

When type is mixed it will conflict with the above customization.


zyklus:
Why does this line exist at all?  It effectively disables type-checking on custom properties; wouldn't it be better to force extension of the Context as in your code snippet?

dunnock:
I suppose it was added to avoid type declaration of third party middleware, though I agree anything used in the code should rather be strictly typed. Would you submit PR?

PR #1033, opened 2017-07-13
Ramda: Fixing curry definition

mwalkerwells:
Fixes this issue: https://github.com/facebook/flow/issues/4364

Adds tons of tests to catch future issues.


thejameskyle:
The flow versions can't overlap:

- `flow_v0.34.x-/` &larr; should be `flow_v0.34.x-v0.39.x/`
- `flow_v0.39.x-/`

mwalkerwells:
@thejameskyle Thanks for your quick feedback on this. I'm surprised that ramda curry was unsafe & untested for so long...

PR #1034, opened 2017-07-13
Fix recompose filename

purge:
remove `-` at end of filename
fixes: `UNCAUGHT ERROR: TypeError: Invalid comparator: ^v0.24.x-`


istarkov:
Isn't it is the bug of flow-typed, based on this
https://github.com/flowtype/flow-typed/wiki/Contributing-Library-Definitions

* Library versions may not specify a semver range, but Flow versions can of the
  following forms:
  * **`flow_v0.22.x-`**: Flow v0.22.x and above

I read this as `x-` can be used for libraries too

GAntoine:
Not quite. If you read carefully, _library versions may not specify a semver range_ means that library versions must have a specific version (disregarding using x.x.x).

@purge Seems CI failed for unrelated reasons, close and reopen the PR to run it again and it should pass.

istarkov:
For me non native speaker `may not` is not the same as `must not`,
`may not` is a very weak version of `must not`.

GAntoine:
I can understand the confusion, I'll update the docs.

GAntoine:
@purge You're gonna have to change the folder name as well, from `recompose_v0.24.x-` to `recompose_v0.24.x`. The spec won't fail until then. My bad, didn't read the entire CI error output.

Issue #490, opened 2016-11-28
Consider replacing `request`.


wtgtybhertgeghgtwtg:
The use for `request` is [pretty simple](https://github.com/flowtype/flow-typed/blob/d0f06ec5404560855065857e78e25f201dfe80d2/cli/src/commands/runTests.js#L135).  Could it be replaced with something lighter, like `got` or `fetch`?

ryyppy:
Hey!
What is the benefit of doing so?

wtgtybhertgeghgtwtg:
Less overall dependencies.  About half of the overall dependencies for `flow-typed` are currently from `request`.  Additionally, there may be some [security concerns](https://github.com/request/request/issues/2230), but I am not too sure if that would actually be a problem for this use case.
The argument has been made, much more articulately, for some other packages, such as [`yarn`](https://github.com/yarnpkg/yarn/issues/1461), [`node-pre-gyp`](https://github.com/mapbox/node-pre-gyp/issues/198), and [`jsdom`](https://github.com/tmpvar/jsdom/pull/1630).

ryyppy:
Makes sense... it seems you have followed the `yarn` discussion... if you have a good replacement and create a PR, I would consider reviewing & merging it!

wtgtybhertgeghgtwtg:
I'll see what I can do, but since the `scripts` aren't Windows friendly, that might not be much.

ryyppy:
True dat, so we also need to make the build steps work for windows as well.. not sure how to proceed for now... if you are motivated to make this happen, I am happy to try it on my machine as well!

wtgtybhertgeghgtwtg:
PR to make scripts work on Windows: #522.

wtgtybhertgeghgtwtg:
I have this complete, but there were never any tests for it.  #526 would help.

Issue #87, opened 2016-05-10
[Test CI] `validate-defs` should ensure that each npm package name is actually an npm package name


jeffmo:
Currently all of our definitions are assumed to be valid npm package names, but we don't automatically verify this. (I've been verifying it as I accept PRs).

It would be good to automate this in the CI tests.

GAntoine:
Re-opening since validation times out specs running on Travis, and I've had to comment out the code.

PR #1045, opened 2017-07-14
Running validate-defs checks NPM for package and fails on 404

GAntoine:
Addresses #87 
Also a replacement PR for #1042 

When running `validate-defs`, sets the `validating` param to true, which attempts to hit [npm.js](http://npmjs.com) and check for the existence of the validated def.

Removes `normalize-styled` since it does not exist as a package, but was added in #1012 

Uses [got](http://npmjs.com/package/got) to make the request. This PR could be the start of removing `request` in favour of `got`  #490


GAntoine:
Come to think of it, maybe this should just print a warning and not fail? Not sure which way to take this, but seeing as we want all the packages in here to be _real_, failing is probably best.

marudor:
Good start.
As I'm main maintainer I tend to do fairly sloppy reviews to manage all the PRs.
Thanks!

PR #1047, opened 2017-07-16
Ability to Ignore Categories of Dependencies when Installing TypeDefs

### LINKED ISSUE #379: Recommended approach for using flow-typed with peerDependencies

### LINKED ISSUE #453: Ability to exclude devDependencies from generating definitions
I can't think of a good reason why devDependencies should be included in the type definitions, so either they should be excluded by default, or an option should allow them to be excluded

### LINKED ISSUE #528: Can't have dependency in peerDeps and devDeps (Error: Found react listed twice in package.json!)
If you clone https://github.com/StreakYC/react-menu-list and then run `flow-typed install` inside of it, then you get this error:

```
• Found flow-bin@v0.36.0 installed. Installing libdefs compatible with this version of Flow...
    at /usr/local/lib/node_modules/flow-typed/dist/lib/npmProjectUtils.js:123:17
    at Array.forEach (native)
    at /usr/local/lib/node_modules/flow-typed/dist/lib/npmProjectUtils.js:121:35
    at Array.reduce (native)
    at getPackageJsonDependencies (/usr/local/lib/node_modules/flow-typed/dist/lib/npmProjectUtils.js:118:30)
    at Object.run$ (/usr/local/lib/node_modules/flow-typed/dist/commands/install.js:177:91)
    at tryCatch (/usr/local/lib/node_modules/flow-typed/node_modules/regenerator-runtime/runtime.js:64:40)
    at Generator.invoke [as _invoke] (/usr/local/lib/node_modules/flow-typed/node_modules/regenerator-runtime/runtime.js:355:22)
    at Generator.prototype.(anonymous function) [as next] (/usr/local/lib/node_modules/flow-typed/node_modules/regenerator-runtime/runtime.js:116:21)
    at tryCatch (/usr/local/lib/node_modules/flow-typed/node_modules/regenerator-runtime/runtime.js:64:40)
```

Its [package.json](https://github.com/StreakYC/react-menu-list/blob/master/package.json) has react listed under both devDependencies and peerDependencies which is legal and sensible as far as I know. (If it's pulled as a dependency of an application, then that application needs to have its own copy of React already; it will never pull in its own copy of React because that would cause conflicts. If react-menu-list is developed on directly, its devDependencies are installed including its own copy of React.)

### LINKED ISSUE #599: Option to not install libdefs for devDependencies?
Most of the time I don't need libdefs for devDepdendencies. It would be nice if `flow-typed install` had an option to not install those.

GAntoine:
Fixes #379 
Fixes #453 
Fixes #528 
Fixes #599

This has been requested so many time, I think it's about time we address it.

Pass the `--ignoreDeps` flag with a list of dependency categories to ignore to `flow-typed install`. Permitted values are `dev`, `bundle`, `peer` and `optional`. For instance, running what's below will only install regular and `optional` dependencies:

```
flow-typed install --ignoreDeps dev bundle peer
```

Still didn't want to mess with the `.flowconfig` file, though seeing as this is the 4th flag I've created, it may be time to talk about integrating with the config file.

### DIFF
--- a/cli/src/commands/__tests__/install-test.js
+++ b/cli/src/commands/__tests__/install-test.js
@@ -129,11 +129,11 @@ describe("install (command)", () => {
           cwd: ROOT_DIR,
           flowVersion: parseFlowDirString('flow_v0.40.0', 'testContext'),
           explicitLibDefs: [],
-          ignoreDeps: [],
+          libdefDir: 'flow-typed',
           verbose: false,
           overwrite: false,
           skip: false,
-          libdefDir: 'flow-typed',
+          ignoreDeps: [],
         });
         expect(result).toBe(1);
         expect(_mock(console.error).mock.calls).toEqual([[
@@ -431,7 +431,6 @@ describe("install (command)", () => {
           overwrite: false,
           verbose: false,
           skip: false,
-          ignoreDeps: [],
         });
 
         // Installs a stub for someUntypedDep
@@ -464,7 +463,6 @@ describe("install (command)", () => {
           overwrite: false,
           verbose: false,
           skip: true,
-          ignoreDeps: [],
         });
 
         // Installs a stub for someUntypedDep
@@ -499,7 +497,6 @@ describe("install (command)", () => {
           overwrite: true,
           verbose: false,
           skip: false,
-          ignoreDeps: [],
         });
 
         // Replaces the stub with the real typedef
@@ -537,7 +534,6 @@ describe("install (command)", () => {
           overwrite: false,
           verbose: false,
           skip: false,
-          ignoreDeps: [],
         });
 
         const libdefFilePath = path.join(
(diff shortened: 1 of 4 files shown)


marudor:
Yeah Although we have to talk with the core flow team about .flowconfig stuff.
I'm with you on adding config though.
Thanks for all this!

I'm still working on getting npm access to publish all your changes.

GAntoine:
@marudor Happy to help! Yeah config would be super nice (eventually). In the meantime, I'll keep the CLI wiki updated and write up a guide on customizing flow-typed through the command like.

wtgtybhertgeghgtwtg:
While this does solve #379, having to add a flag seems like a suboptimal solution.

marudor:
We are communicating with the core flow team to investigate .flowconfig as an option to handle this.

GAntoine:
In the meantime, you can add a custom script to your `package.json` that runs commands with custom arguments.

PR #1051, opened 2017-07-18
Update express$Application.enable() return type

davesidious:
After checking the API documentation for the Express library, I updated the return type for the .enable() method on express$Application to allow chaining without error.


davesidious:
Now that's service!  Thanks!

PR #624, opened 2017-01-21
WIP: Recompose v0.21.x

ekosz:
This is my first foray into creating a serious flow-typing. This PR mostly piggybacks off of the great work of @gcanti at https://github.com/acdlite/recompose/pull/241. This PR takes that work and cleans it up a bit and add tests.

**Testing Checklist**
- [ ] compose,
- [x] mapProps,
- [x] withProps,
- [x] withPropsOnChange,
- [x] withHandlers,
- [x] defaultProps,
- [x] renameProp,
- [ ] renameProps,
- [x] flattenProp,
- [ ] withState,
- [x] withReducer,
- [ ] branch,
- [ ] renderComponent,
- [x] renderNothing,
- [x] shouldUpdate,
- [x] pure,
- [ ] onlyUpdateForKeys,
- [ ] withContext,
- [ ] getContext,
- [ ] lifecycle,
- [ ] toClass,
- [ ] setStatic,
- [ ] setDisplayName,
- [ ] getDisplayName,
- [ ] wrapDisplayName,
- [ ] shallowEqual,
- [ ] isClassComponent,
- [ ] createEagerElement,
- [ ] createEagerFactory,
- [ ] createSink,
- [ ] componentFromProp,
- [ ] nest,
- [ ] hoistStatics

PR #1052, opened 2017-07-18
Add lib defs for redux-entitize

leoselig:


GAntoine:
Hey @leoselig, since `redux-entitize` seems to be your project, would you be interested in adding type definitions directly into it? I could probably find a couple other projects which have done the same thing, without rewriting everything with flow.

leoselig:
@GAntoine
Thank your for your kind suggestion. I could absolutely do that of course.

But I'm a little confused as to what goal flow-typed actually persues. So far, it seems to be the only (and officially recommended) way of providing flow type definitions based both on the package version and the flow version of the dependent project.

I also don't see any outline of a strategy in this repository that declares what packages are "worth" to be managed via flow-typed. I thought it was open for everyone that publishes to the 
registry.

I'd be sincerely eager for the maintainers to clarify that strategy.

GAntoine:
Thanks for your response! To be perfectly honest I don't know of any such strategy, I do think that the typedefs for any package are welcome here.

From what I've gathered, when a user of a package would like flow typings, but the package maintainers don't have them/won't add them, they can use flow-typed instead. 

The opposite case would be if the package itself is either written in flow, or has typings. For example, someone opened a PR to add types for `recompose` on #624, but then recompose [added their own typings](https://github.com/acdlite/recompose/pull/428), so flow-typed was unnecessary.

@calebmer Any thoughts on this?

definitions/npm/redux-entitize_v0.2.x/flow_v0.45.x-/redux-entitize_v0.2.x.js
```diff
@@ -0,0 +1,48 @@
+import { type Schema } from "normalizr";
```
GAntoine: This won't work as you'd expect it to...

leoselig:
I'm afraid I don't have time for this PR at the moment. Will probably get back to it in 1 or 2 weeks.
Thanks for your feedback so far and the discussion on the distribution of type defs.

If you want to tidy up the PR list, feel free to close it. Then I'll open up another one one I'm focused enough.

definitions/npm/redux-entitize_v0.2.x/flow_v0.45.x-/redux-entitize_v0.2.x.js
leoselig: I realized, should have activated the linter to reveal the pointless import

GAntoine:
Thanks for the heads up, I'm happy to keep this one open since you're still around. Ping me when you're back to working on it and we'll get it merged lickity split.

GAntoine:
Updated and merged PR, thanks for the 💪 !

PR #1062, opened 2017-07-21
Add flow types for promise-retry

demands:
https://www.npmjs.com/package/promise-retry


demands:
Hey y'all – I don't understand why travis is failing for the node 6 build. It passes on my machine running node 6! Any pointers? What's the purpose of the "succeeds on well-formed repo" test?

Thanks,
Max

GAntoine:
It's not your code, it's the spec. We've been noticing this specific test failing once in a while. Close the PR and reopen it. If that does not work the first time, try 2 more times, and then we'll see what's going on.

PR #1065, opened 2017-07-23
Lodash groupBy array return value should not be maybe

yangshun:
The return type for the `groupBy` method was changed in #713 into `{[key: V]: ?Array<T>};`. Logically speaking, if a value is not present in the array, it will never be a key in the return value, hence all values in the return object is definitely an array of minimum length 1, and should not be a maybe type. TypeScript's Lodash definition also does not use a maybe type: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/index.d.ts#L7872

cc @ahutchings


yangshun:
Hi maintainers, would you mind having a look? (:

Issue #574, opened 2016-12-28
[redux] Support redux-thunk


remko:
The type definition of `Dispatch` states:

       declare type Dispatch<A: { type: $Subtype<string> }> = (action: A) => A;

Am I right that this doesn't cover support for `redux-thunk` middleware, where functions are dispatched instead of objects? Could this be added?

mtford90:
I changed the definition to the following in my `flow-typed/npm/redux_v3.x.x.js`:

```
declare type Dispatch<A: { type: $Subtype<string> }> = (action: A | (Dispatch<A> => mixed)) => A;
```

aaronjensen:
redux-thunk is not part of redux, it should not be supported as part of the redux libdefs.

Instead, you can do something like this: https://github.com/agentcooper/react-redux-flow-example/blob/master/src/types/index.js#L20

satya164:
That doesn't work. We're exporting a different `Dispatch`, but the dispatch definition inside redux's dispatch is not changed

satya164:
This doesn't work. There are still Flow errors if I use `bindActionCreators`

rafaelkallis:
Is anybody working on this?

satya164:
@rafaelkallis This is the best I could do https://blog.callstack.io/type-checking-react-and-redux-thunk-with-flow-part-2-206ce5f6e705#.jseb9mo7r

rafaelkallis:
Hey @satya164 I've read your blog post earlier while browsing the webs, great work. You should make a PR; the community would appreciate it!

satya164:
@rafaelkallis yeah, just need to figure out how to make `bindActionCreators` work with flow

rafaelkallis:
@satya164 is [this](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/redux_v3.x.x/flow_v0.33.x-/redux_v3.x.x.js#L43-L47) what you're looking for?

satya164:
it's tricky because we are not declaring types for `bindActionCreators` like we do for store, and it doesn't infer the correct thing :(

hmeerlo:
Hi, I got it working, please see my sample repo: https://github.com/hmeerlo/redux-thunk-flow

satya164:
I think better to have a way to specify a types for `bindActionCreators` rather than needed to modify the types for redux. It'll not likely get accepted here because thunk is not a part of redux, so the types will be wrong for people who don't use thunk.

[...]

PR #1071, opened 2017-07-25
Add Dispatch override for redux

pronebird:
Redux `Dispatch` is hardwired to accept only `{ type: $Subtype<string> }`. This makes it particularly hard to use redux with redux-thunk or other middlewares without rolling out your own flow annotations.

This PR adds additional parameter `D` which defaults to original `Dispatch<A>` when not provided by consumer, giving backward compatibility for existing consumers and providing flexibility.

The number of examples on how to use custom Dispatch are available at:
https://flow.org/en/docs/frameworks/redux/#toc-typing-redux-thunk-actions

```
// @flow
import type { Store } from 'redux';

type Dispatch = (action: Action | ThunkAction | PromiseAction) => any;
type GetState = () => State;
type ThunkAction = (dispatch: Dispatch, getState: GetState) => any;
type PromiseAction = Promise<Action>;

// store with default dispatch
type DefaultStore = Store<State, Action>;

// store with custom dispatch
type CustomStore = Store<State, Action, Dispatch>;
```

Related #574.

### DIFF
--- a/definitions/npm/redux_v3.x.x/flow_v0.33.x-/redux_v3.x.x.js
+++ b/definitions/npm/redux_v3.x.x/flow_v0.33.x-/redux_v3.x.x.js
@@ -4,19 +4,21 @@ declare module 'redux' {
 
     S = State
     A = Action
+    D = Dispatch
 
   */
 
-  declare type Dispatch<A: { type: $Subtype<string> }> = (action: A) => A;
+  declare type DispatchAPI<A> = (action: A) => A;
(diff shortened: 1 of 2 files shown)


marudor:
Could you add some tests with a custom Dispatch to ensure we're not breaking it again later?

pronebird:
@marudor good point. I'll add tests and push more commits soon 👍

pronebird:
I have added tests and fixed missing `A` parameter in `bindActionCreators` annotations.

PR #1075, opened 2017-07-27
Styled-Components: Preserving prop types for HOCs

mwalkerwells:
Props information is lost when using HOCs from the styled-components components library.

**Current**
```js
const Component = (props: { name: string }) => <div />
const WrappedComponent = styled(Component)``
const shouldError = <WrappedComponent /> // <- Should error
```

Tests give an idea of the added type soundness.

**With PR**
```js
const Component = (props: { name: string }) => <div />
const WrappedComponent = styled(Component)``
const correctlyErrors = <WrappedComponent /> // <- Correctly errors

const Component2 = (props: { name: string }) => <div />
const WrappedComponent2 = styled(Component2)``
const noError = <WrappedComponent2 name='foo' /> // <- No error
```


omninonsense:
This breaks coverage for things like these:

~~~javascript
const Comp = styled.div`
  color: cyan;
`
~~~

The coverage tool reports them as being uncovered. Am I doing something wrong? Any ideas on how to make the coverage testing to work short of doing something as verbose as

~~~javascript
const Wrapper = styled(() => <div />)`
  color: cyan;
`
~~~

johnhaley81:
Yeah, it seems that anything inside of the tagged template literal is no longer covered.

mwalkerwells:
#1091 Should fix this issue. Let me know if it doesn't.

PR #1079, opened 2017-07-27
Add bound-points definition

vicapow:


definitions/npm/bound-points_v1.x.x/test_bound-points.v1.x.x.js
```diff
@@ -0,0 +1,17 @@
...
+// $ExpectError boundPoints does not return `string`.
+(boundPoints([[10, 20]]): string);
+
+// $ExpectError boundPoints takes an array of arrays.
+boundPoints(['a', 'b']);
+
+// $ExpectError boundPoints takes an array of arrays.
+boundPoints([10, 20]);
+
+// $ExpectError boundPoints returns array of arrays.
```
vicapow: for some reason, when I remove the `// $ExpectError` the tests still pass. any ideas why?

marudor: Works for me. (As in test fails if I remove it)
Should be fine.

PR #1083, opened 2017-07-28
Add types for turf/destination, turf/distance, turf/point and turf/polygon

vicapow:


definitions/npm/turf-point_v2.x.x/flow_all/turf-point_v2.x.x.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+
+type $npm$Turf$Point$Point = {
```
vicapow: I redefine this in a few of the definitions because I assume it's not safe to share types across module definitions safely that may or may not get included in the users project

vicapow:
maybe someone could help? Travis is giving the following errors

> UNCAUGHT ERROR: Error: @turf: '@turf' is a malformed definitions/npm/ directory name! Expected the name to be formatted as <PKGNAME>_v<MAJOR>.<MINOR>.<PATCH>

Does flow-typed not support scoped module names?

GAntoine:
Are you getting the same errors in you run tests locally? The regex permits `@something/package` as a package name.

vicapow:
@GAntoine I'm not getting the same errors when I run locally :/

GAntoine:
Ok, try closing and re-opening the PR, there's a chance re-running the spec on Travis might fix it.

GAntoine:
Interesting, seems it's still failing...

GAntoine:
@vicapow I've opened a PR which should fix your issue. If/when it's merged, rebase and run the tests again on Travis.

GAntoine:
Fix is shipped! Fetch and rebase, and you should be good to go.

PR #1111, opened 2017-08-03
WIP: Properly type node-fetch API

maxjoehnk:
The node-fetch package has some differences compared to the browser fetch api.
This PR improves the node-fetch typings.

We need to add custom Response and Request Objects because they differ from the default Browser and node env classes.

This is the first time I'm typing a package with flow so there are probably things I could improve.

### DIFF
--- a/definitions/npm/node-fetch_v1.x.x/flow_v0.25.x-/node-fetch_v1.x.x.js
+++ b/definitions/npm/node-fetch_v1.x.x/flow_v0.25.x-/node-fetch_v1.x.x.js
@@ -1,4 +1,95 @@
+import type { Gunzip } from 'zlib';
+
+declare type fetch$Redirect = 'follow' | 'error' | 'manual';
+
+declare type fetch$RawHeaders = {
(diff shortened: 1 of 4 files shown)


KenanY:
Feel free to use [my changes to the existing test file](https://github.com/flowtype/flow-typed/pull/1113/files#diff-ea8ba5b8204119435a503a54e2aa36e0) 👍

maxjoehnk:
I actually feel like we should use your PR as a base and maybe add some of the properties you were missing (like blob) from my PR. How do you feel about this?

KenanY:
Sounds good to me, but do note that `blob()` is _not_ defined [here](https://github.com/bitinn/node-fetch/blob/1.x/lib/body.js), hence why I didn't have it in my version.

maxjoehnk:
Oh, maybe i should've checked the version of the code I was typing.
Do you wanna reopen your PR? I'll keep my progress for when node-fetch v2 is released then.

maxjoehnk:
Closing until node-fetch v2 is released.

PR #1113, opened 2017-08-03
Add node-fetch's non-spec API

KenanY:
`node-fetch@1` implements some non-spec methods, [such as `Body#buffer()`](https://github.com/bitinn/node-fetch/blob/1.x/lib/body.js#L67-L76). Furthermore, methods like `entries()`, `keys()`, and `values()` are **not** available in `node-fetch@1` (but they will be in `node-fetch@2`). This patch fixes these issues, using a new libdef inspired by the TypeScript equivalent. I raised the required `flow` version to `@0.44` since that's the first version to support [`http.Agent`](https://nodejs.org/api/http.html#http_class_http_agent).


KenanY:
Whoops, duplicate of https://github.com/flowtype/flow-typed/pull/1111. I'm 2 hours too late 😆

KenanY:
Reopening since https://github.com/flowtype/flow-typed/pull/1111 is for `node-fetch@2`.

PR #1123, opened 2017-08-07
Add libdef for node-mime

lewisf:
Project is here: https://github.com/broofa/node-mime


KenanY:
Seems like the flow-version lower bound could be lowered further?

```
Tests for node-mime_v1.x.x/flow_v0.46.x- ran successfully on flow v0.25.0.
        Consider setting v0.25.0 as the lower bound!
```

PR #1133, opened 2017-08-09
Styled-Components: Fixing withTheme implementation + tests

mwalkerwells:
Previously the following would work:
```js
const NeedsFoo3ErrorClass: ReactComponentClass<{ foo: string, theme: Theme }> = withTheme(NeedsFoo1Class);
```
...but ``withTheme`` actually provides the ``{ theme: Theme }`` & the function should not require it as an input. I implemented the inverse.

With these changes, the following works as expected:
```js
const NeedsFoo1Class: ReactComponentClass<{ foo: string }> = withTheme(NeedsThemeReactClass);
```


mwalkerwells:
Not sure why it's only failing on ``Node.js: 6``...

marudor:
Sadly tests are currently flaky.
If one of the two works it usually works after a restart.

PR #1136, opened 2017-08-10
Improved flow types for Jest v20 + Flow >= 0.39

dalongi:
1) More accurate/strict types for things like describe(), test(), it(), ...
2) Include TArguments and TReturn generics on jest.fn() so flow can verify that the params and return types of mocked functions match the original.

The "...args: TArguments" annotation fails on Flow <= 0.38 so I couldn't just add them to the existing definition and test files.


yury:
Please add timeout to `it`. See jasmine [docs](https://jasmine.github.io/api/2.7/global.html#it)

dalongi:
@yury: Interesting, the timeout API isn't documented in Jest (https://facebook.github.io/jest/docs/api.html#testname-fn). It was removed from master as of https://github.com/facebook/jest/pull/3668 but supposedly will come back once https://github.com/facebook/jest/pull/3819 is merged.

Will update.

yury:
Cool, thanks.

yury:
👍

yinanfang:
@dalongi Looks like you added the timeout for some of the function but not "it". The timeout parameter is not in https://github.com/flowtype/flow-typed/blob/master/definitions/npm/jasmine_v2.4.x/flow_all/jasmine_v2.4.x.js and I'm still getting the error from eslint

dalongi:
@yinanfang : This diff is only adding flow types for Jest, not Jasmine.

yinanfang:
@dalongi Oh. My bad. I think the Jamine also have the similar issue. Should I open another pull request and modify the file I pointed out?

dalongi:
@yinanfang: Yes, I think that makes sense. Thanks!

yinanfang:
@dalongi I've created a pull request #1180. Can you take a look when you're free?

PR #1137, opened 2017-08-10
Styled-Components: Cleaned ComponentList formatting & added robustness to withComponent

mwalkerwells:
This now correctly fails...
```js
const error = styled.div.withComponent('NotHere')
```


mwalkerwells:
Object spread is limited to Flow v42+. The version scope has been updated to accommodate.

PR #1144, opened 2017-08-15
fix(lodash): Fix type def for `attempt`

SpainTrain:
- Fixes `unused argument` error when using `_.attempt`

<https://lodash.com/docs/4.17.4#attempt>


marudor:
Could you add a test for this?

SpainTrain:
> Could you add a test for this?

:white_check_mark:

PR #1147, opened 2017-08-16
flow v0.53 definitions for recompose

istarkov:
Having that React types has changed previous types will not work with flow versions above 0.53


istarkov:
Wanna add I gave up trying to run tests locally.

GAntoine:
If I might ask, why did you give up running them locally? Were they taking too long?

istarkov:
Automagically merge conflicts appearing also does not make life easier ;-)
Please merge

istarkov:
BTW I have a question:
In recompose I decided to hold the type definitions inside recompose package
https://github.com/acdlite/recompose/tree/master/types/flow-typed/recompose_v0.24.x/flow_v0.53.x-

It's normal when maintainers of a package hosts their definitions? Isn't it? 
The idea that it allows to hold hand on a pulse, so we can provide typedef immediately on something has changed, and also it's easier to fix typedef as test env is really light and fast.

But what I see for now, every time on sync we have different prettier settings, there are commits with eslint rules here? Why? 

Is any good way now to host libdefs locally and remove them from current package?

istarkov:
Seems like check fail has nothing with this commit

GAntoine:
Would you mind closing and reopening the PR? That should fix the travis job.

PR #1156, opened 2017-08-20
Feat/styled components flow53

NullDivision:
- add v0.53.x flow typedefs for styled-components v2


definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -0,0 +1,332 @@
...
+
+type $npm$styledComponents$Interpolation = ((executionContext: Object) => string) | string | number;
+type $npm$styledComponents$NameGenerator = (hash: number) => string;
+
+type $npm$styledComponents$TaggedTemplateLiteral<R> = {| (Array<string>, $npm$styledComponents$Interpolation): R |};
+
+type $npm$styledComponents$ReactComponentFunctional<Props: {}> = Props => React$Element<*>;
+type $npm$styledComponents$ReactComponentClass<Props, DefaultProps = *> = Class<React$Component<DefaultProps & Props, *>>
+
+type $npm$styledComponents$ReactComponentUnion<Props> =
```
wtgtybhertgeghgtwtg: Is there a reason not to use `React$ComponentType` in `flow@0.53`?

NullDivision: I tried replacing it but ran across several issues that keep compounding. I'd love for someone to take a look at it.

My best guess is that there's a buggy implementation somewhere in there or test mismatch but I've yet to discover it.

wtgtybhertgeghgtwtg: What sort of issues?  I ran into some when I was working on bumping it locally, but they resolved once everything was switched over.

klaasn: I tried to work a bit on it as well, but the code really seems a bit hard to understand so I felt rather uncomfortable. But it really seemed like it could be simplified a lot.

f.e. i rewrote the withTheme stuff like this (following the documentation on the flow site for injection of props), but as this replaces 15 lines with 3 I was afraid I was missing something.

declare function $npm$styledComponents$WithTheme<Props: {}>(
  Component: React$ComponentType<{ theme: $npm$styledComponents$Theme & Props}>
) : React$ComponentType<Props>

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/test_styled-components_v2.x.x.js
```diff
@@ -0,0 +1,260 @@
...
+  testProp: 'foo'
+})`
+  background-color: red;
+`;
+
+const theme: Theme = {
+  background: "papayawhip"
+};
+
+// ---- WithComponent ----
```
michalkvasnicak: Should not there be a test for `styled(Component).withComponent('a')` for example?  I was trying this PR in our codebase and it failed on `withComponent()` calls when there is a custom component wrapped in `styled`

michalkvasnicak: Basically I ended up with [this definition (see gist)](https://gist.github.com/michalkvasnicak/5bd086562487bd3a6c72b2bc078442e4) and it is working just fine. Only thing is that I have to typecast a return type like this `const A: typeof Component = styled(Component).withComponent('a')`.

marudor:
@wtgtybhertgeghgtwtg 
@michalkvasnicak 
Ready to merge?

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/test_styled-components_v2.x.x.js
NullDivision: Your definition seems to throw errors on my end. Could you also post the test files for it?

Starting to think rewriting it would work be easier.

mwalkerwells:
Just now saw this PR... I did a pretty substantial refactor on this originally & added a lot a tests a month or so ago.

Would love to simplify this definition (without removing type safety), but overall, typing this library proves to be a bit complicated.

https://github.com/flowtype/flow-typed/pull/1194

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
mwalkerwells: ``React$ComponentType`` has a ``Props`` type parameter, but your return type omits it.
```js
declare function $npm$styledComponents$WithTheme<Props: {}>(
Component: React$ComponentType<{ theme: $npm$styledComponents$Theme & Props}>
) : React$ComponentType
```

mwalkerwells: As of right now, I'm typing ``withTheme`` as a morphism of both ``Props`` and ``DefaultProps``. If a component requires a theme prop, ``withTheme`` removes the need to provide it when instantiating (``<Component />``) & "adds" (intersection) the prop to the ``DefaultProps`` of the component.

mwalkerwells:
```js
declare var needsComponentWithDefaultProp: Class<React$Component<{}, { theme: $npm$styledComponents$Theme }>> => void

class NeedsThemePropComponent extends React.Component<{ theme: $npm$styledComponents$Theme }, {}> {}

const Foo = withTheme(NeedsThemePropComponent)

needsComponentWithDefaultProp(Foo) // OK
<Foo /> // OK
```

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/test_styled-components_v2.x.x.js
mwalkerwells: @michalkvasnicak What is the type of your custom component? Can you shoe the test case that fails?

mwalkerwells: @michalkvasnicak I was just about to write a test case for this, but I don't think this should work for custom components...
```js
class CustomComponent extends React.Component<{ foo: string }> {
  render() {
    return <div>{this.props.foo}</div>
  }
}
const CustomComponentNewTag = styled(CustomComponent).withComponent('a') // ???
```
How exactly should ``CustomComponentNewTag`` behave in this case?

I understand how it would work when using a default component from ``$npm$styledComponents$StyledComponentsComponentList``, but not for custom components.

https://www.styled-components.com/docs/basics#extending-styles

mwalkerwells: Right now ``styled(CustomComponent)`` has the ``withComponent`` property and shouldn't...

mwalkerwells:
@michalkvasnicak @NullDivision @wtgtybhertgeghgtwtg Any of you get a chance to take a look at https://github.com/flowtype/flow-typed/pull/1194 & my comments on this PR? I'd love to get something pushed...

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/test_styled-components_v2.x.x.js
NullDivision: I think if we remove custom components from the input list in `$npm$styledComponents$ReactComponentStyledStaticPropsWithComponent` we should get what we want.

mwalkerwells: @NullDivision https://github.com/flowtype/flow-typed/pull/1194#issuecomment-327029569

mwalkerwells:
@NullDivision Are you ok with closing this PR in favor of https://github.com/flowtype/flow-typed/pull/1194?

NullDivision:
@mwalkerwells, code seems ok. I like the cleanup, so yes.

PR #1160, opened 2017-08-20
Rename parameter so that it does not conflict with a reserved word

vinsonchuong:
- Fixes "SyntaxError: default is a reserved word" from eslint
- Commit hook has also reformatted this file


vinsonchuong:
Note that a commit hook has reformatted the files in this pull request, causing the diff to look pretty extreme.

PR #1186, opened 2017-08-28
[ramda] .prop() return type should be a non-maybe type

Peeja:
Currently, the return value of of Ramda's `.prop()` (after currying) is `?T`, where `T` is the type of the values of the object it's getting a property from. I propose that, while sound, this keeps Flow from catching mistakes it could otherwise catch, and that it would be better to type it as `T`.

There are two cases here: the case where the object (type `O`) is defined with an indexer property, and the case without.

Without an indexer property, the keys of `O` are known. Thus, `Ramda.prop('foo', {bar: 1})` is an error, because `'foo'` does not match the type `$Keys<{bar: number}>`. Without an indexer property, then, `Ramda.prop()` can never fail to return a `T`. Returning `?T`, then, requires handling a `null`/`undefined` return value which can never occur at runtime.

With an indexer property, Flow doesn't know the keys of `O`. Thus, `Ramda.prop('foo', (o: {[string]: number}))` could return `undefined` if `'foo'` is not in that `o` at runtime. But [Flow has already decided that property access for objects with indexer properties is unsafe](https://flow.org/en/docs/types/objects/#toc-objects-as-maps), and it's the programmer's responsibility to ensure a value exists at the key.

Therefore, I think it makes the most sense for Ramda's `.prop()` to follow the same pattern as normal property access and return `T` instead of `?T`.

(This change should apply to more versions than the one given, but I'll add them later if I get a thumbs up for this change.)

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.39.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.39.x-/ramda_v0.x.x.js
@@ -601,8 +601,8 @@ declare module ramda {
   declare function project<T>(keys: Array<string>, ...rest: Array<void>): (val: Array<{[key:string]: T}>) => Array<{[key:string]: T}>;
   declare function project<T>(keys: Array<string>, val: Array<{[key:string]: T}>): Array<{[key:string]: T}>;
 
-  declare function prop<T,O:{[k:string]:T}>(key: $Keys<O>, ...rest: Array<void>): (o: O) => ?T;
-  declare function prop<T,O:{[k:string]:T}>(key: $Keys<O>, o: O): ?T;
+  declare function prop<T,O:{[k:string]:T}>(key: $Keys<O>, ...rest: Array<void>): (o: O) => T;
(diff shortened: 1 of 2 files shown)


GAntoine:
@Peeja Could you rebase and apply the changes to all the files? IMO this looks good, but I'm no expert so someone else will have to take a look.

@mwalkerwells @mattgstevens Could either of you take a look at this change? I'm out of my element here. 😅  Thanks!

mattgstevens:
Unless Im missing something, this seems like a reasonable change. 

`flow` will be able to pick up the non-"indexer property" case as expected (either inferred, or with a provided type), and based on the official docs linked the "indexer property" case (which would must have been typed as such) makes further checks needed as this is clearly a very permissive type.

It may have been intended to help in the "indexer property" case, but I don't see the reason to always return a `Maybe T` here.

@Peeja it would be worthwhile to add a test showing the "bad" behaviour when using `prop` fn with an `O` type that has the "indexer property" like you did here:
https://github.com/flowtype/flow-typed/pull/1186/commits/5baa1f0366da912a23a5d921c83a25d931086080#diff-b6649b612997ad895acca2c33510ee73R129

GAntoine:
Closed in favour of #1450

PR #1191, opened 2017-08-29
Add a libdef for fetch-jsonp

julienw:


GAntoine:
Hey @julienw, seems the specs are failing on anything below flow v39. If you're still interested in seeing this merged, you'll need to change the folder from `flow_all` to `flow_v0.39.x-`. You're also welcome to add a libdef for versions below 39, but I doubt any version that low is still widely used.

julienw:
Thanks for your guidance, I just changed it.
Let's see what Travis says :)

GAntoine:
Travis can really be grumpy some days... Looks like this time it's `39.0` that's causing issues. Bump it up to `40.0`, and if there's still problems we'll address it differently.

julienw:
Actually I'm pretty sure 39 was broken earlier too :) But I still applied blindly what you suggested ;)

Pushed the fix :)

julienw:
Actually it's not clear to me where I should put `// $ExpectError` (especially in https://github.com/flowtype/flow-typed/pull/1191/files#diff-92b09102b7fddfd22b11579d5986479cR24) and maybe that's the reason why there is a failure on earlier Flow versions. Should I put it above line 23 ?

GAntoine:
Yeah I think it goes above 23. Try the change locally and lowering the version, and if the tests run on your system, go ahead and push a lower floe version.

julienw:
I did it and could lower the lower bound. I noticed that only Flow 0.36 and 0.37 don't handle the error properly; earlier Flow versions actually work fine :)

It's likely not a big deal because as you said earlier few people likely use these earlier verisons.

PR #1194, opened 2017-08-30
Styled-Components: Adding compatibility for React updates in Flow v0.53+ & Bug Fixes

mwalkerwells:
* Added defaultProps typing to Functional components
* Fixed issue with passing an inferred functional type into ``styled()``
* Improved type name consistency
* Switched from a custom functional component type to ``React$StatelessFunctionalComponent<Props>``
* Updated ``React$Element`` references to ``React$Node``
* Removed instances of a type param being passed to ``StyledComponentsComponentListKeys``
* Added type export for ListKeys
* Added the ability to pass a key from ListKeys into ``styled(/* ListKeys  */)``
* Corrected ``withComponent`` type to disallow custom types
* Added new relevant tests

Flow v0.53: https://github.com/facebook/flow/releases/tag/v0.53.0


mwalkerwells:
This is now possible in Flow...
```js
import React from 'react'

declare var Component: {
  <Props: {}>(Props): React$Element<*>,
  defaultProps: {}
}
  
<Component />
```

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -0,0 +1,369 @@
...
+
+// ---- CLASS COMPONENT DEFINITIONS ----
+class $npm$styledComponents$ReactComponent<Props, DefaultProps> extends React$Component<Props> {
+  static defaultProps: DefaultProps
+}
+type $npm$styledComponents$ReactComponentClass<Props, DefaultProps = *> = Class<$npm$styledComponents$ReactComponent<Props, DefaultProps>>
+type $npm$styledComponents$ReactComponentClassUndefinedDefaultProps<Props> = Class<React$Component<Props>>
+
+// ---- COMPONENT FUNCTIONS INPUT (UNION) & OUTPUT (INTERSECTION) ----
+type $npm$styledComponents$ReactComponentUnion<Props> =
```
wtgtybhertgeghgtwtg: Same as #1156, is there a problem with `React$ComponentType`?

mwalkerwells: IIRC, ``React$ComponentType`` did not exist when I wrote this type originally.

This definition references the use of defaultProps, whereas ``React$ComponentType`` does not.

```js
// ---- FUNCTIONAL COMPONENT DEFINITIONS ----
type $npm$styledComponents$ReactComponentFunctional<Props, DefaultProps = *> = {
  (Props): React$Element<*>,
  defaultProps: DefaultProps,
}
```
[#L8](https://github.com/mwalkerwells/flow-typed/blob/0b783539a7d1fda556b973c01c3289f7875be938/definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js#L8)

``$npm$styledComponents$ReactComponentUnion<Props>`` should be contravariant WRT ``React$ComponentType``, which should make it more flexible, but still safe.

That being said, we should update the definition:
```js
// ---- COMPONENT FUNCTIONS INPUT (UNION) & OUTPUT (INTERSECTION) ----
type $npm$styledComponents$ReactComponentUnion<Props> =
  | $npm$styledComponents$ReactComponentFunctional<Props>
  | $npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props>
  | $npm$styledComponents$ReactComponentClass<Props>
  | $npm$styledComponents$ReactComponentClassUndefinedDefaultProps<Props>
  | React$StatelessFunctionalComponent<Props>; // ADDED
```
...and this:
```js
type $npm$styledComponents$ReactComponentClassUndefinedDefaultProps<Props> = Class<React$Component<Props, *>>
```
^^ without this, the  [default ``State`` is void](https://github.com/facebook/flow/blob/master/lib/react.js#L29), & we want it to be inferred.

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -0,0 +1,369 @@
+// @flow
+
+type $npm$styledComponents$Interpolation = (<C: {}>(executionContext: C) => string) | string | number;
+type $npm$styledComponents$NameGenerator = (hash: number) => string;
+
+type $npm$styledComponents$TaggedTemplateLiteral<R> = {| (Array<string>, $npm$styledComponents$Interpolation): R |};
+
+// ---- FUNCTIONAL COMPONENT DEFINITIONS ----
+type $npm$styledComponents$ReactComponentFunctional<Props, DefaultProps = *> = {
+  (Props): React$Element<*>,
```
mwalkerwells: @wtgtybhertgeghgtwtg We actually need to update this to be ``React$Node``, not ``React$Element``.

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
wtgtybhertgeghgtwtg: > IIRC, React$ComponentType did not exist when I wrote this type originally.

It's only been out since `flow@0.53`, I wouldn't be surprised.
> This definition references the use of defaultProps, whereas React$ComponentType does not.

What benefits does `defaultProps` provide here?

mwalkerwells: Tried to summarize it here (https://github.com/flowtype/flow-typed/pull/1156#discussion_r136143193 / https://github.com/flowtype/flow-typed/pull/1156#discussion_r136144500), but I can probably do better...

If you think of a React component as a callable object that takes props and context as input and returns a ``React$Node``, and has a ``defaultProps`` property...a HOC (a function that takes and returns an object), is merely a callable object morphism, typically transforming ``Props`` and ``DefaultProps``.

mwalkerwells:
```js
declare var Component: {
  ({ foo: string }): React$Node,
  defaultProps: {},
}

declare var HOC: ({
  ({ foo: string }): React$Node,
  defaultProps: {},
}) => ({
  ({}): React$Node,
  defaultProps: { foo: string },
})

const Test = HOC(Component)

<Test /> // No ``foo`` prop
```

mwalkerwells: That's why use ``DefaultProps`` in my definition, because it's a lossy type morphism without it.

wtgtybhertgeghgtwtg: While [it isn't explicitly typed](https://github.com/facebook/flow/blob/v0.54.0/lib/react.js#L81-L86), the intention of `defaultProps` seems to be that it'd be a subtype of `Props`.

mwalkerwells: Hmmm... you mean that ``DefaultProps`` should be a supertype of ``Props``? I had read [that](https://github.com/facebook/flow/blob/v0.54.0/lib/react.js#L81-L86) before, but after re-reading it, assuming you meant supertype,  I think you're right...

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -0,0 +1,369 @@
...
+  declare type ReactComponentClass<Props: {}, DefaultProps: ?{} = *>      = $npm$styledComponents$ReactComponentClass<Props, DefaultProps>;
+  declare type ReactComponentClassUndefinedDefaultProps<Props: {}>        = $npm$styledComponents$ReactComponentClassUndefinedDefaultProps<Props>;
+  declare type ReactComponentUnion<Props>                                 = $npm$styledComponents$ReactComponentUnion<Props>;
+  declare type ReactComponentIntersection<Props>                          = $npm$styledComponents$ReactComponentIntersection<Props>;
+  declare type ReactComponentStyledStaticProps<Props>                     = $npm$styledComponents$ReactComponentStyledStaticProps<Props, $Subtype<$Keys<$npm$styledComponents$StyledComponentsComponentList<*>>>>;
+  declare type ReactComponentStyled<Props>                                = $npm$styledComponents$ReactComponentStyled<Props, $Subtype<$Keys<$npm$styledComponents$StyledComponentsComponentList<*>>>>;
+  declare type ReactComponentStyledTaggedTemplateLiteral<Props>           = $npm$styledComponents$ReactComponentStyledTaggedTemplateLiteral<Props, $Subtype<$Keys<$npm$styledComponents$StyledComponentsComponentList<*>>>>;
```
wtgtybhertgeghgtwtg: > $npm$styledComponents$StyledComponentsComponentList<*>

`$npm$styledComponents$StyledComponentsComponentList` doesn't take a type argument.

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -0,0 +1,369 @@
...
+
+  declare module.exports: {
+    <Props>(ReactComponentUnion<Props>): ReactComponentStyledTaggedTemplateLiteral<Props>,
```
wtgtybhertgeghgtwtg: This should also be able to take a string.

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
mwalkerwells: Ah, you're right. Missed updating that...

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
mwalkerwells: What do you mean?

mwalkerwells: Just saw you edited your comment. Can you elaborate on "sounds brittle and annoying"?

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
wtgtybhertgeghgtwtg: Or, better yet, it could just use ` $npm$styledComponents$StyledComponentsComponentListKeys`.

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
wtgtybhertgeghgtwtg: > What do you mean?

`styled('div')` is valid.  That's where [`styled.div` comes from](https://github.com/styled-components/styled-components/blob/master/src/constructors/styled.js#L9-L11), after all.

> Can you elaborate on "sounds brittle and annoying"?

Firstly, it'd have to be updated whenever a new tag is added to the spec.  Far from a daily thing, but still.
Secondly, it'd make factory functions a pain.
```
function makeThing(tag: string) {
  return styled(tag)`/* styles */`;
}
```
wouldn't work, since `tag` isn't a `$Keys<$npm$styledComponents$StyledComponentsComponentList>`.

wtgtybhertgeghgtwtg: Also, it wouldn't work with custom elements.  I don't know why you would use custom elements with React, but I'm sure somebody would complain about it.

mwalkerwells: Ah, you're right about it needing to take a ``string``. I'll update that.

mwalkerwells: I originally added ``Keys<$npm$styledComponents$StyledComponentsComponentList>`` so that you get really helpful typeahead suggestions, but of course that's not relevant to passing in a string.

mwalkerwells: What do you think about exporting the ``$npm$styledComponents$StyledComponentsComponentListKeys`` type for factory functions?

mwalkerwells: This is unsafe 99% of the time... Keeping this library definition up-to-date should be trivial, as new native elements are extremely infrequent.
```js
function makeThing(tag: string) {
  return styled(tag)`/* styles */`;
}
```

wtgtybhertgeghgtwtg: Exporting `StyledComponentsComponentListKeys` would work, yeah.

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -7,25 +7,26 @@ type $npm$styledComponents$TaggedTemplateLiteral<R> = {| (Array<string>, $npm$st
 
 // ---- FUNCTIONAL COMPONENT DEFINITIONS ----
 type $npm$styledComponents$ReactComponentFunctional<Props, DefaultProps = *> = {
-  (Props): React$Element<*>,
+  (Props): React$Node<*>,
   defaultProps: DefaultProps,
 }
 
-type $npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props> = { (Props): React$Element<*> }
+type $npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props> = { (Props): React$Node<*> }
```
wtgtybhertgeghgtwtg: `React$Node` [doesn't take a type argument](https://github.com/facebook/flow/blob/v0.54.0/lib/react.js#L16).

mwalkerwells: Man, you're on top of this! :)

wtgtybhertgeghgtwtg: Wow, that was fast.

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -0,0 +1,372 @@
...
+  declare type ReactComponentClass<Props: {}, DefaultProps: ?{} = *>      = $npm$styledComponents$ReactComponentClass<Props, DefaultProps>;
+  declare type ReactComponentClassUndefinedDefaultProps<Props: {}>        = $npm$styledComponents$ReactComponentClassUndefinedDefaultProps<Props>;
+  declare type ReactComponentUnion<Props>                                 = $npm$styledComponents$ReactComponentUnion<Props>;
+  declare type ReactComponentIntersection<Props>                          = $npm$styledComponents$ReactComponentIntersection<Props>;
+  declare type ReactComponentStyledStaticProps<Props>                     = $npm$styledComponents$ReactComponentStyledStaticProps<Props, ComponentListKeys>;
+  declare type ReactComponentStyled<Props>                                = $npm$styledComponents$ReactComponentStyled<Props, ComponentListKeys>;
+  declare type ReactComponentStyledTaggedTemplateLiteral<Props>           = $npm$styledComponents$ReactComponentStyledTaggedTemplateLiteral<Props, ComponentListKeys>;
+
+  declare module.exports: {
+    <Props, Key: ComponentListKeys>(Key | ReactComponentUnion<Props>): ReactComponentStyledTaggedTemplateLiteral<Props>,
```
wtgtybhertgeghgtwtg: Why take type argument `Key` instead of just taking `(ComponentListKeys | ReactComponentUnion<Props>)`?

mwalkerwells: The funny bit is that I originally had it as ``(ComponentListKeys | ReactComponentUnion<Props>)``, but changed it because I wrongly _remembered_ that ``ComponentListKeys`` wasn't inside a ``$Subtype``.

Updated.

mwalkerwells: I'm actually about to change it anyways to address this: https://github.com/flowtype/flow-typed/pull/1156#pullrequestreview-58414828

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -0,0 +1,372 @@
...
+class $npm$styledComponents$ReactComponent<Props, DefaultProps> extends React$Component<Props> {
+  static defaultProps: DefaultProps
+}
+type $npm$styledComponents$ReactComponentClass<Props, DefaultProps = *> = Class<$npm$styledComponents$ReactComponent<Props, DefaultProps>>
+type $npm$styledComponents$ReactComponentClassUndefinedDefaultProps<Props> = Class<React$Component<Props, *>>
+
+// ---- COMPONENT FUNCTIONS INPUT (UNION) & OUTPUT (INTERSECTION) ----
+type $npm$styledComponents$ReactComponentUnion<Props> =
+  | $npm$styledComponents$ReactComponentFunctional<Props>
+  | $npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props>
```
wtgtybhertgeghgtwtg: Integration tests (i.e. me copypasting this into a project that uses `styled-components`) are showing that Flow can't pick between `$npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props>` and `React$StatelessFunctionalComponent<Props>` when `styled` gets passed a functional component.  Have you not run into this error?

mwalkerwells: Can you show the test?

wtgtybhertgeghgtwtg:
```
styled(() => <div />);
```

definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js
```diff
@@ -11,7 +11,8 @@ type $npm$styledComponents$ReactComponentFunctional<Props, DefaultProps = *> = {
   defaultProps: DefaultProps,
 }
 
-type $npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props> = { (Props): React$Node }
+type $npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props> =
```
wtgtybhertgeghgtwtg: If it's `React$StatelessFunctionalComponent<Props>`, why not just use `React$StatelessFunctionalComponent<Props>`?

mwalkerwells: Only for clarity of the definitions...
```js
type $npm$styledComponents$ReactComponentUnionWithDefaultProps<Props, DefaultProps> =
  | $npm$styledComponents$ReactComponentFunctional<Props, DefaultProps>
  | $npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props>
  | $npm$styledComponents$ReactComponentClass<Props, DefaultProps>
  | $npm$styledComponents$ReactComponentClassUndefinedDefaultProps<Props>
```

mwalkerwells: These global types are already not fun to work with, so trying to make it somewhat easier to understand...

mwalkerwells: I thought about using underscores to make it easier on the eyes, but it's a bit unconventional...
```js
type $npm$styledComponents$ReactComponent__Union__WithDefaultProps<Props, DefaultProps> =
  | $npm$styledComponents$ReactComponent__Functional<Props, DefaultProps>
  | $npm$styledComponents$ReactComponent__Functional__UndefinedDefaultProps<Props>
  | $npm$styledComponents$ReactComponent__Class<Props, DefaultProps>
  | $npm$styledComponents$ReactComponent__Class__UndefinedDefaultProps<Props>
```

wtgtybhertgeghgtwtg: I fail to see how `$npm$styledComponents$ReactComponentFunctionalUndefinedDefaultProps<Props>` is clearer than `React$StatelessFunctionalComponent<Props>`, but it's your call.

mwalkerwells:
Alright, I've addressed everything discussed here, except for the HOC Props/DefaultProps morphism, which I think should be addressed separately (it's a much bigger change and less type-safe, which means removing a lot of tests). I plan on posting on the Flow repo & I'll cc you two there to discuss.

@wtgtybhertgeghgtwtg You notice anything else?
@NullDivision We're good to go regarding your last comment: https://github.com/flowtype/flow-typed/pull/1156#discussion_r136747743
```js
// ---- STYLED FUNCTION ----
// Error: styled(CustomComponent).withComponent('a')
// Ok:    styled('div').withComponent('a')
```
https://github.com/mwalkerwells/flow-typed/blob/846c1ddfebbe40e0f8d389445a244d3a59da3bd8/definitions/npm/styled-components_v2.x.x/flow_v0.53.x-/styled-components_v2.x.x.js#L58-L60

wtgtybhertgeghgtwtg:
I don't see anything else.  I've already switched my factory functions to use `ComponentListKeys` and it's working fine.

mwalkerwells:
Alright, @wtgtybhertgeghgtwtg @NullDivision, I just submitted a PR to the Flow repo...

NullDivision:
I'm still not sure about the continued use of DefaultProps since it's been dropped from Flows lexicology. Also wondering whether to `Class<React$Component>` types could be replaced with `React$ComponentType` but it works fine on my end so as far as I'm concerned it's good for me.

For now I'm more concerned with making the definition work with v53 but I'm wondering if we can (post-release) retype it using the TS definitions in the package itself.

mwalkerwells:
> concerned with making the definition work with v0.53

Let me know if you find something that's not working!

> wondering whether to Class<React$Component> types could be replaced with React$ComponentType

Using ``React$ComponentType`` would just mean less precise typing & changing this definition to use it should be coupled with updates related to updating the handling of ``Props``/``DefaultProps`` outlined here: https://github.com/facebook/flow/pull/4806.

>TS definitions in the package itself.

Can't speak to the Typescript definitions, but I'd imagine they're less type robust.


Ok, glad to hear you are 👍🏻. Independent of the ``Props``/``DefaultProps`` discussion, this library definition is both compatible & more type roubst. Win win.

mwalkerwells:
@marudor Ready to 🛬! 😀

omninonsense:
Not entirely sure why from quickly glancing at the type defs, but this breaks flow coverage for simple cases like these:

~~~js
const Wrapper = styled.div``
~~~

If you use Nuclide and enable flow coverage, this will report that `Wrapper` (and the return value of `styled.div`) is "Not covered by Flow". Sadly, I don't have much time now so I can't look at this myself (I might look at it next week if no one beats me to it, though).

Issue #1140, opened 2017-08-13
Help with Mongoose typings


albertorestifo:
I'm trying to create typings for Mongoose, but I don't know if what I want to do is possible.

I want to start simple, so let's ignore for now all the possible ways to define a model in Mongoose. My goal is to make something like this possible, where looking up a property that wasn't defined in the schema will error in Flow.

```js
import mongoose from 'mongoose';

const UserSchema = new mongoose.Schema({
  email: {
    type: String, 
    required: true 
  },
  createdAt: { 
    type: Date, 
    default: Date.now 
  },
});

const User = mongoose.model('User', UserSchema);

// And let's run a query

User.findOne({ email: 'example@example.com' }).exec()
  .then((user) => {
    // $ExpectError
    user.lol
  })
```

This is what I came up with, which doesn't achieve my goal:

```typescript
declare class Mongoose$Schema<Schema> {}

type Mongoose$Query<Schema> = {| [$Keys<Schema>]: any |}
type Mongoose$Projection<Schema> = { [$Keys<Schema>]: any }
                                    
declare type Mongoose$Result<Schema> = {| [$Keys<Schema>]: any |}                                   
                                    
declare interface Mongoose$QueryInterface<Schema> {
  exec(): Promise<Mongoose$Result<Schema> | void>
}

declare class Mongoose$Model<Schema> {
  findOne(query: Mongoose$Query<Schema>, projection?: Mongoose$Projection<Schema>): Mongoose$QueryInterface<Schema>
}
```

Is what I wish to do even possible?

mwalkerwells:
This works correctly, with one downside: I cannot transform the Type using an intersection (commented out type definition). ☹️

In this example, ``user.email`` is an instance of ``String``, not a ``string`` literal.

```js
// @flow

// type TypeTransform =
//   & (({ type: typeof String }) => string)
//   & (({ type: typeof Date }) => Date)

type TypeTransform = <T>({ type: T }) => T

type MongooseSchemaConfigTransform<Config>
  = $ObjMap<Config, TypeTransform>

declare module 'mongoose' {
  declare class Schema<Config> {
    constructor(config: Config): void
  }
  declare type Model<Config> = {
    findOne: ($Shape<MongooseSchemaConfigTransform<Config>>) =>
      { exec: void => Promise<MongooseSchemaConfigTransform<Config>> },
  }

  declare export default {
    model: <Config>(string, Schema<Config>) => Model<Config>,
[...]

PR #1196, opened 2017-08-30
Add lib defs for mongoose@4.x.x

nodkz:
This libdef covers almost all regular use cases for mongoose models.
Fill free to extend or fix unregular methods and props.

**PLEASE do not add callbacks args (for `.save()`, `.find()` and other methods). 
Promises all the things!!!**

Related #1140
Related https://github.com/entria/flow-typed-wip/issues/1

### DIFF
--- a/definitions/npm/mongoose_v4.x.x/flow_v0.50.x-/mongoose_v4.x.x.js
+++ b/definitions/npm/mongoose_v4.x.x/flow_v0.50.x-/mongoose_v4.x.x.js
@@ -1,0 +1,483 @@
+import mongoose from "mongoose";
+
+type MongoId =
+  | typeof mongoose.Types.ObjectId
+  | {
+      toString(): string
+    };
+
+type MongoOrScalarId = MongoId | string | number;
+
+type SchemaFields = {
+  [fieldName: string]: any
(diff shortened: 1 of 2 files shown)


nodkz:
By the way, how can I enable getters/setters in tests?
https://github.com/nodkz/flow-typed/blob/73f2aa162363ee43f6a6a4e70f21b17a513c9c82/definitions/npm/mongoose_v4.x.x/test_mongoose-v4.js#L44-L64
```js
  // NOT SUPPORTED in flow-typed tests
  // But will work with your code if you add in your .flowconfig
  //   [options]
  //   unsafe.enable_getters_and_setters=true
  //
  // get password(): ?string {
  //   return this._plainPassword;
  // }
  //
  // set password(password: string): void {
  //   this._plainPassword = password;
  //   this.token = this.encryptPassword(password);
  // }
  //
  // get fullname(): string {
  //   const fullname = [];
  //   if (this.lastname) fullname.push(this.lastname);
  //   if (this.name) fullname.push(this.name);
  //   if (this.middlename) fullname.push(this.middlename);
  //   return fullname.join(' ');
  // }
```

PR #1203, opened 2017-09-01
[knex] Update knex.js for Flow v0.54

avaly:
From Flow v0.54.0 release notes:

> Extending a polymorphic class must now explicitly specify the parent
> class's type args. That is, `class A<T> {}; class B extends A {}` is now
> an error, but class `C extends A<string> {}` or `class D<T> extends A<T> {}` is ok.

- [x] Added generic types to `Promise` and all classes which use `Promise`
- [x] Added tests for promise syntax
- [x] Added missing definitions for `transaction` and `into`


avaly:
PS: the `prettier` and `lint-staged` integration on git commit hooks strip away any `mixins` from classes definitions.

```js
declare class Knex$Transaction<R> mixins Knex$QueryBuilder<R>, events$EventEmitter, Promise<R> {
```
becomes this:

```js
declare class Knex$Transaction<R> {
```

I had to commit with `git commit --no-verify` to bypass this.

PR #1205, opened 2017-09-01
Fix prettier errors throw during `./run_def_tests.sh`

johnhaley81:
A number of prettier errors were thrown when running
`./run_def_tests.sh`. This fixes those.

Also when running said command the `definitions/package-lock.json` was
updated with some new versions. I'm not sure why that happened. I can
remove that change if needed.

### DIFF
--- a/cli/src/commands/__tests__/install-test.js
+++ b/cli/src/commands/__tests__/install-test.js
@@ -1,107 +1,107 @@
 // @flow
 
-(require('../../lib/git'): any).rebaseRepoMaster = jest.fn();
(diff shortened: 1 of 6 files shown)


johnhaley81:
Yikes, some hook in the package ran a different version of prettier I think? Which destroyed the formatting and made all of the linter tests fail.

johnhaley81:
So it seems like there are 3 different versions of prettier fighting for dominance. 1 when running `./run_def_tests.sh`, 1 when making a commit and 1 on CI. I'm not sure where they're coming from, I'll look into it more.

johnhaley81:
This should fix #1206

PR #1219, opened 2017-09-05
Update to `react-addons-css-transition-group`

EvHaus:
The `enter`, `leave` and `appear` keys of the `ReactCSSTransitionGroupNames` object are actually all totally optional.

See: https://github.com/reactjs/react-transition-group/blob/v1-stable/src/utils/PropTypes.js#L38 (all properties are missing `.isRequired`)

### DIFF
--- a/definitions/npm/react-addons-css-transition-group_v15.x.x/flow_v0.26.x-v0.52.x/react-addons-css-transition-group_v15.x.x.js
+++ b/definitions/npm/react-addons-css-transition-group_v15.x.x/flow_v0.26.x-v0.52.x/react-addons-css-transition-group_v15.x.x.js
@@ -1,10 +1,10 @@
 declare module 'react-addons-css-transition-group' {
   declare type ReactCSSTransitionGroupNames = {
-    enter: string,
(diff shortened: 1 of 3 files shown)


GAntoine:
@EvHaus If you still want to see this merged in, you'll need to modify the failing spec.

EvHaus:
@GAntoine Where are you checking the prop types? According to the source code that I'm looking at (https://github.com/reactjs/react-transition-group/blob/v1-stable/src/CSSTransitionGroup.js#L8) all three (`enter`, `leave` and `appears` are optional). And the code works correctly when any are omitted.

Rebase + pushed a fix for broken test.

GAntoine:
I was looking at the first link you posted, but the second one seems to be the actual propTypes for `CSSTransitionGroup`. Honestly React and propTypes aren't really my "expertise", so I'll leave it up to someone who knows more to validate this change.

Unfortunately I can't merge this in while the specs are still failing. 😕

EvHaus:
Sorry about that. Tests should be fixed now.

PR #1225, opened 2017-09-05
Fix uuid definition

joeybaker:
The current definition leaves off the default export which, while deprecated is
valid still. https://github.com/kelektiv/node-uuid#deprecated-api

This is currently used in react-native, so using this definition breaks flow
types in RN >= 0.48.
https://github.com/facebook/react-native/blob/7d4a5b67d497ecc9a07447a95f73d8e2ee84573d/Libraries/Blob/Blob.js#L93


joeybaker:
hmmm… this doesn't fail for me locally.

joeybaker:
Tests fixed. Sorry about that.

PR #1226, opened 2017-09-05
Update chai_v4.x.x libdef to include 'message' parameter for expect

castinet:


GAntoine:
Hey @castinet, looks like your change indeed matches the API (and the ts definition)! If you're still keen on getting this merged, would you mind rebasing the change and seeing if it fixes the failing spec?

GAntoine:
Closing as this change has already been made. None the less, thanks for the PR!

PR #1228, opened 2017-09-06
Add library definition helmet_v3.x.x

followbl:
This PR adds library definitions for the [helmet](https://github.com/helmetjs/helmet) package.

This is my first contribution to flow-typed so please let me know if I can adjust/adhere to anything.


definitions/npm/helmet_v3.x.x/flow_v0.38.x-/helmet_v3.x.x.js
```diff
@@ -0,0 +1,93 @@
+import type { Middleware, Request, Response } from 'express'
```
marudor: This doesn't work. Sadly it will result in any.
Could should be able to import that within the module declaration though.

So could you move all the declares into the declare module?

haishengwu-okta: @marudor are you suggesting this TODO can be fixed now as well? Thanks.
https://github.com/flowtype/flow-typed/blob/master/definitions/npm/backbone_v1.x.x/flow_v0.20.x-/backbone_v1.x.x.js#L2

PR #1241, opened 2017-09-13
fix moment.parseZone()

jedwards1211:
According to https://momentjs.com/docs/#/parsing/parse-zone/, `parseZone` may now accept up to four arguments.
Additionally, I think someone failed to mark `parseZone()` `static`.


jedwards1211:
Actually wait, I think I need to make this for 2.3.0+

PR #1242, opened 2017-09-13
fix defs for moment.parseZone in v2.3.x+

jedwards1211:
This will work for `2.4.x`, `2.5.x`, etc. right?

### DIFF
--- a/definitions/npm/moment_v2.3.x/flow_v0.34.x-/moment_v2.3.x.js
+++ b/definitions/npm/moment_v2.3.x/flow_v0.34.x-/moment_v2.3.x.js
@@ -1,0 +1,238 @@
+type moment$MomentOptions = {
+  y?: number|string,
+  year?: number|string,
+  years?: number|string,
+  M?: number|string,
+  month?: number|string,
+  months?: number|string,
+  d?: number|string,
+  day?: number|string,
+  days?: number|string,
+  date?: number|string,
+  h?: number|string,
+  hour?: number|string,
+  hours?: number|string,
+  m?: number|string,
+  minute?: number|string,
+  minutes?: number|string,
+  s?: number|string,
+  second?: number|string,
+  seconds?: number|string,
+  ms?: number|string,
+  millisecond?: number|string,
+  milliseconds?: number|string,
+};
+
+type moment$MomentObject = {
+  years: number,
+  months: number,
+  date: number,
+  hours: number,
+  minutes: number,
+  seconds: number,
+  milliseconds: number,
+};
+
+type moment$MomentCreationData = {
+  input: string,
+  format: string,
+  locale: Object,
+  isUTC: bool,
+  strict: bool,
+};
+
+type moment$CalendarFormat = string | (moment: moment$Moment) => string;
+
+type moment$CalendarFormats = {
+  sameDay?: moment$CalendarFormat,
+  nextDay?: moment$CalendarFormat,
+  nextWeek?: moment$CalendarFormat,
+  lastDay?: moment$CalendarFormat,
+  lastWeek?: moment$CalendarFormat,
+  sameElse?: moment$CalendarFormat,
+};
+
+declare class moment$LocaleData {
+  months(moment: moment$Moment): string;
+  monthsShort(moment: moment$Moment): string;
+  monthsParse(month: string): number;
+  weekdays(moment: moment$Moment): string;
+  weekdaysShort(moment: moment$Moment): string;
+  weekdaysMin(moment: moment$Moment): string;
+  weekdaysParse(weekDay: string): number;
+  longDateFormat(dateFormat: string): string;
+  isPM(date: string): bool;
+  meridiem(hours: number, minutes: number, isLower: bool): string;
+  calendar(key: 'sameDay'|'nextDay'|'lastDay'|'nextWeek'|'prevWeek'|'sameElse', moment: moment$Moment): string;
+  relativeTime(number: number, withoutSuffix: bool, key: 's'|'m'|'mm'|'h'|'hh'|'d'|'dd'|'M'|'MM'|'y'|'yy', isFuture: bool): string;
+  pastFuture(diff: any, relTime: string): string;
+  ordinal(number: number): string;
+  preparse(str: string): any;
+  postformat(str: string): any;
+  week(moment: moment$Moment): string;
+  invalidDate(): string;
+  firstDayOfWeek(): number;
+  firstDayOfYear(): number;
+}
+declare class moment$MomentDuration {
+  humanize(suffix?: bool): string;
+  milliseconds(): number;
+  asMilliseconds(): number;
+  seconds(): number;
+  asSeconds(): number;
+  minutes(): number;
+  asMinutes(): number;
+  hours(): number;
+  asHours(): number;
+  days(): number;
+  asDays(): number;
+  months(): number;
+  asMonths(): number;
+  years(): number;
+  asYears(): number;
+  add(value: number|moment$MomentDuration|Object, unit?: string): this;
+  subtract(value: number|moment$MomentDuration|Object, unit?: string): this;
+  as(unit: string): number;
+  get(unit: string): number;
+  toJSON(): string;
+  toISOString(): string;
+}
+declare class moment$Moment {
+  static ISO_8601: string;
+  static (string?: string, format?: string|Array<string>, locale?: string, strict?: bool): moment$Moment;
+  static (initDate: ?Object|number|Date|Array<number>|moment$Moment|string): moment$Moment;
+  static unix(seconds: number): moment$Moment;
+  static utc(): moment$Moment;
+  static utc(number: number|Array<number>): moment$Moment;
+  static utc(str: string, str2?: string|Array<string>, str3?: string): moment$Moment;
+  static utc(moment: moment$Moment): moment$Moment;
+  static utc(date: Date): moment$Moment;
+  static parseZone(): moment$Moment;
+  static parseZone(rawDate: string): moment$Moment;
+  static parseZone(rawDate: string, format: string | Array<string>): moment$Moment;
+  static parseZone(rawDate: string, format: string, strict: boolean): moment$Moment;
+  static parseZone(rawDate: string, format: string, locale: string, strict: boolean): moment$Moment;
+  isValid(): bool;
+  invalidAt(): 0|1|2|3|4|5|6;
+  creationData(): moment$MomentCreationData;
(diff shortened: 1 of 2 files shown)


marudor:
This is for 2.3.x. So 2.3.0 2.3.1 and so on. Not 2.4.0

jedwards1211:
So...is there a way to put in defs for `2.3.x+`?  If not...facepalm

marudor:
Sadly not.
 I think...
I'm actually not sure.
@jeffmo @GAntoine ?

jedwards1211:
Pleeeease tell me someone thought of this during the initial design phase.

Since one can do `flow-v0.48.x-` to for the flow directories, I assume one can do it for the other package directories - I just haven't seen conclusive evidence anywhere that it does work.

GAntoine:
Thanks for bringing this up, AFAIK there's no way.

What's the breaking change that stops this from being `2.x.x`? Looking at a diff with `2.x.x`, and after reading the [changelog](https://github.com/moment/moment/blob/develop/CHANGELOG.md), it can't see anything that can't be reworked to include `2.0.x - 2.2.x`. Though I may be missing something...

Regardless, I see your point. I guess it was just assumed that breaking changes would be relegated to major versions?

```diff
@@ -52,8 +52,6 @@ type moment$CalendarFormats = {
   sameElse?: moment$CalendarFormat,
 };

-type moment$Inclusivity = '()' | '[)' | '()' | '(]' | '[]'
-
 declare class moment$LocaleData {
   months(moment: moment$Moment): string;
   monthsShort(moment: moment$Moment): string;
@@ -98,25 +96,22 @@ declare class moment$MomentDuration {
   get(unit: string): number;
   toJSON(): string;
   toISOString(): string;
-  isValid(): bool;
 }
 declare class moment$Moment {
   static ISO_8601: string;
   static (string?: string, format?: string|Array<string>, locale?: string, strict?: bool): moment$Moment;
-  static (
-      initDate: ?Object|number|Date|Array<number>|moment$Moment|string,
-      validFormats?: ?Array<string>|string,
-      locale?: ?boolean|string,
-      strict?: ?boolean|string
-  ): moment$Moment;
+  static (initDate: ?Object|number|Date|Array<number>|moment$Moment|string): moment$Moment;
   static unix(seconds: number): moment$Moment;
   static utc(): moment$Moment;
   static utc(number: number|Array<number>): moment$Moment;
   static utc(str: string, str2?: string|Array<string>, str3?: string): moment$Moment;
   static utc(moment: moment$Moment): moment$Moment;
   static utc(date: Date): moment$Moment;
+  static parseZone(): moment$Moment;
   static parseZone(rawDate: string): moment$Moment;
-  parseZone(): moment$Moment;
+  static parseZone(rawDate: string, format: string | Array<string>): moment$Moment;
+  static parseZone(rawDate: string, format: string, strict: boolean): moment$Moment;
+  static parseZone(rawDate: string, format: string, locale: string, strict: boolean): moment$Moment;
   isValid(): bool;
   invalidAt(): 0|1|2|3|4|5|6;
   creationData(): moment$MomentCreationData;
@@ -187,7 +182,7 @@ declare class moment$Moment {
   endOf(unit: string): this;
   local(): this;
   utc(): this;
-  utcOffset(offset: number|string, keepLocalTime?: boolean, keepMinutes?: boolean): this;
+  utcOffset(offset: number|string): this;
   utcOffset(): number;
   format(format?: string): string;
   fromNow(removeSuffix?: bool): string;
@@ -204,20 +199,20 @@ declare class moment$Moment {
   toJSON(): string;
   toISOString(): string;
   toObject(): moment$MomentObject;
-  isBetween(from: moment$Moment|string|number|Date|Array<number>, to: moment$Moment|string|number|Date|Array<number>, units?: string, inclusivity?: moment$Inclusivity): bool;
-  isBefore(date?: moment$Moment|string|number|Date|Array<number>, units?: ?string): bool;
-  isSame(date?: moment$Moment|string|number|Date|Array<number>, units?: ?string): bool;
-  isAfter(date?: moment$Moment|string|number|Date|Array<number>, units?: ?string): bool;
-  isSameOrBefore(date?: moment$Moment|string|number|Date|Array<number>, units?: ?string): bool;
-  isSameOrAfter(date?: moment$Moment|string|number|Date|Array<number>, units?: ?string): bool;
+  isBefore(date?: moment$Moment|string|number|Date|Array<number>): bool;
+  isSame(date?: moment$Moment|string|number|Date|Array<number>): bool;
+  isAfter(date?: moment$Moment|string|number|Date|Array<number>): bool;
+  isSameOrBefore(date?: moment$Moment|string|number|Date|Array<number>): bool;
+  isSameOrAfter(date?: moment$Moment|string|number|Date|Array<number>): bool;
+  isBetween(date: moment$Moment|string|number|Date|Array<number>): bool;
   isDST(): bool;
   isDSTShifted(): bool;
   isLeapYear(): bool;
   clone(): moment$Moment;
   static isMoment(obj: any): bool;
   static isDate(obj: any): bool;
+  static locale(locale: string, localeData?: Object): string;
   static updateLocale(locale: string, localeData?: ?Object): void;
-  static locale(locale?: string, localeData?: Object): string;
   static locale(locales: Array<string>): string;
   locale(locale: string, customization?: Object|null): moment$Moment;
   locale(): string;
```

jedwards1211:
> What's the breaking change that stops this from being 2.x.x?

Dude...breaking changes are not the only thing that matters!

In semver, *minor version increments mean new features were added to the API*.  Think about this.
If I add flow defs for API features in 2.3.0 that weren't present in 2.2.0, then someone who's using 2.2.0 could mistakenly use 2.3.0 features without getting any flow errors.
On the other hand, if I only add flow defs for features that are present in 2.2.0, then someone who's using 2.2.0 will get flow errors as desired if they try to use features added in 2.3.0, but someone who's got 2.3.0 and tries to use features that were added in 2.3.0 will get flow errors.

You guys seriously didn't think about this?

jedwards1211:
And just to clarify, in version 2.3.0+ `parseZone` accepts additional arguments, where in version 2.0.0-2.2.x `parseZone` only accepts a single argument.  The single-argument case works the same as before so it's not a breaking change.

GAntoine:
No need to get snarky, I agree with what you're saying. 😛 

I was curious if we *could* merge them all under `2.x.x` in this case specifically, but seeing as

>  in version 2.3.0+ parseZone accepts additional arguments, where in version 2.0.0-2.2.x parseZone only accepts a single argument.

that's sadly not the case. I was hoping we could work around the issue, not simply ignore it.

So yeah, unless someone knows of an alternative, IMO this will have to be addresses.

jedwards1211:
Really, ideally there should be a separate directory for each minor release of a package, because each minor release has some addition to the API, assuming the package is following semver correctly.  But of course in practice people are rarely going to have time for that, and this will always lag behind package releases.  So `flow-typed` should at least pick the greatest minor version less than the minor version of the package installed.

jedwards1211:
Sorry for the strong words, you're right...I was surprised this didn't occur to anyone.

jedwards1211:
@GAntoine wait, maybe it does already work?
https://github.com/flowtype/flow-typed/blob/master/cli/src/lib/npm/npmLibDefs.js#L284

GAntoine:
Huh, will you look at that. I'm not totally sure if that's the right case, so I'll take a look at it tomorrow. Since you have the code locally, could you try creating a fake `package.json` and testing if this change works?

jedwards1211:
@GAntoine okay, thankfully, it does work as desired, so thankfully, whoever implemented the code to pick the version of defs to install did think about this case.

```
$ flow-typed update moment@2.18.1
• Searching for 1 libdefs...
ERROR: Unable to rebase the local cache repo. Error rebasing the `master` branch of the following repo:
/Users/andy/.flow-typed/repo

undefined

NOTE: Unable to rebase local cache! If you don't currently have internet connectivity, no worries -- we'll update the local cache the next time you do.

• Installing 1 libDefs...
  • moment_v2.3.x.js
    └> ./flow-typed/npm/moment_v2.3.x.js
• The following installed libdefs are compatible with your dependencies, but may not include all minor and patch changes for your specific dependency version:

  • libdef: moment_v2.3.x (satisfies moment@2.18.1)

  Consider submitting a versioned update for this package to 
  https://github.com/flowtype/flow-typed/
```

jedwards1211:
If you ask me, **none** of the projects in this repo should be using `x` for the minor version number, because no one gets this warning, hence no one thinks to make sure to update the defs for minor updates.

johnryan:
@jedwards1211 was it intentional to remove the `units` parameter on the `isSame` def? It seems that according to the docs that is still a valid param.

mjomble:
@jedwards1211 which version of the 2.x.x typedefs did you use as a base for this PR?

In addition to the change mentioned by @johnryan, I also just realised that [this change](https://github.com/flowtype/flow-typed/commit/468280fffbbf6888a6c4f7deb931cc877762d4be) is missing from the 2.3.x defs, even though it was committed to 2.x.x back in August.

Could you make sure that all changes done in 2.x.x are also present in 2.3.x?

jedwards1211:
@johnryan that's weird, I had just started from the existing 2.x.x typedefs, and I don't see how they could have been out of date, but I'll figure that out.

jedwards1211:
gahhh, I had forgotten to pull from upstream...that's why my 2.x.x typedefs were out of date.  Fix coming soon.

jedwards1211:
@GAntoine this was an incidental mistake, but maybe we should just scrap the 2.3.x typedefs and merge this change into the 2.x.x typedefs?  Seems sort of pointless unless I went ahead and made typedefs for each minor version.

Keep in mind that with only 2.x.x typedefs anyone who tries to use the `isValid` method @mjomble added with a version of moment that's too old *will not get flow errors*, and have WTF moments when they get a `TypeError`.

But I guess the operating principle here is for users to keep current with the highest minor version, huh...

Marinolinderhof:
@jedwards1211 i'm confused. Is 2.x.x for moment version > 2.9.0 and 2.3.x for momemt < 2.9.0?

for instance:
```
flow-typed moment 2.3.x>   isBetween(date: moment$Moment|string|number|Date|Array<number>): bool;
flow-typed moment 2.x.x>   isBetween(from: moment$Moment|string|number|Date|Array<number>, to: moment$Moment|string|number|Date|Array<number>, units?: string, inclusivity?: moment$Inclusivity): bool;
moment docs 2.9.0+ >  moment().isBetween(moment-like, moment-like, String, String);
```
I get 2.3.x automatically when``` "flow-typed install -s -o",```  
and i guess ```isBetween()``` always needs 2 arguments?

PR #1243, opened 2017-09-13
Reenable Ramda's `sortBy`

johnhaley81:
`sortBy` was disabled due to failing tests after v0.38. This was because 
flow started to correctly infer a type error in the test.

The test was using the `nth` function which expects an Array but the
output of the `sortBy` test was expecting a Tuple. Flow does not
consider Tuples and Array equal and threw a type check error even though
this is completely valid at runtime.

Since Flow favors soundness, this is a valid type error. The fix is
re-enabling `sortBy` and fixing the test to not use `nth`.


johnhaley81:
Looks like tests are failing on node v6 just like https://github.com/flowtype/flow-typed/pull/1222#issuecomment-327401392

johnhaley81:
I rebased and solved the conflicts. All of them were due to me auto-formatting with prettier before. I removed the prettier formatting changes and just stuck with leaving re-enabling `sortBy`. I can run it through another round of prettier in a different PR if you like.

johnhaley81:
Looks like there was maybe a network error with https://travis-ci.org/flowtype/flow-typed/jobs/276938388

PR #1245, opened 2017-09-14
add humanize-number

deecewan:


GAntoine:
@deecewan Are you still interested in getting this merged in? If that's the case, I'd recommend setting a lower limit to the flow version (by replacing `flow_all`). We can look at the failing specs after that.

definitions/npm/humanize-number_v0.0.x/flow_v0.53.x-/humanize-number_v0.0.x.js
```diff
@@ -0,0 +1,11 @@
+type $npm$HumanizeNumber$Options = {
+  delimiter: string,
+  separator: string
```
GAntoine: You'll need to make the presence of these unrequired, to successfully run `humanize(100, { delimiter: " " });`:

```
type $npm$HumanizeNumber$Options = {
  delimiter?: string,
  separator?: string
};
```

Issue #1151, opened 2017-08-17
Error: flow_v0.38.x <=v0.52.x: Unexpected trailing characters:  <=v0.52.x


etrepum:
We started getting this error when running `flow-typed update` today:

```
$ yarn run flow-typed update
yarn run v0.27.5
$ "/app/node_modules/.bin/flow-typed" "update"
• Found 93 dependencies in package.json to install libdefs for. Searching...
• rebasing flow-typed cache...


UNCAUGHT ERROR: Error: flow_v0.38.x <=v0.52.x: Unexpected trailing characters:  <=v0.52.x
    at validationError (/app/node_modules/flow-typed/dist/lib/validationErrors.js:14:11)
    at parseDirString (/app/node_modules/flow-typed/dist/lib/flowVersion.js:155:45)
    at _callee2$ (/app/node_modules/flow-typed/dist/lib/npm/npmLibDefs.js:543:146)
    at tryCatch (/app/node_modules/regenerator-runtime/runtime.js:64:40)
    at Generator.invoke [as _invoke] (/app/node_modules/regenerator-runtime/runtime.js:355:22)
    at Generator.prototype.(anonymous function) [as next] (/app/node_modules/regenerator-runtime/runtime.js:116:21)
    at tryCatch (/app/node_modules/regenerator-runtime/runtime.js:64:40)
    at invoke (/app/node_modules/regenerator-runtime/runtime.js:154:20)
    at /app/node_modules/regenerator-runtime/runtime.js:164:13
    at <anonymous>
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I guess it's related to these:

```
$ git grep '<=v' .
react-redux_v5.x.x.js:// flow-typed version: 19506e57e6/react-redux_v5.x.x/flow_>=v0.30.x <=v0.52.x
react-router-dom_v4.x.x.js:// flow-typed version: 19506e57e6/react-router-dom_v4.x.x/flow_>=v0.38.x <=v0.52.x
react-router_v4.x.x.js:// flow-typed version: 19506e57e6/react-router_v4.x.x/flow_>=v0.30.x <=v0.52.x
```

Currently we have a CI script that tries to sync our repo with flow-typed, which is where this error surfaced first:

```bash
#!/bin/bash -e

FLOW_CACHE=$HOME/.flow-typed
FLOW_REPO=$FLOW_CACHE/repo
[...]

PR #1251, opened 2017-09-15
Parse serialized upper-bounded ranges that contain a space

ahutchings:
Refs #1151

I noticed after committing that prettier had auto-formatted the changed files, resulting in lots of eslint errors. Are the prettier and eslint configs out of sync?


GAntoine:
Closing since the issue was fixed at #1311.

PR #1261, opened 2017-09-19
Add definitions for combine-section-reducers

iofjuupasli:
https://github.com/ryo33/combine-section-reducers


GAntoine:
Hey @iofjuupasli, I can't accept this until the specs are passing. Looks like L42 is succeeding when it should be failing.

PR #1262, opened 2017-09-19
lodash/fp

iofjuupasli:
- [x] reorder arguments
- [x] curry
- [x] add aliases
- [x] add new methods
- [x] add test
- [x] use some improved definitions from ramda
- [x] flow55 $Compose and $ComposeReverse

to produce curried versions use https://gist.github.com/iofjuupasli/fd36c31abf3cedabbee2db8f218fa0b3


TrySound:
@iofjuupasli flow55 has `$Compose` and `$ComposeReverse` types which can be used for flow/flowRight/pipe/compose functions.

TrySound:
Also would be good to have lodash/fn definitions, since lodash bundle is not treeshakable.

iofjuupasli:
@TrySound would be good. But I don't think I can do it for now

iofjuupasli:
@TrySound @marudor @rgbkrk @yenbekbay @robwise @ryyppy @jdalton
Can someone help to review this, and maybe propose some patches

rgbkrk:
Whoa this is super cool, I'll have to check this out later today.

definitions/npm/lodash_v4.x.x/flow_v0.47.x-/lodash_v4.x.x.js
```diff
@@ -1042,3 +1107,878 @@ declare module "lodash" {
...
+    & (<A, R>((...r: [A]) => R) => CurriedFunction1<A, R>)
+    & (<A, B, R>((...r: [A, B]) => R) => CurriedFunction2<A, B, R>)
+    & (<A, B, C, R>((...r: [A, B, C]) => R) => CurriedFunction3<A, B, C, R>)
+    & (<A, B, C, D, R>((...r: [A, B, C, D]) => R) => CurriedFunction4<A, B, C, D, R>)
+    & (<A, B, C, D, E, R>((...r: [A, B, C, D, E]) => R) => CurriedFunction5<A, B, C, D, E, R>)
+    & (<A, B, C, D, E, F, R>((...r: [A, B, C, D, E, F]) => R) => CurriedFunction6<A, B, C, D, E, F, R>)
+
+  declare type UnaryFn<A,R> = (a: A) => R;
+
+  declare type Flow = (<A,B,C,D,E,F,G>(ab: UnaryFn<A,B>, bc: UnaryFn<B,C>, cd: UnaryFn<C,D>, de: UnaryFn<D,E>, ef: UnaryFn<E,F>, fg: UnaryFn<F,G>, ...rest: Array<void>) => UnaryFn<A,G>)
```
robwise: Down below where you define Flow you don't use this type? Is it not ready yet?

robwise: duh I missed this in the checklist

robwise:
This is incredible, so impressive. I skimmed it but I'm not good enough to be able to spot problems just by reading, do you need me to write some tests or something?

iofjuupasli:
It would be awesome to try it on real project with good flow coverage and lodash/fp usage.
`Flow` type should be used here, yes. `$Compose` can be used only on flow v55+. So I decided to finish v47+, and then add copy with compose used

rgbkrk:
Love all the test coverage here for the flow definition.

robwise:
Hmm, all my projects are already up to 55 now :(

iofjuupasli:
@robwise isn't 47+ works with 55?

robwise:
@iofjuupasli I've got it working on my repo but there are no errors, I think maybe because I very heavily rely on `flow`?

iofjuupasli:
@robwise good to know, thank you.
I'll try to finish this PR today

iofjuupasli:
@marudor comments?

definitions/npm/lodash_v4.x.x/flow_v0.55.x-/lodash_v4.x.x.js
```diff
@@ -0,0 +1,1955 @@
...
+      value: T1,
+      defaultValue: T2
+    ): T1,
+    // NaN is a number instead of its own type, otherwise it would behave like null/void
+    defaultTo<T1: number, T2>(value: T1, defaultValue: T2): T1 | T2,
+    defaultTo<T1: void | null, T2>(value: T1, defaultValue: T2): T2,
+    flow: $ComposeReverse,
+    flow(funcs?: Array<Function>): Function,
+    flowRight: $Compose,
+    flowRight(funcs?: Array<Function>): Function,
```
TrySound: Why this still defined?

iofjuupasli: Does $Compose work for both `(...fn)` and `(fn[])`?

TrySound: Ah, okay. Just confused a bit :)

maximderbin:
Wow. That's awesome. Could we get it merged?

robwise:
@iofjuupasli So far I'm finding a bunch of good stuff with this. I just realized that I can try this on prettier-atom since I'm using lodash fp a bunch there as well, and this will allow playing around with it in a public repo. 

So far, I think the argument order of `_.defaultTo` is backward?

robwise:
Another issue:

```js
const buildFoo = _.flow(_.uniqueId, _.set('id', _, {}));
```

^ this is going to error because it thinks that call to `_.set` returns an Object instead of a function and so you are trying to pass an Object to flow which is illegal. I suspect this would come up as a problem for any methods where the `_` placeholder is applicable?

robwise:
Conforms is also duplicated several times?

iofjuupasli:
@robwise

> ^ this is going to error because it thinks that call to _.set returns an Object instead of a function and so you are trying to pass an Object to flow which is illegal. I suspect this would come up as a problem for any methods where the _ placeholder is applicable?

https://github.com/flowtype/flow-typed/pull/1188/files
I think in general it's too hard to implement it on all methods. But it's possible to add placeholder support for some methods that used with placeholder the most.

> So far, I think the argument order of _.defaultTo is backward?

Yes. Is there a problem?

https://github.com/iofjuupasli/flow-typed/blob/e517881f2255fb64d1976965ee63d2e1b28b595d/definitions/npm/lodash_v4.x.x/flow_v0.47.x-v0.54.x/lodash_v4.x.x.js#L1900-L1906

> _.conforms is also duplicated several times?

:+1: need to fix

iofjuupasli:
I've commited `conforms` clean up.

I think I can't work on tricky parts (placeholder, curry) for now. So if anyone can handle this after this PR, it would be awesome. It mostly can be copied from Ramda definition

iofjuupasli:
@marudor CI randomly fails on https://github.com/flowtype/flow-typed/blob/master/cli/src/lib/npm/__tests__/npmLibDefs-test.js#L289-L291

GAntoine:
@iofjuupasli Close and re-open the PR, it should fix the CI failure. I'm working on a fix for that test, but it'll be a while.

robwise:
> So far, I think the argument order of _.defaultTo is backward?

>Yes. Is there a problem?

Oh I think I was looking at the vanilla lodash defs, my bad. I forgot they were in the same file!

marudor:
Wow this is amazing.
I've read through it but couldn't spot anything by reading it.
Considering some of you used it and it looked fine I will merge this.
It's better than nothing for sure.
I'm hoping once it is merged we get some more PRs for this!

Mr-Wallet:
Is it deliberate that there's no `?` on the arguments of the FP functions? For example, in a project I work on, the primary purpose of `fp.getOr` is so that a `?Array` or `?Object` can be accessed in a safe method without a separate guard. Right now the definition is `Object | Array<any>` but what we really need is something more like `Object | Array<any> | void | null`. This would apply to quite a lot of functions, to say the least.

robwise:
@Mr-Wallet Agreed, I think they even fixed this for vanilla lodash but not FP (although I might be wrong).

Rodeoclash:
This is great, however we use a lodash/fp with types generated from Relay Modern which is causing some problems. Array types in Modern are defined as `$ReadOnlyArray<>` which is incompatible the types in this library for collections (just regular `Array<T>`). 

It looks like adding `$ReadOnlyArray<T>` to the union type accepted by collections fixes the issue.

Mr-Wallet:
In general `fp` doesn't mutate input arrays, so rather than union it in, virtually all of the Arrays in the fp definitions ought to be outright replaced by ReadOnlyArrays. `Array<T>` is basically a supertype of `$ReadOnlyArray<T>`, so if the read-only version is allowed then you don't need to also include the mutable version.

iofjuupasli:
This PR is closed
Open new PRs if you want any improvements  😉

PR #1264, opened 2017-09-20
feat/ramda latest assoc and assoc path mixed source object

mattgstevens:
**Note**
I see that some definitions have not been formatted by `prettier` yet and ran that over the latest ramda definitions in the first commit with the command:
`node_modules/.bin/prettier --write npm/ramda_v0.x.x/flow_v0.39.x-/*`

**Description**
This change allows for mixed types in the source object and using `assoc` and `assocPath` to update such a source without flow errors.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.39.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.39.x-/ramda_v0.x.x.js
@@ -1179,40 +1179,34 @@ declare module ramda {
   declare function assoc<T, S>(
     key: string,
     ...args: Array<void>
-  ): ((
-    val: T,
-    ...rest: Array<void>
-  ) => (src: { [k: string]: S }) => { [k: string]: S | T }) &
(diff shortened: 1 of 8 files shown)


mattgstevens:
@mwalkerwells I see your name for the commit that introduced the test that fails with the changes introduced in this commit:

```javascript
//$ExpectError
const a1:{[k:string]:number|boolean} = _.assoc('c', 's', { a: 1, b: 2 })
```

What I would like to have is a Map that has different values in different keys, like this:
```javascript
const mixedA: { x: { [string]: number }, y: Array<string> } = _.assoc(
  "x",
  { x1: 11 },
  { y: ["y1"] }
);
```

Is there any reason to demand that the Map is "homogenous" types and not allow for the example above?

mattgstevens:
@marudor any thoughts on this?

mattgstevens:
seeing other MRs be accepted, guessing triage includes failed tests; removed the `$ExpectError` test that does not fail as this change allows for mixed values in an object.

mattgstevens:
thanks @marudor

@mwalkerwells if you do have some thoughts on this at a later time, I'm open to hear them

PR #1267, opened 2017-09-21
redux-actions: Add support for nested reducers

xMihai:
redux-actions supports nested reducers in `handleActions`, just like `createActions` supports nested actions


Jarlakxen:
mmhh..... looks like for some reason 
```
    // $ExpectError
    'REQUIRERS_REDUCER_VALUE': {},
```
This is valid?? Because the logs says

```
  34:     // $ExpectError
```

PR #1270, opened 2017-09-22
Flow v0.55 definitions for recompose

istarkov:
use $Compose instead of Compose

new tests, and tests fixes


istarkov:
Error in travis is not about this PR

PR #1274, opened 2017-09-22
Make lodash memoize retain type information

aaronjensen:
Looks like prettier did its thing on the test file, but the new tests are at the bottom.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.55.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.55.x-/lodash_v4.x.x.js
@@ -1,52 +1,135 @@
 declare module "lodash" {
-  declare type __CurriedFunction1<A, R, AA: A> =
-    & ((...r: [AA]) => R)
-  declare type CurriedFunction1<A, R> = __CurriedFunction1<A, R, *>
+  declare type __CurriedFunction1<A, R, AA: A> = (...r: [AA]) => R;
+  declare type CurriedFunction1<A, R> = __CurriedFunction1<A, R, *>;
 
-  declare type __CurriedFunction2<A, B, R, AA: A, BB: B> =
-    & ((...r: [AA]) => CurriedFunction1<BB, R>)
-    & ((...r: [AA, BB]) => R)
-  declare type CurriedFunction2<A, B, R> = __CurriedFunction2<A, B, R, *, *>
+  declare type __CurriedFunction2<A, B, R, AA: A, BB: B> = ((
+    ...r: [AA]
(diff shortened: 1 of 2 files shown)


definitions/npm/lodash_v4.x.x/flow_v0.47.x-/test_lodash-v4.x.x.js
```diff
@@ -329,43 +321,58 @@ directStrings = _.map(nums, function(num) {
...
 // $ExpectError string. This type is incompatible with number
-var strings : string[] = _.times(5);
-timesNums = _.times(5, function(i: number) { return i + 1; });
+var strings: string[] = _.times(5);
+timesNums = _.times(5, function(i: number) {
+  return i + 1;
+});
 // $ExpectError string. This type is incompatible with number
-timesNums = _.times(5, function(i: number) { return JSON.stringify(i); });
+timesNums = _.times(5, function(i: number) {
```
GAntoine: Try putting this back on a single line, it might fix the build error.

aaronjensen: Thanks for pointing this out, I missed the failure. Putting it on one line isn't an option due to prettier, but I did move the comment to the right place.

aaronjensen:
looks like this stuff all changed out from under me. I'm trying to get it updated, but for some reason I can't get $ExpectError tests to fail anymore.

aaronjensen:
Ok, looks like this may be good now, let's see if it passes CI

aaronjensen:
Am I doing something wrong that is causing these to get run through prettier?

GAntoine:
Looks like CI passed this time.

> Am I doing something wrong that is causing these to get run through prettier?

Are you saying that you're not running prettier locally, and that the changes are happening after you open the PR?

aaronjensen:
I believe it is running in a precommit hook?

GAntoine:
Hmm, there doesn't seem to be a post-commit hook in the project, so I'm assuming your system is setup to run it. Regardless, I don't see a problem with prettification, as long as CI is happy and the changes are valid.

aaronjensen:
It does via `lint-staged` afaict

aaronjensen:
https://github.com/flowtype/flow-typed/blob/498ec718356ce17cdcbf9ddbf2e8efd522a78fd1/definitions/package.json#L9

aaronjensen:
ugh. Did not mean to close.

GAntoine:
Whoops, I totally missed that. Well I'm no flow pro, but your change seems to match the lodash API, so, approved!

PR #1275, opened 2017-09-22
Express req.query values can be arrays.

serhalp:
The express v4 libdef has express's `req.query` has being an object with string values, but the values can also be arrays of string. The [express docs](http://expressjs.com/en/api.html#app.set) claim that:
> The simple query parser is based on Node’s native query parser, [querystring](https://nodejs.org/api/querystring.html).
> The extended query parser is based on [qs](https://www.npmjs.com/package/qs).

Both modules support query string array syntax:

```
> querystring.parse('foos=one&foos=two')
{ foos: [ 'one', 'two' ] }
> qs.parse('foos=one&foos=two')
{ foos: [ 'one', 'two' ] }
```

It's unclear to me how to write a good test for this. The one I wrote doesn't seem great, and appears to pass before my change, so I must be doing something wrong. Little help?

Thanks.


marudor:
Don't know how to write a better test either :/

PR #1292, opened 2017-09-29
Modify typing for redux v4 to propagate type of Dispatch

billschaller:
This PR modifies the types of MapDispatchToProps, connect, and Provider to allow the type of Dispatch to be passed through. This allows use of thunk actions, promise actions, etc.

I would have done the same treatment for v5, but there's currently a bug in the type defs for redux v5 that makes it fail with recent versions of flow, and I don't have time to figure out how to fix that at the moment.


billschaller:
@calebmer @stryju @cobywhite

GAntoine:
@billschaller Are you interested in bringing this PR up to date, and fixing the specs?

PR #1293, opened 2017-09-30
[rxjs] Add support for combineLatest with a single argument

Li0liQ:
We are missing a typedef for [`Observable.combineLatest(a)`](
http://reactivex.io/rxjs/test-file/spec-js/observables/combineLatest-spec.js.html#lineNumber448).
While the usage seems odd, it is valid and makes sense when you expect new observables to be added later to the list of arguments.


Li0liQ:
Can someone, please, help me understand what is the failing "fails if libdef not published on npm" node v6 unit test is about?

GAntoine:
There's an issue with some specs failing randomly after opening a PR. If the tests are passing locally, closing and re-opening the PR seems to fix it temporarily.

PR #1294, opened 2017-09-30
[lodash] Add single function import typings, update tests

Li0liQ:
[Lodash](https://github.com/lodash/lodash) supports cherry-picking methods.
It is useful if you don't want to bring the whole library for a single method.
This PR adds (exposes existing) type definitions for all `lodash` 4.17.4 single methods and updates tests to test those methods instead.


Li0liQ:
Updated the branch to resolve conflicts. Should be good to merge.

GAntoine:
@iofjuupasli If you have the time, I could use a hand reviewing this!

PR #1295, opened 2017-10-01
Add bufferCount() to Observable.

fmoo:
Docs for bufferCount here:
https://github.com/btroncone/learn-rxjs/blob/master/operators/transformation/buffercount.md

Unlike buffer(), it takes one required and one optional argument.  The
return value is the same.


frantic:
cc @wbinnssmith, @matthewwithanm, @nmote

PR #1297, opened 2017-10-01
Add libdef for raven-js at v3.17.x

mruzekw:
Connected to https://github.com/getsentry/raven-js/issues/717

This basically takes the existing Typescript definitions and translates them into flow definitions. I am trusting that the TS defs are accurate.

Also, it is possible these types work for 3.x.x, but I don't know know enough history of the project to say for sure.

The tests could be more extensive, but I covered the basics.


skovhus:
Thanks! 🚀

Issue #1279, opened 2017-09-25
What version of 'react-i18next' should I use?


MichaelDeBoey:
I thought that the version in this directory is the same as the version of the package itself, but when I was searching for the [`react-i18next`](https://github.com/i18next/react-i18next) libdef, I only found `v2.x.x ` (`flow` version: `>=v0.30.x <=v0.39.x`) & `v1.10.x` (`flow` version: `>=v0.30.x <=v0.52.x`).
And there aren't any libdef's exposed from de library 😕

I use the latest `flow-bin` (`^0.55.0`)

marudor:
With 0.53 there were some breaking Changes.
So I most likely restricted it to <=v0.52.x to ensure it is working.
As the typedefs for react-i18next v1.1 do not work for flow > 0.52.x

MichaelDeBoey:
OK I understand that, but shouldn't libdef for `v5.x` also be in the repository?
Now there's only `v1.10.x` & `v2.x.x`, but the package already is [`v5.4.1`](https://github.com/i18next/react-i18next/blob/master/CHANGELOG.md)... 😕

I suppose that it's not provided and a PR would be welcome?

If so, I'd like to help with that, but I'm not experienced in making one, so any help would be awesome then 🙂

PR #1298, opened 2017-10-02
Add libdef for react-i18next v6.x.x for flow v0.53.x- (#1279)

TLadd:
Copied over the v2.x.x libdef, added the additional exports that have been added to match the latest release of react-i18next, and made modifications to be 0.53.x- compatible.

### DIFF
--- a/definitions/npm/react-i18next_v6.x.x/flow_v0.53.x-/react-i18next_v6.x.x.js
+++ b/definitions/npm/react-i18next_v6.x.x/flow_v0.53.x-/react-i18next_v6.x.x.js
@@ -1,0 +1,91 @@
+declare module "react-i18next" {
+  declare type TFunction = (key?: ?string, data?: ?Object) => string;
+  declare type Locales = string | Array<string>;
+
+  declare type TranslatorProps = {
+    t: TFunction,
+    i18nLoadedAt: Date,
+    i18n: Object
+  };
+
+  declare type Translator<OP, P> = (
+    component: React$ComponentType<P>
+  ) => Class<React$Component<OP, *>>;
+
+  declare type TranslateOptions = $Shape<{
+    wait: boolean,
+    nsMode: "default" | "fallback",
+    bindi18n: false | string,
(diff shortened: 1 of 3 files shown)


definitions/npm/react-i18next_v6.x.x/test_react-i18next_v6.x.x.js
```diff
@@ -0,0 +1,175 @@
...
+class FlowErrorClassComp extends React.Component<Props> {
+  render() {
+    // $ExpectError - wrong argument type
+    const { s, t } = this.props;
+    return <div prop={t({})} />; // misuse of t()
+  }
+}
+
+// $ExpectError - too few arguments
+translate();
```
stunaz: why is that an error, we can have `translate()` which means we are using defaultNS ==> https://react.i18next.com/components/translate-hoc.html

definitions/npm/react-i18next_v6.x.x/flow_v0.53.x-/react-i18next_v6.x.x.js
```diff
@@ -0,0 +1,91 @@
...
+    nsMode: "default" | "fallback",
+    bindi18n: false | string,
+    bindStore: false | string,
+    withRef: boolean,
+    translateFuncName: string,
+    i18n: Object
+  }>;
+
+  declare function translate<OP, P>(
+    locales: Locales,
```
stunaz: locales is an optional argument too ==> https://react.i18next.com/components/translate-hoc.html

stunaz:
@TLadd I found an issue with the flow-typed , `translate` can have no parameters which means it is using defaultNS , have a look at https://react.i18next.com/components/translate-hoc.html or better at the source code https://github.com/i18next/react-i18next/blob/master/src/translate.js#L20

TLadd:
You are correct. Would you like to submit a PR correcting the issue?

stunaz:
will do :-) will ping you to review

PR #1300, opened 2017-10-03
update react-typekit typings for flow 0.53

chrisdrackett:


GAntoine:
There's a typo in the folder name, it should be `flow_v0.53.x-`, with the `-` at the end. You'll also need have to move the current `test_react-typekit_v1.x.x.js` file into `flow_v0.26.x-v0.52.x`.

PR #1301, opened 2017-10-03
Added missing constuctor

jankalfus:
Added missing constructor that is available since moment.js version 2.3


masylum:
This would actually fix: https://github.com/flowtype/flow-typed/issues/1299

Issue #1234, opened 2017-09-08
Crash with backslash path separator in cli/run-tests


HyperBrain:
After checkout, when executing `./cli/dist/cli.js run-tests` the CLI crashes in case `path.sep` is a backslash (`\`). For me that's the case in my Git bash.

```
 cli/dist/cli.js run-tests
C:\Projects\flow\flow-typed\cli\dist\commands\runTests.js:70
var basePathRegex = new RegExp('definitions' + _node.path.sep + 'npm' + _node.path.sep + '(@[^' + _node.path.sep + ']*' + _node.path.sep + ')?[^' + _node.path.sep + ']*' + _node.path.sep + '?');
```

The reason is, that the use of path.sep in the RegExp is not correct in the backslash case. In that case the separator has to be doubled, as backslash as a regexp character has to be escaped properly.

Instead of using `path.sep` directly, the possibly escaped path.sep should be build before and then inserted into the cli.

yizhoucao:
Same issue here on my windows laptop

pablowatson:
Same issue on windows. Cannot run flow-type at all.

PR #1315, opened 2017-10-06
Closes #1313

### LINKED ISSUE #1313: flow-typed v2.2.0 fails on windows
```
npx flow-typed
...\node_modules\flow-typed\dist\cli.js
Invalid regular expression: /definitions\npm\(@[^\]*\)?[^\]*\?/: Unterminated character class
```

It doesn't matter what commands I run and what folder I run them in - end result is the same.

* node v8.1.2
* Windows 10 x64

jkomyno:
From NodeJS' docs:

> path.sep provides the platform-specific path segment separator:
> - \ on Windows
> - / on POSIX

> Note: On Windows, both the forward slash (/) and backward slash (\) are accepted as path segment separators; however, the path methods only add backward slashes (\).

Backward slashes resulted in breaking flow-typed for Windows. This PR fixes it.

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -47,7 +47,7 @@ type TestGroup = {
  * directory.
  */
 const basePathRegex = new RegExp(
-  `definitions${path.sep}npm${path.sep}(\@[^${path.sep}]*${path.sep})?[^${path.sep}]*${path.sep}?`,
+  'definitions/npm/(\@[^/]*/)?[^/]*/?',
 );
 async function getTestGroups(
   repoDirPath,
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -46,9 +46,7 @@ type TestGroup = {
  * structs. Each TestGroup represents a Package/PackageVersion/FlowVersion
  * directory.
  */
-const basePathRegex = new RegExp(
-  'definitions/npm/(\@[^/]*/)?[^/]*/?',
-);
+const basePathRegex = new RegExp('definitions/npm/(\@[^/]*/)?[^/]*/?');
 async function getTestGroups(
   repoDirPath,
   onlyChanged: boolean = false,
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -46,7 +46,7 @@ type TestGroup = {
  * structs. Each TestGroup represents a Package/PackageVersion/FlowVersion
  * directory.
  */
-const basePathRegex = new RegExp('definitions/npm/(\@[^/]*/)?[^/]*/?');
+const basePathRegex = new RegExp('definitions/npm/(@[^/]*/)?[^/]*/?');
 async function getTestGroups(
   repoDirPath,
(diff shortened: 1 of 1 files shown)


GAntoine:
Thanks for the PR! Do you think this would also fix #1234 ?

cli/src/commands/runTests.js
```diff
@@ -47,7 +47,7 @@ type TestGroup = {
  * directory.
  */
 const basePathRegex = new RegExp(
-  `definitions${path.sep}npm${path.sep}(\@[^${path.sep}]*${path.sep})?[^${path.sep}]*${path.sep}?`,
+  'definitions/npm/(\@[^/]*/)?[^/]*/?',
```
GAntoine: Place everything on one line to make the specs pass:
```
const basePathRegex = new RegExp(REGEX);
```

jkomyno: Yeah I just noted, it was a PR on the fly so I didn't clone the project and tested the lint by myself.
Edit incoming

jkomyno:
@GAntoine it should be fine now

GAntoine:
@kulmajaba Since you were gonna open a PR, would you mind taking a look at this? I'm happy with it, but let me know if you see any glaring issues.

kulmajaba:
Looks good to me, I've nothing to add.

GAntoine:
Then it's accepted, thanks again for your help.

cli/src/commands/runTests.js
```diff
@@ -46,9 +46,7 @@ type TestGroup = {
  * structs. Each TestGroup represents a Package/PackageVersion/FlowVersion
  * directory.
  */
-const basePathRegex = new RegExp(
-  `definitions${path.sep}npm${path.sep}(\@[^${path.sep}]*${path.sep})?[^${path.sep}]*${path.sep}?`,
-);
+const basePathRegex = new RegExp('definitions/npm/(@[^/]*/)?[^/]*/?');
```
HyperBrain: @GAntoine Does this still work with `\` as path separator? I think the checks for `^/` will miss a backslash on Windows as well as the terminating optional `/`.
The correct fix should be to use `path.sep`, but to double it in case it is a backslash (correct Regex escape).

E.g.
```
...\flow-typed\definitions\npm\mylib\
```
would not match

GAntoine: It no longer works with a backslash, but according to

> Note: On Windows, both the forward slash (/) and backward slash (\\) are accepted as path segment separators;

that shouldn't be an issue.

HyperBrain: Yes, Windows works with forward slashes for all fs operations, but my point was, that the regex is used to _parse_ incoming paths, which still have backslashes embedded on Windows, so the forward slash only regex is unsuitable to parse Windows paths.

GAntoine: Oh I see, I wasn't aware of that... Would you mind opening a new issue? I'll ping the right people and we can discuss it there. I don't work on windows so I can't really verify either change, more deeply then how sound they look.

PR #1316, opened 2017-10-06
Updated Reselect type definitions flow v0.54-

thehogfather:
This PR contributes the following:
* type definitions that is valid for flow `v0.54` and above ( consequently renaming old type defs to be valid for `v0.28.x - v0.53.x`
* additional test for nested structured selectors

cc @julienw


definitions/package-lock.json
```diff
@@ -1112,6 +1112,14 @@
           "resolved": "https://registry.npmjs.org/sprintf-js/-/sprintf-js-1.0.3.tgz",
```
GAntoine: Undo these changes, we shouldn't need to modify `package-lock` anymore.

definitions/npm/reselect_v3.x.x/flow_v0.54.x-/reselect_v3.x.x.js
```diff
@@ -0,0 +1,1327 @@
...
+    resultFunc: C,
+    recomputations: () => number,
+    resetRecomputations: () => number
+  };
+
+  declare type ParametricSelector<S, P, R> = (
+    state: S,
+    props: P,
+    ...args: any[]
+  ) => R;
```
julienw: I see it's derived from the typescript version, but I don't get why we need to have both Selectors and ParametricSelectors. Wouldn't it be enough to have only the ParametricSelectors ? (and then rename them Selectors :) ). This is what the old libdef does.

julienw: Also I think the "ParametricSelector" part just doesn't work :) I added a test to prove it:

```javascript
// TEST: props as second argument
const selector = createSelector(
  (state: State, props) => "foo",
  (state) => "bar",
  (result1, result2) => result1 + result2
);
// END TEST
```

You'll get this error in flow v0.54 => 0.56:

```
   Error: 0-test_reselect_3.x.x.js:107
                         v--------------
   107: const selector = createSelector(
   108:   (state: State, props) => "foo",
   109:   (state) => "bar",
   110:   (result1, result2) => result1 + result2
   111: );
        ^ function call. Could not decide which case to select
   107: const selector = createSelector(
                         ^^^^^^^^^^^^^^ callable object type
     Case 3 may work:
              v--------------
      41:     <S, R1, R2, T>(
      42:       selector1: Selector<S, R1>,
      43:       selector2: Selector<S, R2>,
      44:       combiner: (res1: R1, res2: R2) => T
      45:     ): OutputSelector<S, T, (res1: R1, res2: R2) => T>,
              -------------------------------------------------^ polymorphic type: function type. See lib: reselect_v3.x.x.js:41
     But if it doesn't, case 4 looks promising too:
              v-----------------
      47:     <S, P, R1, R2, T>(
      48:       selector1: ParametricSelector<S, P, R1>,
      49:       selector2: ParametricSelector<S, P, R2>,
      50:       combiner: (res1: R1, res2: R2) => T
      51:     ): OutputParametricSelector<S, P, T, (res1: R1, res2: R2) => T>,
              --------------------------------------------------------------^ polymorphic type: function type. See lib: reselect_v3.x.x.js:47
     Please provide additional annotation(s) to determine whether case 3 works (or consider merging it with case 4):
     108:   (state: State, props) => "foo",
                           ^^^^^ parameter `props`
```

definitions/npm/reselect_v3.x.x/test_reselect_3.x.x.js
```diff
@@ -4,7 +4,7 @@ import {
   defaultMemoize,
```
julienw: please add the test for our bug :) For example:
```javascript
// TEST: Should work when using another selector as functions
const simpleSelector = (state: State): string => "foo";
const resultSelector = (arg: string): string => "foo";

const combinedSelector1 = createSelector(simpleSelector, resultSelector);

const combinedSelector2: (state: State) => string = createSelector(
  combinedSelector1,
  resultSelector
);
// END TEST
```
This test doesn't pass with Flow 0.28 => 0.33 so you'll need to change the lower bound of the old libdef to 0.34.

thehogfather:
@julienw thanks for the comments. Although I've addressed them, i have also unintentionally pushed a `git rebase` of my local working copy :S so this PR is now pretty messy.

I'll reopen a fresh one with a reference to this.

PR #1317, opened 2017-10-07
Handle default props in react-redux

xdissent:
This makes `connect()` work with component classes that have default props. It uses a technique found in https://github.com/facebook/flow/issues/4644 to handle those classes specially. The connected component class has a `static defaultProps` which doesn't actually exist at runtime, but I couldn't get it to work any other way. I couldn't get it to work at all in 0.53 so that version is unchanged. Also fixes the existing tests for flow >0.55 by adding return types to the `mapStateToProps` fns.


GAntoine:
Nice work 💪 

@kalley You seem to have worked on the react-redux libdefs, if you have a little time could you look this over quickly and give us your thoughts? Cheers!

kalley:
This looks good to me. :tada: Very similar to what I've been working on [for `styled-components`](https://gist.github.com/kalley/22859a23a70e5800d47e3cdd43ad4fbc) as well

definitions/npm/react-redux_v5.x.x/flow_v0.54.x-/react-redux_v5.x.x.js
```diff
@@ -0,0 +1,132 @@
+// flow-typed signature: 8db7b853f57c51094bf0ab8b2650fd9c
+// flow-typed version: ab8db5f14d/react-redux_v5.x.x/flow_>=v0.30.x
```
GAntoine: You'll need to remove the signatures.

GAntoine:
Awesome, thanks a ton!

villesau:
@xdissent @kalley seems that these types are currently pretty badly broken. There was a bug which didn't run the libdef tests properly so basically none of the $ExpectError tests didn't test anything: https://github.com/flowtype/flow-typed/issues/1385

xdissent:
Yeah, I think the broken ones are all defaultProps-related too, so, my bad! =/
But after a second look, I'm pretty sure most of the other connector props are incorrect too. I think we want `$Subtype` rather than `$Supertype`, or at least we want the wrapped component's props to be a subtype of state props & dispatch props (or merge props). Right now the intersection `$Supertype`s are kinda forcing the rest of the wrapped component's props into `OP`, which is a clever way to avoid diffing them out manually. But because `({a: 'x'}: $Supertype<{a: number}>)`, it basically lets you `connect()` any component, regardless of prop types. I have a bunch of new (failing) test cases I'll push shortly.

I put together a version that I think gets all those things right, but for some reason it's broken for functional components. I hope I'm missing something super obvious, but can anyone please tell me why the following checks?

```
((props: {a: string}) => 1: ($Subtype<{a: number}>) => any)
```

It becomes a big issue when you try to wrap a function:

```
// P must be a subtype of {a: number}
function wrap<P: {a: number}>(fn: P => any): void {}
wrap((props: {a: string}) => 1) // <-- should error, right?
```

[More examples here](https://flow.org/try/#0PTAEAEDMBsHsHcBQiCWBbADrATgF1AFSgCGAzqAEoCmxAxvpNrGqAOTY32vK4CeGVUAAUmGcgF5QAb2IAuUADsArmgBGVbAF9EfAaABCxACYjYY0JJnzSubCgUBzbYhCgA8pFC1YS7KUG4ABYo5JDEKNCkAHQxLmAA3AAUVmwAHqya8gAkAMpKqrpUADw5AHwAlKCuRQC0NaAaTNjIrjFRqkr48IG8oEawVKQKrPhBIQD8iN4KNqCqxgBiChagiRiipPKGJhuV4qWgAIyISfNGS-KJufmFRaZiFRYHxAq85S1gQqCkgT7QRnNBKofAoAbhYN8bvxBLBPPdorEjFRaNBiBxQJAlAp6ChYMt4NhiBglnd5PDSok1hsyXtnq9yvIAG6wFBGOKgNodUbBcg2CLQUDwHAAa1IkwJRKWiTOS3e7IAEmRFLAxo5QOC+hD4Cggn1kaj0YzyBxiALIDNZIhMdjcLjliboAB1QnEhSk4QbCnmy7rMybYS0kj0pksgFSbQO52ShTSxYKOUosjkbYAYWYGAaqVwVFB5GodFwUTTmDxOdwRW25OkiFAoA4oI0q0qUhrtbrVFwvmWRSMKEZVVKre0zlcAEl8P4qGhyBqhdhhV4cBx6NBepAcIuSwoy15UaR-AioogkYn0dacXjBS7ixh3eTEvn6EX06WFLgACrQu4VEOs9kSjAb0SVN00qao6gabAmnZTlOmVBgsQvBRTU3LBtzfMVEAAoCZXjRAgA)

Am I losing my mind or is that a flow bug?

villesau:
@xdissent to me it looks lot like a bug. Though, there is no documentation for $Subtype $Supertype.

Seems that $Subtype doesn't return errors in any case, when $Supertype has only one situation when it doesn't error:
[example](https://flow.org/try/#0PTAEAEDMBsHsHcBQiCWBbADrATgF1AFSgCGAzqAEoCmxAxvpNrGqAOTY32uIgAkAygFcMVPAE8RiABRSMTDKQBcoAN7FlpXNhQA7AOYBfAJSgAvAD5QARmVSBw0bglUAPGo1bdegDSgARso6gmh+ogbmJhbWRgDcPMCiTNi2svJKquqgQSFhkZY2oHZCIuIibpma2vq+AVnBodjhedFx8Yk4KXKwCsrudTnYNR5Vhs0FRQ6lrn2VXkP9DU1m+bFt2Emdab2Z2Q3zu7nL1rb2JU5lM57V-oH1YRFHVqsyXT0Zw3M3C4dR46eOznKH2utQOjQev1WIHayUKqW66T6YP2d0aYxOxQBFwqVx8XzBS0hrWh6w6cNeiJ2qJRA2Mjwxk3O0xxI0JKziJI25K272+g3xqLpvwZZ0BSMFEPZyD4Qj8TMU0nhb0urPRhXscrFLM+oIlY2eSspt1paommuxwLxutpkpayBePPFA3ms30QvyDPNzMtNMWtqecQdCO2xr2ApN9PVsqZQNArqtofu+taQeVVOd4cWpo1MZVXjZdtTRr5vp+Hqjgi9sYJ-tWiCAA)

maxsalven:
Hi all, react-redux continues to be broken, despite being one of the most used 3rd party libraries for React. Is there any clarity on whether this is a Flow bug or a libdef bug?

PR #1322, opened 2017-10-08
[react-intl] Remove requirement for intl prop

Li0liQ:
When `injectIntl` HOC is used it injects `intl` property into component.  So there is no reason for resulting component to require consumers to provide it.


GAntoine:
@Li0liQ You'll need to fix the specs before I merge this in.

PR #1327, opened 2017-10-09
fix translate in react-i18next_v6.x.x

stunaz:
`translate` can have no parameters. 
As discussed in https://github.com/flowtype/flow-typed/pull/1298#issuecomment-335154361

@TLadd please review


TLadd:
This looks good to me. It was a mistake in the original libdef; locales should be optional as it is here.

Note that I don't have merge/approval power, so someone else will need to look at this too.

Issue #409, opened 2016-10-23
[libdef][express4] Missing router.param() method definition


kombucha:
Hi,

The Router class definition is missing the `.param(name, callback)` method, resulting in a check error.

<img width="449" alt="screen shot 2016-10-23 at 16 55 58" src="https://cloud.githubusercontent.com/assets/1584370/19627191/c09d5192-9941-11e6-8510-81c123229a11.png">

See [express documentation](http://expressjs.com/en/4x/api.html#router.param)

PR #1336, opened 2017-10-10
Adds express$Router.param

kennethaasan:
I saw that express missed express$Router.param (http://expressjs.com/en/api.html#app.param). I'm not sure if I typed it 100% correct, but it works in my project.

Best Regards,
Kenneth


GAntoine:
Related to #409

GAntoine:
Judging from the failing specs, it looks like your def isn't compatible with flow 32 or 33..

kennethaasan:
Hi @GAntoine,

Tests are passing now. I've never worked with such a old version of flow, so I didn't know this syntax :)

Anything else I should add/fix?

GAntoine:
Nope, this looks good to me, I'll go ahead and merge it. Thanks for the fix 💪

PR #1343, opened 2017-10-12
Escapes backslashes when on windows for path regex

### LINKED ISSUE #1335: Regex change in #1315 leads to failure of matches for Windows paths
In #1315 , a path matching regex was changed to prevent crashes on Windows where an invalid regex would be created (due to missing escapes for backslashes).

The changed code was in `runTests.js`:
Original regex:
`definitions${path.sep}npm${path.sep}(\@[^${path.sep}]*${path.sep})?[^${path.sep}]*${path.sep}?`

The change just replaced all `path.sep` with `/`.

The problem now is, that the regex will not match any paths returned by Windows anymore, because they natively contain backslashes.

In my opinion there are 2 feasible approaches to fix the "fix" properly:

(1) Use the original regex, but escape `path.sep` properly in case it is a backslash:
```
const escapedPathSep = path.sep === '\\' ? '\\\\' : path.sep;
... `definitions${escapedPathSep}npm${escapedPathSep}(\@[^${escapedPathSep}]*${escapedPathSep})?[^${escapedPathSep}]*${escapedPathSep}?`
```

(2) Normalize the path so that it only contains slashes:
```
`definitions${path.sep}npm${path.sep}(\@[^${path.sep}]*${path.sep})?[^${path.sep}]*${path.sep}?`.replace('\\', '')
```

(1) would be the most compatible solution as (2) might break if paths explicitly contain backslashes on Unix filesystems.

GAntoine:
Fixes #1335 

/cc @HyperBrain @ebrearley

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -46,7 +46,10 @@ type TestGroup = {
  * structs. Each TestGroup represents a Package/PackageVersion/FlowVersion
  * directory.
  */
-const basePathRegex = new RegExp('definitions/npm/(@[^/]*/)?[^/]*/?');
+const escapedPathSep = path.sep === '\\' ? '\\\\' : path.sep;
+const basePathRegex = new RegExp(
+  `definitions${escapedPathSep}npm${escapedPathSep}(\@[^${escapedPathSep}]*${escapedPathSep})?[^${escapedPathSep}]*${escapedPathSep}?`,
+);
 async function getTestGroups(
   repoDirPath,
   onlyChanged: boolean = false,
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -46,10 +46,7 @@ type TestGroup = {
  * structs. Each TestGroup represents a Package/PackageVersion/FlowVersion
  * directory.
  */
-const escapedPathSep = path.sep === '\\' ? '\\\\' : path.sep;
-const basePathRegex = new RegExp(
-  `definitions${escapedPathSep}npm${escapedPathSep}(\@[^${escapedPathSep}]*${escapedPathSep})?[^${escapedPathSep}]*${escapedPathSep}?`,
-);
+const basePathRegex = new RegExp('definitions/npm/(@[^/]*/)?[^/]*/?');
 async function getTestGroups(
   repoDirPath,
   onlyChanged: boolean = false,
@@ -71,9 +68,7 @@ async function getTestGroups(
     libDefs = libDefs.filter(def => changedDefs.includes(def.pkgName));
   }
   return libDefs.map(libDef => {
-    const groupID =
-      `${libDef.pkgName}_${libDef.pkgVersionStr}${path.sep}` +
-      `${libDef.flowVersionStr}`;
+    const groupID = `${libDef.pkgName}_${libDef.pkgVersionStr}/${libDef.flowVersionStr}`;
     return {
       id: groupID,
       testFilePaths: libDef.testFilePaths,
@@ -283,8 +278,7 @@ async function runTestGroup(
 ): Promise<Array<string>> {
   // Some older versions of Flow choke on ">"/"<"/"="
   const testDirName = testGroup.id
-    .split(path.sep)
-    .join('--')
+    .replace('/', '--')
     .replace(/>/g, 'gt')
     .replace(/</g, 'lt')
     .replace(/=/g, 'eq');


GAntoine:
It would be `cli run-tests --onlyChanged`. You'd need to change a definition (just tweak it so that the hash is different then the one on github). It should then only run tests for that definition.

GAntoine:
Thanks for investigating, I'll wait on your results from your end.

So I've dug into this quite a bit, and from what I can tell the current code on master should work just fine on Windows. CAVEAT: If you're passing custom `testPatterns` from the command line, you'll need to pass forward slashes, or replace any backslashes with forward ones.

Could you pull my latest changes to this PR, made a change to the libdef for `array-uniq_v1.x.x`, and run these two lines:

`yarn run build && node dist/cli.js run-tests array-uniq_v1.x.x/flow_v0.25.x`
`yarn run build && node dist/cli.js run-tests array-uniq_v1.x.x\flow_v0.25.x`

HyperBrain:
Will do in a minute

HyperBrain:
Here's the result:
```
PS C:\Projects\experiment\flow-typed-gantoine> node cli\dist\cli.js run-tests array-uniq_v1.x.x\flow_v0.25.x
Running definition tests in C:\Projects\experiment\flow-typed-gantoine\definitions\npm...


All tests passed!
PS C:\Projects\experiment\flow-typed-gantoine> node cli\dist\cli.js run-tests array-uniq_v1.x.x/flow_v0.25.x
Running definition tests in C:\Projects\experiment\flow-typed-gantoine\definitions\npm...

Fetching all Flow binaries...
Finished fetching Flow binaries.

Testing array-uniq_v1.x.x/flow_v0.25.x- (flow-v0.30.0)...
Testing array-uniq_v1.x.x/flow_v0.25.x- (flow-v0.31.0)...
Testing array-uniq_v1.x.x/flow_v0.25.x- (flow-v0.32.0)...
Testing array-uniq_v1.x.x/flow_v0.25.x- (flow-v0.33.0)...
Testing array-uniq_v1.x.x/flow_v0.25.x- (flow-v0.34.0)...
```

I tested it in PowerShell only, because there I'm free to choose the path separator.
The backslash version did not work, but the forward slash version did.

I'm not sure if it's important: Letting the second variant run til the end leads to a lot of flow errors. But that could be a pure setup issue on my side (I think it is unrelated to the whole path handling).

GAntoine:
Alright, I was hoping to see that as the output. I'm getting the same thing on my end, minus the flow errors. What's your feeling on forcing windows users to use forward slashes when passing `testPatterns`? Obviously it's not ideal, but an ideal fix would be a lot more complex, and might just break more stuff unknowingly...

Out of curiosity, can you post the error messages you were getting? I'm assuming the flow errors were caused by the change made to the libdef, but I'd like to double check.

HyperBrain:
> What's your feeling on forcing windows users to use forward slashes when passing testPatterns? 

Don't ask me - I usually use Git Bash or the Ubuntu shell because PowerShell sucks 😄 I'm completely fine to have only forward slashes in `testPatterns`.

The output:
```
ERROR: array-uniq_v1.x.x/flow_v0.25.x-
 * array-uniq_v1.x.x/flow_v0.25.x- (flow-v0.30.0): Unexpected Flow errors(2):
   0-test_array-uniq_v1.x.x.js:1
     1: import arrayUniq from 'array-uniq';
                              ^^^^^^^^^^^^ This modules resolves to "<<PROJECT_R
OOT>>\\../../node_modules/array-uniq/package.json", which is outside both your r
oot directory and all of the entries in the [include] section of your .flowconfi
g. You should either add this directory to the [include] section of your .flowco
nfig, move your .flowconfig file higher in the project directory tree, or move t
his package under your Flow root directory.

   0-test_array-uniq_v1.x.x.js:1
     1: import arrayUniq from 'array-uniq';
                              ^^^^^^^^^^^^ array-uniq. Required module not found


   0-test_array-uniq_v1.x.x.js:9
     9: // $ExpectError


   Found 3 errors

   null
```

These repeat for each flow version.  The `null` output at the end looks weird.

GAntoine:
> I'm completely fine to have only forward slashes in testPatterns.

Then I'm happy merging this change and addressing issues from Windows users as they appear. 
As for the error, I'm gonna setup *everything* on my Windows system tonight, to verify the fix on my end as well.

Thanks again for the bug report and for your help debugging this fix. 🔥

GAntoine:
Tested on windows in powershell, and I'm also getting the errors. However I get the errors even on master, so unless master is also broken somewhere else, it should be fine.

PR #1344, opened 2017-10-13
Fix compatibility with redux-form v7

jpolo:


definitions/npm/redux_v3.x.x/flow_v0.33.x-/redux_v3.x.x.js
```diff
@@ -8,82 +8,82 @@ declare module 'redux' {
...
 
-  declare type StoreEnhancer<S, A, D = Dispatch<A>> = (next: StoreCreator<S, A, D>) => StoreCreator<S, A, D>;
+  declare export type StoreEnhancer<S, A, D = Dispatch<A>> = (next: StoreCreator<S, A, D>) => StoreCreator<S, A, D>;
 
-  declare function createStore<S, A, D>(reducer: Reducer<S, A>, enhancer?: StoreEnhancer<S, A, D>): Store<S, A, D>;
-  declare function createStore<S, A, D>(reducer: Reducer<S, A>, preloadedState: S, enhancer?: StoreEnhancer<S, A, D>): Store<S, A, D>;
+  declare export function createStore<S, A, D>(reducer: Reducer<S, A>, enhancer?: StoreEnhancer<S, A, D>): Store<S, A, D>;
+  declare export function createStore<S, A, D>(reducer: Reducer<S, A>, preloadedState: S, enhancer?: StoreEnhancer<S, A, D>): Store<S, A, D>;
 
   declare function applyMiddleware<S, A, D>(...middlewares: Array<Middleware<S, A, D>>): StoreEnhancer<S, A, D>;
```
GAntoine: Did you also mean to export `applyMiddleware`?

jpolo: That's very true ! it should be for every member

jpolo:
When using redux-form flow complains about not finding Dispatch from redux.
The only diff between flow doc and the redux def was the export keyword.
I tried my change locally and it stops complaining.

The redux-form project host its dot flow defs that's why it is not easy to
test ( it might be a flow-bin related issue) Nevertheless my PR seems a
straightforward workaround.

PR #1346, opened 2017-10-13
query-string's extract() method will never return null

ZhangYiJiang:
The function looks like this 

```js
exports.extract = function (str) {
	return str.split('?')[1] || '';
};
```

https://github.com/sindresorhus/query-string/blob/master/index.js#L116

It's plainly obvious that this will never return `null` or `undefined`.


GAntoine:
Thanks for the fix 👋

PR #1354, opened 2017-10-14
Added sinon v3

doberkofler:


GAntoine:
Any reason you're only setting these for flow version 28 instead of 28+? Eg: `flow_v0.28.x` VS `flow_v0.28.x-`.

doberkofler:
@GAntoine my mistake: sure it should be 0.28 or later!

PR #1355, opened 2017-10-14
Added ms

doberkofler:


doberkofler:
@GAntoine my mistake: sure it should be 0.25 or later!

PR #1357, opened 2017-10-14
Change shelljs to allow all flow versions > 0.28

doberkofler:


doberkofler:
@GAntoine my mistake: sure it should be 0.28 or later!

PR #1361, opened 2017-10-15
Add Flow-Typed tests for Unified.js library.

SirYummy:
Basic core tests for Unified.js.  See also: [https://github.com/unifiedjs/unified](https://github.com/unifiedjs/unified).


definitions/npm/unified_v6.1.x/flow_v0.2.x/test_unified.js
```diff
@@ -0,0 +1,10 @@
+/* @flow */
```
GAntoine: File name should be `test_unified_v6.1.x.js`

definitions/npm/unified_v6.1.x/flow_v0.2.x/unified_v6.1.x.js
```diff
@@ -0,0 +1,20 @@
+type Unified$processor = {
```
GAntoine: Might want to consider using a higher version of flow, and adding a trailing version dash. `.../flow_v0.38.x-/...`

Issue #1099, opened 2017-08-02
A lot of functions don`t use $ReadOnlyArray type in the Lodash type definitions


mtnt:
A `$ReadOnlyArray` type is a subtype of `Array`. Many declarations, where incoming array will not be modified, has just an `Array` as type. It should be `$ReadOnlyArray` for consistent use.

GAntoine:
Would you be up to open a PR that fixes this?

Schedulaar:
Also a lot of functions should use `$ReadOnly<{ [id: any]: T }>` instead of just `{ [id: any]: T }` as input parameter. For example `omitBy` should have the type
```js
    omitBy<A, T: $ReadOnly<{ [id: any]: A }>>(
      predicate: OPredicate<A>
    ): (object: T) => Object;
    omitBy<A, T: $ReadOnly<{ [id: any]: A }>>(predicate: OPredicate<A>, object: T): Object;
```
instead of
```js
    omitBy<A, T: { [id: any]: A }>(
      predicate: OPredicate<A>
    ): (object: T) => Object;
    omitBy<A, T: { [id: any]: A }>(predicate: OPredicate<A>, object: T): Object;
```

PR #1369, opened 2017-10-17
lodash: Use $ReadOnlyArray as an input to `find*` functions

irnc:
This change solves part of an issue #1099. It doesn't touch functions other than find. Other immutable functions should be changed separately.

Change was straightforward, I just replaced `Array`  with `$ReadOnlyArray` on `find*` functions and added a single test case, but then `prettier` decided to convert every single quote to a double quote. Please let me know if I should submit these changes differently.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.55.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.55.x-/lodash_v4.x.x.js
@@ -226,12 +226,12 @@ declare module "lodash" {
       end?: number
     ): Array<T | U>;
     findIndex<T>(
-      array: ?Array<T>,
+      array: ?$ReadOnlyArray<T>,
       predicate?: Predicate<T>,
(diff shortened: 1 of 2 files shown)


GAntoine:
@Mtnt Would you mind reviewing this?

Mtnt:
@GAntoine I can, but I`m not sure it will be soon - maybe week, maybe two

GAntoine:
@irnc After taking another look at this I think it's a fine change. If you can rebase it (and fix the conflicts) I can get it merged in.

irnc:
@GAntoine I tried to rebase my branch on master, conflicts was due to different class member separators. It was `,` before #1446 introduced `;`. I changed my code to use `;` but then prettier converted everything back to `,`.

I felt that something is wrongly configured on my side, so I reinstalled all dependencies with no luck, prettier still changes everything to `,`. If this style is desired, I would be glad to submit prettier changes in a separate pull request.

Issue #1371, opened 2017-10-17
Unable to run any test because the flow 0.57.2 binary has a build error on Windows


doberkofler:
Since flow 0.57.2 was release it is practically impossible to run any tests on the definition files because they all will fail due to an error in the flow 0.57.2 binary.

```
 * shelljs_v0.7.x/flow_v0.28.x- (flow-v0.57.2): Unexpected Flow errors(2):
   Library parse error:
   <BUILTINS>\INDEX:1
     1: 101 file bom.js
            ^^^^ Unexpected identifier

   Found 1 error
```

PR #1374, opened 2017-10-17
Fix for issue #1371

doberkofler:
Skip testing flow versions 0.57 before 0.57.3 on the Windows platform


GAntoine:
I'll take your word that it's happening on Windows (I can't test it ATM). Your fix looks right, and I don't expect it to break anything, so I'll get it merged in.

PR #1378, opened 2017-10-17
Improve typing for Ramda #pick.

AndrewSouthpaw:
There were some issues with the current definition for Ramda's `#pick`:

- Would not correctly detect invalid keys, so the following would not fail

```js
const pck2: Object = _.pick([ 'a', 'b' ], { a: 1 })
```

- Did not allow passing in typed object literals (a very common use case), so something like this would fail when the keys had different types:

```js
const pickedObj: { a: string, b: number, c: boolean } = { a: 'hello', b: 1, c: true }
const pck3: { a: string, b: number } = pick(['a', 'b'], pickedObj)
```

This PR addresses both issues.


GAntoine:
Sweet fix! Could you rebase it when you get the chance?

AndrewSouthpaw:
Done!

GAntoine:
Beautiful, approved.

PR #1384, opened 2017-10-18
Added cuid module

JWebCoder:


KenanY:
What about `cuid.slug()` and `cuid.fingerprint()`?

PR #1393, opened 2017-10-23
Glamorous types (WIP)

aaronjensen:
This is a work in progress, but I wanted to get a PR up for visibility.

This adds types for glamorous. At the moment, it only adds a couple css properties because having all of the properties in there makes the file cumbersome to work with.

@GAntoine is it possible to split a libdef into multiple files?

Dev with:

```bash
fswatch -l 0.1 -0 . | xargs -0 -n1 -I {} bash -c 'clear && flow --include-warnings'
```

Glamorous tracking issue here: https://github.com/paypal/glamorous/issues/348

### DIFF
--- a/definitions/npm/glamorous_v4.x.x/.flowconfig
+++ b/definitions/npm/glamorous_v4.x.x/.flowconfig
@@ -1,0 +1,5 @@
+[libs]
+flow_v0.57.x-/glamorous_v4.x.x.js
+
+[options]
+suppress_comment=\\(.\\|\n\\)*\\$ExpectError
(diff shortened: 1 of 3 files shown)


definitions/npm/glamorous_v4.x.x/.flowconfig
```diff
@@ -0,0 +1,5 @@
+[libs]
```
aaronjensen: I'll rebase this out or remove it before this is done, this just makes development much more convenient than running tests.

villesau:
@aaronjensen what's the status of this? Quite a few lines of libdefs, would be cool to get this merged! :)

aaronjensen:
I’d say the status is that glamorous is deprecated and emotion has great types and is easy to migrate to. I neither use flow nor glamorous any more so I’ll close this out. If someone else wants to take it over that’d be great. Cheers!

PR #1394, opened 2017-10-23
Add jest v21.x.x definitions

Couto:
Add definitions for jest v21.x.x
This is mostly a copy from previous versions but with the jest methods added.


Couto:
I didn't noticed about #1386 so I'm closing this.

PR #1395, opened 2017-10-23
Add initial formik types 

aaronjensen:
Still a work in progress, but can be merged as a start. Any help is appreciated. Thanks!

https://github.com/jaredpalmer/formik/issues/78


sibelius:
I think we can merge this half-baked and then improve it later on

what do u guys think?

aaronjensen:
@sibelius I'd be down, but we should probably type the HOC as Function

aaronjensen:
Ok, this should be ready to merge now. /cc @GAntoine

aaronjensen:
@GAntoine Sure, done.

ZwaarContrast:
I'm having issues using types of formik.
I get these errors:
```
[flow] property `dirty` is not readable.
[flow] property `initialValues` is not readable.
[flow] property `isValid` is not readable.
```
With this code:
```
<Formik
            initialValues={{ comments: '' }}
            onSubmit={values => {
                console.log('Submit', values)
            }}
            render={(props: FormikProps<Values>) => (
                <RequestForm  {...props} />
            )}
        />

type Values = {
    comments: string,
}
```

Am i doing something wrong?

aaronjensen:
@ZwaarContrast I think I reversed the read-only/write-only. If you change the `-` to a `+` in `FormikComputedProps`, does it fix your problem? If so, could you PR that along w/ your test case? Thanks!

PR #1406, opened 2017-10-26
Reselect libdef for flow v0.54 -- try 2 

julienw:
Follow-up to https://github.com/flowtype/flow-typed/pull/1329

Continuing working on migrating to Flow v0.54 in our project, I stumbled over this additional issue. Reordering the declarations in SelectorCreator fixes the issue, yet triggers the exact same issue for older Flow versions. So I kept the old libdef for those older Flow versions. I also reverted the patch from https://github.com/flowtype/flow-typed/pull/1329 for those versions as they don't need it, but I kept it for the current versions because otherwise they show another error (even if I still haven't understood it :) ).

This seems to show that Flow used to have the same issue as the current ones for function intersections, but in the opposite direction ;)


definitions/npm/reselect_v3.x.x/test_reselect_3.x.x.js
```diff
@@ -66,15 +66,32 @@ createSelector(
...
+  const simpleSelector = (state: State): string => "foo";
+  const compoundSelector = createSelector(simpleSelector, () => 42);
+  const func: State => number = createSelector(
+    simpleSelector,
+    compoundSelector,
+    resultFunc
+  );
+
+  const result: number = func({ x: 42, y: 42 });
+}
```
julienw: Note: I used scopes for these tests so that we don't pollute the global space with variables -- this makes it easier to not fortuitously redeclare an existing variable.

PR #1407, opened 2017-10-26
fixes: 1385 fail test if not valid success message

villesau:
This is naive solution for the https://github.com/flowtype/flow-typed/issues/1385 . Flow returns errCode 0 if it only detects warnings so it can't be used unless someone makes changes to Flow it self.

Failing on all the warnings might not be bad idea either as type definitions should follow as good practices as possible imo.

Testing the type definitions would benefit a lot of tests too. This bug could perhaps have been caught by them.

Fixing this might reveal quite a lot of broken type definitions.

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -275,6 +275,25 @@ async function getCachedFlowBinVersions(): Promise<Array<string>> {
   return versions.map(version => `v${version}`);
 }
 
+async function writeFlowConfig(testDirPath, libDefPath) {
+  const destFlowConfigPath = path.join(testDirPath, ".flowconfig");
+  const flowConfigData = [
+    "[libs]",
+    path.basename(libDefPath),
+    "",
+    "[options]",
+    "suppress_comment=\\\\(.\\\\|\\n\\\\)*\\\\$ExpectError",
+    "include_warnings=true",
+    "",
+
+    // Be sure to ignore stuff in the node_modules directory of the flow-typed
+    // CLI repository!
+    "[ignore]",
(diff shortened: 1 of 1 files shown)


aaronjensen:
It looks like include_warnings isn't valid in older flow versions. You'll probably need to write a different flowconfig based on the version

villesau:
That's actually the tricky part! seems that the type definition tests are run in parallel and they all uses the same flowconfig. Flow doesn't allow defining the .flowconfig path separately so i can't define it differently for each version.

The range can overlap between 0.52.x and 0.53.0 when the include_warnings was introduced so using batch version range is not feasible.

villesau:
Now the runner writes .flowConfig for each version separately, but that means that the tests are run in series which most likely makes them way too slow.

villesau:
Now it only creates new .flowconfig when needed and run tests in batches again.

GAntoine:
I was working on another solution, but stopped when notices how similar it was to this one. I'll take a look at it this evening.

cli/src/commands/runTests.js
```diff
@@ -354,61 +415,28 @@ async function runTestGroup(
...
+    );
+    const lowerVersions = flowVersionsToRun.filter(flowVer =>
+      semver.lt(flowVer, "0.53.0")
+    );
 
-            child.on("close", errCode => {
-              res({ stdErrOut, errCode, execError: null });
-            });
-          });
+    await writeFlowConfig(testDirPath, testGroup.libDefPath, true);
```
GAntoine: Could you switch there around to run the lower versions first?

villesau:
I also changed the version hinting check to work in the same way as the actual tests.

PR #1409, opened 2017-10-27
Add types for persist-path v1.x.x

Bannerets:
https://github.com/hobbyquaker/persist-path
https://www.npmjs.com/package/persist-path


Bannerets:
Am I doing something wrong?

GAntoine:
Try switching from `flow_v0.22.x-` to `flow_v0.25.x-`, that should make the specs pass.

PR #1411, opened 2017-10-27
Write test for 1385 regression

villesau:
Drafting a test case to prevent https://github.com/flowtype/flow-typed/issues/1385 from happening again.

@GAntoine if you have an idea how to run this only against latest version of Flow, that would probably be good. Now the solution checks the latest console log and error code.

This would also be the first end to end test for libdef tests :)


cli/src/commands/runTests.js
```diff
@@ -593,7 +593,8 @@ export async function run(argv: Argv): Promise<number> {
   const onlyChanged = Boolean(argv.onlyChanged);
 
   const cwd = process.cwd();
-  const cwdDefsNPMPath = path.join(cwd, "definitions", "npm");
+  const basePath = argv.path ? String(argv.path) : cwd;
+  const cwdDefsNPMPath = path.join(basePath, "definitions", "npm");
```
GAntoine: Is path a new argument you just created for this purpose?

villesau: Yep I added it only to be able to adjust the libdef path for tests.

GAntoine: Not much, add an entry for it in the wiki, and add a `setup` method to `runTests`, so that the  new parameter appears when running `flow-typed run-tests --help`.

You'll need to change the `run` method to accept the new `Args` instead of `Argv`:

```javascript
import typeof Yargs from 'yargs';

export type Args = {
  _: Array<string>,
  path?: string,
};

export function setup(yargs: Yargs) {
  return yargs.usage(`$0 ${name} - ${description}`).options({
    path: {
      alias: 'p',
      describe: 'Some description...',
      type: 'string',
    }
  })
})

export async function run(args: Args) { ...
```

villesau: Thanks! Added the setup function and wiki entry.

GAntoine:
Alright, everything looks good and the specs seem to run correctly. My only concern is the run time taking ~20 seconds on my system. I'm happy to merge this in, and we can take a look at improving that at a later time.

villesau:
@GAntoine yeah that's my concern too. When the run time becomes a problem, might be worth thinking how to run only single version of flow. Now it runs every version from 0.54.x ->. This also means that the run time will increase on every flow release.

PR #1422, opened 2017-10-30
Add direct exports declarations

Leeds-eBooks:
I'm following the technique used [here](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/lodash_v4.x.x/flow_v0.55.x-/lodash_v4.x.x.js) to declare types for direct exports (e.g. `import pure from 'recompose/pure'`)


Leeds-eBooks:
It doesn't look like any of these failures are related to my changes?

GAntoine:
Seems like the current definitions don't work with flow 57+. I can't merge this in while the build is broken, so you'll either need to fix the current definition, or cap the flow version at 56, ie `flow_v0.55.x-v0.56.x`. If you're comfortable enough fixing the definition, that would be :100:.

Leeds-eBooks:
@GAntoine I've had a go at fixing the def but it is so unbelievably complex I just can't get my head around it. So I went for the easier option ¯\_(ツ)_/¯

GAntoine:
Haha I don't blame you. I'm perfectly happy with the easy option, but let me ping some people who could maybe fix it for 57+. If we don't get anywhere I'll merge this in, as it's still a useful change.

GAntoine:
@istarkov You seem to be the recompose guy. If you have the time, could you look into why this definition is failing with flow v57? I took a look at the flow changelog, but couldn't see any changes that would break the current definition.

Leeds-eBooks:
@GAntoine Thanks for merging – do I need to do anything / wait to get install the updated definition? `flow-typed install recompose@0.24` is not yet including these changes.

GAntoine:
What version of flow are you on? If you're on 57, I just merged a definition for 57 last night, so just try `flow-typed install`. Maybe try removing the current `recompose_v0.24.x.js` file from flow-typed folder first?

Leeds-eBooks:
@GAntoine My mistake, I was on a version of flow lower than 0.55, and this PR only applies to 0.55 & 0.56! Thanks for your help.

Issue #286, opened 2016-08-30
[Ecosystem] Vendoring in libraries?


ryyppy:
I realized that more and more libdefs are about to be shipped with the related npm package... which is awesome, but.... isn't this going to be a huge problem with breaking flow versions? 

Flow-Typed at least has a formal way of handling flow dependencies!

There needs to be a definitive recommendation how to handle libdefs for library authors... IMO it makes more sense to have a readme template instructing the users how to get libdefs for the target library and keep the libdefs in one place...

Talking especially about these PRs in [redux](https://github.com/reactjs/redux/pull/1887) and [avajs](https://github.com/avajs/ava/pull/1007)

Other libraries I could think of are react-native, draftjs, immutablejs ,...

@gabelevi @jeffmo @thejameskyle

thejameskyle:
@gabelevi @jeffmo and I met a few weeks back to discuss the different workflows we wanted to recommend.

For libraries there are two scenarios: Ones written in Flow and ones that are not written in Flow.

For ones written in Flow the recommendation in our document is to ship them as `.js.flow` files. For ones that are not, the recommendation is to contribute back to flow-typed.

I'm still not a fan of this, I want it to be that flow-typed is always the source of truth and we should build tools around that. I think that makes for a better ecosystem.

montogeek:
What about something similar as https://github.com/typings/typings but for Flow?

jrnail23:
@thejameskyle: for a lib written in flow that also uses flow-typed libs, how should we deal with those flow-typed defs?  Do you also ship the flow-typed defs as `.js.flow` files?

And when using my lib in a project, will that lib's flow-typed defs conflict with any that my project may already be using? (i.e., let's say both the lib and my project use bluebird's flow-typed defs)

Brianzchen:
After 5 years this issue seems solved.
- typed libs ship with `.js.flow`
- untyped are added to flow-typed
- for libs written in flow that use flow-typed it's been solved recently by #4143

PR #1435, opened 2017-10-31
feature: install defs from node_modules/<package>/flow-typed if it exists for each package

jedwards1211:
This is a workaround for https://github.com/facebook/flow/issues/4917.  (Problem: It's impossible to provide different `.js.flow` files for different versions of flow in the same package.)

One can currently rely on PRing def updates for the various targeted flow versions to the central `flow-typed` repository instead of using `.js.flow` files, but this has downsides:
* You have to wait for the PR to be merged
* You can't do this for scoped private packages, at least not without revealing information about them

That's why I made this PR: this makes it possible to store definitions (which can target multiple versions of flow) in a `flow-typed` directory in the published package instead of making PRs to the central repo.

### Notes
* self-provided defs will only be found for packages that are already installed in `node_modules`!
* if an installed package provides its own `flow-typed` directory it will override anything in `flow-typed`'s central repo.

### Directory structure
A package's `flow-typed` directory has the same structure as `flow-typed/definitions/npm/<package>_v<version>/` would,
except that the `.js` files must use the exact version of the package:
```
node_modules/underscore/
  flow-typed/
    flow_v0.13.x-v0.37.x/
      underscore_v1.8.3.js
    flow_v0.38.x-/
      underscore_v1.8.3.js
```


mruzekw:
Wow! I hadn't thought of this. My only concern is that it might get confusing using `flow-typed` to both consume and publish types. How are the two cases differentiated?

jedwards1211:
@mruzekw well, we would need to update the wiki to document this way of publishing types for a library if this PR gets merged.

mruzekw:
Yes, that sounds important. But I'm unclear on the differences now...could you clarify? The path and the directory tree in "Directory Structure" don't match.

jedwards1211:
@mruzekw I meant that path to be within this repo, not within the package in question.  I tweaked the wording

mruzekw:
So the published package tree needs to have the following?
```
node_modules/underscore/
  flow-typed/
    flow_v0.13.x-v0.37.x/
      underscore_v1.8.3.js
    flow_v0.38.x-/
      underscore_v1.8.3.js
```

jedwards1211:
ah no, it just needs to have
```
  flow-typed/
    flow_v0.13.x-v0.37.x/
      underscore_v1.8.3.js
    flow_v0.38.x-/
      underscore_v1.8.3.js
```
(as a sibling of `package.json`)

mruzekw:
Got it. Thanks for clarifying.

It's unlikely a project would check these into git and just generate it as part of the build.

jedwards1211:
@mruzekw yes, though if one needs to support multiple versions of flow, I don't think there's a very good way to autogenerate all the defs anyway...

jedwards1211:
@mruzekw oh I think I misunderstood, you mean you would prefer to generate the `flow-typed` directory in `prepublish` rather than checking it into git?

mruzekw:
I have something in the works with respect to generating flow types across multiple versions. It's not easy... In the proof of concept stage.

I mean you have to update all the flow types for every version with each change in package version anyway. Why not automate that?

jedwards1211:
Oh, that's cool.  Yeah by all means if you have a way to automate it, go for it!

mruzekw:
Cool, that seems like a separate choice from this PR anyway. Great work! Hope the maintainers and community will consider it.

mruzekw:
Oh! One more question. Can I test my package(d) types with different versions of flow?

jedwards1211:
@mruzekw you mean using `flow-typed run-tests`?  I believe that won't work with this since `run-tests` is designed to be run on the flow-typed repository itself.  I'll have to think about the best way to support running tests on the packaged types.

GAntoine:
Oh wow, nice work! 💪 I'm gonna take some time to look over this change, and reach out to the others since this is a significant architectural change.

@marudor Thoughts on this change? 💭

jedwards1211:
@GAntoine do you think we should allow the `.js` files in the package not to use the exact package version?  That was just the way it came out when I reused some existing functions in `flow-typed`, but I figured it might be a good thing as it would force users to think about keeping their defs up-to-date.

TiddoLangerak:
I'm just a passer-by, but I have my doubts about the folder structure: if you're a library publisher that wants to publish flow types, then chances are you're already using `flow-typed` for your own dependencies. Currently the `flow-typed` folder contains only 3rd party code, which you don't want to publish. This folder is likely in your `.npmignore`, `.gitignore`, or omitted from the files array in your `package.json`. Chances are also that this folder is treated differently by other tooling (e.g. excluded from linters).

If we now add our own type defs in there as well then we need to instruct our tools to treat half of the folder differently from the rest. Especially for include patterns this isn't very nice (such as the files array in `package.json`), since we now either need to enumerate all versions, or use glob patterns to match all the `flow_*` folders.

Now this is all not impossible to do, but I think it would be more ergonomically to have a single folder dedicated to own types. This could either be a new top-level folder, or a single folder within `/flow-typed` (e.g. `/flow-typed/package/flow_v0.13.x-v0.37.x/underscore_v1.8.3.js`).

jedwards1211:
That's a good point, I'm trying to think what other names would make sense for the own types folder...

jedwards1211:
@GAntoine @marudor I'm thinking `flow-self-typed` might be a good name for a package's own types folder. What do you think?

jedwards1211:
@GAntoine what do you think about calling the types folder `flow-self-typed`?

GAntoine:
In all honesty I'm not too sure about this solution.

At first glance I have to agree with the points Tiddo brought up, in that we can't hijack the `flow-typed` library for this use. As for `flow-self-typed`, I can't justify making such a change, or installing that as a best practice, without input from the core flow team. Unfortunately it's a busy time of year for everyone, so this will have a linger for a while.

At the very least I'll keep it open and we can keep discussing it, since it's an interesting idea to a real problem people are having. I'm sorry I can't be of more help, I just don't think I'm in a position to call the shots. 😞

jedwards1211:
I'm not sure if you fully understand the points that Tiddo brought up.  Publishing I a package with its typedefs in a `flow-typed` folder is awkward for people's workflow.  But I could change this PR to look for self-provided types in a package's `flow-self-typed` folder instead, and it wouldn't interfere with any existing functionality or be awkward to use.  In other words what Tiddo mentioned doesn't mean this approach is doomed, just needs one little tweak.  The only concern is the name for that folder, there are no other ramifications.

GAntoine:
Oh no I understand what the purpose of `flow-self-typed` is, I wasn't saying that his comments were why I was unsure of this change. From what I can tell your solution would work just fine in the wild.

My reservations are towards making such a change without input from the flow/flowtype team, or from library maintainers who have flow typedefs in their libraries. This would require changing the way people interact with flow and flow-typed, and I don't feel I'm in a position to make that call.

mruzekw:
I'm just happy this issue is getting some attention. I wish the flow team
were more proactive on this.

jedwards1211:
To be perfectly clear, here is how it would work with what I'm proposing that avoids the problem Tiddo mentioned.

Let's say you're developing a package `foo` that depends on `underscore`.  Here is what your repo would look like:

```js
.npmignore
flow-self-typed/
  flow_v0.13.x-v0.37.x/
    foo_v1.0.0.js
  flow_v0.38.x-/
    foo_v1.0.0.js
flow-typed/
  npm/
    underscore_v1.x.x.js
lib/
  index.js
package.json
```

`flow-typed` controls the `flow-typed/` directory and it still works for flow checking.
`.npmignore` would ignore the `flow-typed` directory, but `flow-self-typed` would get published with the package.

Then someone installs `foo` in their project, which also uses `flow-typed/`.  At that point, their project looks like this:
```js
node_modules/
  foo/
    flow-self-typed/
      flow_v0.13.x-v0.37.x/
        foo_v1.0.0.js
      flow_v0.38.x-/
        foo_v1.0.0.js
    lib/
      index.js
    package.json
  underscore/
    ...
package.json
```
Now they run `flow-typed install`.  When looking for defs for `foo`, `flow-typed` sees that `node_modules/foo/flow-self-typed` is a directory; it then copies `node_modules/foo/flow-self-typed/flow_v0.38.x-/foo_v1.0.0.js` to the project's `flow-typed/npm/foo_vx.x.x.js`.  For other packages, like `underscore`, it still copies the defs from the central repository.

Does that clarify things?

jedwards1211:
Also to be clear, we wouldn't have to make any changes to `flow` itself to support this.  You do realize that?

jedwards1211:
And remember, with self-provided type defs, you guys wouldn't have to waste as much time going through PRs for type def updates to this repo.

GAntoine:
Right, I think I'm getting a much better idea of what you're proposing here. 👀  Give me the weekend to pour over it, and I'll try my best to get someone else to look over it was well. In the meantime any input from others would be greatly appreciated!

jedwards1211:
@GAntoine cool, thanks :smile: I think the way `flow-typed` was designed works best for packages whose maintainers are unwilling to provide flow types, and that was obviously what it was designed for.  But for private packages and packages whose maintainers do want to provide flow types, I think self-provided defs like this are more convenient.  The only real alternative is changing flow itself to support multiple `.js.flow` files per `.js` file for different versions of flow, which would be a way more significant change than this.

GAntoine:
Funny enough, someone just asked that question as an issue: #1451. I've asked caleb if there's a convention in place for them, and maybe his answer will help clarify what to do with this PR.

jedwards1211:
@GAntoine as far as I know the only convention is `.js.flow` files...and given their limitations I've mentioned, I hope to make this the preferred way for libraries to provide their own libdefs.

GAntoine:
Right, so I gave this some thought, and after thinking about the use cases I don't feel like having the library maintainer version flow themselves is the best/should be the recommended approach. I'd wager most maintainers are happy keeping definitions for the latest version of flow they're using, and won't take the effort to version their typedefs according to our requirements (our being flow-typed, not the flow. project)

IMO our approach should be to have them keep definitions for the latest flow version they're using, and if they'd like to version by flow release, they can submit PRs to this repo. This has the added benefit that the community can update these definitions along with the library's author(s), keeping definitions up to date and preventing ones from projects that no longer use flow from deprecating.

The downside to what I'm proposing what brought up in the other thread: *private libraries*. Organizations with private code may not want to expose their type definitions, and with good reason. They could fork the `flow-typed` project and tweak the cli to point to that private fork:

> This would be a viable option. The flow-typed CLI already contains the logic to resolve libdefs with different versions of flow.  

What I don't agree with is the following:

> However, the company would need to set up and maintain their own repository and tools. It's possible the resources aren't there to do this, or it's not worth it for one library.

The work and maintenance needed would be minimal, and we can absolutely find ways to ease the process.

To wrap my post up: I think this is a real problem some flow users/library maintainers are facing, and your solution has opened up some great discussion, but I don't feel like it's an issue `flow-typed` needs to solve. What we can do instead is alleviate the burden with clear and easy to find documentation.

BTW I don't plan on closing this PR, since I don't want to lose track of the work put into it. If anything it's a great reminder for future discussions.

mruzekw:
@GAntoine Thanks for giving this some thought.

I think what would make using a private repository easier is allow us to add those via configuration. That way the team doesn't need to fork and maintain it for their needs.

In the case of a library definition that matches exactly in name, flow version, and library version, flow-typed would privilege the private repo definition.

jedwards1211:
The downside of maintaining a private flow-typed repo for right now is keeping it in sync with the public flow-typed repo.  Unless it layers on top of the public repo like @mruzekw suggests, we have to set up some kind of task to periodically pull from the public repo to our private repo.

And for private packages, it really makes sense to release the defs at the same time as you release the package, in contrast to the way we want to be able to update defs for defunct public projects.  Releasing typedefs separately is more of a headache.  If you don't have several dozen private packages you're using in multiple different projects, where updating to the latest version of Flow for all at once is a major ordeal, I don't think you really realize how much easier this would make the process of gradual upgrades throughout the organization's codebase.  I still prefer having this additional option (perhaps allowing the defs in flow-typed to always take priority, in case a project stops being maintained).

GAntoine:
Related: #286

jedwards1211:
Yeah, there are no alternative plans for releasing defs for private packages are there? No plans to make versionable .js.flow files?

jedwards1211:
I may just end up putting the functionality of this PR into a standalone command that replaces the stubs created by `flow-typed`.

GAntoine:
> Yeah, there are no alternative plans for releasing defs for private packages are there? No plans to make versionable .js.flow files?

Not as far as I can tell. 😕

chrisgervang:
I really like the proposed solution, and I'd prefer adding more flexibility into `flow-typed` over requiring private projects to fork and maintain. Wanting to use published types from private packages is not limited to companies, since it's only $7/m now for individuals to use that feature.

Regarding,

> I don't feel like having the library maintainer version flow themselves is the best/should be the recommended approach

I feel the same, but can empathize with a package author who does want to maintain multiple versions of their types. To satisfy both, this could look for specific flow version directories first and then look for libdefs in the `flow-self-typed` directory. 

That way, it's harder to get blocked. If package authors don't want to maintain more than one version of libdef, they don't have to. If someone needs to add backward compatibility, they can.

```
.npmignore
flow-self-typed/
  foo_v1.0.0.js
  flow_v0.13.x-v0.37.x/
    foo_v1.0.0.js
  flow_v0.38.x-/
    foo_v1.0.0.js
flow-typed/
  npm/
    underscore_v1.x.x.js
lib/
  index.js
package.json
```

chrisgervang:
FWIW, I'm using the script written by @A-Tokyo for now: https://github.com/facebook/flow/issues/4323

jedwards1211:
@chrisgervang That's a good idea.  I wish types from the `flow-typed` directory would override `.js.flow` types (https://github.com/facebook/flow/issues/5233); if that were the case instead of having a `flow-self-typed/foo_v1.0.0.js` file, people could just continue relying on `lib/index.js.flow` for the flow-version-agnostic types, and it wouldn't interfere when people want to add flow-version-specific defs to flow-typed.

jedwards1211:
@GAntoine I didn't notice, looks like you guys have decided you'd like to adopt this approach in 3.0?  If so I'll update the code to use the `flow-self-typed` directory.

jedwards1211:
Also @GAntoine make sure no one mistakenly thinks this is ready and merges it -- it still uses the `flow-typed` directory inside a package, not the `flow-self-typed` directory.

jedwards1211:
@GAntoine are you able to run the chrome node inspector on Jest?  I'm seeing errors like "Unable to open devtools socket: address already in use."

GAntoine:
@jedwards1211 I've never been able to use the debugger successfully, I usually resort to `console.log`.

jedwards1211:
@GAntoine it was painful but I finally solved my issues with `console.log`.  I just pushed changes: now this branch looks for the `flow-self-typed` directory in an installed package, and the tests are now correct.

Notes:
* I had to add a `selfTyped` flag to self-provided `NpmLibDef`s to handle two edge cases:
  * If you have `underscore^1.6.0` in your `dependencies`, `flow-typed` will not install defs for `1.8.3` (for example) from the central repo.  But if you have `1.8.3` installed and it provides its own types (at `node_modules/underscore/flow-self-typed/flow_v0.38.x-/underscore_v1.8.3.js`), they will get installed as `flow_typed/npm/underscore_v1.8.3.js`.
  * In `getNpmLibDefVersionHash`, instead of trying to get a git commit hash for self-provided type defs, I use the library package version instead.

PR #1439, opened 2017-11-01
Document cli's --ignoreDeps flag

A-Tokyo:
- :books: Document the CLI's `--ignoreDeps` flag

#### Reason:
Most of the time people don't want to install the `devDependancies`,  I faced the problem myself and found a lead at https://github.com/flowtype/flow-typed/issues/599 I think it's a good addition to the `README`; by intuition, I checked there first.


GAntoine:
Yeah I can agree with this change. I'm usually hesitant of adding more into the README, but in this case I think it's important enough to have it in an obvious place.

PR #1441, opened 2017-11-01
[redis] Added initial library definition for Redis v2 API

yeiniel:
The redis Node.js package has a big surface. This library definition just address a few method. I hope in the future other contributors extends the library definition as required by their needs.


yeiniel:
Sure no problem. I will go back to this as soon as possible.

yeiniel:
I'm seeing errors on Travis CI but they don't appear to be related to the new pull request

PR #1450, opened 2017-11-02
Improve typing for Ramda #prop.

AndrewSouthpaw:
This builds on work in #1186, regarding `prop` returning a non-maybe type. That PR was approved but has not been updated, so I included them here in addition to my own changes.

A current shortcoming of `prop` is that it does not allow for objects with more than one type for values, due to it's definition: `declare function prop<T, O: { [k: string]: T }>`. A perfectly common use case like this would fail:

```js
const myName = prop('name', { id: 1, name: 'Andrew' }: { id: number, name: string })
```

Using the `$ElementType` introducing in flow 0.49 allows to gain advantages of type checking the accuracy of keys *and* types of values.

Most of the PR is simply renaming files:

- Changed the latest folder to be `flow_v0.39.x-v0.48.x`
- Created a new folder `flow_v0.49.x-`

The actual changes are in [this commit](https://github.com/flowtype/flow-typed/commit/a1d498220fcebd281da6dc85f777815ec38e05ef).


AndrewSouthpaw:
Thanks @GAntoine ! Shall I go through the rest of the library and fix it everywhere? It seems like there's a lot of places where it uses `O: { [k: string]: T }`.

GAntoine:
If it's not too much trouble that would be sweet! In all honesty I don't understand much about this typedef, so I trust that any changes you make are accurate. If you feel like these changes need to be mane, and you're up for it, I'll review and merge in whatever you send out way.

AndrewSouthpaw:
👍

Issue #1272, opened 2017-09-22
flow 0.55.0 issues with `bluebird`


EvHaus:
After upgrading to Flow 0.55.0, `bluebird` now complains about:

```
[No file]:0
inconsistent use of library definitions
     v-----------------------------------------------------
 53: declare class Bluebird$Promise<+R> extends Promise<R>{
 54:   static Defer: Class<Bluebird$Defer>;
 55:   static PromiseInspection: Class<Bluebird$PromiseInspection<*>>;
...:
202: }
     ^ Bluebird$Promise. Covariant property `bind` incompatible with invariant use in. See lib: flow-typed/npm/bluebird_v3.x.x.js:53
     v-----------------------------------------------------
 53: declare class Bluebird$Promise<+R> extends Promise<R>{
 54:   static Defer: Class<Bluebird$Defer>;
 55:   static PromiseInspection: Class<Bluebird$PromiseInspection<*>>;
...:
202: }
     ^ Function. See lib: flow-typed/npm/bluebird_v3.x.x.js:53

[No file]:0
inconsistent use of library definitions
136:   static bind(ctx: any): Bluebird$Promise<void>;
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. This type is incompatible with. See lib: flow-typed/npm/bluebird_v3.x.x.js:136
 98:     bind: Function$Prototype$Bind; // (thisArg: any, ...argArray: Array<any>) => any;
               ^^^^^^^^^^^^^^^^^^^^^^^ function type. See lib: /private/tmp/flow/flowlib_24bda01a/core.js:98

```

capaj:
If it helps anyone, commenting this out fixed my flow check. It's probably going to error out if you use `bind()` method in your code, but if you're not, it won't be an issue.

jcready:
See https://github.com/facebook/flow/issues/4953

SimenB:
@capaj what line did you comment out?

sporto:
133 `static bind(ctx: any): Bluebird$Promise<void>;`

SimenB:
Thanks!

jcready:
This is a [regression in flow](https://github.com/facebook/flow/issues/4953).

philipp-sorin:
4 months had been passed but issue still exists? Anyone want to do anything with it?

PikachuEXE:
As mentioned before this is reported as a regression in flow
[...]

PR #1457, opened 2017-11-03
Bluebird definition Flow v55 + Updates

GAntoine:
This makes some tweaks to the existing bluebird definitions. It also caps the latest definition at flow v54, and creates one for flow v55, but with `static bind()` commented out. This sorta addresses #1272 and facebook/flow#4953 (but not really), in that users of bluebird can update their flow versions to the latest, until a real fix is made.


GAntoine:
@EvHaus would you mind taking a look at this? I'd like your opinion on it.

EvHaus:
Did you run the code through `prettier` in the PR as well? It makes it pretty hard to see where the actual code changes were. Any chance the code formatting can be moved to a separate PR, so this one is just for addressing the actual type changes?

PR #1000, opened 2017-07-01
Add storybook-router

pascalduez:
Hi,

https://github.com/gvaldambrini/storybook-router

Is importing types from other lib declarations supposed to work? 
`import type { LocationShape, GetUserConfirmation } from 'react-router';`
Does not seem to do the validation job, but default to `any`.

PR #1459, opened 2017-11-03
Add storybook-router (replaces #1000)

GAntoine:
Credit to @pascalduez for the work

### DIFF
--- a/definitions/npm/storybook-router_v0.x.x/flow_v0.25.x-/storybook-router_v0.x.x.js
+++ b/definitions/npm/storybook-router_v0.x.x/flow_v0.25.x-/storybook-router_v0.x.x.js
@@ -1,0 +1,35 @@
+type LocationShape = {
+  pathname?: string,
+  search?: string,
+  hash?: string,
+  state?: any
+};
+
+type GetUserConfirmation = (
+  message: string,
+  callback: (confirmed: boolean) => void
+) => void;
+
+type Renderable = React$Element<*>;
+type RenderFunction = () => Renderable;
+type StoryDecorator = (story: RenderFunction) => Renderable;
+
+declare module "storybook-router" {
+  declare type Links = {
+    [key: string]: (kind: string, story: string) => Function
+  };
+
+  declare type RouterProps = {
+    initialEntry?: Array<string>,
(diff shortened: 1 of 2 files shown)


definitions/npm/storybook-router_v0.x.x/flow_v0.25.x-/storybook-router_v0.x.x.js
```diff
@@ -0,0 +1,35 @@
...
+  hash?: string,
+  state?: any
+};
+
+type GetUserConfirmation = (
+  message: string,
+  callback: (confirmed: boolean) => void
+) => void;
+
+type Renderable = React$Element<*>;
```
sibelius: can't we import this from React ?

import type { Element } from 'react'?

or is it better to use private def?

GAntoine: Yeah I think that's safe. Would it be `type Renderable = Element<*>` or `type Renderable = Element`?

definitions/npm/storybook-router_v0.x.x/flow_v0.25.x-/storybook-router_v0.x.x.js
```diff
@@ -0,0 +1,37 @@
+import type { Element } from "react";
```
pascalduez: Isn't ` import type { Element } from "react"; ` only available from Flow >= 0.53?
In such case Element would resolve `any` for all other lower versions.

GAntoine: Correct, I figured it's not a huge deal if it does resolve to `any`. If you're keen to fix it, a PR would be much appreciated!

pascalduez: Your right, that's fine as is I guess. ;)

pascalduez: Thanks for all the hard work!

PR #1467, opened 2017-11-05
Allow a dependency to be listed in multiple package.json dep fields

### LINKED ISSUE #379: Recommended approach for using flow-typed with peerDependencies

GAntoine:
This PR does 2 things
- allows multiple entries of a package in `package.json`, as long as the versions match
- only installs definitions for packages under `dependencies`, `devDependencies` and `optionalDependencies`.

Dependencies listed under `bundleDependencie` and `peerDependencies` are no longer installed. In the case of `bundleDependencie`, we expect the package-version to be listed under another dependencies. In the case of `peerDependency`, packages under `peerDependencies` aren't actually used/installed, so it should be safe to ignore them.

Replaces and enhances #587 
Closes #379
Credit to @ahutchings for most of the work

### DIFF
--- a/cli/src/commands/__tests__/install-test.js
+++ b/cli/src/commands/__tests__/install-test.js
@@ -222,10 +222,8 @@ describe("install (command)", () => {
             }
           })
         ]);
-        await gitInit(FAKE_CACHE_REPO_DIR), await gitAdd(
-          FAKE_CACHE_REPO_DIR,
-          "definitions"
-        );
+        await gitInit(FAKE_CACHE_REPO_DIR),
+          await gitAdd(FAKE_CACHE_REPO_DIR, "definitions");
         await gitCommit(FAKE_CACHE_REPO_DIR, "FIRST");
 
         setCustomCacheDir(FAKE_CACHE_DIR);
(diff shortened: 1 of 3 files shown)


jedwards1211:
@GAntoine I think in most cases where one uses `peerDependencies`, one will have the same package in `devDependencies` with not necessarily the same version but a semver-compatible version.  Would you accept a change to make this PR tolerate that?

GAntoine:
@jedwards1211 Absolutely! If you have the change in code, you can post it here. Likewise I added you as a collaborator to my fork, so you're welcome to make the change right in branch as well, whatever you prefer.

If you don't have the code handy, what would the change entail?

jedwards1211:
Sorry, I'm just now getting to this, taking a look right now

jedwards1211:
Hmm, it is trickier than I expected because `semver` doesn't have a function to tell you if a version range satisfies another version range (which would be a common case where `^1.2.3` is in `devDependencies` and `^1.0.0` is in `peerDependencies`).  But it may still be possible to solve this problem.

Is it really `flow-typed`'s job to error out if your versions are incompatible though?  Why not just have it warn but install the defs for the highest of the specified versions?

jedwards1211:
There is a `semver.intersects` function that can tell us if ranges intersect. It looks like they don't provide any function to tell if a range is fully contained in another range.

Brianzchen:
@GAntoine can we continue with this PR (or I can open a fresh one myself?). It is rather confusing being told I have duplicate dependencies when one is a peer dependency that isn't used. I'd rather it not warn me and not install anything if specified in peerDependencies. 

I also agree mostly with @jedwards1211 that we shouldn't error, it's really not flow-typed's job to ensure a project is well formed though in regards to which version to install, just install whatever the `devDependencies` or `dependencies` section wants. 

Side note: I think we'd need a separate change in the future that focuses on picking the right single version if we actually find installable duplicates because of these two changes https://github.com/flow-typed/flow-typed/pull/4140 & https://github.com/flow-typed/flow-typed/pull/4143

PR #1473, opened 2017-11-06
add ObjectId for mongodb

sibelius:
I'm having this error

```jsx
 * mongodb_v2.x.x/flow_v0.50.x- (flow-v0.51.1): Unexpected Flow errors(2):
   Error: 0-test_mongodb-v2.js:5
     5: mongodb.ObjectId.isValid('000000000000000000000000');
                         ^^^^^^^ property `isValid`. Property not found in
     5: mongodb.ObjectId.isValid('000000000000000000000000');
        ^^^^^^^^^^^^^^^^ MongoDB$ObjectID
```

I would like to use `mongodb` def on `mongoose` to fix this one https://github.com/flowtype/flow-typed/issues/1469


mjomble:
This new definition causes problems on projects that import things other than `ObjectId` from the `mongodb` package.

For example, this fails now: `const { MongoClient } = require('mongodb')` but didn't with the auto-generated stub.

While adding full typedefs for the entire `mongodb` package would be a larger undertaking, perhaps the remaining exports of the package could be typed as `any` in the meantime?

sibelius:
@mjomble would you mind adding some of them based on typedef ts 

send them on a PR and we can improve it together

mjomble:
Ok, I'll look into it

PR #1475, opened 2017-11-06
data.maybe - fromEither, fromValidation, ap

StefanoMagrassi:
Add to `data.maybe` lib missing definitions:

- **[fromEither](http://docs.folktalejs.org/en/latest/api/data/maybe/Maybe.html#Maybe.fromEither)** - static constructor method
- **[fromValidation](http://docs.folktalejs.org/en/latest/api/data/maybe/Maybe.html#Maybe.fromValidation)** - static constructor method
- **[ap](http://docs.folktalejs.org/en/latest/api/data/maybe/Maybe.html#Maybe.prototype.ap)** - applicative functor method


definitions/npm/data.maybe_v1.x.x/flow_v0.32.x-/data.maybe_v1.x.x.js
```diff
@@ -1,7 +1,28 @@
-// flow-typed signature: 9ed0f65c3fef605d7b40569f85a56b00
-// flow-typed version: <<STUB>>/data.maybe_v1.2.2/flow_v0.35.0
+// @flow
 
-declare module 'data.maybe' {
+import type Either from "data.either";
```
GAntoine: This will just resolve to `any` type, it won't actually validate that the `Either` passed to `fromEither` is of the right type. If that's ok with you I can merge this in anyways.

StefanoMagrassi: @GAntoine can I do something to fix this or it is a feature not yet implemented in flow-typed?

GAntoine: This is just a hunch, but can you moving it into `declare module 'data.maybe'`, running the tests on your system and reporting back?

StefanoMagrassi: I tried moving it into the module declaration:

```javascript
declare module "data.maybe" {
  import type Either from "data.either";
  ...
```

and expecting for an error in my tests, but it didn't work:

```
// $ExpectError
const _em = Maybe.fromEither('this is not an Either Monad');
```

GAntoine: Hmm, well I'm stumped. Just move everything back the way it was and I'll merge it in. I'd rather have a somewhat working definition then no definition at all!

StefanoMagrassi: Yes, I agree.

PR #1479, opened 2017-11-07
formik, installs dummy expected error to fix eslint error

pcwinters:
PR https://github.com/flowtype/flow-typed/pull/1395 introduced a definition test file that fails an eslint rule requiring a negative test. To be honest, I'm not sure how the travis build was successful for it. Is there something missing in the travis config?

Clone master to reproduce:
```
cd definitions
yarn install
yarn test

/Users/patrick.winters/projects/hack/flow-typed/definitions/npm/formik_v0.9.x/test_formik.js
  1:1  error  Test has to include at least 1 negative Tests  flow-typed/negative-tests
```

### DIFF
--- a/definitions/npm/formik_v0.9.x/test_formik.js
+++ b/definitions/npm/formik_v0.9.x/test_formik.js
@@ -1,15 +1,30 @@
 // @flow
 
 import * as React from "react";
-import { Formik, Form, Field, type FieldProps } from "formik";
+import {
+  Formik,
+  Form,
+  Field,
+  type FieldProps,
+  type FormikActions,
+  type FormikConfig
+} from "formik";
 
 <Formik
   initialValues={{ foo: "hi" }}
-  onSubmit={(values, formikActions) => {
+  onSubmit={(values: *, formikActions: FormikActions<*>) => {
     values.foo;
     values.bar;
   }}
 />;
+
+const config1: FormikConfig = {
+  onSubmit: (values: *, formikActions: FormikActions<*>) => {}
+};
+// $ExpectError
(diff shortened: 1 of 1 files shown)


definitions/npm/formik_v0.9.x/test_formik.js
```diff
@@ -3,6 +3,9 @@
 import * as React from "react";
 import { Formik, Form, Field, type FieldProps } from "formik";
 
+// $ExpectError
+const foo: number = "bar";
```
GAntoine: I know it's/I'm nitpicking here, but can you replace this with a broken version of what's below it? The entire point is to test that incorrect use of `Formik` will throw errors. Something like:

```javascript
<Formik
  // $ExpectError
  initialProperties={{ foo: "hi" }}
/>;
```

pcwinters: Not nit picking at all. I must have been feeling really lazy. Updated

pcwinters:
@GAntoine, up to date. Can you think of how the MR that introduced this would have passed the build in the first place? Running `./travis.sh` on a fresh clone looks to be broken.

GAntoine:
Great, thanks for changing that. I have a hunch, the flow team recently changed the way `$ExpectError` works, and we had to tweak a couple things to work around it no longer working. One of them dealt with warnings vs. errors on build, though my memory is fuzzy. I don't have the time to look into it now, but if you can open a quick issue about it, that would be 🔥 .

pcwinters:
@GAntoine , https://github.com/flowtype/flow-typed/pull/1496. It was something much more straightforward.

PR #1496, opened 2017-11-08
run 'yarn test' in defintions, which enforces eslint rules

pcwinters:
Recently I found that I was unable to successfully run `./run_def_tests.sh` after updating a ramda defintion because another definition test was committed without a negative test (in violation of the "flow-typed/negative-tests" rule in `definitions/.eslintrc`. https://github.com/flowtype/flow-typed/pull/1479

FWIW, this commit [https://github.com/flowtype/flow-typed/commit/fb7612683c8a2546d95bfe58f91f50c654ecd066](https://github.com/flowtype/flow-typed/commit/fb7612683c8a2546d95bfe58f91f50c654ecd066), a year ago, commented out `yarn test` in `/definitions`, which enforces the rule and is also run by `./run_def_tests.sh`.

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -2,8 +2,8 @@
 #set -o errexit
 
 cd definitions && \
-#yarn install && \
-#yarn test && \
+yarn install && \
+yarn test && \
 cd ../cli && \
 yarn && \
 yarn run flow && \


travis.sh
```diff
@@ -2,8 +2,8 @@
 #set -o errexit
 
 cd definitions && \
-#yarn install && \
-#yarn test && \
+yarn install && \
+yarn test && \
```
GAntoine: Change to `yarn run test-quick`, we don't need it to build. Also consider changing it to `yarn run lint`, if you're just looking to run eslint without running jest.

pcwinters: @GAntoine, tbh, this PR made me a little uneasy because the root of the problem is that the travis script does not run what the Contributing guide tells contributors to run (ie `run_def_tests.sh`). Why not have it do that instead? Any discrepancy between the class of errors/problems that Travis alerts contributors to and the errors/problems they find during development will cause confusion.

GAntoine: For the sake of consistency, I agree with the change and I'll merge this PR. In the future I plan on rewriting the contributing guide, and move away from `run_def_tests.sh`.

PR #1499, opened 2017-11-08
use stricter typing on ramda's propEq

LoganBarnett:
`propEq` now requires the key to be on the object supplied, and the comparison value must be the same type as the value of the property on that key.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
@@ -369,16 +369,16 @@ declare module ramda {
   }
 
   /**
-  * DONE:
-  * Function*
-  * List*
-  * Logic
-  * Math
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw:
@LoganBarnett can you speak more to the workaround with `$Ramda_ElementType`? Why do you use it sometimes and not others within the def for `PropEq`?

I ask because I'm having my own battles with `$ElementType` for other parts of Ramda like `values`.

LoganBarnett:
@AndrewSouthpaw I think I am in error not using the `$Ramda_ElementType` workaround for the other arities. My tests as they are don't cover the negative cases for other arities. I'm strengthening them now - I'll bring my findings back here and perhaps open another PR if need be. As sloppy as it might be, I just added the workaround to in a desperate attempt to make the type check pass for the negative condition.

Did you check out the ticket linked in the comment? I think the folks working on immutable.js ran into some issues similar to the ones we're having with ramda.

PR #1504, opened 2017-11-09
Add type refinement to Ramda #isNil.

AndrewSouthpaw:
The Ramda `isNil` function is a useful predicate that should refine the type of a variable. The test added in the diff did not pass originally, but does not with the updated definition.

It's not possible to provide the type refinement while still specifying a boolean literal, so something like this isn't possible:

```js
declare function isNil(x: void | null): true %checks (x === undefined || x === null);
//1:   declare function isNil(x: void | null): true %checks (x === undefined || x === null);
//                                                                              ^ boolean. Expected boolean literal `true`
//1:   declare function isNil(x: void | null): true %checks (x === undefined || x === null);
//                                             ^ boolean literal `true`
```

Having a literal type isn't important though, since usually we'd be using this for something like type refinement, so returning `boolean` seems fine here and more useful than the previous definition.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
@@ -430,8 +430,8 @@ declare module ramda {
   declare function type(x: ?any): string;
   declare function isArrayLike(x: any): boolean;
 
-  declare function isNil(x: void | null): true;
-  declare function isNil(x: mixed): false;
+  declare function isNil(x: mixed): boolean %checks(x === undefined ||
+    x === null);
 
   // *List
   declare function adjust<T>(
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/test_ramda_v0.x.x_misc.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/test_ramda_v0.x.x_misc.js
@@ -1,6 +1,15 @@
 /* @flow */
 /*eslint-disable no-undef, no-unused-vars, no-console*/
-import _, { compose, pipe, curry, filter, find, repeat, zipWith } from "ramda";
+import _, {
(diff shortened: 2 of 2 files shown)


definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
```diff
@@ -430,8 +430,8 @@ declare module ramda {
   declare function type(x: ?any): string;
   declare function isArrayLike(x: any): boolean;
 
-  declare function isNil(x: void | null): true;
-  declare function isNil(x: mixed): false;
+  declare function isNil(x: mixed): boolean %checks(x === undefined ||
```
GAntoine: Is `%` meant to be a `//`?

AndrewSouthpaw: Nope. It's a special notation used within flow to [add refinement of types](https://github.com/facebook/flow/issues/4723#issuecomment-325157852). Since we don't have the actual function definition here, we have to provide some additional hints as to what exactly it's checking. Fortunately in the case of `isNil`, that's quite simple.

GAntoine: Well TIL!

AndrewSouthpaw: Neat, right? The docs for more advanced features of flow are disappointingly sparse. 😦

GAntoine: Super neat! It's a shame about the docs, but with flow changing so often it's hard to blame the team. Heck I have a hard time documenting at work, and I'm paid to do it. Hopefully once everything settles and they release a V1, we'll get some solid docs 📖

PR #1513, opened 2017-11-11
support ramda's propEq's === nature

LoganBarnett:
I ran into some problems with the `$ElementType` utility type. `$Keys<T>` is a union of property names. I added some test cases to include objects with multiple names and the typedef we had broke.  

After exploring with a repl I've found `propEq` is totally fine with comparing values of varied types. The behavior of `propEq` is that of `===` ([example](http://ramdajs.com/repl/#?const%20foo%20%3D%20%7B%20bar%3A%20%27string%27%7D%0AR.propEq%28%27bar%27%2C%205%2C%20foo%29)), and in Flow you can compare using `===` with anything and that type checks ok ([example](https://flow.org/try/#0KwAgvBIOTFBQcDGB7AdgZwC4lQVwLbggCMSaWIWATkVAEzxkbZUCm6uANtmDgeJGpMKyTAAtWNXnkKDMVOEA)). This means we can treat the comparison value as `mixed` and we can drop the use of `$ElementType`. No more workaround needed, and the type signature more accurately reflects the runtime behavior of `propEq`.

Also we found out `propEq` supports treating `number` as a property for `Array` ([example](http://ramdajs.com/repl/#?const%20xs%20%3D%20%5B1%2C%202%2C%203%5D%0A%2F%2F%20Is%20element%201%20the%20value%20of%202%3F%0AR.propEq%281%2C%202%2C%20xs%29%20%2F%2F%20true)). This PR also adds the array support.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
@@ -1034,24 +1034,21 @@ declare module ramda {
   ): (y: A) => boolean;
   declare function eqBy<A, B>(fn: (x: A) => B, x: A, y: A): boolean;
 
-  // Workaround for $ElementType.
-  // See https://github.com/facebook/flow/issues/4804
-  declare type $Ramda_ElementType<A, B> = $ElementType<A, B>;
   // Flow cares about the order in which these appear. Generally function
   // siguatures should go from smallest arity to largest arity.
   declare type PropEq = (<T>(
     prop: $Keys<T>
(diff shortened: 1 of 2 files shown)


LoganBarnett:
@AndrewSouthpaw this might be relevant to some of your endeavors with `$ElementType` and `$Keys`. I'm happy to answer any questions you might have about it.

GAntoine:
@LoganBarnett Want me to merge this in, or wait for feedback from Andrew?

LoganBarnett:
@GAntoine sorry - go ahead and merge it. I wanted to give @AndrewSouthpaw some visibility since we have some overlap with these specific types within Ramda.

Thanks!

AndrewSouthpaw:
Thanks @LoganBarnett for mentioning it. Nice to see the library continuing to improve. 😄

PR #1520, opened 2017-11-13
add types to react relay

sibelius:


AugustinLF:
Hey @sibelius, are the createFragmentContainer types correct? Because you typed it as taking and returning TBase, which, if I'm correct, implies that their props are identical. However, for a container taking a `viewer` fragment, the parent will pass `viewer={||}`, but the "contained" will receive `viewer: {/* whatever is define in the fragment*/}`, no? I'm on flow `0.63.1`

Because it seem that I have errors on all my fragment since updating my libdefs. Since I'm in the middle of big flow version upgrade, maybe the errors are coming from somewhere else (I still have > 150 errors), so feel free to correct me, if this did work for you.

If needed, I could probably push a repro example if needed :)

sibelius:
@AugustinLF create an issue in this repo to track this

maybe @calebmer or @kassens could help on this one

AugustinLF:
@sibelius sorry, dunno why I didn't make an issue in the first place. Issue is created, with a simple code example which looks like a case I have.

PR #1532, opened 2017-11-14
prevent ramda's find from searching an object

LoganBarnett:
Ramda's documentation for [find](http://ramdajs.com/docs/#find) as well as a [quick example](http://ramdajs.com/repl/#?R.find%28o%20%3D%3E%20o%20%3D%3D%20%27bar%27%2C%20%7Bfoo%3A%20%27bar%27%7D%29) shows that it cannot meaningfully perform a search over an object. This corrects the type and adds an assertion in the test that expects an error when trying to call `find` with an object.

Ramda handles these situations gracefully but in a Flow ecosystem I don't believe it will be desired to pass objects into `find` and expect that to type check successfully.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
@@ -621,14 +621,13 @@ declare module ramda {
   ): ((xs: string) => string) & ((xs: T) => ?V);
   declare function nth<T: string>(i: number, xs: T): T;
 
-  declare function find<V, O: { [key: string]: * }, T: Array<V> | O>(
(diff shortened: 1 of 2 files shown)


LoganBarnett:
From the Travis logs:

```
30.78s$ ./travis.sh
yarn install v0.27.5
info No lockfile found.
[1/4] Resolving packages...
error An unexpected error occurred: "https://registry.yarnpkg.com/husky: ETIMEDOUT".
```

Perhaps a hiccup? Can we run it again?

Also, @GAntoine is there something I should do to address the `needs-updates`?

GAntoine:
> is there something I should do to address the needs-updates

Nah, needs-updates in my way of tracking broken PRs. In this case it's the issue with the failing specs. I'll restart the build and see if it fixes the issue.

PR #1546, opened 2017-11-17
Headers issue in node-fetch.

romeovs:
Consider the following code using `node-fetch`:

```
// @flow

import fetch from "node-fetch"

fetch("https://requestb.in/xzxek2xz", {
  headers: {
    Authorization: "Bearer foo",
  },
})
.then(function (res) {
  console.log(res)
})
```

This works fine as it should. Yet checking the flow types returns the following error:

```
example.js:6
                v
  6:   headers: {
  7:     Authorization: "Bearer foo",
  8:   },
       ^ object literal. This type is incompatible with an argument type of
 28:     headers?: HeaderInit | {
                   ^^^^^^^^^^ array type. See lib: flow-typed/npm/node-fetch_v1.x.x.js:28

example.js:6
                v
  6:   headers: {
  7:     Authorization: "Bearer foo",
  8:   },
       ^ property `Authorization`. Property not found in
 28:     headers?: HeaderInit | {
                   ^^^^^^^^^^ Headers. See lib: flow-typed/npm/node-fetch_v1.x.x.js:28


Found 2 errors
```

When looking into the `node-fetch` definition, I saw the `RequestInit` type is an interface instead of a simple type and after fooling around a bit, I concluded that was the issue.

I'm not 100% sure why it should've been an interface in the first place (it only has properties, no behaviour).

I've added the bug to the the test and fixed the definition.

Here's a [Try Flow](https://flow.org/try/#0CYUwxgNghgTiAEBLAdgFxDAZlMCBKIAjgK4gDOqAksoqvAN4BQ88AkABYhSgxkD8ALgbNWrANoUYKAOYBdIZJkAaRqwC+jDYzAB7ZBXhxC8IQRLkqNOgF5hLTtwxkhTFiwCCxVOx1SAXlCoiHpCAEQAQlxwMPCYOjqhKiwaakA) that shows the issue, and [here's one with the fix applied](https://flow.org/try/#0CYUwxgNghgTiAEAXAngBwQJRARwK4gGdEBJAOwEtF4BeeAbwCh54BIACxClBgIH4AuekxYsA2kRjlSAcwC6giVOkAaBiwC+DTQzAB7UkXhxs8QVjyESFKrUbMOXED0F3m8AIK5EbXZIBeUIjk+oIARABCnHAw8ABmurqhqsya6kA).

### DIFF
--- a/definitions/npm/node-fetch_v1.x.x/flow_v0.44.x-/node-fetch_v1.x.x.js
+++ b/definitions/npm/node-fetch_v1.x.x/flow_v0.44.x-/node-fetch_v1.x.x.js
@@ -20,11 +20,13 @@ declare module 'node-fetch' {
     size: number
   }
 
+  declare type HeaderObject = {
+    [index: string]: string
+  }
+
   declare interface RequestInit {
     method?: string,
-    headers?: HeaderInit | {
-      [index: string]: string
-    },
+    headers?: HeaderObject,
     body?: BodyInit,
     redirect?: RequestRedirect,
 
--- a/definitions/npm/node-fetch_v1.x.x/test_node-fetch_v1.js
+++ b/definitions/npm/node-fetch_v1.x.x/test_node-fetch_v1.js
@@ -10,7 +10,10 @@ import type { Readable } from 'stream';
 (nodeFetch(123): Promise<Response>);
 
 nodeFetch('foo', {
-    method: 'GET'
+    method: 'GET',
+    headers: {
+      Authorization: "Foo",
+    },
 });
 
 nodeFetch('foo', {


romeovs:
Failing tests are not caused by me. Seems like something else is going wrong.

KenanY:
This patch **is** causing the following fails:

```
 * node-fetch_v1.x.x/flow_v0.44.x- (flow-v0.52.0): Unexpected Flow errors(2):
   Error: 0-test_node-fetch_v1.js:75
    75:   (res.bodyUsed: boolean);
               ^^^^^^^^ property `bodyUsed`. Property not found in
    75:   (res.bodyUsed: boolean);
           ^^^ Response
   
   Error: 0-test_node-fetch_v1.js:76
    76:   (res.body: Readable);
               ^^^^ property `body`. Property not found in
    76:   (res.body: Readable);
           ^^^ Response
   
   Error: 0-test_node-fetch_v1.js:77
    77:   (res.text(): Promise<string>);
               ^^^^ property `text`. Property not found in
    77:   (res.text(): Promise<string>);
           ^^^ Response
   
   Error: 0-test_node-fetch_v1.js:78
    78:   (res.buffer(): Promise<Buffer>);
               ^^^^^^ property `buffer`. Property not found in
    78:   (res.buffer(): Promise<Buffer>);
           ^^^ Response
```

Do note that `Request` and `Response` **do** mixin/inherit from `Body`. See, for example: https://github.com/bitinn/node-fetch/blob/1.x/lib/request.js#L66

romeovs:
How can I reproduce these errors locally? `./run_def_tests.sh` hangs indefinitely in some jest test, ` node ./cli/ run-tests --onlyChanged` also hangs and `flow focus-check definitions/npm/node-fetch_v1.x.x/test_node-fetch_v1.js` gives me no errors:

```
[2017-11-20 10:21:53] Checking 1 files
[2017-11-20 10:21:53] executable=/usr/local/lib/node_modules/flow-bin/flow-osx-v0.59.0/flow
[2017-11-20 10:21:53] version=0.59.0
[2017-11-20 10:21:53] Parsing
[2017-11-20 10:21:55] Building package heap
[2017-11-20 10:21:55] Loading libraries
[2017-11-20 10:21:56] Resolving dependencies
[2017-11-20 10:21:57] to_merge: Focused: 1, Dependents: 0, Dependencies: 0
[2017-11-20 10:21:57] Calculating dependencies
[2017-11-20 10:21:57] Merging
[2017-11-20 10:21:57] Done
[2017-11-20 10:21:57] Checked set: Focused: 1, Dependents: 0, Dependencies: 0
```

It might be my limited knowledge of flow types, but these errors seem unrelated to my change (although they must be since this change introduced them). They seem more related to the fact that `Response` does not mixing `Body` properly. Assistance would be appreciated!

romeovs:
In any case, I've updated the PR to make the tests work but I'm pretty sure this breaks in other ways since the tests are very limited.

KenanY:
I'm running:

```
node cli/dist/cli.js run-tests node-fetch
```

romeovs:
Okay, the tests were breaking again because the `post-commit` hook altered the code in a way that breaks it.

Post-commit hooks that alter code are not really principle-of-least-surprise, so I'd do away with them (but that's not the point in this pr).

KenanY:
Ah, I see. Perhaps you can commit your changes with the `-n` flag, so that the post-commit hook is skipped?

romeovs:
Yes, did just that!

KenanY:
Awesome 👍

PR #1551, opened 2017-11-19
Add libdefs for pinojs v4

taylon:
Here is a libdef for [pinojs](https://github.com/pinojs/pino) v4.

I was confused if I should declare stuff globally using a prefix (like pino$Logger) or if I should keep it as I did declaring everything inside the `declade module` body and thus requiring a `import type` statement (which for me feels less natural). 

I saw both patterns in different libdefs so I'm not sure what is the best practice, so let me know if there is any necessary adjustment to be made and I'll do it =D


GAntoine:
@taylon Thanks for the PR! Looks like the existing definition for v3 is broken. Would you be able to take a crack at fixing it?

KenanY:
Yeah for the v3 definition I couldn't figure out why line 18 _wasn't_ throwing an error 🤷‍♂️

taylon:
I took a look at the v3 definition and what I found is that it is missing the `import type` statement.
After adding it the tests succeed in flow version 0.57 and newer, but it still fails in 0.54 to 0.56. I'm not sure if these tests ever passed but if they did I don't know why.

From: 

```javascript
import pino from "pino";

const p: pino.Logger = pino();
```

to:

```javascript
import pino from "pino";
import type { Logger } from "pino";

const p: Logger = pino();
```

Is it ok to increase the minimum flow version supported? If so I can make a commit with the changes real quick =D

Also as I mentioned previously, I'm not sure if we should be declaring the types globally or requiring the import type statement as both v3 and v4 currently are.

GAntoine:
> I found is that it is missing the import type statement.

Nice find! 👍 

> Is it ok to increase the minimum flow version supported? If so I can make a commit with the changes real quick

Absolutely! Anyone who wants to use an older version of flow can take the time to fix it later.

taylon:
Cool, just pushed the change, let me know if there is anything else =D

PR #1562, opened 2017-11-21
Make styled-components ThemeProvider theme prop covariant

JedWatson:
Object properties are invariant by default which means that we cannot pass a type like:

```js
type Theme = { stuff: { ... } };
```

into `<ThemeProvider theme={}>` because in ThemeProvider could modify our object, but it does not so making it covariant is valid.


thejameskyle:
Hmm, all the errors in here are in the styled-components 2.x definitions which this doesn't touch. I guess they need to be fixed up?

PR #1568, opened 2017-11-22
Bluebird.join: handler is optional.

StreetStrider:


StreetStrider:
@GAntoine what updates from my side are needed?

GAntoine:
@StreetStrider The current definition of bluebird is broken, and so merging this in would break the master build. I took a crack at fixing it on #1457, but couldn't get it working. You're welcome to try and fix it, I won't have the time to address it for a little while.

StreetStrider:
@GAntoine, I'll take a look at your changes.
The definitions for bluebird which I got from flow-typed repo was already broken, I've commented out `static bind` (locally). So I don't get your position on this PR, and how it increases the mess here. Does it conflict with your work?, if so, we can apply it after your PR would be merged (or apply it ontop of your branch).

villesau:
Looks like this is conflicting. Is this still valid PR @StreetStrider @GAntoine?

StreetStrider:
@villesau if it's not done in trunk, it's still actual PR, cause Bluebird allows handler to be optional.

villesau:
@StreetStrider would you mind updating the PR? The build is so old that it can't be found anymore.

StreetStrider:
@villesau I got a bunch of PRs to flow-typed, but they are stale for a long time. The maintainers basically don't care and accept only from their cohort. When I originally prepared this, everything was OK. I will not waste any time here. I recommend you take patches you need, apply them onto your local `flow-typed/` defs and commit them. That's way too faster than expect any traction here.

PR #1569, opened 2017-11-23
[rxjs] Adds types to Observable.pipe()

radex:
`Observable.pipe()` is not a very commonly used (or well documented) method, but it's crucial for users of Flow, because with Observable.pipe, it's possible to define your own custom Rx operators as functions. This way (unlike adding custom methods to Observable prototype), the whole chain can remain strongly typed.

The purpose of `rxjs$OperatorFunctionLast` is so that functional operators returning e.g. ConnectableObservable can be used with pipe without losing return type information.


radex:
Thank you!

PR #1570, opened 2017-11-23
Added typedef for xml-escape

mjomble:


KenanY:
@mjomble Can you limit the flow range to `v0.25.x-`? This patch currently fails on versions of flow older than v0.25.

mjomble:
Done, Travis build passed now.

Issue #1519, opened 2017-11-13
TypeError: Invalid Version: ow-v0.24.2


sibelius:
when running `run_def_tests.sh`

GAntoine:
Weird, I just fetched and I'm not getting that error. What's the exact command you're running?

XaveScor:
I have the some error in debian/windows10/windows10WSL.

GAntoine:
@XaveScor Could you post the error message?

XaveScor:
@GAntoine 
```
UNCAUGHT ERROR: TypeError: Invalid Version: 0.30.0.zip
    at new SemVer (C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\node_modules\semver\semver.js:2
93:11)
    at compare (C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\node_modules\semver\semver.js:566:
10)
    at Object.lt (C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\node_modules\semver\semver.js:60
0:10)
    at C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\dist\commands\runTests.js:412:27
    at Array.sort (native)
    at Array.sort (C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\node_modules\core-js\modules\es
6.array.sort.js:21:15)
    at getCachedFlowBinVersions$ (C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\dist\commands\ru
nTests.js:411:20)
    at tryCatch (C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\node_modules\regenerator-runtime\
runtime.js:65:40)
    at Generator.invoke [as _invoke] (C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\node_modules
\regenerator-runtime\runtime.js:303:22)
    at Generator.prototype.(anonymous function) [as next] (C:\Users\andrey.vassilyev\AppData\Roaming\npm\node_modules\flow-typed\node_modules\regenerator-runtime\runtime.js:117:21)
```

XaveScor:
```
UNCAUGHT ERROR: TypeError: Invalid Version: ow-v0.36.0
cli.js:58
[...]

PR #1578, opened 2017-11-25
remove call native `unzip` on non-windows

XaveScor:
If user hasn't `unzip` util on Linux or MacOS then flow-typed run-tests failed. I replace native call to unzip module.
I filter files in bin dir, if we have trash in bin folder.
#1519 should be solved

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -99,10 +99,6 @@ async function getOrderedFlowBinVersions(): Promise<Array<string>> {
       const QUERY_PAGE_SIZE = 100;
       const OS_ARCH_FILTER_RE = new RegExp(BIN_PLATFORM);
 
-      if (!await fs.exists(BIN_DIR)) {
-        await fs.mkdir(BIN_DIR);
-      }
-
       let binURLs = new Map();
       let apiPayload = null;
       let page = 0;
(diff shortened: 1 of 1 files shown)


XaveScor:
@GAntoine here?)

GAntoine:
This one I'll need a little time to comb over, but I don't expect there to be any issues.

cli/src/commands/runTests.js
```diff
@@ -260,15 +240,24 @@ async function getOrderedFlowBinVersions(): Promise<Array<string>> {
...
  * Return the sorted list of cached flow binaries that have previously been retrieved from github
  * and cached in the `.flow-bins-cache` directory.  This function is usually called when a failure
  * has occurred when attempting to refresh the flow releases from github, i.e. offline or over
  * API limit.
  */
 async function getCachedFlowBinVersions(): Promise<Array<string>> {
-  // read the files from the bin dir and remove the leading `flow-v` prefix
-  const versions = (await fs.readdir(path.join(BIN_DIR))).map(dir =>
+  // read the files with name `flow-vx.x.x` from the bin dir and remove the leading `flow-v` prefix
+  const versions = (await fs.readdir(path.join(BIN_DIR))).filter(checkFlowFilename).map(dir =>
```
GAntoine: What kind of trash are you trying to filter out? As in, which files are causing issues?

cli/src/commands/runTests.js
```diff
@@ -99,10 +99,6 @@ async function getOrderedFlowBinVersions(): Promise<Array<string>> {
       const QUERY_PAGE_SIZE = 100;
       const OS_ARCH_FILTER_RE = new RegExp(BIN_PLATFORM);
 
-      if (!await fs.exists(BIN_DIR)) {
```
GAntoine: ~What's the reason for removing this check?~ Nevermind, I see you moved it below.

cli/src/commands/runTests.js
```diff
@@ -208,38 +204,22 @@ async function getOrderedFlowBinVersions(): Promise<Array<string>> {
...
-            await new Promise((res, rej) => {
-              const unzipExtractor = unzip.Extract({ path: flowBinDirPath });
-              unzipExtractor.on("error", function(err) {
-                rej(err);
-              });
-              unzipExtractor.on("close", function() {
-                res();
-              });
-              fs.createReadStream(zipPath).pipe(unzipExtractor);
+          await new Promise((res, rej) => {
```
GAntoine: This change makes sense 👍

GAntoine:
@XaveScor Sorry for taking so long to get to this, did a review and I have a couple questions.

cli/src/commands/runTests.js
XaveScor: If `flow-typed run-tests` crushed BIN_DIR can contain 'garbage files'(such as flow-0.x.x.zip, etc). Without filter `runTests.js` tries check semver and failed. Semver returns undefined.

PR #1593, opened 2017-11-28
Export `StyleDefinition` type from aphrodite

jakewtaylor:
Make the `StyleDefinition` type an export from the Aphrodite definitions.
I needed to use `StyleDefinition` in order for a component to receive custom classes:

```JSX
<MyComponent
  classes={[styles.class1, styles.class2]}
/>
```

```JSX
import { css } from 'aphrodite';
import type { StyleDefinition } from 'aphrodite';

type Props = {
  classes: Array<StyleDefinition>,
};

class MyComponent extends React.PureComponent<Props> {
  render () {
    return (
      <p className={css(styles.mainClass, ...this.props.classes)}>
        Hello, world!
      </p>
    );
  }
}
```

### DIFF
--- a/definitions/npm/aphrodite_v0.5.x/flow_v0.28.x-/aphrodite_v0.5.x.js
+++ b/definitions/npm/aphrodite_v0.5.x/flow_v0.28.x-/aphrodite_v0.5.x.js
@@ -11,7 +11,7 @@ declare module "aphrodite" {
     [key: string]: Object
   };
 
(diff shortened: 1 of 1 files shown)


jakewtaylor:
Will need guidance on how to fix the build failure - it looks unrelated to my changes?

GAntoine:
> it looks unrelated to my changes?

Correct, I've restarted the build and will merge this in once it passes.

PR #1595, opened 2017-11-28
(Ramda) Add missing `UnaryFn<C, D>` in `Pipe` definition

mobily:
There was a missing UnaryFn argument that prevented types to flow correctly.


mobily:
Thanks! :)

PR #1597, opened 2017-11-28
add express.json() middleware support (requires 4.16.x)

LoganBarnett:
This adds the ability to use `express.json()` per the [documentation](http://expressjs.com/en/4x/api.html#express.json). This was introduced in 4.16.0 so there's a new version section for express that covers 4.16.x and above.

### DIFF
--- a/definitions/npm/express_v4.16.x/flow_v0.32.x-/express_v4.16.x.js
+++ b/definitions/npm/express_v4.16.x/flow_v0.32.x-/express_v4.16.x.js
@@ -1,0 +1,274 @@
+import type { Server } from "http";
+import type { Socket } from "net";
+
+declare type express$RouterOptions = {
+  caseSensitive?: boolean,
+  mergeParams?: boolean,
+  strict?: boolean
+};
+
+declare class express$RequestResponseBase {
+  app: express$Application;
+  get(field: string): string | void;
+}
+
+declare type express$RequestParams = {
+  [param: string]: string
+};
+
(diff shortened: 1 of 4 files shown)


LoganBarnett:
Is there some pre-commit hook stuff I should be looking at here? I ran the tests, they passed, and then I formed up the PR. Once I saw this error, I ran them again locally and got the same error the CI server is seeing.

LoganBarnett:
Ok, it looks like the precommit hook was removing `mixins` from the definition. I've forced a commit without the hook. I expect this run will pass.

GAntoine:
@LoganBarnett I'm concerned about how nicely this will play with the current definition of `4.x.x`...

LoganBarnett:
@GAntoine Is this in reference to the 4.x.x vs 4.16.x versions? If so, I was mistaken in not including some research I'd done before putting this together.

Here's what I've found: #322 covers functionality that basically means `flow-typed` should use the highest matching version when providing lib defs. @jeffmo's [comment](https://github.com/flowtype/flow-typed/issues/322#issuecomment-249053787) goes over the scenarios, and then this was implemented in #326.

My understanding here is if you are you using 4.15.x or lower, it will use the 4.x.x defs. If it's 4.16.x or higher, you get the 4.16.x defs.

If that doesn't address your concerns or I'm just off base in what your concern is to begin with, I'm happy to address it :)

GAntoine:
Fantastic, that's just the info I was looking for! Accepted and merged, great work.

PR #1608, opened 2017-12-02
Add type definition for joi-to-json-schema v3.2.x

EdwardDrapkin:
Hey there!

I'm back with another lib def, this one is real small.

I couldn't type hint the first arg which should be `npm$joi$ObjectSchema` but the joi libdef is not in scope and it doesn't seem there's any way to get it added to the test that I can find.  Let me know if there's something I can do to get the Joi type definition in scope in this libdef.

Thanks!


GAntoine:
Great PRs, keep-em coming!

PR #1609, opened 2017-12-02
Libdefs for angular $scope

villesau:
Fixes this: https://github.com/flowtype/flow-typed/issues/1580

ping @LoganBarnett 

Only $scope types added, forEach parameters named.

here is documentation with type hints for angular scope: https://docs.angularjs.org/api/ng/type/$rootScope.Scope


definitions/npm/angular_v1.4.7/flow_v0.47.x-/angular_v1.4.7.js
```diff
@@ -0,0 +1,220 @@
...
+    url: string,
+    defaultParams?: Object,
+    actions?: Object,
+    options?: Object,
+  ): AngularResource
+
+  declare function AngularCompileService(
+    a: JqliteElement
+  ): JqliteElement
+
```
villesau: My changes below

LoganBarnett:
Solid! @GAntoine I'm happy with this if you are.

@villesau thanks for the improvements!

PR #1611, opened 2017-12-03
express v4.16.x: Fix code coverage issue by putting type imports inside module declaration

forticulous:
Since flow 0.59 there has been a coverage issue with express `app.listen` method. I've seen issues like this before where the solution was to make sure the import type statements were inside the module declaration, so what's what i've done here


forticulous:
I'll reopen this when i've looked at it some more

PR #1612, opened 2017-12-03
Better angular 1.4.7 types

villesau:
Plenty of enhancements for angular 1.x libdefs. These are the types we currently use at Smartly for angular. So not perfect, but better. Some utility function types missing etc.
Some of the commits have link to appropriate angular documentation to make reviewing slightly easier.

ping @LoganBarnett

### DIFF
--- a/definitions/npm/angular_v1.4.7/flow_v0.47.x-/angular_v1.4.7.js
+++ b/definitions/npm/angular_v1.4.7/flow_v0.47.x-/angular_v1.4.7.js
@@ -119,7 +119,7 @@ declare module angular {
 
   declare function module(
     name: string,
-    deps: ?Array<Dependency>
+    deps?: ?Array<Dependency>
   ): AngularModule
 
   declare function element(html: string | Element | Document): JqliteElement
--- a/definitions/npm/angular_v1.4.7/flow_v0.47.x-/angular_v1.4.7.js
+++ b/definitions/npm/angular_v1.4.7/flow_v0.47.x-/angular_v1.4.7.js
@@ -94,7 +94,7 @@ declare module angular {
 
   declare type ValueDeclaration = (
     name: string,
-    di: $npm$angular$DependencyInjection<(...a: Array<mixed>) => Object>,
+    value: mixed,
   ) => AngularModule
 
   declare type ConstantDeclaration = (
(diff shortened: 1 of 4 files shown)


definitions/npm/angular_v1.4.7/flow_v0.47.x-/angular_v1.4.7.js
```diff
@@ -7,9 +7,16 @@
...
+  declare type ScopeBindings = '<' | '=' | '&' | '<?' | '=?' | '&?' | '=ngModel';
+  declare type Scope = {[key: string]: ScopeBindings};
+  declare type ControllerFunction = (...a: Array<*>) => void;
+
   // I'm not sure how represent this properly: Angular DI declarations are a
   // list of strings with a single function at the end. The function can vary,
   // so it is a type param.
+  //
+  // NOTE: if you use compile step to mangle array, replace below with
+  // declare type $npm$angular$DependencyInjection<T> = T
```
villesau: This is something i would like to be able to handle in better way than just writing a comment here. We don't use array dependency injection syntax with angular, and i think it's pretty common approach.

definitions/npm/angular_v1.4.7/flow_v0.47.x-/angular_v1.4.7.js
```diff
@@ -22,12 +29,12 @@ declare module angular {
...
 
-  declare function AngularLinkFunction(
-    scope: mixed,
+  declare type AngularLinkFunction = (
+    scope: $Scope<*>,
     element: JqliteElement,
     attrs: mixed,
     controller: mixed
-  ): void
+  ) => void
```
villesau: any idea why flow couldn't find `$Scope<*>` when AngularLinkFunction was declared as `declare function`? And whats the actual difference with `declare function fn():void` and `declare type fn = () => void`?

LoganBarnett: I've found it's easier to declare overrides with `declare type` and just intersection them with `&`. Could that be related to what you're looking at?

villesau: Seems that with `declare function` ordering matters, but not with `declare type fn = () => {}`. Do you know if there are any other differences? One what i'm really curious about is that is it somehow possible to type function `this`. Naturally that's not possible with arrow function, but with regular function libdefs that would make much sense.

GAntoine:
@villesau Ping me when these are ready to merge.

villesau:
@GAntoine added type for angular $timeout. Good to go from my perspective.

PR #1613, opened 2017-12-03
make changed libdef message welcoming to contribute changes back

villesau:
Many projects might have enhanced libdefs which they are jsut too lazy to contribute back, or then they don't exactly know how to do it. Suggest contributing changed libdefs back to flow-typed when flow detected that libdef is changed.

precommit-pretteir did it's thing, I just changed the message.

### DIFF
--- a/cli/src/commands/install.js
+++ b/cli/src/commands/install.js
@@ -1,46 +1,46 @@
 // @flow
 
-import {signCodeStream} from '../lib/codeSign';
+import { signCodeStream } from "../lib/codeSign";
 
-import {copyFile, mkdirp} from '../lib/fileUtils';
+import { copyFile, mkdirp } from "../lib/fileUtils";
 
-import {findFlowRoot} from '../lib/flowProjectUtils';
+import { findFlowRoot } from "../lib/flowProjectUtils";
 
 import {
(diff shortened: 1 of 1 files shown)


cli/src/commands/install.js
```diff
@@ -417,35 +417,39 @@ async function installNpmLibDef(
...
+        "  • %s\n" + "    %s\n    %s\n    └> %s",
         colors.bold(
           colors.red(
             `${terseFilePath} already exists and appears to have been manually ` +
-              `written or changed!`,
-          ),
+              `written or changed!`
+          )
+        ),
+        colors.green(
```
villesau: My changes are this and next line. I think flow-typed prettier is not maybe in best possible shape :D

cli/src/commands/install.js
```diff
@@ -417,35 +417,39 @@ async function installNpmLibDef(
...
+          `Consider contributing your changes back to flow-typed repository :)`
         ),
-        'Use --overwrite to overwrite the existing libdef.',
+        `Read more from https://github.com/flowtype/flow-typed/wiki/Contributing-Library-Definitions`,
```
GAntoine: `from` -> `at`

PR #1614, opened 2017-12-04
Added typedef for icepick

tompuric:


tompuric:
Checked the logs and I don't see how my changes could have caused this issue. This seems to also be an issue in the master builds looking at its history.

```
Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
```

Let me know if there is anything I can do, thanks.

GAntoine:
Yeah it's a timeout issue with that specific spec, I have a mind to fix it someday... Anyways awesome PR, keep 'em coming!

PR #1615, opened 2017-12-04
include $ReadOnlyArray in the array types that reduce accepts

LoganBarnett:
Prompted by a discussion [here](https://github.com/facebook/flow/issues/5457), I came across some issues with feeding `$ReadOnlyArray` to `reduce`. Since Ramda assumes immutability, I've added support of `$ReadOnlyArray` to `reduce`.

Side effects of this PR include:
1. Renaming output variables of `reduceRight` and `scan` to not overlap with `reduce`'s output variables.
2. Adding a negative test for `reduce`.
3. Using an example from the documentation for `reduce`.
4. Using the `declare type Foo = curried1_1_1 & curried1_2 & curried2_1 & curried_3 ...` form over `declare function`.
5. Removing the rest args as a means of ignoring unused arguments in `reduce`.

CC @AndrewSouthpaw due to recent work with Ramda's libdefs.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
@@ -976,21 +976,21 @@ declare module ramda {
 
   declare function reverse<T, V: Array<T> | string>(xs: V): V;
 
-  declare function reduce<A, B>(
-    fn: (acc: A, elem: B) => A,
-    ...rest: Array<void>
-  ): ((init: A, xs: Array<B>) => A) &
-    ((init: A, ...rest: Array<void>) => (xs: Array<B>) => A);
-  declare function reduce<A, B>(
-    fn: (acc: A, elem: B) => A,
-    init: A,
-    ...rest: Array<void>
-  ): (xs: Array<B>) => A;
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw:
Looks good to me. 

- What's the advantage of `declare type` over `declare function`?

In the tests could you also add one where the function changes the type from the `xs`? Something like `reduce((acc, x) => acc + parseInt(x), 0, ['1', '2', '3'])`? Based on the type definition it should work, but wouldn't hurt to add the test to make sure we maintain that flexibility in the def.

LoganBarnett:
I've found `declare type foo: Foo` to be better for declaring lots of overrides (which Ramda is full of due to being curried). Maybe one thing we should try is actually using `CurriedFunction` for these and see if we can just provide one type signature per function. We'd still need the `declare type` form for that though.

re: type change - good suggestion! I'll add that today.

GAntoine:
@LoganBarnett With Andrew's LGTM, let me know when this is ready to be merged in!

LoganBarnett:
TIL "LGTM" doesn't mean "Let's Get This Merged" ¯\_(ツ)_/¯ I've only seen it on PRs!

This adds the test requested. Switching to full `CurriedFunction` mode is a battle for another day.

PR #1628, opened 2017-12-08
Adjust R.reduce to accept objects.

AndrewSouthpaw:
Per the docs.

@LoganBarnett

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
@@ -978,17 +978,17 @@ declare module ramda {
 
   declare type Reduce = (<A, B>(
     fn: (acc: A, elm: B) => A
-  ) => ((init: A) => (xs: Array<B> | $ReadOnlyArray<B>) => A) &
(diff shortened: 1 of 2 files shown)


LoganBarnett:
Hrm, objects are iterable, yeah? When I try it in the repl this is what I get:
```
reduce: list must be array or iterable
```

[Here's what I tried using](http://ramdajs.com/repl/?v=0.25.0#?R.reduce%28%28acc%2C%20i%29%20%3D%3E%20%7B%0A%20%20console.log%28acc%29%0A%20%20return%201%0A%7D%2C%200%2C%20%7Bfoo%3A%202%7D%29):

``` javascript
R.reduce((acc, i) => {
  console.log(acc)
  return 1
}, 0, {foo: 2})
```

I believe that example matches the proposed type def. Am I doing something wrong here?

AndrewSouthpaw:
Ugh. That's embarrassing. Somewhere in my work yesterday evening it appeared I was using an object in Ramda's reduce, and adding that line fixed the issue. 😅 You're totally right, reduce does not work on objects, thanks for catching it.

LoganBarnett:
It's all good :) I know I've had to do some day-after fixes for at least one of my PRs here because of something simple I missed. When I saw the your PR I was kicking myself because I know Ramda supports objects and arrays in some places and maybe I'd missed `reduce`! It could have easily gone either way.

PR #1635, opened 2017-12-11
WIP: flow-types for bem-cn

s-diez:
I have tried to implement the flow types for https://github.com/albburtsev/bem-cn.
One part does not work as expected, any help would be appreciated as this is my first module definition.

Faulty part:
When intersecting the "ToStringFunction" with the callable object "InnerFunction" all checks for the additional properties of the callable object are ignored. Therefore line 34, 36 etc. do not throw errors.


villesau:
Any progress with this @s-diez ? :)

s-diez:
> Any progress with this @s-diez ? :)

Hi sorry I did not really attempt to continue with this. I'll try to test it with a newer flow version soon.

s-diez:
@villesau I did a quick test and the problem seems to be the same as before. I don't really know whats wrong here. As there seems to be no other interest in the definitions of bem_cn, I guess I'll close it next if no one else comments.

villesau:
@s-diez Shouldn't `toString` return  only a string? thus `toString: () => string` should be enough?

s-diez:
@villesau You're right it should be `toString: typeof ToStringFunction`.
`() => string` or `declare function ToStringFunction(): string;` should be the same or am I missing something? When testing it makes no difference.
But the main problem described in the first comment remains:

> When intersecting the "ToStringFunction" with the callable object "InnerFunction" all checks for the additional properties of the callable object are ignored. Therefore line 34, 36 etc. do not throw errors.

No matter how I declare it, which should all result in the same anyway, as I understand it:
- `declare export type BemFunction = InnerFunction & () => string;`
- `declare export type BemFunction = InnerFunction & typeof ToStringFunction;`
- `declare export type BemFunction = InnerFunction & ToStringFunction;` and `declare type ToStringFunction = () => string;`

s-diez:
Closed because of lack of interest.

Issue #1639, opened 2017-12-11
Most Lodash functions accept null & undefined 


villesau:
Lodash libdefs does not always accept null or undefined when Lodash most of the time is totally fine with them. I made a small script to skim trough Lodash and check out which ones does _not_ accept null or undefined and there were only handful of them. here is the script i used on flow documentation page:

```
_.map(_, (value, key) => {
  if (typeof value !== 'function') {
    return;
  }
  try {
    value(..._.times(value.length, () => null));
  } catch (e) {
    return key;
  }
}).filter(v => v);

```
and here are the actual functions which failed when passing nulls:

```
[
  "after",
  "ary",
  "before",
  "bind",
  "curry",
  "curryRight",
  "debounce",
  "defer",
  "delay",
  "flip",
  "memoize",
  "negate",
  "once",
  "partial",
  "partialRight",
  "rearg",
  "rest",
  "spread",
  "tap",
  "throttle",
  "thru",
  "unary"
]
```

Are Lodash libdefs strict on nulls by design, or is it just inaccuracy? All the other functions returned something more or less meaningful when passing in nulls: null, undefined, empty array or object. Some returned Window or -1 etc. Return values vary a bit depending if passing in null or undefined

villesau:
And below is what lodash returned with this script for each function when passing desired amount of nulls. 

Script:

```
JSON.stringify(_.map(_, (value, key) => {
  if (typeof value !== 'function') {
    return;
  }
  try {
    return {
      v: value(..._.times(value.length, () => null)),
      k: key
    };
  } catch (e) {
    //return key;
  }
}).filter(v => v && !(v.v instanceof Window)), null, 2);
```

result:

```
[
  {
    "v": {},
    "k": "assign"
  },
  {
    "v": {},
    "k": "assignIn"
  },
  {
    "v": {},
    "k": "assignInWith"
  },
  {
    "v": {},
    "k": "assignWith"
  },
  {
    "v": [],
    "k": "at"
  },
  {
    "k": "bindAll"
  },
  {
    "k": "bindKey"
  },
  {
    "v": [],
    "k": "castArray"
  },
  {
    "v": null,
    "k": "chain"
  },
  {
    "v": [],
    "k": "chunk"
  },
  {
    "v": [],
    "k": "compact"
  },
  {
    "v": [],
    "k": "concat"
  },
  {
    "k": "cond"
  },
  {
    "k": "conforms"
  },
  {
    "k": "constant"
  },
  {
[...]

PR #1643, opened 2017-12-12
R.indexOf accepts void type.

AndrewSouthpaw:
Although undocumented, `R.indexOf` accepts `null` and `undefined` as search terms. It is actually useful to check for a void type inside an array, so I think loosening the definition in this case makes sense.

@LoganBarnett

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/ramda_v0.x.x.js
@@ -640,9 +640,9 @@ declare module ramda {
     input: A
   ): R;
 
-  declare function indexOf<E>(x: E, xs: Array<E>): number;
+  declare function indexOf<E>(x: ?E, xs: Array<E>): number;
   declare function indexOf<E>(
-    x: E,
+    x: ?E,
     ...rest: Array<void>
   ): (xs: Array<E>) => number;
 
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/test_ramda_v0.x.x_list.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-/test_ramda_v0.x.x_list.js
@@ -118,7 +118,6 @@ const str: string = "hello world";
 
   const ind: number = _.indexOf(1, ns);
   const ind1: number = _.indexOf(str)(ss);
-
   const ind2: { [key: string]: { [k: string]: number | string } } = _.indexBy(
     x => "s",
     os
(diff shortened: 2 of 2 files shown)


LoganBarnett:
Hmm. It's worth noting that Ramda has defensive checks in a lot of places (it's built without a type system like Flow in mind, so it kind of has to). With that mindset, this catches my eye:

``` javascript
declare function indexOf<E>(x: E, xs: Array<E>): number;
```

Here if we say `E` is `string` then that means it's not possible for anything in that array to be `void`. So providing some `void` value for `x` would always return false. If the `E` is `string?` then automatically the definition above will allow `x` to be a `void` value which has the potential to be true.

This is kind of a design decision where the valid data sets we can compare against don't provide a clear answer. It's kind of on us as maintainers to figure out what kind of strict checking we want.

In this case `indexOf` can provide an answer but at type check time we can also kind of know an answer (`void` on a non-`void` array will always be `-1`). Just spitballing here, but maybe a test we can ask ourselves is do we want to allow knowable comparisons like that? I think the answer is no, but I'm not sure how convicted I feel about it. @AndrewSouthpaw How do you feel about it? Would you rather passing a `void` value check as an error? If we say it's an error, and you legitimately have `void` you'd need to unwrap it like you did any other maybe type before using it. That would be the cost. The gain would be if you accidentally pass a `void` value in there, Flow will catch it and tell you.

GAntoine:
@AndrewSouthpaw Good to merge?

AndrewSouthpaw:
Sorry, I was out for a couple weeks.

I've read through your point a couple times, @LoganBarnett, and I think I understand what you're getting at, but in case I miss it feel free to reiterate.

I hear what you're saying about the tradeoff, and the value of throwing an error when accidentally passing a `void` in for the value, and the question of whether we want Flow to be really intelligent about comparisons (e.g. `void` in a non-`void` array is -1). I tend to agree, which is why I didn't go full out and write something like

```js
declare function indexOf<E>(
  & (x: E, xs: Array<E>): number
  & (x: ?E, xs: Array<E>): -1
)
```

I think there are legitimate cases where you should allow a void to be passed in rather than creating a special typing case. Common case would be lookup functionality, e.g. you have some `maybe` search term and its looking in a list of terms. 

Of course, the same could be said for a lot of the definitions. It's similar to #1639, there's lots of Ramda functions that don't error on a `void` value I'm sure, and do we want to make all those flexible? Probably not. So the question is, if we draw a line in the sand, where do we draw it?

I'd say the line should be drawn where it's useful, as justified by use cases. For example, `R.max` technically allows voids: 

```js
R.max(1, undefined) // 1
```

But that seems a very unlikely use case. I'd say, by contrast, that using a `null` search term is not terribly uncommon and should be allowed. BUT, as before, we don't want to make Flow libdefs *too* intelligent, so Flow typing `(x: ?E, xs: Array<E>): -1` seems excessive, and it's better to just allow a `maybe` type.

Thoughts?

LoganBarnett:
@AndrewSouthpaw whoa, glad to hear you're ok!

The term "search term" you used I think has me swayed here. You're searching for something, and it might be possible to not have that thing? Would the "Flow keeping me honest" part mean before you search you check to see if you have a valid search term? I'm not in the trenches where you're using `indexOf`. I think that's where it's really going to shine.

@GAntoine this looks good to me! Let's get it merged. This is more art than science at this point and if we need to revisit it, it's not like there's a lot of friction here.

PR #1644, opened 2017-12-13
Mysql: query values can be object

jaredh159:
Libdef for `mysql` was not correct, it gave an error for one of the documented examples in the README.md:

https://github.com/mysqljs/mysql#escaping-query-values

The diff for the test file is mostly prettier noise, but I'll leave a comment highlighting the two lines I added testing the exact example from the readme.

### DIFF
--- a/definitions/npm/mysql_v2.x.x/test_mysql_v2.x.x.js
+++ b/definitions/npm/mysql_v2.x.x/test_mysql_v2.x.x.js
@@ -1,21 +1,21 @@
 // @flow
 
-import mysql from 'mysql';
-import type { Connection, Pool } from 'mysql';
+import mysql from "mysql";
+import type { Connection, Pool } from "mysql";
 
 // Connection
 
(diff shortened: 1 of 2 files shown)


definitions/npm/mysql_v2.x.x/test_mysql_v2.x.x.js
```diff
@@ -1,139 +1,143 @@
...
+connection.query("SELECT 1");
+connection.query("SELECT * FROM `table` WHERE id = ?", [1]);
+connection.query({
+  sql: "SELECT * FROM `table` WHERE id = ?",
+  nestTables: true
+});
+connection.query("", [1, "d", { a: 1 }, [{ b: 1 }]]);
+
+let post = { id: 1, title: "Hello MySQL" };
+connection.query("INSERT INTO posts SET ?", post);
```
jaredh159: The above _two lines_ were the only ones I added in the test file, the rest is `prettier` auto-formatting from the commit hook.

PR #1647, opened 2017-12-13
Awilix fix

EdwardDrapkin:
There's a bug in how I was defining a few things in this awilix libdef that I've since fixed.  Thanks for your time!

### DIFF
--- a/definitions/npm/awilix_v2.x.x/flow_v0.56.x-/awilix_v2.x.x.js
+++ b/definitions/npm/awilix_v2.x.x/flow_v0.56.x-/awilix_v2.x.x.js
@@ -1,5 +1,5 @@
 declare module "awilix" {
-  declare module .exports: {
+  declare module.exports: {
     createContainer(?awilix$ContainerOptions): awilix$Container<{}>,
     ResolutionMode: Class<awilix$ResolutionMode>,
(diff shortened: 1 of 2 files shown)


EdwardDrapkin:
Whoops, my bad.  Good news though, the fix is in! (see what I did there)

EdwardDrapkin:
Sorry about those last two commits, I guess Flow doesn't seem to capture strings as literals when they're parameterized. :(

So it looks like there's no way to reliably model this library completely.

container.register('foo', asValue(123)): flow sees "string" and not "'foo'" :(

definitions/npm/awilix_v2.x.x/flow_v0.56.x-/awilix_v2.x.x.js
```diff
@@ -1,5 +1,5 @@
 declare module "awilix" {
-  declare module.exports: {
+  declare module .exports: {
```
GAntoine: Pretty sure this is a typo

EdwardDrapkin: Jetbrains does that automatically and they haven't fixed it yet, edited and pushed, sorry

PR #1649, opened 2017-12-14
Use `declare module.exports` syntax instead of deprecated `declare var exports`

samwgoldman:
In [this commit](https://github.com/facebook/flow/commit/1c91714e54893ab610ab02c45220e73376df90f4), Flow added a new syntax to define a CommonJS modules `module.exports` type. Instead of writing `declare var exports: T`, you can write `declare module.exports: T`.

This new syntax was released in Flow v0.25, nearly 2 years ago. Our intention was to eventually remove the special-case-y support for `declare var exports` within a few versions, but late is better than never.

This PR does a two things:

1. Change all definitions to have a minimum version of 0.25. There was a single library definition, `mori`, which only supported Flow versions less than 0.25. I tried that libdef on versions greater than 0.25, but there were errors so I dropped it.

2. Change all occurrences of `declare var exports` and `declare function exports` with the new syntax. There were no occurrences of `declare class exports`.


samwgoldman:
I ran a bunch of them locally, yeah. Also, this change should affect any typing rules.

samwgoldman:
Also, thanks!

PR #1650, opened 2017-12-15
Update test_react-dnd-v2.js

wmonk:


wmonk:
Any idea why this isn't working? I want to add a test to demostrate that the `react-dnd` typings don't work correctly for the `DragDropContext` function.

GAntoine:
@wmonk It's unrelated to your change, if specs are succeeding on your end I'll merge this in.

PR #1666, opened 2017-12-19
react-dnd definition: Add DragDropContextProvider

B-Reif:
Adds a definition for react-dnd's DragDropContextProvider component.

Here's the component's docs:
http://react-dnd.github.io/react-dnd/docs-drag-drop-context-provider.html

And its source:
https://github.com/react-dnd/react-dnd/blob/master/packages/react-dnd/src/DragDropContextProvider.js

The prop typing mirrors the source's PropTypes. Let me know if anything is missing.

### DIFF
--- a/definitions/npm/react-dnd_v2.x.x/flow_v0.53.x-/react-dnd_v2.x.x.js
+++ b/definitions/npm/react-dnd_v2.x.x/flow_v0.53.x-/react-dnd_v2.x.x.js
@@ -198,6 +198,16 @@ declare module "react-dnd" {
   // Drag Drop Context
   // ----------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


AndrewSouthpaw:
Probably fell through the cracks. :-) You'll need to fix the CI errors before it can be merged.

janpaul123:
Maybe rebase so the travis build runs again (right now it's a dead link)? 😄

AndrewSouthpaw:
Yeah, this PR is from before the org change, so it'll need to be rerun. A rebase would suffice.

villesau:
This now exists in the react-dnd_v2.x.x.js, thanks for the initial work @B-Reif !

PR #1674, opened 2017-12-21
yamljs: create definition

forticulous:
Initial incomplete definition for yamljs, adds definitions for parse and stringify


bwolfe:
This breaks 90% of use cases.  What about load? register? dump? parseFile?

PR #1678, opened 2017-12-22
Add load method to yamljs

bwolfe:
Using "load" is a very common use-case and at the very least is needed in here to stop compile errors.


GAntoine:
Thanks for this, I saw your comment in the other PR. If you want to open an issue to populate this definition, that would be much appreciated.

bwolfe:
Opened ticket #1680

PR #1679, opened 2017-12-23
Add definitions for material-ui (v1-beta)

rsolomon:
https://github.com/mui-org/material-ui/pull/9453 removed Flow type definitions from all components in the `material-ui@v1.0.0-betax` library. This pulls definitions as they existed (more or less) from https://github.com/mui-org/material-ui/pull/9453 and places them into a version flow-typed definition file.

I know the tests are sparse, but this is mostly not new code. It's also massive, and it took a very long time to extract all of the definitions. It would be great if others could contribute to fix any weakly typed areas (`Object` and `Function` abound), and to add tests.


oliviertassinari:
@rsolomon Thanks so much for working on it. We will definitely document the new definitions on Material-UI side once this PR is merged.

PR #1681, opened 2017-12-23
Allow tuple types for lodash map

villesau:
Lodash map allows mapping over tuples.


villesau:
Oh, prettier..

villesau:
I think it's not possible to save the tuple type over to maps return value. Though, I'm glad to be proven wrong!

PR #1685, opened 2017-12-28
Swap order of lodash flowRight definitions

RobinClowers:
This works around this issue: https://github.com/facebook/flow/issues/2360.


RobinClowers:
Scratch that, it doesn't actually fix the issue.

PR #1686, opened 2017-12-29
Add definition for jest.restoreAllMocks().

harrykao:
https://facebook.github.io/jest/docs/en/jest-object.html#jestrestoreallmocks


silvenon:
Awesome!

PR #1687, opened 2017-12-29
Fix redux-devtools-extension_v2.x.x.js

0xcaff:
`devToolsEnhancer`'s options argument is optional. See here for reference: https://github.com/zalmoxisus/redux-devtools-extension#13-use-redux-devtools-extension-package-from-npm


GAntoine:
@0xcaff Looks like there's a failing test: `Error: 0-test_redux-devtools-extension_v2.x.x.js:84`

0xcaff:
@GAntoine Sorry about that. It works now.

PR #1692, opened 2017-12-31
Add styled-normalize definitions

sergeysova:


definitions/npm/styled-normalize_v2.2.x/flow_v0.59.x-/styled-normalize_v2.2.x.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+import type { TaggedTemplateLiteral, Interpolation } from "styled-components";
```
GAntoine: This will fail here, declare it inside `declare module "styled-normalize"`

sergeysova: I can't define it types inside styled-normalize. Because in real project, flow see difference. And on each update of styled-components, I should update styled-normalize.

PR #1697, opened 2018-01-04
React-test-renderer definitions added.

alpcanaydin:
Type definitions for react-test-renderer v16.x.x

I ported it from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-test-renderer/index.d.ts


gre:
This definition is not correct.

```diff
+  declare function create(
 +    nextElement: React$ElementType,
 +    options?: TestRendererOptions
 +  ): ReactTestRenderer;
```

should be

```
+  declare function create(
 +    nextElement: React$Element<any>,
 +    options?: TestRendererOptions
 +  ): ReactTestRenderer;
```


test uses

```diff
+const Example = () => <div>Example</div>;
+const wrapper = renderer.create(Example);
```

but it should be


```diff
+const Example = <div>Example</div>;
+const wrapper = renderer.create(Example);
```

create takes an element, not a component. https://www.npmjs.com/package/react-test-renderer

romanlv:
It's missing definitions for `react-test-renderer/shallow`, [usage](https://reactjs.org/docs/shallow-renderer.html)

PR #1703, opened 2018-01-06
Fix Number/number definition in promise-retry

crccheck:
Like most of JavaScript, `number` is a literal value and not the wrapper `Number`.

Some formatting changes came in from husky/prettier.


crccheck:
I give up

PR #1706, opened 2018-01-08
Add _.pick() and _.isEmpty() array argument types

andrewrota:
Modifications to the underscore library definition so that `_.isEmpty()` supports an array, and `_.pick()` supports an array of keys.


matthugs:
I think some of these updates may be in error; here are some amendments https://github.com/matthugs/flow-typed/tree/underscore_updates

All the test failures appear to be attributable to the prettier commit; @andrewrota didn't touch the corresponding portions of the libdef.

GAntoine:
@matthugs Would you be interested in opening a up-to-date PR with your amendments? I'll be closing this one since it's stagnant and incorrect.

PR #1707, opened 2018-01-08
add select() and put.resolve() to redux-saga definitions

chrisnojima:
Hi,

This adds support for https://redux-saga.js.org/docs/api/#selectselector-args
```
yield select()
```
and https://redux-saga.js.org/docs/api/#putresolveaction
```
yield put.resolve({type: 'myAction'})
```


chrisnojima:
Prettier is complaining on files I didn't touch so I assume master doesn't actually pass either

chrisnojima:
any pointers on what updates are needed here?

chrisnojima:
@GAntoine can i get some guidance here

GAntoine:
Looks like the definitions for v11 and v13 are failing on flow v0.62+. There's a bug currently that stops v62+ from getting fetched properly locally, so I can't debug this until then. Let me get that fixed first, then I'll take a look as to why this is failing.

chrisnojima:
Thanks!

chrisnojima:
@GAntoine got a second to take a look at this? I run-tests locally and set the number of flow versions to maximum and it still tells me it passes locally. Looking at CI its complaining about other libs

GAntoine:
It seems to be related to #2025. I'm putting this on hold till that's addressed.

PR #1708, opened 2018-01-09
Fix type of redisClient.set()

stouf:
- The second argument must be a string (it used to be `any`)
- There is an option third argument which is a callback invoked when Redis is done saving the provided data

Example:

```js
const client = redis.createClient({ host: '127.0.0.1', port: 6379 });
client.set('some-key', 'Some value!', (error) => {
  console.log('Operation terminated!');
  console.log('Error?', error);
});

// Outputs:
// Operation terminated!
// Error? null
```

### DIFF
--- a/definitions/npm/redis_v2.x.x/flow_v0.34.x-/redis_v2.x.x.js
+++ b/definitions/npm/redis_v2.x.x/flow_v0.34.x-/redis_v2.x.x.js
@@ -19,7 +19,7 @@ declare module "redis" {
     hgetall: (topic: string, key: string) => Array<string> | void;
     hdel: (topic: string, key: string) => number;
     get: (key: string) => any;
-    set: (key: string, value: any) => void;
+    set: (key: string, value: string, cb?: (error: Error | null) => void) => void;
     del: (...keys: Array<string>) => void;
     rpoplpush: (source: string, destination: string) => string | void;
     publish: (topic: string, value: any) => void;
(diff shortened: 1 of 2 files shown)


jedwards1211:
@stouf 
> The second argument must be a string (it used to be any)

Did you actually check this against the docs or code of the `redis` package?

According to [the docs](http://redis.js.org/#redis-a-nodejs-redis-client-usage-example-sending-commands) non-string values are allowed and merely coerced to a string:
> Please be aware that sending null, undefined and Boolean values will result in the value coerced to a string!

stouf:
@jedwards1211 

> according to the docs of the client package non-string values are allowed and merely coerced to a string

Yes, you are totally right. However, the client package will not try to serialize the data you pass in. For example,

```js
const client = redis.getClient();
client.set('some-key', {
  name: 'my name',
  surname: 'my surname'
});
```

should output the following to the standard output of your node process:

```
node_redis: Deprecated: The SET command contains a argument of type Object.
This is converted to "[object Object]" by using .toString() now and will return an error from v.3.0 on.
Please handle this in your code to make sure everything works as you intended it to.
```

If you manually check in Redis, you will see that your data got saved as the string `[object Object]`.

So, strictly speaking, you are totally right when you write that the client accepts any type value. However, not serializing your data before passing it to the client can lead to some unexpected results.
Based on that, I thought that updating the definition of `set()` to something more strict would be safer than keeping it more flexible, even if it could be a little less convenient sometimes.

I also considered being a little more permissive and allowing `number`, `boolean` and `null` as well.  However, even if `set()` turns those types into `string`, `get()` doesn't try to parse them from `string` to their original type. Therefore, I thought it would be more consistent to keep only `string` for both `get()` and `set()`.

What do you think about it? If you think we should keep the type of `set()` so it accepts any type of data, let's open an issue to discuss about it since it would get more attention than discussion in this closed pull request :)

Issue #695, opened 2017-03-12
Bug in Rxj.combineLatest definition


szagi3891:
Currently defined :

```
  combineLatest<A, B>(
    a: rxjs$Observable<A>,
    resultSelector: (a: A) => B,
  ): rxjs$Observable<B>;

  combineLatest<A, B, C>(
    a: rxjs$Observable<A>,
    b: rxjs$Observable<B>,
    resultSelector: (a: A, b: B) => C,
  ): rxjs$Observable<C>;

...
```

Should be defined

```
  combineLatest<A, B>(
    a: rxjs$Observable<A>,
    resultSelector: (t:T, a: A) => B,
  ): rxjs$Observable<B>;

  combineLatest<A, B, C>(
    a: rxjs$Observable<A>,
    b: rxjs$Observable<B>,
    resultSelector: (t:T, a: A, b: B) => C,
  ): rxjs$Observable<C>;
...
```

As for the missing argument:
resultSelector: (**t:T,** a: A) => B,
resultSelector: (**t:T,** a: A, b: B) => C,

szagi3891:
In this definition is the same error:

```
  combineLatest<A, B>(
    a: rxjs$Observable<A>,
    resultSelector: (a: A) => B,
  ): rxjs$Observable<B>;

  combineLatest<A, B, C>(
    a: rxjs$Observable<A>,
    b: rxjs$Observable<B>,
    resultSelector: (a: A, b: B) => C,
  ): rxjs$Observable<C>;
```

PR #1717, opened 2018-01-10
Rxjs fix signatures(withLatestFrom, combineLatest, zip)

berlysia:
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-withLatestFrom
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-combineLatest
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-combineLatest
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-zipProto
http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#static-method-zip

# The last argument is now properly optional
These operators receive a projection function as the last argument, and it is optional.
Under the previous signature, It's not optional in case of 2 or more Observables are given as arguments.

Actual behavior is described in document of `withLatestFrom`:
> Projection function for combining values together. Receives all values in order of the Observables passed, where the first parameter is a value from the source Observable. (e.g. a.withLatestFrom(b, c, (a1, b1, c1) => a1 + b1 + c1)). If this is not passed, arrays will be emitted on the output Observable.

It is the same for `combineLatest`, `zip` and static version of these operators.

# Fix projection function's argument signature
In operators as instance methods, projection function should receive the value from source Observable as the first argument, but ignored in previous signature.

from actual behavior:
```
declare class Observable<+T> {
  // ...
  withLatestFrom<A, B, C>(
    a: Observable<A>,
    b: Observable<B>,
    resultSelector: (t: T, a: A, b: B) => C
  ): Observable<C>;
}
```

previous signature:
```
declare class Observable<+T> {
  // ...
  withLatestFrom<A, B, C>(
    a: Observable<A>,
    b: Observable<B>,
    resultSelector: (a: A, b: B) => C //  source Observable's value is missing
  ): Observable<C>;
}
```

It is the same for instance method version of `combineLatest` and `zip`.

### DIFF
--- a/definitions/npm/rxjs_v5.0.x/flow_v0.25.0-v0.33.x/rxjs_v5.0.x.js
+++ b/definitions/npm/rxjs_v5.0.x/flow_v0.25.0-v0.33.x/rxjs_v5.0.x.js
@@ -1054,43 +1054,6 @@ declare class rxjs$Observable<+T> {
     closingSelector: () => rxjs$Observable<any>
   ): rxjs$Observable<rxjs$Observable<T>>;
 
-  withLatestFrom<A>(a: rxjs$Observable<A>): rxjs$Observable<[T, A]>;
-
-  withLatestFrom<A, B>(
-    a: rxjs$Observable<A>,
-    resultSelector: (a: A) => B
-  ): rxjs$Observable<B>;
-
-  withLatestFrom<A, B, C>(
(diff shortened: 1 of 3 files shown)


berlysia:
This PR seems to resolve issue #695

PR #1719, opened 2018-01-11
Fix options for react-redux_v5

PepijnSenders:
In react-redux v5 we have a new or different options object, the one currently described in flow-typed is the options object for `connectAdvanced`. The current options according to the API are described [here](https://github.com/reactjs/react-redux/blob/master/docs/api.md#connectmapstatetoprops-mapdispatchtoprops-mergeprops-options). Please bare with me since this is my first PR to flow-typed and I don't use flow myself so I had a bit of trouble running the tests, but I think they should be fine. (Just making this PR to do a PR to a library I'm using).

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.53.x-v0.53.x/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.53.x-v0.53.x/react-redux_v5.x.x.js
@@ -56,9 +56,33 @@ declare module "react-redux" {
     subKey?: string
   ): Provider<*, *>;
 
+  declare function areStatesEqual(
+    nextState?: Object,
+    state?: Object,
+  ): boolean;
+
+  declare function areStatesEqual(
+    nextOwnProps?: Object,
+    ownProps?: Object,
+  ): boolean;
+
+  declare function areStatePropsEqual(
+    nextStateProps?: Object,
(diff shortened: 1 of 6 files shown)


PepijnSenders:
So I see the tests are failing on the new functions I created, if someone could help me out there that would be really appreciated!

definitions/npm/react-redux_v5.x.x/flow_v0.53.x-v0.53.x/react-redux_v5.x.x.js
```diff
@@ -56,9 +56,33 @@ declare module "react-redux" {
...
+  declare function areMergedPropsEqual(
+    nextMergedProps?: Object,
+    mergedProps?: Object,
+  ): boolean;
+
   declare type ConnectOptions = {
     pure?: boolean,
-    withRef?: boolean
+    areStatesEqual?: areStatesEqual,
+    areOwnPropsEqual?: areOwnPropsEqual,
```
wanderley: You have to provide a definition for `areOwnPropsEqual` as you did for `areStatesEqual`.

villesau:
Looks like these changes are for older version of react-redux. @PepijnSenders If this is still problem with up to date libdefs, can you submit a fix for them instead? I'm going to close this one since this conflicts and seems to be pretty stale.

PR #1720, opened 2018-01-11
Added Support for Radium wrapping  with the new Flow versions

dolma:
Version 0.53.0 of Flow introduced new React types like StatelessFunctionalComponent<Props> and changed the definition of Components to React.Component<Props, State> instead of React.Component<DefaultProps, Props, State>.

The libdef for Radium still used the old versions of these types. It was breaking the type checking on presentational components being wrapped by Radium.


dolma:
The test file doesn't pass the CI build because of a conflict in imported module versions. I didn't really understand how the test files were run at first. I will now move each test file in the appropriate version sub-folder.

dolma:
The Travis build still fails and the errors displayed are from another version of the libdef which was deployed in the past so I assumed it didn't fail the tests back then.
How come it fails now despite not having been touched at all ? Is it due to newer versions of Flow ?

dolma:
Ok so it appears to be fixed for stateless components but now the React.Component is suffering from the same issue: The type check is broken when the component is being wrapped by Radium.

I'm trying to find a fix for it at the moment but no luck so far. It is making the tests fail because of some of the `// $ExpectError` calls that don't get any error because of the broken type check...

I could, as a workaround, remove the broken parts of the tests but it wouldn't solve the underlying problem so I don't really fancy doing it. Does anyone have any idea on how to solve this one ?

dolma:
Here's the code you need to reproduce the error:

```
// @flow
import React, { Fragment, type StatelessFunctionalComponent } from 'react';
import Radium from 'radium';

type Props = { text: string };

const Foo: StatelessFunctionalComponent<Props> = ({ text }: Props) => <div>{text}</div>;

class Bar extends React.Component<Props, void> {
  render() {
    return (
      <div>{this.props.text}</div>
    );
  }
}

const WrappedFoo = Radium(Foo);
const WrappedBar = Radium(Bar);

const RadiumError = () => (
  <Fragment>
    <Foo />
    <WrappedFoo />
    <WrappedFoo text={3} />
    <Bar />
    <WrappedBar />
    <WrappedBar text={true} />
  </Fragment>
);
```

All of these components should give an error either for missing a prop or for having the wrong prop type. However, only the unwrapped components report an error. The wrapped components are not properly type checked.

With the changes I committed so far, the `WrappedFoo` functional component properly reports errors, but the `WrappedBar` class component still doesn't report any error...

I also reproduced the error in Flow Try: [https://flow.org/try/#0PTAE...](https://flow.org/try/#0PTAEAEDMBsHsHcBQBLAtgB1gJwC6gEoCmAhgMY4A0oA3qAGJbEDmqhAdpaDgJ7qGgBlHMRyFohAM4S6AVzblksNsWgBhWBiXs8AX1CQsG0AHIsJcsYDciRCFAB1fAEEACi4Ci+G5DkKlodgALYnlCAAp7RnQ+ABN1TTZtAC5QUiUJHCwZcmwAShpQUERC0mhiKQJiGOQZVHc2YNCsAIAPUTYYiQco2PjMRI4aYsLQMw7CLDD86mGR0cIcGSw2UAkZPiwAOjGYianZ0B1ho8KzReXK6tr6xtIJ6yPEH3kcRRX8KprUCJ7COI1+slUulMtkcHkhqcFksVkEQncfsRon8+loOLkHjZdqViGYuLx+KoylJUQMcAAeFxUIQAPlAAF5QETyhJyUQyDhNqTtJTqTgaTTrNiynicRUPldUJDQJSaWE0gkgUIRGJJNJfK8lCpuRxZbkUsrROIpLIXm9tQC0RSXILhrzBPz5ZaySlmSTnTyqQ6afqmcSJDrrXzbY87C58AB5FwCGw8PigFyGdBdRm0URtFIZLDINhMQ7WWxgOgRiM2NJsDL0WCwA3CI1q01+ZRqD26xOwZN0xlhNOENqHFLt5P5el08nVABuNOo6ZwOnJwEntsLoAAQk4vIgxV1V7jWu1OgRzJzA5SkxIqBPYMgYnSZlDxpNpgczjDQGEDoVx8gpzPAsgJE2dBz02Wd50XH8aQODFjkQUMwEcVwPHwUAABV3AEVCYy3EFuiRWI6GrBlLi+MJCNgGDy0rSJ8L+XdmkZCVSPomCcIrPAaORLBVxkPNuxHOkwlAOwBECWAZGgGJQGA2AYmyfgJkMLAugAIxkPAlGgbguECfg5HgX4pIVQEOC6NIsDMcgtOkww5LuUAQgCCzsDtBhmFYDgoLmclyOErzQAASHJTiCKI4B-KCkK-l82d6WoABmPRwoOcl6L8lKopiNLku8zK0timcskIJL-IXNyWG0KCMSAA)

I wrote a simplified version of the `Radium` wrapper code since I can't import it there.

**Note**: Interestingly, my type definition fix doesn't work there, both `WrappedFoo` and `WrappedBar` fail to report errors properly. I'm assuming it's because of the simplification to the Radium wrapper I made. Doesn't change the underlying problem though.

Any opinion on what's going on would be highly appreciated. Thanks

dolma:
Related Issue posted #1726

dolma:
@GAntoine  Any idea about why this is failing to type check ?

villesau:
I assume this is not moving forward -> closing. Feel free to reopen!

PR #1736, opened 2018-01-16
fix koa-router Context type (fixes #1698)

### LINKED ISSUE #1698: koa-router_v7.2.x ctx is of any type instead of KOA Context type
In type KoaRouter$Middleware the ctx type is defined as "any" when it should ideally be KOA Context type.

saadq:
I think I've found a workaround to fixing #1698 by getting the `Context` type from the `koa` module. It's pretty hacky, so if anyone has any suggestions on doing this in a better way please feel free to improve upon this.


saadq:
I don't believe there is currently a way to have dependencies between library definitions, so will close this out for now.

PR #1737, opened 2018-01-16
Replace `any` by `mixed` in express_v4.x.x.js

nvie:
Using `any` will effectively disable the type checker anywhere you use the `req.body` property, which is dangerous. By declaring it as a `mixed` type, it allows you to use it, but you'll have to make sure that any assumptions you make about its contents are checked first, which is the way to go.


nvie:
@GAntoine Any feedback for me on this PR? I'm not sure what the label "needs-updates" means here. Do you require something from me? Accepting this PR would be greatly appreciated.

GAntoine:
@nvie The build was failing the first time around, but I restarted it. Looks like this is good to merge.

PR #1739, opened 2018-01-17
Add type for redis.llen()

stouf:
Reference: https://redis.io/commands/llen

Demonstration:

```js
'use strict';

const client = redis.createClient();

const returned = client.llen('key', (error, length) => {
  console.log('error', error);
  console.log('length', length);
});
console.log('returned', returned);
```

Outputs

```
returned false
error null
length 0
```


GAntoine:
@stouf I just merged in your other PR, and so this needs a rebase.

stouf:
The pull request has been updated :+1:

PR #1747, opened 2018-01-20
[temp] Change return type of callback to mixed for async functions.

bolinfest:
Because the current return type is declared as `void`, Flow complains
when an `async` function is used in the callback because it returns a
`Promise`.


zertosh:
👍 lgtm – I'd merge this, but I'm not a member here

GAntoine:
On it.

PR #1750, opened 2018-01-22
Added optional resolver to merge

tompuric:
Added a missing parameter


tompuric:
This PR takes reference from Lodash's merge implementation

https://github.com/flowtype/flow-typed/blob/master/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js#L1147
`mergeWith<T: Object, A: Object>(
      object: T,
      customizer?: (
        objValue: any,
        srcValue: any,
        key: string,
        object: T,
        source: A
      ) => any | void
    ): Object;`

There was discussion with colleagues about converting the current T and S1 types to Object and Object respectively. We will discuss in future and submit a PR if necessary.

PR #1754, opened 2018-01-23
fix(redis): fix hgetall arguments

jedwards1211:
The [`HGETALL` command](https://redis.io/commands/hgetall) only takes one argument, not two.


jedwards1211:
closing in favor of another PR

PR #1755, opened 2018-01-24
fix numerous faulty definitions for redis

jedwards1211:
There were some pretty low quality method definitions for `redis` in here, hopefully mainly because the node `redis` package doesn't document many commands itself, and defers to the official Redis documentation.

Primarily, someone mistakenly thought that a number of methods return results.  As the `redis` package docs say, "Note that the API is entirely asynchronous. To get data back from the server, you’ll need to use a callback."

The `hgetall` method def also had an extra argument that shouldn't be there.

Although I'm pretty sure no methods of `RedisClient` require a callback, I marked the callbacks for getter-type methods as required since forgetting to include a callback would almost certainly be a mistake.

### DIFF
--- a/definitions/npm/redis_v2.x.x/flow_v0.34.x-/redis_v2.x.x.js
+++ b/definitions/npm/redis_v2.x.x/flow_v0.34.x-/redis_v2.x.x.js
@@ -1,35 +1,102 @@
 /* This module definition is by no means complete. A lot of methods of the RedisClient class are missing */
 declare module "redis" {
   declare class RedisClient extends events$EventEmitter mixins RedisClientPromisified {
-    hmset: (key: string, map: any, callback: (?Error) => void) => void;
-    rpush: (key: string, value: string, callback: (?Error) => void) => void;
-    lpush: (key: string, value: string, callback?: (?Error, number) => void) => void;
+    hmset: (
+      key: string,
+      map: any,
+      callback?: (error: ?Error) => void
+    ) => void;
+    rpush: (
+      key: string,
+      value: string,
+      callback?: (error: ?Error) => void
+    ) => void;
+    lpush: (
+      key: string,
+      value: string,
+      callback?: (error: ?Error, newLength: ?number) => void
+    ) => void;
     lrem: (
       topic: string,
       cursor: number,
-      value: string
-    ) => Array<string> | void;
+      value: string,
+      callback?: (error: ?Error, entries: ?Array<string>) => void
+    ) => void;
     lrange: (
       topic: string,
       cursor: number,
       cursor2: number,
-      (error: Error | null, entries: Array<string>) => void
-    ) => boolean;
-    llen: (key: string, (error: Error | null, length: number) => void) => boolean;
-    hset: (topic: string, key: string, value: string) => number;
-    hget: (topic: string, key: string, value: string) => string | void;
-    hgetall: (topic: string, key: string) => Array<string> | void;
-    hdel: (topic: string, key: string) => number;
-    get: (key: string, (Error | null, string | null) => void) => void;
-    set: (key: string, value: string, cb?: (error: Error | null) => void) => void;
-    setex: (key: string, timeout: number, value: string, callback?: (error: ?Error, result: ?string) => void) => void;
-    ttl: (key: string, callback: (error: ?Error, ttl: ?number) => void) => void;
-    del: (keys: Array<string>, cb?: (Error | null) => void) => void;
-    mget: (keys: Array<string>, (Error | null, Array<string | null>) => void) => void;
-    mset: (keysAndValues: Array<string>, cb?: (Error | null) => void) => void;
-    rpoplpush: (source: string, destination: string) => string | void;
-    flushall: (cb?: (Error | null) => void) => void;
-    publish: (topic: string, value: any) => void;
+      callback: (error: ?Error, entries: ?Array<string>) => void
+    ) => void;
+    llen: (
+      key: string,
+      callback: (error: ?Error, length: ?number) => void
+    ) => void;
+    hset: (
+      topic: string,
+      key: string,
+      value: string,
+      callback?: (error: ?Error, result: ?(0 | 1)) => void
+    ) => void;
+    hget: (
+      topic: string,
+      key: string,
+      value: string,
+      callback: (error: ?Error, result: ?string) => void
+    ) => void;
+    hgetall: (
+      topic: string,
+      callback: (error: ?Error, result: ?Array<string>) => void
+    ) => void;
+    hdel: (
+      topic: string,
+      key: string,
+      callback?: (error: ?Error, numRemoved: ?number) => void
+    ) => void;
+    get: (
+      key: string,
+      callback: (error: ?Error, value: ?string) => void
+    ) => void;
+    set: (
+      key: string,
+      value: string,
+      callback?: (error: ?Error, result: ?(0 | 1)) => void
+    ) => void;
+    setex: (
+      key: string,
+      timeout: number,
+      value: string,
+      callback?: (error: ?Error, result: ?string) => void
+    ) => void;
+    ttl: (
+      key: string,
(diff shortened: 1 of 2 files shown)


definitions/npm/redis_v2.x.x/flow_v0.34.x-/redis_v2.x.x.js
```diff
@@ -1,35 +1,102 @@
...
+      key: string,
+      callback: (error: ?Error, ttl: ?number) => void
+    ) => void;
+    del: (
+      keys: Array<string>,
+      callback?: (error: ?Error, numRemoved: ?number) => void
+    ) => void;
+    mget: (
+      keys: Array<string>,
+      callback: (error: ?Error, values: ?Array<?string>) => void
```
stouf: Isn't `values` *always* going to be an Array? If so, shouldn't we not use the maybe type here?

jedwards1211: I thought that `values` would be `undefined` in the case that the callback is called with an `error`, am I wrong?

Otherwise I think you're right that `null` is always used instead of `undefined`, but I'm wondering why that matters to you.  I generally prefer to guard against something being `null` or `undefined` even if one of the two is never expected (which I think is the  philosophy behind maybe types encompassing both). How are you doing your null checks?

stouf: > I thought that `values` would be `undefined` in the case that the callback is called with an `error`, am I wrong?

Nop, that's correct 👍 However, because of the strong convention about how one should handle errors passed to callbacks in Node.js, wouldn't it be more convenient to assume that one would always check `error` before doing anything with the result? What do you think?
(I came to wonder about it the other day, I'm glad that I can discuss it with someone else 🙂)

> I'm wondering why that matters to you

When Flow checks the equivalence of two function signatures, it can cause errors.
See the following example:

```js
/* @flow */
'use strict';

const client = redis.createClient({
  host: '127.0.0.1',
  port: 6379
});


/*::
type fetch$fetcher = (Array<string>, (Error | null, ?Array<?string>) => void) => void;
*/
function fetch(fetcher/*: fetch$fetcher*/, keys/*: Array<string>*/) {
  return new Promise((resolve, reject) => {
    fetcher(keys, (error, values) => {
      if (error) {
        reject(error);
        return;
      }
      resolve(values);
    });
  });
}


fetch(client.mget, ['a', 'b', 'c'])
  .then(console.log)
  .catch(console.error);
```

Flow returns the following error:

```
Cannot call fetch with client.mget bound to fetcher because null or undefined [1] is incompatible with null [2] in the
first argument of the second argument.

     index.js
 [2] 13│ type fetch$fetcher = (Array<string>, (Error | null, ?Array<?string>) => void) => void;
       :
     25│ }
     26│
     27│
     28│ fetch(client.mget, ['a', 'b', 'c'])
     29│   .then(console.log)
     30│   .catch(console.error);
     31│

     flow-typed/npm/redis_v2.x.x.js
 [1] 84│       callback: (error: ?Error, values: ?Array<?string>) => void
```

There is also the following case:

```js
/* @flow */
'use strict';

const client = redis.createClient({
  host: '127.0.0.1',
  port: 6379
});


function getLengths(keys/*: Array<string>*/)/*: Promise<Array<number>>*/ {
  return new Promise((resolve, reject) => {
    client.mget(keys, (error, values) => {
      if (error) {
        reject(error);
        return;
      }
      if (!values) {
        throw new Error('Unexpected error');
      }
      const lengths = values.map(value => {
        return !value ? 0 : value.length;
      });
      resolve(lengths);
    });
  });
}

exports.getLengths = getLengths;
```

where Flow wouldn't say anything as long as you don't have the linter enabled. If the linter is enabled, Flow tells the following:

```
Sketchy null check on string [1] which is potentially an empty string. Perhaps you meant to check for null or
undefined [2]? (sketchy-null-string)

        index.js
        20│         throw new Error('Unexpected error');
        21│       }
        22│       const lengths = values.map(value => {
        23│         return !value ? 0 : value.length;
        24│       });
        25│       resolve(lengths);
        26│     });

        flow-typed/npm/redis_v2.x.x.js
 [2][1] 84│       callback: (error: ?Error, values: ?Array<?string>) => void
```

It can be worked around by changing

```js
return !value ? 0 : value.length;
```

to

```js
return value == null ? 0 : value.length;
```

but I would like to avoid using `==` as much as possible and be more strict on my types (although this is probably purely personal)

jedwards1211: @stouf I tend to error on the side of caution with the type of the result of a callback since IIRC I tried modeling a callback this way in hopes flow would know the result is defined if `error` is not, and I don't think it worked in all cases:
```js
callback: ((error: Error) => any) & ((error: null | undefined, values: Array<?string>) => any)
```

jedwards1211: The case where you need the `value == null` check doesn't actually have to do with whether `values` is an `Array` or an `?Array`, does it?

stouf: > It's definitely less convenient to have to null check `values` if you know `error` is `null`, but `(?Error, Array<?string>) => any` wouldn't guarantee null safety.

I see. I would personally go for convenience for this particular case, but your point is also valid in my opinion.
I took a quick look at how Flow annotates the Node.js core functions and it seems they maybe-type the error and don't maybe-type the result:
https://github.com/facebook/flow/blob/82f35b7a5a232ea17d7cb539a9353940d4720ab3/lib/node.js#L900-L917
Maybe doing the same would be a good idea for the sake of being consistent?

That said, I think the question of what is a good way of annotating callback functions is not particular to this function nor this library, but something global to this whole repository. Maybe getting the community to agree on one way of doing it and "enforcing" it would be a good idea? @GAntoine Any input on it?

> The case where you need the `value == null` check doesn't actually depend on whether `values` is an `Array` or an `?Array`, does it?

Nop, indeed. It was about the result passed through the callback being an array of `?string` instead of `string | null`. Sorry if it was confusing 😅

>  I don't really see why it should require the callback to accept only `Error | null` instead of just `?Error`. To me handling both `null` and `undefined` the same way is so simple that one might as well play it safe and always do it instead of writing code that can tolerate one and not the other.

Yep, it sounds good to me too :+1:

jedwards1211: Huh, I didn't know that about the core flow types, I'm kind of surprised, but that's a valid point as well!  I wouldn't mind being consistent with the core types.  Maybe someday they'll come up with a solid way to indicate that one or the other argument is defined.

stouf: Yeah, I wish the way you annotated callback in one of your previous comments worked, that would be great. I haven't looked for this, but maybe some people brought that up on the Flow repository.

Anyway, thanks for your input 🙂
I'm currently not in the need for this file to change but if it changes in the future or if someone else needs it to change, you wouldn't mind a type definition such as the following?

```js
function whatever(cb/*: (?Error, ResultType) => void */) {
  // ...
}
```

Just confirming to avoid useless back and forth ^^

jedwards1211: Yeah, that sounds good to me.  I wouldn't object if you made a PR right now (though I'm too busy at the moment :smile:)

stouf: Okay :+1: I'm little busy too but I'll try to find some time sending a pull request soon-ish ;-)

stouf: https://github.com/flowtype/flow-typed/pull/1884 :)

PR #1756, opened 2018-01-24
Ramda#pathOr accepts numbers for path.

AndrewSouthpaw:
@LoganBarnett 

Per [the docs](http://ramdajs.com/docs/#pathOr). Also refactors the signature using the pattern

```
declare type Foo = ...
declare var foo = Foo
```

per discussion in #1615.

Also, tries out a different spacing structure to better illustrate the nesting of intersected functions, for better reading. I'd really like your opinion on this one, hopefully we could standardize. I like what you started with `Reduce`, and it's impossible for me to keep track of the nested parens.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.49.x-v0.61.x/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.49.x-v0.61.x/ramda_v0.x.x.js
@@ -1533,21 +1533,41 @@ declare module ramda {
   declare function path<V, A: null | void>(p: Array<string>, o: A): void;
   declare function path<V, A: mixed>(p: Array<string>, o: A): ?V;
 
-  declare function pathOr<T, V, A: NestedObject<V>>(
-    or: T,
-    ...rest: Array<void>
-  ): ((p: Array<string|number>, ...rest: Array<void>) => (o: ?A) => V | T) &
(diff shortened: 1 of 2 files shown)


LoganBarnett:
Are those legitimate failures with the test suite? This is for `propOr` but it looks like it's about `propEq`.

Code otherwise LGTM. Thanks @AndrewSouthpaw!

Since we're talking build failures - I've run into memory issues when running the test suite sometimes, especially the whole suite. I should file a bug at some point, but haven't had time to put something good together. Work is pretty crazy for me right now.

AndrewSouthpaw:
Yeah I dunno what the build failures are about, and I can't test them until #1759 works itself out. I'm guessing there's something that happens in the most recent versions of flow.

LoganBarnett:
@AndrewSouthpaw is this blocking you right now from downstream activity (like outside of this repo)? If we can wait for a fix and then iterate on the tests that would be ideal. I assume a fix is coming soonish?

GAntoine:
#1770 is out to address the issue, I'm looking to merge it in soon™.

AndrewSouthpaw:
Yeah I'm not worried. Usually I make changes in my own libdef anyway, try them out for a while, and then merge up the branch changes.

AndrewSouthpaw:
Bump @GAntoine, updated the branch, tests now pass.

PR #1762, opened 2018-01-25
fix(redis): fix hgetall result type

jedwards1211:
Argh, I didn't notice the section in the docs about how the client's `hgetall` builds an object out of the returned key-value pair array


GAntoine:
Seems one of the tests is failing, you might want to take a look at `test_redis_v2.x.x.js:118`.

GAntoine:
🔥

PR #1765, opened 2018-01-25
[react-relay] Replace FlowFixMe with *

vnctaing:
That should fix that error  

screen shot 2018-01-25 at 11 41 28 am

### DIFF
--- a/definitions/npm/react-relay_v1.x.x/flow_v0.47.x-/react-relay_v1.x.x.js
+++ b/definitions/npm/react-relay_v1.x.x/flow_v0.47.x-/react-relay_v1.x.x.js
@@ -349,7 +349,7 @@ declare module "react-relay" {
     connectionConfig: ConnectionConfig
   ): TBase;
 
-  declare export type Variables = { [name: string]: * };
+  declare export type Variables = { [name: string]: boolean | string | number | null | void };
   declare export type DataID = string;
 
   declare type TEnvironment = Environment;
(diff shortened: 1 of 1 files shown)


GAntoine:
Is `*` the right type thing to use here?

vnctaing:
I'm not totally sure, but I don't think Flow is able to determine the variables type in advance to the `RelayRuntime` package. 

I think the original flow-typed definition was coming from that [commit](https://github.com/facebook/relay/commit/0a4027d4c2328549e2d314dbb5eb8ad1eeaeb051#diff-acbf22bc09d64e2ad9b25c12298f9859R30)

I've tried to analogically look into [how they check the graphql variables](https://github.com/apollographql/react-apollo/blob/76fc936f6292ef214a6eeb6cfd9963d573b871d8/src/types.ts#L15) typing in react-apollo. 

```
export type OperationVariables = {
  [key: string]: any; // I try to avoid using `any` in favor of `*`
};
```
Maybe input from someone more familiar with relay codebase would be more than welcome ?

GAntoine:
@sibelius Could I get your input on this?

sibelius:
I think variables can have any kind of value, or maybe only primitives types

vnctaing:
You're right, we shouldn't let pass functions or class through variables. I guess an union type of primitive would work. like `boolean | string | number | null | void`

vnctaing:
Actually we want to be able to send plain object to the union type too. I've forgotten to add it to the union type... opening a PR ...

PR #1767, opened 2018-01-25
fix(express): add optional Server return to listen method

rogeriopvl:
This is a temporary fix for express typedefs to a known issue when declaring a class with overloaded methods (see more info here: https://github.com/facebook/flow/issues/4198). 

Since this issue has no fix in sight and causes coverage to fail reaching 100%, perhaps this seems an acceptable workaround?

### DIFF
--- a/definitions/npm/express_v4.16.x/flow_v0.32.x-/express_v4.16.x.js
+++ b/definitions/npm/express_v4.16.x/flow_v0.32.x-/express_v4.16.x.js
@@ -210,15 +210,15 @@ declare class express$Application extends express$Router mixins events$EventEmit
     hostname?: string,
     backlog?: number,
     callback?: (err?: ?Error) => mixed
-  ): Server;
+  ): ?Server;
   listen(
     port: number,
     hostname?: string,
     callback?: (err?: ?Error) => mixed
-  ): Server;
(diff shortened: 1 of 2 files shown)


GAntoine:
@LoganBarnett I think you'd know more then I would, if this is an acceptable workaround?

LoganBarnett:
We have this exact workaround for one of our repos but haven't had a chance to put together a PR for it. @rogeriopvl thanks doing that!

@GAntoine yeah this is totally fine. Having to do an extra null check is pretty harmless.

LoganBarnett:
Actually, let's provide some comments as to why we're doing this. The workaround itself is fine.

Here's the comment we left in our codebase:
``` javascript
/*
With flow-bin ^0.59, express app.listen() is deemed to return any and fails flow type coverage.
Which is ironic because https://github.com/facebook/flow/blob/master/Changelog.md#misc-2 (release notes for 0.59)
says "Improves typings for Node.js HTTP server listen() function."  See that?  IMPROVES!
To work around this issue, we changed Server to ?Server here, so that our invocations of express.listen() will
not be deemed to lack type coverage.
*/
```
Feel free to use verbatim, mangle, or simply provide your own.

GAntoine:
Awesome, thanks for the hand. 💪I went ahead and added the comment as-is, it's too real not to include verbatim.

PR #1768, opened 2018-01-25
Diff HOC props with router props

matthewrfindley:
An issue was found using HOC with callbacks as props where the callbacks appears to be inferring the type from the value passed in rather than the prop definition.  You can see an example here. https://github.com/facebook/flow/issues/5720

This PR addresses this error:

```
Error: packages/xxx/src/components/CloseButton.js:37
 37: class CloseButton extends React.Component<CloseButtonProps, CloseButtonState> {
                                               ^^^^^^^^^^^^^^^^ CloseButtonProps. This type is incompatible with
146:     Component: React$ComponentType<{| ...ContextRouter, ...P |}>
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type. See lib: packages/flow-types/npm/react-router-dom_v4.x.x.js:146
  Property `closeAlert` is incompatible:
     35:           closeAlert={closeAlert}
                                     ^^^^^^^^^^^^^^^^ function. Too few arguments passed to. See: packages/xxx/src/components/CloseButton.test.js:35
     25:   closeAlert: (string | null) => void,
                             ^^^^^^^^^^^^^^^^^^^^^^^ function type

```


matthewrfindley:
This does not appear to solve the problem quite yet.  There are still specs failing.

I am having difficulty combining the exact prop match and the changes introduced in this PR.  If anyone has more context that can shed light on a solution to the error listed in the description, it would be greatly appreciated.

PR #1772, opened 2018-01-27
Adding startsWith function

functionalStoic:
Added startsWith function as it was missing from the type definition. Here is the documentation.

Here is the ramda docs
http://ramdajs.com/docs/#startsWith

Here is the actual implementation within the Ramda Library
https://github.com/ramda/ramda/blob/v0.25.0/source/startsWith.js


GAntoine:
@functionalStoic With the other PR just merged, if you still want to see these changes implemented, I'd recommend closing this PR, and making your changes off a fresh `master`.

functionalStoic:
@GAntoine Ok Will do. Thanks!

PR #1779, opened 2018-01-29
Make handleActions use ReducerDefinition

xeophin:
In `handleAction()` the definition has already been changed to used the ReducerDefinition, now `handleActions()` should follow, too.

As far as I understand the code, this definition should also be able to handle nested reducers.


GAntoine:
@xeophin You'll need to update the specs, specifically `test_handleActions.js:42`. It's currently passing with your changes, whereas it should be failing.

PR #1789, opened 2018-01-30
improve type def for lodash unzip

maartenschumacher:
This change brings the unzip definition to be in line with the zip definition, using flow's tuple syntax. I've removed the ability to pass in `null` or `undefined` because it's much easier to null-check the input than check the output for an empty array. Also like this you will be able to use destructuring syntax like `const [numbers, chars] = unzip([[1, 'a'], [2, 'b']])` without flow complaining about accessing undefined tuple indexes.


definitions/npm/lodash_v4.x.x/flow_v0.55.x-v0.62.x/lodash_v4.x.x.js
```diff
@@ -408,7 +408,14 @@ declare module "lodash" {
     uniq<T>(array?: ?Array<T>): Array<T>;
     uniqBy<T>(array?: ?Array<T>, iteratee?: ?ValueOnlyIteratee<T>): Array<T>;
     uniqWith<T>(array?: ?Array<T>, comparator?: ?Comparator<T>): Array<T>;
-    unzip<T>(array?: ?Array<T>): Array<T>;
+    unzip<A, B>(array: Array<[A, B]>): [Array<A>, Array<B>];
```
GAntoine:
```
   319: unzip([[1, "a"], [2, "b"]])[0].map(x => x * 10);
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^ function call. Could not decide which case to select
   319: unzip([[1, "a"], [2, "b"]])[0].map(x => x * 10);
        ^^^^^ polymorphic type: function type
     Case 1 may work:
     411:     unzip<A, B>(array: Array<[A, B]>): [Array<A>, Array<B>];
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type. See lib: lodash_v4.x.x.js:411
     But if it doesn't, case 2 looks promising too:
     412:     unzip<A, B, C>(array: Array<[A, B, C]>): [Array<A>, Array<B>, Array<C>];
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ polymorphic type: function type. See lib: lodash_v4.x.x.js:412
     Please provide additional annotation(s) to determine whether case 1 works (or consider merging it with case 2):
     319: unzip([[1, "a"], [2, "b"]])[0].map(x => x * 10);
                ^^^^^^^^^^^^^^^^^^^^ inferred union of array element types (alternatively, provide an annotation to summarize the array element type)
```

maartenschumacher:
My laptop crashed when I tried running tests locally...

PR #1792, opened 2018-01-30
improve koa-router libdef (fixes #1698)

### LINKED ISSUE #1698: koa-router_v7.2.x ctx is of any type instead of KOA Context type

saadq:
I don't know if you'd be okay with this, but I basically needed to copy all the types from `koa` into this middleware (which I put into a separate `declare module "koa"`) in order to get proper type coverage for `ctx`. If you'd be okay with doing something like this, I can push out a bunch of other `koa` middleware libdefs that I've been using in my personal projects which basically does the same thing with the `koa` types copied into the middleware libdef.

### DIFF
--- a/definitions/npm/koa-router_v7.2.x/flow_v0.25.x-/koa-router_v7.2.x.js
+++ b/definitions/npm/koa-router_v7.2.x/flow_v0.25.x-/koa-router_v7.2.x.js
@@ -1,11 +1,10 @@
 /**
  * A lot of type definitions are copied from Koa because we can't `import type` yet
  * Comments from types removed for brevity. Read the actual `koa` libdef to learn more about these types.
- *
- * TODO: Remove the "koa" module from here when flow-typed v3 allows for dependencies between libdefs
  */
 
-declare module "koa" {
+declare module "koa-router" {
+  // Begin `koa` types
   declare class Server extends net$Server {
     listen(
       port?: number,
@@ -168,7 +167,7 @@ declare module "koa" {
       ((name: string) => Context)
   };
 
-  declare export type Context = {
+  declare type Context = {
     accept: $PropertyType<Request, "accept">,
     app: Application,
     cookies: Cookies,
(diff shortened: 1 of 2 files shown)


saadq:
Update: I realized it probably wasn't a good idea to create a `koa` module because if someone installed `koa-router` types without installing `koa` types, it would give them a false positive and allow them to import stuff from `koa`.

So instead of putting it in a separate module, I just copied the types into the module and just stuck the types in like so:

```
declare module "koa-router" {
  // begin koa types
  ... koa types
  // end koa types

  ... koa-router types
}

File: definitions/npm/koa-router_v7.2.x/flow_v0.25.x-/koa-router_v7.2.x.js [hunk 1]
```diff
@@ -1,20 +1,279 @@
...
-  ctx: any,
-  next: () => Promise<void>
-) => Promise<void> | void;
+  declare type JSON = string | number | boolean | null | JSONObject | JSONArray;
+  declare type JSONObject = { [key: string]: JSON };
+  declare type JSONArray = Array<JSON>;
 
-declare module "koa-router" {
-  declare class Router {
+  declare type SimpleHeader = {
```
GAntoine: @saadq Would it make sense to prepend all of these imported `koa` types, with something like `Koa$SimpleHeader`? It might help an internal type from one meant to be exported. Would that break anything?

saadq: Yeah that's a good idea, will do that soon.

saadq:
Hmm, I'm actually noticing issues (even without the changes you requested) if I use a type from Koa in my application like so:

```
import Router from 'koa-router'
import type { Middleware } from 'koa'

function errorHandler(): Middleware {
  return async (ctx, next) => {
    const { response } = ctx
    try {
      await next()
    } catch (err) {
      response.status = err.status || 500
      response.body = err.message
      ctx.app.emit('error', err, ctx)
    }
  }
}

const router = new Router()
router.use('test', errorHandler()) // Throws Flow error when it shouldn't
```

I think there may be an issue due to Flow being nominally typed...

So for now, I think this approach might be too unreliable.

PR #1793, opened 2018-01-31
lodash add %checks to isNil

danielbartsch:
I had to change the `lodash class` to a module, because `%checks` seems [not to be allowed in classes](https://github.com/facebook/flow/issues/4723#issuecomment-340513481)

This ensures that flow gets that `isNil` checks if the parameter is null or undefined. (Same as [in ramda](https://github.com/flowtype/flow-typed/pull/1504))

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
@@ -1,5 +1,5 @@
 declare module "lodash" {
-  declare type HigherOrderFunctionOptional = (funcs?: Array<Function>) => Function
+  declare type HigherOrderFunctionOptional = (...funcs?: Array<Function>) => Function
 
   declare type __CurriedFunction1<A, R, AA: A> = (...r: [AA]) => R;
   declare type CurriedFunction1<A, R> = __CurriedFunction1<A, R, *>;
@@ -737,8 +737,7 @@ declare module "lodash" {
   declare export function before(n: number, fn: Function): Function;
   declare export function bind(func: Function, thisArg: any, ...partials: Array<any>): Function;
   declare export function bindKey(obj?: ?Object, key?: ?string, ...partials?: Array<?any>): Function;
-  declare export var curry: Curry;
-  declare export function curry(func: Function, arity?: number): Function;
+  declare export var curry: Curry | (func: Function, arity?: number) => Function;
   declare export function curryRight(func: Function, arity?: number): Function;
   declare export function debounce<F: Function>(func: F, wait?: number, options?: DebounceOptions): F;
   declare export function defer(func: Function, ...args?: Array<any>): TimeoutID;
@@ -1362,8 +1361,8 @@ declare module "lodash" {
     // NaN is a number instead of its own type, otherwise it would behave like null/void
   declare export function defaultTo<T1: number, T2>(value: T1, defaultValue: T2): T1 | T2;
   declare export function defaultTo<T1: void | null, T2>(value: T1, defaultValue: T2): T2;
-  declare export var flow: HigherOrderFunctionOptional;
-  declare export var flowRight: HigherOrderFunctionOptional;
+  declare export var flow: HigherOrderFunctionOptional | $ComposeReverse;
+  declare export var flowRight: HigherOrderFunctionOptional | $Compose;
   declare export function identity<T>(value: T): T;
   declare export function iteratee(func?: any): Function;
   declare export function matches(source?: ?Object): Function;
@@ -1408,7 +1407,7 @@ declare module "lodash" {
 }
 
 declare module "lodash/fp" {
-  declare type HigherOrderFunction = (funcs: Array<Function>) => Function
+  declare type HigherOrderFunction = (...funcs: Array<Function>) => Function
 
   declare type __CurriedFunction1<A, R, AA: A> = (...r: [AA]) => R;
   declare type CurriedFunction1<A, R> = __CurriedFunction1<A, R, *>;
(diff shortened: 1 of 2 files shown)


GAntoine:
@aaronjensen Could you give this a look over, and let me know if it makes sense?

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
```diff
@@ -3111,10 +3110,10 @@ declare module "lodash/fp" {
   declare export function defaultTo<T1: number, T2>(defaultValue: T2, value: T1): T1 | T2;
   declare export function defaultTo<T1: void | null, T2>(defaultValue: T2): (value: T1) => T2;
   declare export function defaultTo<T1: void | null, T2>(defaultValue: T2, value: T1): T2;
-  declare export var flow: HigherOrderFunction;
-  declare export var pipe: HigherOrderFunction;
-  declare export var flowRight: HigherOrderFunction;
-  declare export var compose: HigherOrderFunction;
+  declare export var flow: HigherOrderFunction | $ComposeReverse;
```
aaronjensen: I don't think this does what you want, this means that any invocation of this function must be compatible with both `HigherOrderFunction` and `$ComposeReverse`. Perhaps `&` is what you're looking for? in other words, I'm not sure how this doesn't fail tests/make this function less usable, but maybe I'm missing something?

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
```diff
@@ -1,5 +1,5 @@
 declare module "lodash" {
-  declare type HigherOrderFunctionOptional = (funcs?: Array<Function>) => Function
+  declare type HigherOrderFunctionOptional = (...funcs?: Array<Function>) => Function
```
aaronjensen: this is a big change, is this correct?

GAntoine:
Thanks a ton for helping out! 💪  If you happen to know someone you feel would be more qualified, feel free to ping them.

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
danielbartsch: [Before](https://github.com/flowtype/flow-typed/pull/1793/files#diff-a986497c6415236202a140ff65ef7c06L3114) it was written as two functions with the same name as class methods.

I tried to recreate that functionality with `declare`s inside a module. But duplicated entries aren't possible. So I thought that's the equivalent. Are you sure that `HigherOrderFunction & $ComposeReverse` are the equivalent of above link?

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
danielbartsch: Oops, yes. Just checked the lodash docs. Before it was also possible with `...funcs`, but now it seems to only work with `funcs?: Array`. I originally wanted to make it more compatible with `$Compose` and `$ComposeReverse`.

Looking into it again.

danielbartsch: `compose` and `pipe` tests fail, if I remove the var args again. I'm not sure why. Probably in connection with your other comment in line 3113.

In the class before, compose was defined like this:
```javascript
  compose: $Compose;
  compose(funcs: Array<Function>): Function;
```
now I tried to recreate it in a module with `declare` like this:
```javascript
declare type HigherOrderFunction = (funcs: Array<Function>) => Function

declare export var compose: HigherOrderFunction | $Compose;
```
But that doesn't seem to be the equivalent of the above class-way

danielbartsch: 001de7a that worked! thanks

aaronjensen:
Cool, the only other thing is that this may break `const _ = require('lodash')`. Could you add a test to ensure that at least one lodash function still works using that? If it does break, I'm not sure what to do, you may not be able to have both the isNil working and the single object export at the same time w/o lots of duplication/boilerplate...

danielbartsch:
It worked with `import _ from 'lodash'`, but not with `const _ = require('lodash')`

aaronjensen:
Yeah, we probably don't want to merge this then as it'll break the type for anyone not using the module syntax.

I wonder if there's something else that can be done, like adding the `isNil` function using the class property syntax? Or maybe you could `&` an object type with the `isNil` declaration with the lodash class? I don't know if `%checks` only doesn't work w/ class methods or if it also doesn't work with object types... some things worth playing with.

danielbartsch:
@aaronjensen does it have to work with `const _ = require('lodash')`? Because I just tried, and it *does* work with `const _ = require('lodash').default`

aaronjensen:
I would say so, yes, but i'm no authority. I'm not sure that `const _ = require('lodash').default` would actually work in node and it'd be strange to force flow users to do that even if it did.

aaronjensen:
That's too bad. Probably not much we can do right now then 😢 

One thing is that I think `lodash-es` may have a resurgence when webpack 4 comes out now that `sideEffects: false` is set. Perhaps you can apply your types to `lodash-es`, since that one is actually meant to be typed as a module.

danielbartsch:
@GAntoine does the work-in-progress label mean that you're working on it?

GAntoine:
Its more of a reminder for me not to close this PR by accident. I'd be a shame to lose the work done here, and it can be picked up again once the issues with `%checks` are addressed in flow.

AndrewSouthpaw:
Closing for now... bummer that it fell stale. 🙁

PR #1794, opened 2018-01-31
Introduce react-navigation libdef

Ashoat:
We (`react-navigation` maintainers) are switching from in-package types to hosting our types on `flow-typed` :)

### DIFF
--- a/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
+++ b/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
@@ -1,28 +1,57 @@
 // @flow
 
-import * as React from 'react';
-import { Animated, Image } from 'react-native';
-import type {
-  StyleObj,
-} from 'react-native/Libraries/StyleSheet/StyleSheetTypes';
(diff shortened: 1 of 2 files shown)


definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
```diff
@@ -0,0 +1,1013 @@
+// @flow
+
+import * as React from 'react';
+import { Animated, Image } from 'react-native';
+import type {
+  StyleObj,
+} from 'react-native/Libraries/StyleSheet/StyleSheetTypes';
+import type { Layout } from 'react-native-tab-view/src/TabViewTypeDefinitions';
```
Ashoat: Are any of these imports problematic? Things seem to work, but in a conversation on Slack @calebmer seemed to indicate that it shouldn't be possible to import from packages like this.

Ashoat:
Okay, I've removed all import statements. This PR should be ready to merge.

saadq:
I had to do something similar for [koa-router](https://github.com/flowtype/flow-typed/pull/1792) in order to get types from `koa` in there. I think there was talk about dependencies between libdefs for flowtyped v3.0 so I guess we can look forward to that.

PR #1796, opened 2018-02-01
React relay variables

alfonsoperez:
Currently, setting arrays as variables doesn't work.
Also, for instance passing function won't trigger any error
I thought this could be done as a recursive type, I think this covers every type of variable?, but perhaps I am missing something

### DIFF
--- a/definitions/npm/react-relay_v1.x.x/flow_v0.47.x-/react-relay_v1.x.x.js
+++ b/definitions/npm/react-relay_v1.x.x/flow_v0.47.x-/react-relay_v1.x.x.js
@@ -349,7 +349,7 @@ declare module "react-relay" {
     connectionConfig: ConnectionConfig
   ): TBase;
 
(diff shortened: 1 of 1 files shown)


mandx:
Instead of:
```javascript
   declare type Variable = string | void | number | Variables | void | Array<Variables>; 
```
Shouldn't it be:
```javascript
   declare type Variable = string | null | number | Variables | void | Array<Variables>; 
```
(Replace one of the two `void`s with `null`)?

GAntoine:
Oh man, I can't believe I missed that. 😅  @mandx Would you be up for opening a quick PR to fix it? If not, let me know and I'll open one tonight.

mandx:
Here it is! See #1803.

However, this changes are problematic beyond the typo, can you take a look at #1802? This is also breaking type-checking in my mutation inputs because of optional (`?`) values...

PR #1811, opened 2018-02-03
Improve hoist non react statics defs

maxmalov:
- Add support for hoisted statics (https://github.com/flowtype/flow-typed/pull/1783#issuecomment-362682624)
- Expand library version to 2.x.x

### DIFF
--- a/definitions/npm/hoist-non-react-statics_v2.x.x/flow_v0.54.1-/hoist-non-react-statics_v2.x.x.js
+++ b/definitions/npm/hoist-non-react-statics_v2.x.x/flow_v0.54.1-/hoist-non-react-statics_v2.x.x.js
@@ -1,0 +1,12 @@
+declare module 'hoist-non-react-statics' {
+  /*
+    S - source component statics
+    TP - target component props
+    SP - additional source component props
+  */
+  declare module.exports: <TP, SP, S>(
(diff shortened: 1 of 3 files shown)


definitions/npm/hoist-non-react-statics_v2.x.x/flow_v0.54.1-/hoist-non-react-statics_v2.x.x.js
```diff
@@ -0,0 +1,12 @@
...
+  /*
+    S - source component statics
+    TP - target component props
+    SP - additional source component props
+  */
+  declare module.exports: <TP, SP, S>(
+    target: React$ComponentType<TP>,
+    source: React$ComponentType<TP & SP> & S,
+    blacklist?: { [key: $Keys<S>]: boolean }
+  ) => React$ComponentType<TP> & S; // TODO: handle blacklisted properties
```
maxmalov: @marcind do you have an idea how the resulting statics keys here can be filtered according to blacklist type?

maxmalov:
@GAntoine I'm not sure we can make these defs stronger, can this PR be merged as is?

GAntoine:
I don't see why not, it's already an improvement over the old definitions.

PR #1815, opened 2018-02-05
data.either: fix wrong static modifier in interface definition 

StefanoMagrassi:


StefanoMagrassi:
a new PR will be opened from a cleaner fork

PR #1824, opened 2018-02-06
Add libdef for detox

coderjonny:
Library definition for Detox. 🙏🏽
https://github.com/wix/detox

related issue: https://github.com/wix/detox/issues/275

### DIFF
--- a/definitions/npm/detox_v7.0.x/test_detox_v7.0.x.js
+++ b/definitions/npm/detox_v7.0.x/test_detox_v7.0.x.js
@@ -1,0 +1,160 @@
+/* @flow */
+
+const detox = require("detox");
+const {device, expect, element, by, waitFor} = require("detox");
+
+const configs = {
+    configurations: {
+        "ios.sim.debug": {
+            binaryPath:
+                "packages/apps/Example/ios/build/Build/Products/Debug-iphonesimulator/example.app",
+            build:
+                "xcodebuild -project packages/apps/Example/ios/Example.xcodeproj -scheme Example -configuration Debug -sdk iphonesimulator -derivedDataPath packages/apps/Example/ios/build",
+            type: "ios.simulator",
+            name: "iPhone X"
+        }
+    }
+};
+
+
+// Detox Object API
+// ----------------
+detox.cleanup();
+// $ExpectError
+detox.init({configurations: ''}, {initGlobals: 123, launchApp: false});
+detox.init(configs, {initGlobals: false});
+
+
+// Device Object API
+// -----------------
+device.launchApp()
+device.relaunchApp()
+device.terminateApp()
+device.reloadReactNative()
+device.sendToHome()
+device.installApp()
+device.uninstallApp()
+// $ExpectError
+device.openURL({url: 'https://flow.org', sourceApp: false})
+device.openURL({url: 'https://flow.org'})
+
+// $ExpectError
+device.sendUserNotification()
+device.sendUserNotification({})
+
+// $ExpectError
+device.setOrientation('normal')
+device.setOrientation('portrait')
+
+// $ExpectError
+device.setLocation()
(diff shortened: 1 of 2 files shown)


definitions/npm/detox_v7.0.x/test_detox_v7.0.x.js
```diff
@@ -0,0 +1,170 @@
...
+// $ExpectError
+expect(element(by.id('UniqueId204'))).toHaveText(123)
+expect(element(by.id('UniqueId204'))).toHaveText('some text')
+// $ExpectError
+expect(element(by.id('UniqueId204'))).toHaveValue(true)
+expect(element(by.id('UniqueId204'))).toHaveValue("1")
+expect(element(by.id('UniqueId204'))).toNotExist()
+
+// waitFor API
+waitFor(element(by.id('UniqueId204'))).toBeNotVisible()
```
rotemmiz: waitFor only works with `withTimeout(Number)`, the below lines will never trigger anything. is there a way to enforce it with flow ?

definitions/npm/detox_v7.0.x/test_detox_v7.0.x.js
```diff
@@ -0,0 +1,170 @@
...
+by.label(123)
+by.type('RCTImageView')
+by.traits(['button'])
+// $ExpectError
+by.traits('button')
+
+
+// Actions on element
+// ------------------
+ element(by.id('testElement')).tap()
```
rotemmiz: The whole API is async (promise based), can flow enforce async usage across the API ?

definitions/npm/detox_v7.0.x/test_detox_v7.0.x.js
```diff
@@ -0,0 +1,170 @@
...
+    NATIVE_BUTTON: "NATIVE_BUTTON",
+    WELCOME_VIEW: "WELCOME_VIEW"
+}
+
+const welcome = element(by.id(e2e.WELCOME_VIEW));
+expect(welcome).toBeVisible();
+const welcome_text = "Shake your phone to open the developer menu.";
+expect(element(by.text(welcome_text))).toBeVisible();
+const btn = element(by.id(e2e.NATIVE_BUTTON));
+btn.tap();
```
rotemmiz: Seems like everything is here. I have a few questions though, written above.

Thanks for taking care of Detox :)

definitions/npm/detox_v7.0.x/test_detox_v7.0.x.js
skovhus: I cannot really see anything we can do here using Flow.

definitions/npm/detox_v7.0.x/test_detox_v7.0.x.js
skovhus: We cannot enforce the using to `away` on these. 

But I think all the `=> {}` and `=> void` should be replace by `=> Promise<void>`.

skovhus:
Really nice work here. 

I do suggest that all async function should return `Promise<void>` instead of `void` and `{}`.

Let me know if you need any help landing this. : )

AndrewSouthpaw:
I'll wait on that to happen before we merge then. :)

adamterlson:
Any motion here?

AndrewSouthpaw:
Just waiting on someone to do the work suggested by @skovhus.

coderjonny:
Hey guys, 

Sorry for the delay .. I've updated name of the folder to test `detox 7.x.x` and updated some more flowTypes. Let me know if that works 👍

definitions/npm/detox_v7.0.x/test_detox_v7.0.x.js
coderjonny: changed return types to `Promise<void>`. 👍

PR #1825, opened 2018-02-08
Update react-navigation libdef

brentvatne:
We just released 1.0, which included a couple of small API changes from our beta release which was previously on here. There is no reason for anyone to continue to use the beta release and so v1.x.x still applies to the version. Additionally, we formatted the libdef using prettier

### DIFF
--- a/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
+++ b/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
@@ -1,7 +1,6 @@
 // @flow
 
 declare module 'react-navigation' {
-
   /**
    * First, a bunch of things we would love to import but instead must
    * reconstruct (mostly copy-pasted).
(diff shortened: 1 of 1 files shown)


definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
```diff
@@ -144,9 +146,37 @@ declare module 'react-navigation' {
...
 
+  declare export type NavigationReplaceAction = {|
+    +type: 'Navigation/REPLACE',
+    +key: string,
+    +routeName: string,
+    +params?: NavigationParams,
+    +action?: NavigationNavigateAction,
+  |};
+  declare export type NavigationPopAction = {|
+    +type: 'Navigation/POP',
```
GAntoine: Are these leading `+` correct?

brentvatne: I believe so, the intention is for them to be immutable cc @ericvicenti

GAntoine: Then this LGTM, give me the go-ahead and I'll merge it in.

PR #1832, opened 2018-02-11
Copy v6 react-i18next to v7

TLadd:
Mostly, this is just copying over the v6 files to v7 for react-i18next. The only breaking change in that release was to, by default, not render a parent tag around the output of the Trans component. Since the parent prop was being set to a default before, the types didn't change any.

A couple of minor updates were added that I modified the types to handle as well:
1. Add function namespace argument to translate HOC (https://github.com/i18next/react-i18next/commit/28a841b7598855a9049d6d07dbd6d10467b9a5a8)
2. Add usePureComponent to TranslateOptions (https://github.com/i18next/react-i18next/commit/39e9468bd4a716757564176257958c088fa9c22c)

### DIFF
--- a/cli/src/lib/npm/__tests__/npmLibDefs-test.js
+++ b/cli/src/lib/npm/__tests__/npmLibDefs-test.js
@@ -300,6 +300,7 @@ describe('npmLibDefs', () => {
   describe('findNpmLibDef', () => {
     describe('when no cached libDefs found', () => {
       it('returns null', async () => {
+        jest.setTimeout(10000);
         const pkgName = 'jest-test-npm-package';
         const pkgVersion = 'v1.0.0';
         const flowVersion = {kind: 'all'};
--- a/definitions/npm/react-i18next_v7.x.x/flow_v0.53.x-/react-i18next_v7.x.x.js
+++ b/definitions/npm/react-i18next_v7.x.x/flow_v0.53.x-/react-i18next_v7.x.x.js
@@ -1,0 +1,82 @@
+declare module "react-i18next" {
+  declare type TFunction = (key?: ?string, data?: ?Object) => string;
+
+  declare type TranslatorProps = {
+    t: TFunction,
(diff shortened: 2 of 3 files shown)


definitions/npm/react-i18next_v6.x.x/test_react-i18next_v6.x.x.js
```diff
@@ -50,16 +50,12 @@ class ClassComp extends React.Component<Props> {
 }
 
 // $ExpectError - wrong argument type
-const FlowErrorComp = ({ s, t }: Props) =>
```
TLadd: This became an unused suppression in flow v65 only. I believe it was just because the line the error occurs on changed.

TLadd:
Build keeps failing on Node 6 only on the test

```
npmLibDefs › findNpmLibDef › when no cached libDefs found › returns null
    Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.
      
      at node_modules/jest-jasmine2/build/queue_runner.js:72:21
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:633:19)
```

Not sure why.

MichaelDeBoey:
CC @GAntoine

cli/src/lib/npm/__tests__/npmLibDefs-test.js
```diff
@@ -300,6 +300,7 @@ describe('npmLibDefs', () => {
   describe('findNpmLibDef', () => {
     describe('when no cached libDefs found', () => {
       it('returns null', async () => {
+        jest.setTimeout(10000);
```
GAntoine: Makes sense 👍

GAntoine:
That specific specs fails on a bell curve. Setting the timeout to 10K won't guarantee it'll always pass, but as long as there are no side effects I think it's a good change. Accepted, will merge in.

PR #1843, opened 2018-02-14
libdef firebase: Add auth().setPersistence

adambrgmn:
Hi, first PR here. And thanks for all previous work!

[firebase.auth().setPersistence](https://firebase.google.com/docs/reference/js/firebase.auth.Auth#setPersistence) was missing in the firebase_v4.x.x libdef. But it's quite easy to add. So instead of opening an issue first I decided to implement it and send a PR. Hope that's okey!

Tests are added and passing.

Have a nice day!


adambrgmn:
☺️👍

PR #1851, opened 2018-02-16
libdef lodash: Fix flow, flowRight, pipe, and Compose

omeid:
For flow 55.x and over, the lodash typedefs are using $Compose
and $ComposeReverse but it was overrriden by a member type.

This commit actually use function overloading to support both
$Compose, $ComposeReverse, and lodash array variant.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
@@ -1371,10 +1371,8 @@ declare module "lodash" {
     // NaN is a number instead of its own type, otherwise it would behave like null/void
     defaultTo<T1: number, T2>(value: T1, defaultValue: T2): T1 | T2;
     defaultTo<T1: void | null, T2>(value: T1, defaultValue: T2): T2;
-    flow: $ComposeReverse;
-    flow(funcs?: Array<Function>): Function;
-    flowRight: $Compose;
-    flowRight(funcs?: Array<Function>): Function;
+    flow: ($ComposeReverse & (funcs: Array<Function>) => Function);
+    flowRight: ($Compose & (funcs: Array<Function>) => Function);
     identity<T>(value: T): T;
     iteratee(func?: any): Function;
     matches(source?: ?Object): Function;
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw:
@villesau Want to provide context on this one? I work less in lodash.

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
```diff
@@ -3130,13 +3128,10 @@ declare module "lodash/fp" {
...
-    flow(funcs: Array<Function>): Function;
-    pipe: $ComposeReverse;
-    pipe(funcs: Array<Function>): Function;
-    flowRight: $Compose;
-    flowRight(funcs: Array<Function>): Function;
-    compose: $Compose;
+    flow: ($ComposeReverse & (funcs: Array<Function>) => Function);
+    pipe: ($ComposeReverse & (funcs: Array<Function>) => Function);
+    flowRight: ($Compose & (funcs: Array<Function>) => Function);
+    compose: ($Compose & (funcs: Array<Function>) => Function);
```
villesau: Funny fact: Flow doesn't have such function as compose in 4.x.x version, it's flowRight now :)

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
```diff
@@ -3130,13 +3128,10 @@ declare module "lodash/fp" {
...
     defaultTo<T1: void | null, T2>(defaultValue: T2, value: T1): T2;
-    flow: $ComposeReverse;
-    flow(funcs: Array<Function>): Function;
-    pipe: $ComposeReverse;
-    pipe(funcs: Array<Function>): Function;
-    flowRight: $Compose;
-    flowRight(funcs: Array<Function>): Function;
-    compose: $Compose;
+    flow: ($ComposeReverse & (funcs: Array<Function>) => Function);
+    pipe: ($ComposeReverse & (funcs: Array<Function>) => Function);
```
villesau: Pipe actually doesn't exist either.

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
omeid: Yeah, I wasn't sure about that either, I just fixed up what was there. So remove it altogether?

omeid: Actually, for the `lodash/fp`, (4.x.x) `compose` is an alias to `flowRight`, so it is there.

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
omeid: Same as my previous comment, this is for `lodash/fp` and `pipe` is basically an alias `flow`.

villesau: Neat! I wasn't aware that there are some aliased functions. IN this case everything seems good!

omeid: S no longer `needs-udpates`? ;)

villesau: @omeid yep! Unfortunately I'm still lacking rights for the repo so there is not much I can do except pinging @AndrewSouthpaw :D

PR #1854, opened 2018-02-17
Fix failing regression spec

GAntoine:
From the release notes in version 66: `Error message redesign to help you better debug typing issues in your programs.`


GAntoine:
Needs to wait till travis starts using flow_v66

PR #1855, opened 2018-02-17
Add babel-core peer dependency, fix Jest breaking.

AndrewSouthpaw:
Based on fix [here](https://github.com/facebook/jest/issues/5542#issuecomment-366433503).

Likely that the regression issue will still cause the build to break (#1854), so this PR is gated on that one.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -42,7 +42,9 @@
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
+    "babel-core": "^6.26.0",
     "babel-eslint": "^8.2.1",
+    "babel-jest": "^22.2.2",
     "babel-plugin-syntax-async-functions": "^6.5.0",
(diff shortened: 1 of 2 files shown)


cli/package.json
```diff
@@ -54,15 +56,21 @@
     "eslint-plugin-prettier": "^2.5.0",
     "flow-bin": "0.64.0",
     "jest": "^22.1.4",
-    "prettier": "^1.9.2"
+    "prettier": "^1.9.2",
+    "regenerator-runtime": "^0.11.1"
```
SimenB: this shouldn't be needed for yarn

GAntoine:
1854 has been shipped, but I expect the build to fail for other reasons. Try rebasing on master; if the spec here fails on an issue with ramda (and not the regression), you're god to merge it in.

PR #1860, opened 2018-02-19
Fix react-dom/test-utils annotations for functions that throw exceptions

### LINKED ISSUE #1852: Invalid type annotations in flow lib: `react-dom/test-utils`
Lots of functions in `react-dom/test-utils` are throwing therefore annotating return values with `?Element` is wrong. This is not the case with definitely typed annotations though.

Such as for example this code must compile without any dance:

```
const domNode = ReactTestUtils.findRenderedDOMComponentWithTag(component, 'input');

Simulate.mouseDown(domNode);
```

Yet this results into `domNode – null or undefined. This type is incompatible with the expected param type of`.

pronebird:
Closes #1852


pronebird:
Actually it seems that his has to be changed too: 

https://github.com/facebook/flow/blob/master/lib/react-dom.js

villesau:
I assume this is not moving forward -> closing. Feel free to reopen!

PR #1865, opened 2018-02-20
sha3 definition (fixes eslint error in master)

EdwardDrapkin:
Add a definition for the SHA3 library here: https://github.com/phusion/node-sha3

### DIFF
--- a/definitions/npm/sha3_v1.2.x/flow_v0.56.x-/sha3_v1.2.x.js
+++ b/definitions/npm/sha3_v1.2.x/flow_v0.56.x-/sha3_v1.2.x.js
@@ -5,7 +5,7 @@ declare module "sha3" {
     digest(type?: 'hex' | 'binary'): string;
   }
 
-  declare module .exports: {
+  declare module.exports: {
     SHA3Hash: Class<SHA3Hash>
   }
 }
--- a/definitions/npm/sha3_v1.2.x/flow_v0.56.x-/sha3_v1.2.x.js
+++ b/definitions/npm/sha3_v1.2.x/flow_v0.56.x-/sha3_v1.2.x.js
@@ -1,8 +1,8 @@
 declare module "sha3" {
   declare class SHA3Hash {
-    constructor(hashLength: 224 | 256 | 384 | 512): SHA3Hash;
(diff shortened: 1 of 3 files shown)


EdwardDrapkin:
Build is gonna fail because eslint in master is broken

EdwardDrapkin:
I dunno if it's a bug with eslint or what but eslint was erroring because another typedef had raw JSX in the same scope as a function declaration, so I fixed that by just wrapping the JSX in a function call.  Shouldn't affect that def's flow tests at all.

definitions/npm/sha3_v1.2.x/flow_v0.56.x-/sha3_v1.2.x.js
```diff
@@ -0,0 +1,11 @@
+declare module "sha3" {
+  declare class SHA3Hash {
+    constructor(hashLength?: 224 | 256 | 384 | 512): SHA3Hash;
+    update(data: any, inputEncoding?: 'binary' | 'ascii' | 'utf8'): void;
+    digest(type?: 'hex' | 'binary'): string;
+  }
+
+  declare module .exports: {
```
GAntoine: Remove the space in `module .exports`

GAntoine:
> I dunno if it's a bug with eslint or what but eslint was erroring because another typedef had raw JSX in the same scope as a function declaration, so I fixed that by just wrapping the JSX in a function call. 

❤️

definitions/npm/sha3_v1.2.x/flow_v0.56.x-/sha3_v1.2.x.js
```diff
@@ -5,7 +5,7 @@ declare module "sha3" {
     digest(type?: 'hex' | 'binary'): string;
   }
 
-  declare module .exports: {
+  declare module.exports: {
```
GAntoine: 👍

definitions/npm/sha3_v1.2.x/flow_v0.56.x-/sha3_v1.2.x.js
EdwardDrapkin: fixed, I blame setting the save hotkey to 'format and save' in jetbrains >.>

PR #1866, opened 2018-02-20
jsonwebtoken definition

EdwardDrapkin:
Adds a definition for jsonwebtoken from here: https://github.com/auth0/node-jsonwebtoken

### DIFF
--- a/definitions/npm/jsonwebtoken_v8.1.x/flow_v0.56.x-/jsonwebtoken_v8.1.x.js
+++ b/definitions/npm/jsonwebtoken_v8.1.x/flow_v0.56.x-/jsonwebtoken_v8.1.x.js
@@ -1,0 +1,109 @@
+declare module "jsonwebtoken" {
+  declare module.exports: {
+    sign: jwt$Sign,
+    decode: jwt$Decode,
+    verify: jwt$Verify,
+    JsonWebTokenError: Class<jwt$WebTokenError>,
+    NotBeforeError: Class<jwt$NotBeforeError>,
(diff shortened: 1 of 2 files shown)


definitions/npm/jsonwebtoken_v8.1.x/flow_v0.56.x-/jsonwebtoken_v8.1.x.js
```diff
@@ -0,0 +1,109 @@
+declare module "jsonwebtoken" {
+  declare module.exports: {
```
GAntoine: Up to you, but I'd place this declaration near the bottom of the file, below the declarations for `jwt$Sign`, `jwt$Decode`, etc...

EdwardDrapkin: I just noticed your comment here; it's not at all important to me, it just works out that way naturally when I write the file (I start with the module declaration and fill out the file from the top down).  Is it stylistically encouraged to put the module declaration at the bottom of the file?  I hadn't given it a single brainwave before reading your comment, so it doesn't matter to me at all.   You already merged this though, so you let me know if you want me to refactor and open a new PR.

GAntoine: Nah, I mean it's a question of personal preference. I don't enforce minor stylistic choices like those, it's mostly out of habit when writing reviews. I merged it in since I was happy with it, regardless of the location of the `export`. 👍

PR #1867, opened 2018-02-21
add react-color

saadq:
Edit: Am I doing something wrong to make Travis fail? Tests seem to pass for me.


GAntoine:
@saadq You'll need to rebase on the latest version of master, that should make the specs pass.

PR #1868, opened 2018-02-21
libdef: react-router-v4: Improve withRouter HOC

omeid:
withRouter is basically a HOC component that injects three props,
history, location, and match. This commit improves the type def for
withRouter so that it returns a React$ComponentType that accepts
Props minus any of the injected props, it uses $Exact to avoid any
extraneous props.


omeid:
I have update the commit, withRouter no longer enforces `$Exact` on returned comp prop.

definitions/npm/react-router-dom_v4.x.x/flow_v0.53.x-/react-router-dom_v4.x.x.js
```diff
@@ -73,16 +73,16 @@ declare module "react-router-dom" {
...
     url: string
   };
 
   declare export type ContextRouter = {|
-    history: RouterHistory,
-    location: Location,
-    match: Match
+    history: RouterHistory | void,
+    location: Location | void,
+    match: Match | void,
```
saadq: I think the convention is to use [optional types](https://flow.org/en/docs/types/objects/#toc-optional-object-type-properties) if the property may not exist.

For example,

```js
history: RouterHistory | void,
```

would be

```js
history?: RouterHistory,
```

omeid: The [`$Diff`](https://flow.org/en/docs/types/utilities/#toc-diff) documentation and [HOC](https://flow.org/en/docs/react/hoc/#toc-injecting-props-with-a-higher-order-component) guide seems to prefer the `| void` notation. I also found lots of definitions using the same pattern so I went with it, happy to change it.

saadq: Oh, it's perfectly fine with me, I just thought the `?` was more convenient so just wanted to mention it in case you didn't know. Think you just need to get tests to pass.

AndrewSouthpaw: [Both seem to work fine](https://flow.org/try/#0MYewdgzgLgBAhgLhgEgCIEsBmmA8BvAXwBoY8ZMQQkwBXAWwCMBTAJxgB8YA3EdAExgEAfDAC8pAgChQkWAyRosuQiTIUqMAPy1GrQSPGFJQA), and I'd prefer using maybe types, i.e.

```
history: ?RouterHistory
```

@saadq note that there's a difference between optional props (`foo?: number`) and maybe types (`foo: ?number`). We want the latter.

@omeid if you're willing to change it, please do so. I'm fine with leaving it in if you hate it though. 😉

AndrewSouthpaw: Also checking with flow. [Issue](https://github.com/facebook/flow/issues/5902)

omeid: Thanks @AndrewSouthpaw but I believe we do not want Maybe Types, the props in question here is never null.

And as for `property? : T` vs  `property: T | void`, they're actually handled differently, the first one also includes undefined, so you get weird behavior with diff.

```jsx
 // the error is not caught!
const a1: $Diff<{|p: number|}, {p?: number}> = {p: 1}

 // errors as expected: 
 // Cannot assign object literal to `a2` because property `p` is missing in
 // object type but exists in object literal.
const a2: $Diff<{|p: number|}, {p: number | void}> = {p: 2}
```

AndrewSouthpaw: Yep, agreed.

AndrewSouthpaw:
It looks like the build is still failing, can you look into it?

omeid:
Yes, I am trying to figure out why.

omeid:
The new code only works on flow 0.57.0 and above. Do I create
`definitions/npm/react-router{,-dom,-native}_v4.x.x/flow_v0.57.x-`?

And move
`definitions/npm/react-router{,-dom,-native}_v4.x.x/flow_v0.53.x-` to 
`definitions/npm/react-router{,-dom,-native}_v4.x.x/flow_v0.53.x-v0.57.x`?

AndrewSouthpaw:
Correct, you'll need to cap the directories.

Looks like it's still failing?

omeid:
Yup, I will dig into it.

FezVrasta:
I just wanted to let you know that the new `withRouter` definition is still not correct. 

On my code I get this error pointing to the `enhance(Configure)` code.

```
[flow] Cannot call `enhance` with `Configure` bound to `a` because number [1] is incompatible with undefined [1] in property `bodyIndex` of the first argument. (References: [1])
```

If I get rid of `withRouter` and leave other Recompose HOCs it works just fine...

omeid:
@FezVrasta Can you give a more complete example is what is enhance and Configure?

FezVrasta:
It's any code like:

```
import { compose, defaultProps, withProps, type HOC } from 'recompose';
import { withRouter } from 'react-router';

const enhance = compose(
  withRouter,
  defaultProps({ foo: 'bar' }),
  withProps(props => ({ baz: 1 }),
);

const Comp = props => null;

export default enhance(Comp);
```

heavenshell:
Any update?

omeid:
Sorry guys, I have been rather busy to update this, will get on to soonish.
If you would like to contribute, let me know and I will give you push permission to the fork.

mike1808:
I'm using this annotation and it works for me

```
  declare export function withRouter<C: React$ComponentType<*>>(
    Component: C
  ): React$ComponentType<$Diff<React$ElementConfig<C>, ContextRouter>>;

```

GAntoine:
Closing since the definition for `react-router-native` has changed quite a bit.

omeid:
@GAntoine Can you please link me to what has changed and whatever it addresses the problem this pull request fixes?

FezVrasta:
I don't see how react-router-native relates to this pr..

GAntoine:
There are conflicts with the latest set of changes in `react-router` and `react-router-native.` You'll need to address these before this PR can move forward.

omeid:
I no longer need this.

PR #1870, opened 2018-02-21
react-navigation_v1.x.x: add Header.HEIGHT property

EdwardDrapkin:
The component Header in `react-navigation` has a static HEIGHT property that wasn't in the flow definition.

### DIFF
--- a/cli/src/commands/__tests__/runTests-test.js
+++ b/cli/src/commands/__tests__/runTests-test.js
@@ -24,7 +24,7 @@ describe('run-tests (command)', () => {
     });
 
     it('console logs about unused suppression', async () => {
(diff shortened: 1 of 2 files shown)


EdwardDrapkin:
I'm up to date in master and the PR from earlier today built successfully, not sure why this one is failing... there haven't been any changes to master outside of definitions since then!

EdwardDrapkin:
Looks like the output format changed for Flow and/or Travis was truncating output and/or the test was truncating output running in Travis (passed locally) so I changed  the test to test for "Unused suppression" instead of "Unused suppression comment."  This should fix the build and I can't imagine a scenario where this false positives or false negatives as a result of this change.

PR #1500, opened 2017-11-08
Update redux-form_v5.x.x.js

lvdang:
Change to be compatible with latest flow v58.0

PR #1872, opened 2018-02-22
[redux-form] Update redux-form for flow 53+

GAntoine:
Replaces #1500 

Based of the work of @lvdang


GAntoine:
@villesau Can you give me a hand with this last error?

villesau:
@GAntoine you might want to use ComponentType instead? https://flow.org/en/docs/react/types/#toc-react-componenttype

definitions/npm/redux-form_v5.x.x/flow_v0.53.x-/redux-form_v5.x.x.js
```diff
@@ -0,0 +1,156 @@
...
+  };
+
+  declare export type FieldInputProps = {
+    input: InputProps,
+    meta: MetaProps,
+  };
+
+  declare type FunctionComponent<P, S> = (props?: P, void) => ?React$Element<any>;
+  declare type ClassComponent<P, S> = Class<React$Component<P, S>>;
+  declare type Component<P> = FunctionComponent<P, any> | ClassComponent<P, any>;
```
villesau: I think all of these three could be replaced with builtin react types: https://flow.org/en/docs/react/types/

definitions/npm/redux-form_v5.x.x/flow_v0.53.x-/test_redux-form_v5.x.x.js
```diff
@@ -0,0 +1,49 @@
...
+    )
+}
+
+const form = (): React$Element<any> => (
+  <form name="name">
+    <input name="foobar" />
+  </form>
+);
+
+const connected: typeof React.Component = reduxForm({
```
villesau: Also here instead of React.Component, React.ComponentType<SomeProps or *> would probably be the way to go.

villesau:
Looks good to me! Seems that redux-form is already in version 7.2.3 though. It has some api changes compared to this.

GAntoine:
Thanks for the review! Yeah this is mostly a stepping stone to get us to 7.x.

lvdang:
@GAntoine @villesau  good job guys. I was on vacation, got little distracted with other stuff. Thanks for your help.

ryami333:
Nice, looking forward to types for 7.X. The maintainer has made a token attempt to publish built-in types, but his build-step is a bit broken so you can't import `FormProps` for example. I have offered to fix this, but he seems a bit AWOL at the moment.

PR #1552, opened 2017-11-19
Added lib defs for apollo-link-http@1.2.x

ctgardner:
Seeking clarity about a few things before merging. (Sorry. First-time lib def author/contributor.)

PR #1873, opened 2018-02-22
[apollo-link-http] Partial definition for apollo-link-http

GAntoine:
Replaces #1552


GAntoine:
@ctgardner I've built types for `ApolloLink` and `RequestHandler`, based on the `ts` ones. Could I get you to look over the changes?

PR #1659, opened 2017-12-18
Added typedef for nextjs v4 and Update typedef for next-routes

Jarlakxen:

PR #1878, opened 2018-02-23
[next][next-routes] New and Updated definitions

GAntoine:
Updates and fixed the great work done by @Jarlakxen on #1659

- Adds a definition for next_v4
- Updates the definition for next_v2, and increases lowest flow version
- Updates the definitions for next-routes


GAntoine:
Thanks for the reviews!

PR #1884, opened 2018-02-24
Redis: update callback type annotations

stouf:
Update all the callback annotations so the result argument isn't a maybe type when it doesn't need to.

Please see https://github.com/flowtype/flow-typed/pull/1755#discussion_r169025106 for the discussion that lead to this pull request.

cc @jedwards1211


GAntoine:
@stouf You'll need tor rebase and fix the conflicts.

stouf:
@GAntoine Done :+1:

PR #1889, opened 2018-02-25
Libdefs for React bootstrap

villesau:
Types derived from here: https://react-bootstrap.github.io/components/alerts/


dperetti:
Thanks for sharing this. Unfortunately FormControl's inputRef is wrong. I guess it should be of type `?HTMLInputElement`.

dperetti:
Thanks for sharing this. Unfortunately FormControl's inputRef is wrong. I guess it should be of type `React.Ref<typeof HTMLInputElement>`. (cf https://flow.org/en/docs/react/types/#toc-react-ref)
Same kind of issue for Checkbox and Radio.

PR #1897, opened 2018-02-26
Fix mixed/any problems in socket.io-client

denizdogan:
Fix problems where `mixed` should have been `any`


GAntoine:
Can you explain the change with an example? Generally we try to avoid changes types to `any`, where possible.

denizdogan:
@GAntoine Basically `send` can take any arguments of any type in socket.io, same for `emit` save for the first argument. The callback function in `on` can also take any arguments of any types, because they're determined by what the server sends to the client. Maybe I've misunderstood the difference between `mixed` and `any`, but I think I got this right.

The alternative to using `any` would be generic types I guess, but it's not entirely clear to me how this would be implemented right now. Either way, in the current state of the libdef, type errors are being produced.

PR #1900, opened 2018-02-26
add a few koa libdefs

saadq:
Would be nice if it was possible to use types from other libdefs, but I guess this is better than having no libdefs for them at all.


villesau:
@saadq It's actually possible but the test suite just does not support that at the moment, flow-typed can't find them. In your own projects you can import them inside module declaration.

saadq:
Oh, I see... I just tried it out and you're right.

Aside from the test suite not being able to handle the `import` statements, I imagine this would also require the user to have the necessary types installed as well right? For example if you have `koa-bodyparser` + its libdef installed but you don't have `koa` libdef installed, there would probably be an issue if the `koa-bodyparser` libdef was importing types from `koa`, right?

It would be nice if somehow `flow-typed` would be able to install "dependencies" of libdefs, so that if you do something like `flow-typed install koa-bodyparser` it would also install the `koa` libdef as well if it doesn't already exist.

PR #1907, opened 2018-03-01
libdef: react-router: fix Match.params

omeid:
React-Router params are never null, they either exists or not.


omeid:
Making this as part of #1868 as they're related.

PR #1911, opened 2018-03-03
Missing ramda definitions

bluedaniel:
Fixes already broken unit tests, allow currying for some functions, and adding some missing ramda definitions, including:

- [x]  `endsWith `
- [x]  `reduceWhile `
- [x]  `innerJoin `
- [x]  `lens`
- [x]  `lensIndex`
- [x]  `lensPath`
- [x]  `lensProp`
- [x]  `view`
- [x]  `set`
- [x]  `over`
- [x]  `mergeDeepLeft`
- [x]  `mergeDeepRight`
- [x]  `mergeDeepWith`
- [x]  `mergeDeepWithKey`
- [x]  `memoizeWith `
- [x]  `applyTo`
- [x]  `o`

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
@@ -1566,8 +1566,9 @@ declare module ramda {
 
   declare function over<T, V, U>(lens: Lens, x: (any) => mixed, val: V): U;
   declare function over<T, V, U>(
-    lens: Lens
-  ): (x: (any) => mixed) => (val: V) => U;
+    lens: Lens,
+    ...rest: Array<void>
+  ): ((x: (any) => mixed, ...rest: Array<void>) => (val: V) => U) & ((x: (any) => mixed, val: V) => U);
 
   declare function path<V>(
     p: Array<mixed>,
@@ -1691,7 +1692,10 @@ declare module ramda {
   ): Array<$ElementType<O, T>>;
 
   declare function set<T, V, U>(lens: Lens, x: T, val: V): U;
-  declare function set<T, V, U>(lens: Lens): (x: T) => (val: V) => U;
+  declare function set<T, V, U>(
+    lens: Lens,
+    ...rest: Array<void>
+  ): ((x: (any) => mixed, ...rest: Array<void>) => (val: V) => U) & ((x: (any) => mixed, val: V) => U);
 
   declare function toPairs<T, O: { [k: string]: T }>(
     o: O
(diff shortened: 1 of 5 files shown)


AndrewSouthpaw:
Looks like some new defs are missing tests, please cover them all. Great work!

functionalStoic:
@bluedaniel Do you have any input or suggestions that you might want to make for issue in this PR 

https://github.com/flowtype/flow-typed/issues/1899

Thanks for the contribution!

wpcarro:
Thanks for adding these! I'm having trouble with the `over` function when consuming these defns.

```javascript
return over(Lens.sources, append(initialSource), state)
// => Cannot call `over` because no more than 1 argument is expected by function type [1].
```

I'm new to Flow, but it seems to be having trouble with the partially applied `append` function.

I'm having similar issues with `set`.

```javascript
return set(Lens.activeSourceIndex, action.index, state)
// => Cannot call `set` because no more than 1 argument is expected by function type [1].
```

If I was more familiar with Flow and/or type systems, I would've taken a deeper dive to figure out why this is happening. For now I can only report the errors... :(

AndrewSouthpaw:
@bluedaniel I'll let you decide whether you want to put @wpcarro's comment in scope for this PR.

bluedaniel:
@wpcarro can you post a full gist that includes what Lens.sources/initialSource and State is?

Here are the tests I added for lenses and view/set/over -
 https://github.com/flowtype/flow-typed/pull/1911/commits/12cddc544251bd87794b03e39a42536348334ce7#diff-ef51b6042b66ae83063751a2cce60c21

wpcarro:
```javascript
// Lens.sources
const sources: Lens<State, Array<Source>> = lensProp('sources')

// State
type State = {|
  activeSourceIndex: ?number,
  sources: Array<Source>,
|}

// initialSource
const initialSource: Source = {
  name: null,
  description: null,
  fileType: 'csv',
  recordSets: [],
  fields: {
    identifiers: [],
    attributes: [],
    retired: [],
  },
}
```

A lot of the errors seemed to go away, when I removed the overloaded defns for `over`, `view`, `set`. I'm not sure if that was the cause of the problem. I think those are there to support currying, right?

Sorry I don't have a smaller example to repro these errors.

bluedaniel:
@wpcarro I'm not sure to be honest, although this bit of code shouldn't have to be defined because this PR contains definitions for all lenses. So your code could read:

```js
// const sources: Lens<State, Array<Source>> = lensProp('sources')
const sources = lensProp('sources'); // already has definition
```

@functionalStoic This is my first experience with Flow really so that pipe function is a bit out of my depth right now. Seems like Flow could really use some partial/currying features to make these definitions easier.

bluedaniel:
@AndrewSouthpaw I'm happy for the PR to be merged as is

AndrewSouthpaw:
K. You'll need to rebase first.

AndrewSouthpaw:
Nice work @bluedaniel, congrats on your first PR to flowtype. 🎉 Thanks for providing a lot more coverage to Ramda!

wpcarro:
WRT the lens definition. I'm not sure merging this was the right move. The lenses don't seem to work with `compose(..)` and presumably not `pipe(..)`... I'm just not sure the test coverage was extensive enough. I'll report back when I have concrete examples of the shortcomings.

AndrewSouthpaw:
Awesome. Thanks for getting involved, @wpcarro. :) Definitely we can't keep apace with carefully checking everything, so it's up to everyone to constantly evolve the types.

PR #1919, opened 2018-03-06
fix inputRef type annotations for react-bootstrap

villesau:
Better types for inputRefs. based on https://flow.org/en/docs/react/types/#toc-react-ref and https://flow.org/en/docs/react/types/#toc-react-elementref

Pointed out by @dperetti here: https://github.com/flowtype/flow-typed/pull/1889#issuecomment-370586750
fixes this: https://github.com/flowtype/flow-typed/issues/1917


villesau:
Seems that unrelated (older react-redux) libdefs breaks the build :/ Maybe it's because i'm the author of previous merged PR where the broken (old) tests were?

PR #1920, opened 2018-03-06
Only take the last commit to diff

villesau:
Only compare to parent, not grand parent. Not sure why it's previously been ~2.
You can try `git diff HEAD~2 --name-only` and `git diff HEAD~1 --name-only` in master to see that the first one actually returns names for two last merged libdefs instead of one.

Noticed this behaviour here: https://github.com/flowtype/flow-typed/pull/1919 See the travis output where unrelated libdefs fails: https://travis-ci.org/flowtype/flow-typed/jobs/350025576

### DIFF
--- a/cli/src/lib/git.js
+++ b/cli/src/lib/git.js
@@ -75,7 +75,7 @@ export async function getDiff() {
       // We are probably already on master, so compare to the last commit.
       const {stdout: headDiff} = await child_process.spawnP(gitPath, [
         'diff',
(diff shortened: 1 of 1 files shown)


AndrewSouthpaw:
Seems reasonable to me, merging. We can change it back if necessary. @GAntoine

GAntoine:
No need, AFAIk this change makes sense. 👍

PR #1922, opened 2018-03-07
Support jest-enzyme version 5

theneva:
`jest-enzyme` version 5 [has been released](https://github.com/FormidableLabs/enzyme-matchers/pull/211/files?short_path=4ac32a7#diff-4ac32a78649ca5bdd8e0ba38b7006a1e) :tada: and it changes a few of the assertions whose types are bundled in `jest`'s definitions (since the library augments the `expect` object provided by `jest`):

- `toBeEmpty()` and `toBePresent()` have been removed
- `toExist()` and `toBeEmptyRender()` have been added
- `toHave{Style,Prop,State}()` now accept an `Object` as their first argument

Breaking changes in a plugin whose types are bundled inside the definitions of another library (Jest, in this case) are complicated. `toBeEmpty()` and `toBePresent()` will be false positives with `jest-enzyme@5`. `toExist()` as well as `Object` as the first argument to `tohHave{Prop,State,Style}()` will be false positives with `jest-enzyme@4`.

Two questions:

1. Is there any sensible way to indicate that some functions only exist in a specific version of jest-enzyme?
2. Should I add these changes to all of jest's definitions?

PS: npm was yelling at me, so I've also synchronized package.json with package-lock.json.


AndrewSouthpaw:
Looks like there are still some build issues; ping again once it passes. :)

definitions/npm/jest_v22.x.x/flow_v0.39.x-/jest_v22.x.x.js
```diff
@@ -123,14 +123,16 @@ type EnzymeMatchersType = {
...
   toBeDisabled(): void,
   toBeEmpty(): void,
+  toBeEmptyRender(): void,
   toBePresent(): void,
   toContainReact(element: React$Element<any>): void,
+  toExist(): void,
   toHaveClassName(className: string): void,
   toHaveHTML(html: string): void,
-  toHaveProp(propKey: string, propValue?: any): void,
+  toHaveProp: ((propKey: string, propValue?: any) => void) & ((props: Object) => void),
```
GAntoine: You can probably change this to `toHaveProp(propKey: string | Object, propValue?: any): void,` to match the changes in https://github.com/flowtype/flow-typed/pull/1932/files

GAntoine:
@AndrewSouthpaw It seems the changes in https://github.com/flowtype/flow-typed/pull/1928 are breaking definitions for test runners.

GAntoine:
@theneva Can you rebase this against the latest version of master, and possibly make the review changes (if they make sense to you)? That should fix the failing specs.

theneva:
@GAntoine Hello, yes, thanks a lot! I'll get to is ASAP, probably tomorrow.

definitions/npm/jest_v22.x.x/flow_v0.39.x-/jest_v22.x.x.js
theneva: I did that initially, but this version yells at you if you both provide an object as the first argument and a string as the second. I'll change it to match my changes in the other file if that's okay with you :)

theneva:
@GAntoine I've rebased against master, and everything passes now.

I've also updated the files changed in #1932 to match this approach, namely:

1. Only allow a single argument if the first argument to `toHave{Prop,State,Style}` is an object
2. Retain `toBeEmpty()` and `toBePresent()` (which will be false positives for users of `jest-enzyme@5`, but in my opinion is better than false negatives for users of `jest-enzyme@4` considering that the release is still very new).

Don't hesitate to let me know if anything needs to be changed!

definitions/npm/jest_v20.x.x/flow_v0.39.x-/test_jest-v20.x.x.js
```diff
@@ -227,11 +231,13 @@ expect(wrapper).toHaveHTML(true);
...
 expect(wrapper).toHaveProp("test");
 expect(wrapper).toHaveProp("test", "test");
-expect(wrapper).toHaveProp({ test: "test" });
 // $ExpectError
 expect(wrapper).toHaveProp();
 // $ExpectError
 expect(wrapper).toHaveProp(true);
+expect(wrapper).toHaveProp({ test: "test" });
+// $ExpectError
+expect(wrapper).toHaveProp({ test: "test" }, "test");
```
theneva: Making this error is the reason for the complicated type signature, instead of just `toHaveProp(propKey: string | Object, propValue?: any): void`

theneva:
Thanks!

PR #1923, opened 2018-03-07
Temporary test runner workaround for broken v0.67 flow build.

### LINKED ISSUE #1921: Master is broken: tests failing because flow bins don't appear to be fetched
Hey there,

Master appears to be broken in that it doesn't look like it's pulling Flow versions and actually running tests.

To reproduce:
`$ docker run -it node:8 bash`
`[docker]$ git clone https://github.com/flowtype/flow-typed.git &&  cd flow-typed/ &&  ./run_def_tests.sh`
(tests fail)
`ls cli/.flow-bins-cache/`

AndrewSouthpaw:
Fixes #1921, at least temporarily.

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -130,6 +130,14 @@ async function getOrderedFlowBinVersions(
         });
 
         apiPayload.forEach(rel => {
+          // Temporary fix for https://github.com/facebook/flow/issues/5922
+          if (rel.tag_name === 'v0.67.0') {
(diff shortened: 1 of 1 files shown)


villesau:
LGTM. We can revisit the situation when the issue in Flow has been fixed.

villesau:
The pr failed because of this: https://github.com/flowtype/flow-typed/pull/1920

AndrewSouthpaw:
Merged. I'll let that build, then run the test here again.

PR #1927, opened 2018-03-08
Fix broken Ramda tests.

AndrewSouthpaw:
[Review: @LoganBarnett ]

There were a couple tests that were breaking for some versions of flow, which are now fixed. Key changes:

- `omit` tests were rewritten to be more straightforward. The currying isn't always supported in all versions of flow, and I kind of don't care to fix the tests to make it so.
- `propOr` should not use `$Keys` for the prop input, since it shouldn't require that the prop exist on the object; that's the point of `propOr`.

This was done with minimal effort and not more restructuring of the tests, that will come later. I just want the build to work. 😀


LoganBarnett:
@GAntoine This looks solid!

Issue #1842, opened 2018-02-14
Organizing and namespacing tests.


AndrewSouthpaw:
It's been a long-time beef of mine that tests are written as giant, single files with all constants globally namespaced and no descriptors for tests generally. What do people think of adopting a convention in lieu of a proper TDD framework to better organize tests? Something like...

```js
// #add
{
  // it should accept numbers
  {
    const a: number = add(1 + 1)
  }
  
  // it should error when we pass null
  {
    // $ExpectError
    const a: number = add(1 + null)
  }
}
```

And so on. Thoughts? If we did adopt any convention, it'd be important that it could be incrementally adopted.

AndrewSouthpaw:
@LoganBarnett any opinions?

LoganBarnett:
Either works for me, as both of them get their own scope. An advantage of @villesau's approach is you can add external data as parameters and you don't have to necessarily build that data yourself. You could do the `((1: any): TypeToHardToMakeHere)` hijinks, but we want to avoid that IMO. Functions also demand a name. We should definitely comment to elaborate, but a function name is a great start!

Okay, as I was writing this I think I convinced myself I'm leaning to @villesau.

One thing I like doing for a baseline of assertions is to lift examples right out of the docs of the lib. The examples had better pass a type check :) I don't think that's exclusive to either of these choices, but perhaps worth a mention.

AndrewSouthpaw:
Yeah I was noticing that as well, it's convenient to be able to pass in parameters that are already typed. Cool, I'll make a note in the contributing readme.

AndrewSouthpaw:
Thinking about this some more... I don't like how function names are hard to read as natural language, but they are useful for testing. How about:

```js
{ // it should do something
  const test = (a: HardToConstructType) => {
    ...
  }
}

LoganBarnett:
Whoops, slow on my email. I was a little surprised about the artificial scope, but I see that it lets you reuse the `test` variable.

I'm all about being expressive with documentation, but a pure comment is _discipline_ based. A function name requires you to say _something_ about the test even if it's limited in its capability to do so.

But if you still think it's a better route to go, let's do it! I don't have a super strong opinion here, and it's not like we're locked out of changing our minds later :) My general preference is to move us forward and either route will provide that.

villesau:
[...]

PR #1928, opened 2018-03-08
Add describe and it verbs as TDD building blocks for tests.

AndrewSouthpaw:
Based on #1842. Adds `describe` and `it` as building blocks to test to allow for more familiar TDD test writing, reducing scope, and being more descriptive. Simple enough solution. I'm not wedded to it, but I wanted to see if it'd work out, and get people's feedback.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,112 +1,18 @@
-## Library Definition Best Practices
+# Library Definition Best Practices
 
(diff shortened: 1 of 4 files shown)


AndrewSouthpaw:
Humbug. I didn't realize the test files are actually copied to a different location... that's part of why imports get messed up. I'll need to think about this some more.

AndrewSouthpaw:
Okay, fixed! @GAntoine can you have another look?

PR #1931, opened 2018-03-08
Adds definitions for package redux-rsaa-middleware.

alessioprestileo:


alessioprestileo:
I looked at the redux definitions for inspiration, and it looks like they only have tests for what is defined as 'declare export function'.
I don't have any functions among my definitions, so I'm not sure about how I should test them... Could you please point me to an example?

alessioprestileo:
I tried to add a test file, does that make sense?

definitions/npm/redux-rsaa-middleware_v1.x.x/flow_v0.66.x-/redux-rsaa-middleware_v1.x.x.js
```diff
@@ -0,0 +1,29 @@
...
+    finishedSendingAction?: RSAAAction,
+  };
+
+  declare export type RSAAState = Object;
+  
+  declare export type DispatchAPI<A> = (action: A) => A;
+  
+  declare export type Dispatch<A: { type: $Subtype<string> }> = DispatchAPI<A>;  
+  
+  declare export var RSAAMiddleware : Middleware<RSAAState, RSAAAction, Dispatch<RSAAAction>>;
```
GAntoine: Change to `module.exports`

definitions/npm/redux-rsaa-middleware_v1.x.x/flow_v0.66.x-/redux-rsaa-middleware_v1.x.x.js
```diff
@@ -0,0 +1,29 @@
+declare module 'redux-rsaa-middleware' {
+  import type { Middleware } from 'redux';
```
GAntoine: This won't work, as imports aren't currently functional. You can create a non-exported type here which mocks `Middleware`. Something like:

```
type Middleware = (req: http$IncomingMessage, res: http$ServerResponse, next: (err?: ?Error) => mixed) => mixed;
```

definitions/npm/redux-rsaa-middleware_v1.x.x/flow_v0.66.x-/redux-rsaa-middleware_v1.x.x.js
AndrewSouthpaw: I'm curious, why was this necessary?

GAntoine: Well it wasn't, that was my sleepy state talking. I just changes it back in a commit. -.-

PR #1935, opened 2018-03-08
Only mangle test file names when a name collision will occur.

### LINKED ISSUE #1331: No possibility to test behavior with different modules?
Sometimes `flow` works different in the following cases:

1. code in one file
2. code in different files

But it seems that we can't write tests right now that can check the second case.

I tried to fix `react-redux` libdefs. I need to declare a connected React component in one file and check its usage in another to reproduce a bug. I tried to create `TestComponent.js` file near `test_connect.js`, but checker says that I need to use name for file that starts with `test_`. But when I renamed a file to `test_component.js`, I got an error 
```
* react-redux_v5.x.x/flow_v0.54.x- (flow-v0.55.0): Unexpected Flow errors(2):
   Error: 2-test_connect.js:4
     4: import CC3 from './test_CC3';
                        ^^^^^^^^^^^^ ./test_CC3. Required module not found
```

Can we handle the situation somehow?
Can I help?

AndrewSouthpaw:
Mostly closes #1331, by only mangling the test file name when a collision is going to occur. This probably covers 99% of use cases, and can re-visit the issue if we hit the edge cases.

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -132,10 +132,16 @@ async function getOrderedFlowBinVersions(
         apiPayload.forEach(rel => {
           // Temporary fix for https://github.com/facebook/flow/issues/5922
           if (rel.tag_name === 'v0.67.0') {
-            console.log('==========================================================================================')
-            console.log('We are tempoarily skipping v0.67.0 due to https://github.com/facebook/flow/issues/5922')
-            console.log('==========================================================================================')
-            return
+            console.log(
+              '==========================================================================================',
+            );
+            console.log(
+              'We are tempoarily skipping v0.67.0 due to https://github.com/facebook/flow/issues/5922',
+            );
+            console.log(
+              '==========================================================================================',
+            );
+            return;
           }
 
           // We only test against versions since 0.15.0 because it has proper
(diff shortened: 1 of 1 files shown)


cli/src/commands/runTests.js
```diff
@@ -509,7 +516,13 @@ async function runTestGroup(
...
           // i.e. underscore/v1.x.x/test-underscore.js
           //      underscore/v1.x.x/flow-v0.22.x/test-underscore.js
-          const destBasename = idx + '-' + path.basename(filePath);
+          //
+          // Only mangles the name when there's a naming collision. Otherwise, uses the original.
+          const fileName = path.basename(filePath);
+          const destBasename = copiedFileNames.has(fileName)
+            ? `${idx}-${fileName}`
+            : fileName;
+          copiedFileNames.add(destBasename);
```
GAntoine: Whatever happens with `copiedFileNames`?

AndrewSouthpaw: Nothing; it's used as a way to track which file names have been copied over to watch for collisions. See line 522.

GAntoine: Then I'd remove it, if its not being printed and is only for debugging purposes.

GAntoine: Forget all of what I said, this LGTM.

AndrewSouthpaw: Resolved offline, leaving it in. Salient takeaway:

screen shot 2018-03-09 at 3 24 44 am

AndrewSouthpaw:
Rebased.

PR #1937, opened 2018-03-09
Add Chromecast Receiver CAF Framework typedefs

craigrbruce:


craigrbruce:
Hi @GAntoine .. i have made the change, apologies.

craigrbruce:
@GAntoine If the build is failing because there is no NPM package, there actually will never be one. See [here](https://developers.google.com/cast/docs/caf_receiver_basic) .. you always reference the script from their CDN. This may be why this has not been tackled before.

AndrewSouthpaw:
Interesting. Do we actually pull down the npm package for some reason?

You might also try putting it under `definitions/browser` instead of `definitions/npm`, that might do what you want. I don't know for sure, this is uncharted territory for me too.

craigrbruce:
@AndrewSouthpaw The test script finishes up with a check as to whether a corresponding npm package exists. There isn't one. Maybe i'll try the `definitions/browser` option later this weekend.

craigrbruce:
ok, all passing now. Thanks for the tip @AndrewSouthpaw

GAntoine:
@AndrewSouthpaw As you guessed it, definitions without an npm package should be placed under `definitions/browser`. We pull the npm package to verify that it does indeed exist, since we want a 1-1 relationship between a definition under `definitions/npm` and a package on `npm`.

jukben:
Hey, may I ask you how I'm supposed to install this and use it? Via CLI I can't find it. Thanks in advance.

PR #1941, opened 2018-03-09
Jest: type `objectContaining` and peers as if they returned their described types

rattrayalex-stripe:
Coming from https://github.com/facebook/jest/pull/5751
cc @SimenB


GAntoine:
@rattrayalex-stripe Can you rebase this against the latest version of master? That should allow the failing specs to pass.

rattrayalex-stripe:
Done, thanks for the tip @GAntoine

PR #1942, opened 2018-03-09
Removes describe and it from global scopes

GAntoine:
Should fix failing specs in #1922 and #1941

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -306,6 +306,7 @@ async function writeFlowConfig(testDirPath, libDefPath, includeWarnings) {
   const flowConfigData = [
     '[libs]',
     path.basename(libDefPath),
+    path.join(__dirname, '..', 'lib', 'tddFramework.js'),
     '',
     '[options]',
     'suppress_comment=\\\\(.\\\\|\\n\\\\)*\\\\$ExpectError',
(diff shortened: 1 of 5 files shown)


cli/src/commands/runTests.js
```diff
@@ -318,6 +318,13 @@ async function writeFlowConfig(testDirPath, libDefPath, includeWarnings) {
     path.join(testDirPath, '..', '..', 'node_modules'),
   ].join('\n');
   await fs.writeFile(destFlowConfigPath, flowConfigData);
+
+  const tddFilePath = path.join('src', 'lib', 'tdd_framework.js');
```
villesau: This should most likely be `const tddFilePath = path.join(__dirname, '..', 'lib', 'tdd_framework.js');`

cli/src/commands/runTests.js
```diff
@@ -318,6 +318,13 @@ async function writeFlowConfig(testDirPath, libDefPath, includeWarnings) {
...
+  const tddFrameworkFile = (await fs.readFile(tddFilePath)).toString();
```
villesau: I think `const tddFrameworkFile = await fs.readFile(tddFilePath, 'utf-8');` is enough here

villesau:
I was also thinking of approach of having `flow-typed-tdd`, or `flow-typed-test-framework` or something and magnling that into libs? Like `import { describe, it } from 'flow-typed-tdd';`. Would it make sense? From the user perspective slightly more convenient than `import { describe, it } from './tdd_framework';` which might raise questions where the test file comes from as you need to point it with path `./` and your folder does not contain that file.

GAntoine:
From a quick search, the following tests use either one, but define it themselves:
- `angular_v1.4.7`
- `jasmine_v2.4.x`
- `js-verify_v0.8.x`
- `ramda_v0.x.x`

So I don't expect this PR to break anything, and it will fix the broken `jest` tests.

AndrewSouthpaw:
Yeah originally I had gone with the importing approach, but ran into the issue with relative paths in the temporary test folder as @villesau described. Nice idea importing it as a libdef rather than a global scope.

GAntoine:
> Nice idea importing it as a libdef rather than a global scope.

You can thank villesau for that! I had it in the global scope as well, but he put together a demo gist which did the trick: https://gist.github.com/villesau/3acbd1c7d01b1118c9ef3b307aa209f1

PR #1944, opened 2018-03-10
[new-def] Adds definitions and tests to `ua-parser-js`

marioestrada:
Adds Flow definitions to https://github.com/faisalman/ua-parser-js

### DIFF
--- a/definitions/npm/ua-parser-js_v0.7.x/flow_v0.42.x-/ua-parser-js_v0.7.x.js
+++ b/definitions/npm/ua-parser-js_v0.7.x/flow_v0.42.x-/ua-parser-js_v0.7.x.js
@@ -1,0 +1,55 @@
+declare module 'ua-parser-js' {
+  declare type UABrowser = {
+    name: string,
+    version: string,
+  };
+
+  declare type UACpu = {
+    architecture: string,
+  };
+
+  declare type UADevice = {
+    model: string,
+    type:
+      | 'console'
+      | 'mobile'
+      | 'tablet'
+      | 'smarttv'
+      | 'wearable'
+      | 'embedded',
+    vendor: string,
+  };
+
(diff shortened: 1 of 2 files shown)


mjomble:
This typedef doesn't seem to work when using `require` instead of `import`.

As even the README of ua-parser-js uses `var parser = require('ua-parser-js');` in an example, it would be nice if this worked with flow-typed as well.

GAntoine:
@mjomble Try `var parser = require('ua-parser-js').default` and let me know if that works.

mjomble:
@GAntoine if you try it in Node.js REPL:

```
> require('ua-parser-js').default
undefined
```

whereas

```
> require('ua-parser-js')
{ [Function: UAParser]
  VERSION: '0.7.18',
  BROWSER: { NAME: 'name', MAJOR: 'major', VERSION: 'version' },
  CPU: { ARCHITECTURE: 'architecture' },
  DEVICE:
   { MODEL: 'model',
     VENDOR: 'vendor',
     TYPE: 'type',
     CONSOLE: 'console',
     MOBILE: 'mobile',
     SMARTTV: 'smarttv',
     TABLET: 'tablet',
     WEARABLE: 'wearable',
     EMBEDDED: 'embedded' },
  ENGINE: { NAME: 'name', VERSION: 'version' },
  OS: { NAME: 'name', VERSION: 'version' },
  UAParser: [Circular] }
```

GAntoine:
Yeah I was hoping you were using a processor like babel, which AFAIK would work when using `.default`. There's no way to cleanly export both in CommonJS and ES6, and ES6 is the "future", so I'm not keen on switching this definition to a CJS export.

mjomble:
Ah, I didn't know the two modes were mutually exclusive.
In that case, I can just suppress the Flow error until switching to import/export.

PR #1947, opened 2018-03-12
Fix type of Knex$Transaction.rollback

stouf:
# Demo

```js
'use strict';

const knex = require('knex');

const db = knex({
  client: 'mysql',
  connection: {
    host: '127.0.0.1',
    port: 3306,
    user: 'root',
    password: 'root',
    database: 'test-transaction'
  }
});

db.transaction((trx) => {
  return trx.rollback();
})
.then((...args) => {
  console.log('then()', args);
})
.catch((error) => {
  console.error('catch()', error);
});
```

Outputs

```
catch() Error: Transaction rejected with non-error: undefined
  <stack trace>
```

However, passing an Error object to `trx.rollback()` such as

```js
'use strict';

const knex = require('knex');

const db = knex({
  client: 'mysql',
  connection: {
    host: '127.0.0.1',
    port: 3306,
    user: 'root',
    password: 'root',
    database: 'test-transaction'
  }
});

db.transaction((trx) => {
  return trx.rollback(new Error('Something went wrong'));
})
.then((...args) => {
  console.log('then()', args);
})
.catch((error) => {
  console.error('catch()', error);
});
```

gives us the following output:

```
catch() Error: Something went wrong
  <stack trace>
```

### DIFF
--- a/definitions/npm/knex_v0.13.x/flow_v0.38.x-/knex_v0.13.x.js
+++ b/definitions/npm/knex_v0.13.x/flow_v0.38.x-/knex_v0.13.x.js
@@ -1,7 +1,7 @@
 declare class Knex$Transaction<R>
   mixins Knex$QueryBuilder<R>, events$EventEmitter, Promise<R> {
   commit(connection?: any, value?: any): Promise<R>;
-  rollback(): Promise<R>;
+  rollback(?Error): Promise<R>;
   savepoint(connection?: any): Promise<R>;
 }
 
--- a/definitions/npm/knex_v0.13.x/test_knex-v0.13.js
+++ b/definitions/npm/knex_v0.13.x/test_knex-v0.13.js
@@ -69,6 +69,11 @@ knex
   .then(function(result) {})
   .catch(function(err) {});
 
+knex
+  .transaction(function(trx) {
+    trx.rollback();
+  });
+
 /**
  * knex is also an event emitter,
  * See : http://knexjs.org/#Interfaces-Events


AndrewSouthpaw:
So is passing an error into `rollback` optional? If so, you'll want to update the def. Right now you're making it required.

stouf:
Yep, I made it required on purpose. Since knex throws an error if one doesn't pass an `Error` to `rollback()`, it seems to me it should be required no? What do you think?
Knex's documentation is unfortunately not helping us much here =/

AndrewSouthpaw:
So, based on what you said in the original post, it sounds like the rollback works either way, and in the console it specifies whether it's rejected with an Error or a non-error. 

A simple test would be to use rollback without an error and see if it actually rolls back.

stouf:
> based on what you said in the original post, it sounds like the rollback works either way

I don't think one can assume the rollback worked either way in that example since there is no check for that. However, I wrote the following to check it:

```js
'use strict';

const knex = require('knex');
const uuid = require('uuid');

const db = knex({
  client: 'mysql',
  connection: {
    host: '127.0.0.1',
    port: 3306,
    user: 'user',
    password: 'password'
    database: 'dbname'
  }
});

const email = `${uuid.v4()}@test.com`;

db.transaction((trx) => {
  db('user').insert({ email }).transacting(trx)
    .then(() => trx.rollback());
})
  .then(() => {
    console.log('then() called');
  })
  .catch((error) => {
    console.log('catched() called with error', error);
  })
  .then(() => {
    return db('user').count('* as count').where('email', email);
  })
  .then((rows) => {
    console.log(`number of rows: ${rows[0].count}`);
    return db.destroy();
  })
  .catch(console.error);
```

and the output is

```
catched() called with error Error: Transaction rejected with non-error: undefined
    <Stack track>
number of rows: 0
```

With this, I think we can confirm the rollback did indeed work, even without giving an `Error` object to `rollback()`.

> in the console it specifies whether it's rejected with an Error or a non-error.

As we can see in the above as well, my code is the one outputting something to the console. What Knex actually does is explicitly creating an error and throwing it in the promise chain.
I wish I could confirm with the documentation if that error is created because Knex is not being used as expected or just as a default value since calling `transaction.rollback()` returns a rejected Promise, but the only references to `transaction.rollback()` in the documentation are examples that can be interpreted in differently ways.

I don't like to rely on the source code when it comes to getting to know how one should use a given library, but looking at the bellow, it seems that Knex will create a default `Error` object if none is given to `transaction.rollback()`.

https://github.com/tgriesser/knex/blob/fbef9fcec770b5669024f454e38a845989cd7f0c/src/transaction.js#L127-L132

Based on all the above, I think the type of `transaction.rollback()` should be `(?Error) => Promise<R>`.
Let me know what you think about it.

AndrewSouthpaw:
Yep, go ahead and update the diff and this will be good to go. 😊

stouf:
> Yep, go ahead and update the diff and this will be good to go. 😊 Also, nice investigative work.

Updated! Thanks for helping out 😊

>  You should also submit a PR to their docs so others benefit from your work

Definitely a good idea 👍

stouf:
> You should also submit a PR to their docs so others benefit from your work

https://github.com/knex/documentation/pull/98 👍

PR #1952, opened 2018-03-13
Adjust type definitions for react-intl-with-message-id_v2.x.x

marcelmokos:
- renamed because enzyme-react-intl will not work in the project


AndrewSouthpaw:
Build is still failing. :)

GAntoine:
Closed as the changing of the module name isn't correct, or supported.

Issue #1954, opened 2018-03-13
Importing React in definitions break React types


osener:
I noticed some types such as `React.Node` resolving to `any` after starting to use flow-typed, and narrowed it down to this definition: https://github.com/flowtype/flow-typed/blob/b8dd3c3a2040ff2701e1de7b8502f3bcd1459002/definitions/npm/react-helmet_v5.x.x/flow_v0.53.x-/react-helmet_v5.x.x.js#L1

Since almost all definitions use `React$Node`, I wasn't sure if this is the expected behavior, and whether I should report this to Flow or update the type definitions in flow-typed to not import React.

I created a small repo to demonstrate the issue: https://github.com/osener/flow-typed-issue

AndrewSouthpaw:
You're welcome to change it. It's a documented issue, and I've added some further notes in #1956 to clarify further. :)

osener:
@AndrewSouthpaw FYI the FAQ says otherwise https://github.com/flowtype/flow-typed/wiki/FAQs#how-do-i-build-type-definitions-that-depend-on-other-definitions

What surprised me the most here wasn't the lack of safety within libdefs themselves though. The definition I linked above is also overwriting `React.Node` with `type Node = any` in my app code, and it took a while to figure out that an innocuous looking libdef could actually do this.

AndrewSouthpaw:
> What surprised me the most here wasn't the lack of safety within libdefs themselves though.

Word, it's not good. Want to make a PR? 😉

zerobias:
Just not do imports outside of the module declaration itself

This one will works perfectly
```js
declare module 'react-helmet' {
  import type { Node, Element } from 'react'
}```

Also please use global types as little as possible; it leads to incompatibility with actual component types

AndrewSouthpaw:
Yes, it does, *and* we have no smart way to track dependencies between libdefs, which is why we generally discourage it.

Issue #1857, opened 2018-02-18
Is it currently possible to have dependencies between library definitions?


zaynv:
I am pretty confused about this. Issues like #16 and #1494 seem to suggest that this isn't currently doable, but definitions like `react-redux` seem to be doing it. Is `redux` a special case or something that allows for this, or is this actually possible for any library?

EdwardDrapkin:
AIUI, it's not currently possible.

react-redux seems to only be using `React$Component` and `React$Element` which are shipped with Flow itself, e.g. https://github.com/facebook/flow/blob/v0.66.0/lib/react.js#L26

There's a lot more shipped inside Flow itself than you might expect, so I always send my developers a link to the [cheatsheet](https://www.saltycrane.com/flow-type-cheat-sheet/latest/) which is programmatically generated from source so it should be complete.

zaynv:
Hmm, but `react-redux` seems to use types from `redux` like here:

https://github.com/flowtype/flow-typed/blob/2132a22c83c2c7d7a96fe0f977604131758cf18f/definitions/npm/react-redux_v5.x.x/flow_v0.54.x-/react-redux_v5.x.x.js#L1

AndrewSouthpaw:
Hmm, I'm not sure, but my guess would be it's possible, especially since it seems to be working in the case of `react-redux`. It could also create a snarled, complicated mess of interdependencies (which quickly turns into circular ones). What's your use case?

saadq:
I'm actually starting to think this isn't currently possible, and the libdefs that are currently doing it right now like `react-redux`, `body-parser`, etc might actually just be incorrect/broken.

I tried out an extremely simple case by creating a fake module in the `npm/definitions` called `fakemodule`, and just having this:

**fakemodule_v1.x.x.js**

```js
import type { Dispatch } from "react-redux";

declare module "fakemodule" {
  declare export var thing: Dispatch;
}
```

**test_fakemodule_v1.x.x.js**

```js
import { thing } from "fakemodule";

// This stuff shouldn't pass, but it does
(thing: string);
(thing: number);
(thing: boolean);

// Just to make sure tests are actually running
const num: number = '10';
```

The only error message that appears is about `num` being an incorrect type. 

![screen shot 2018-02-22 at 8 24 31 am](https://user-images.githubusercontent.com/5678694/36540831-d377015e-17a9-11e8-8885-b0901e6633d9.png)

[...]

PR #1956, opened 2018-03-13
Clarify that raw React types can be used in libdefs.

### LINKED ISSUE #1795: Unexpected `any`-typing of imports
It seems that many possible `import` and `import type` statements in libdefs can fail silently and end up typing their imports as `any`. Some of these are widely using in existing libdefs.

For instance, `import * as React from 'react';` results in `React` and all of its properties being `any`-typed. Apparently you have to use `React$ComponentType` etc., but I can't find any docs that clarify this. The `import * as React from 'react';` pattern is used extensively in many existing libdefs.

`import type` statements will fail without `declare module` blocks, but outside of those blocks they will also result in `any` types.

`import` statement (not `import type`) are only allowed outside `declare module` blocks, and seem to be guaranteed to result in an `any` type.

AndrewSouthpaw:
Closes #1795, #1954.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,8 +8,8 @@ to existing libdefs.
 
 * [Contributing to the definitions repository](#contributing-to-the-definitions-repository)
 * [Writing libdefs tips](#writing-libdefs-tips)
-  * [Avoid `any` when possible](#avoid-any-when-possible)
   * [Don't import types from other libdefs](#dont-import-types-from-other-libdefs)
+  * [Avoid `any` when possible](#avoid-any-when-possible)
   * [Always prefix global variables that aren't really meant to be global](#prefix-global-variables-that-arent-really-meant-to-be-global)
 * [Writing tests](#writing-tests)
   * [Use `describe` and `it` blocks to limit scope](#use-describe-and-it-blocks-to-limit-scope)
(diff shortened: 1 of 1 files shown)


saadq:
Doing `import type` works inside the `module` (as long as it is a built-in type) though right? For example doing stuff like

```js
declare module 'other-module' {
  import type { Node } from 'react';
  import type { Server } from 'http';

  declare var server: Server;
  declare var node: Node;
}
```

Should that be mentioned or should it not be to avoid complexity? 

As a side note it might be worth putting [this cheatsheet](https://www.saltycrane.com/flow-type-cheat-sheet/latest/) somewhere in the `CONTRIBUTING.md` so people know which internal types they can use.

AndrewSouthpaw:
Catch up on #1857 before continuing, in case you haven't. 

> Unfortunately flow-typed does not currently support the second approach as it can't depend on other libdefs during the tests. Other problem is nasty dependencies between library definitions which might become very hard to maintain.

As I interpreted this, it doesn't work full stop. Admittedly, I didn't test it. Have you found otherwise?

saadq:
Yeah, I have done so in some of the definitions I've submitted. Just double checked:

screen shot 2018-03-13 at 2 35 47 pm

screen shot 2018-03-13 at 2 35 49 pm

AndrewSouthpaw:
Ah, good to know.

Well, based on the conversation in #1857, the current ruling is to not encourage dependencies between libdefs, because that could get gnarly very, very quickly. I propose we move forward with this PR, continue the conversation there, and update if needed?

saadq:
Sounds good 👍

AndrewSouthpaw:
Can I get a formal 👍 from @villesau or @GAntoine?

villesau:
It's actually possible to import types which Flow has baked in, like everything from Node and React.

zaynv:
@villesau I think that's what was being discussed above your comment.

PR #1958, opened 2018-03-13
Corrected Redux v3.x.x createStore and Reducer definitions

### LINKED ISSUE #1397: redux.Reducer missing necessary | void
The type Reducer in redux_v3.x.x.js should be:

```JS
declare type Reducer<S, A> = (state: S | void, action: A) => S;
```

A reducer may accept `undefined` as its first argument in order to return an initial state.

e.g.
```JS
const INIT_STATE: StateType = { a: 'b' };
const reducer: Reducer<StateType, Action> = (state: StateType = INIT_STATE, action: Action): StateType => {
  /* etc */
  return state;
};
```

If the user initiates the state with `reducer(undefined, '@@redux/INIT')` then `INIT_STATE` is returned.

Benjamin-Dobell:
Closes #1397 

The two definitions go hand-in-hand, as `createStore` takes a `Reducer` as a parameter and has side-effects specifically surrounding how that `Reducer` will be called. 

# `createStore` type proof

As described in the official [documentation](https://github.com/reactjs/redux/blob/master/docs/api/createStore.md), `createStore`'s `preloadedState` is optional.

# `Reducer` type proof

`createStore`'s `preloadedState` is optional (possibly `undefined`/`void`), as above, it is sent directly to a `Reducer`, specifically the `reducer` provided in the same `createStore` call. Redux is a small library so this can be verified in a straight-forward fashion by visual code inspection, as follows:

1. `reducer` and `preloadedState` are set as `currentReducer` and `currentState` respectively:

    https://github.com/reactjs/redux/blob/v3.7.2/src/createStore.js#L57

2. `currentReducer` called with `currentState` (inside `dispatch(<some action>)`):

    https://github.com/reactjs/redux/blob/v3.7.2/src/createStore.js#L170

### DIFF
--- a/definitions/npm/redux_v3.x.x/flow_v0.55.x-/test_createStore.js
+++ b/definitions/npm/redux_v3.x.x/flow_v0.55.x-/test_createStore.js
@@ -5,7 +5,7 @@ import { createStore } from 'redux'
 type State = Array<number>;
 type Action = { type: 'A' };
 type Store = ReduxStore<State, Action>;
-const reducer = (state: State, action: Action): State => state
+const reducer = (state: State = [], action: Action): State => state
 
 //
 // createStore API
@@ -24,6 +24,7 @@ const store6: Store = createStore(reducer, [1]);
 const store7: Store = createStore(reducer, [1], 'wrong'); // wrong enhancer
 declare var myEnhancer: StoreEnhancer<State, Action>;
 const store8: Store = createStore(reducer, [1], myEnhancer);
+const store9: Store = createStore(reducer, undefined, myEnhancer);
 
 //
 // store members
--- a/definitions/npm/redux_v3.x.x/flow_v0.25.x-v0.27.x/redux_v3.x.x.js
+++ b/definitions/npm/redux_v3.x.x/flow_v0.25.x-v0.27.x/redux_v3.x.x.js
@@ -2,7 +2,7 @@ declare module "redux" {
   declare type State = any;
   declare type Action = Object;
   declare type AsyncAction = any;
-  declare type Reducer<S, A> = (state: S, action: A) => S;
+  declare type Reducer<S, A> = (state: S | void, action: A) => S;
   declare type BaseDispatch = (a: Action) => Action;
   declare type Dispatch = (a: Action | AsyncAction) => any;
   declare type ActionCreator = (...args: any) => Action | AsyncAction;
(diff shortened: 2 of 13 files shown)


Benjamin-Dobell:
Sorry, this is a WIP at this point. Need to clean up a bit.

Benjamin-Dobell:
I had to fix the `combineReducer` tests as they were incorrectly using redux (i.e. omitting initialState).

However, that's done now, so this is now ready to be reviewed and/or merged.

AndrewSouthpaw:
@villesau you might have an opinion here. :)

definitions/npm/redux_v3.x.x/flow_v0.55.x-/test_createStore.js
```diff
@@ -5,9 +5,9 @@ import { createStore } from 'redux'
 type State = Array<number>;
 type Action = { type: 'A' };
 type Store = ReduxStore<State, Action>;
-const reducer = (state: State, action: Action): State => state
+const reducer = (state: State = [], action: Action): State => state
 
-//
+//s
```
Benjamin-Dobell: Aww, man, my `Cmd` key is on the fritz, so I'm leaving random "s" all over the place when I intend to save. I'll force push this away.

villesau:
@Benjamin-Dobell @AndrewSouthpaw Looks good to me :)

colbydehart:
So does this mean that everyone who uses redux types now in their reducers has to do something at the top like 
```
if (state === undefined) {
  return initialState;
}
```
This just broke everything for me and i wanted to know what the suggestion was.

Benjamin-Dobell:
@colbydehart Yes, that's correct. If you weren't already doing so, then unfortunately what you were doing was incorrect.

This is documented in Redux's official documentation, https://redux.js.org/basics/reducers

Excerpt 1.

```
import { VisibilityFilters } from './actions'
​
const initialState = {
  visibilityFilter: VisibilityFilters.SHOW_ALL,
  todos: []
}
​
function todoApp(state, action) {
  if (typeof state === 'undefined') {
    return initialState
  }
​
  // For now, don't handle any actions
  // and just return the state given to us.
  return state
}
```

However, if you're using Babel, there is "sugar" syntax you can use, also demonstrated on the same page.

Excerpt 2:

```
function todoApp(state = initialState, action) {
  // For now, don't handle any actions
  // and just return the state given to us.
  return state
}
```

colbydehart:
oh yeah, default params, duh. That makes sense. Was just wondering. Thanks!

PR #1959, opened 2018-03-13
Do not require component usage to type check in react-redux

villesau:
Fixes issue where non-used components does not type check, see https://github.com/flowtype/flow-typed/issues/1946

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/test_connect.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/test_connect.js
@@ -39,7 +39,7 @@ function testPassingPropsToConnectedComponent() {
   connect(mapStateToProps)('');
 }
 
-function doesNotRequireDefinedComponentToTypeCheck() {
+function doesNotRequireDefinedComponentToTypeCheck1case() {
   type Props = {
     stringProp: string,
   };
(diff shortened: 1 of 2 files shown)


villesau:
@AndrewSouthpaw @maxsalven Do you mind to try this out? Note that this does not fix the potential problem with options, but only the issue where the component is not used.

definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/test_connect.js
```diff
@@ -39,6 +39,100 @@ function testPassingPropsToConnectedComponent() {
   connect(mapStateToProps)('');
 }
 
+function doesNotRequireDefinedComponentToTypeCheck1case() {
```
villesau: I could restructure the tests soonish, but I rather do it in separate PR to keep current changes clear.

AndrewSouthpaw:
I'm curious, how does this address the issue? It looks like you just inlined a few things, did that change the behavior of flow somehow?

villesau:
@AndrewSouthpaw That's a very good question! This is one of those complicated parts of Flow, and I don't have very solid answer for it. I _think_ it shouldn't matter, but in practise it does. It apparently "stores" the type information to type parameter (CP in this case) and "executes" the type check at that time, when previously it "flowed" to outbound component definition when the type is only applied when the type/component is used. So the type check is actually kind of made where it's needed, and when assigning type to type parameter it is considered to be the place to do the check when in case of component it is the usage of component. I haven't experimented, but changing the place of type definition might in some cases have an effect on the place where the error is shown. Sorry if my explanation is not understandable, it's difficult concept which I haven't grasped well enough to explain it properly :D

AndrewSouthpaw:
Nah, that makes plenty sense. :) 

Looks like the build is failing, is that a regression or a fluke?

villesau:
Build failure seems to be blast from the past: It's again the previous libdefs. Which is weird as this PR does not have anymore anything to do with them. Apparently the code picking the libs based on changes is a bit loose.

villesau:
ping @AndrewSouthpaw

AndrewSouthpaw:
Build was failing, restarting...

AndrewSouthpaw:
(Sorry for the delay.) So it looks like the tests are still failing, and now it's properly related to the connect libdefs. @villesau can you dig into it more?

villesau:
@AndrewSouthpaw the test failures are related to _old_ libdef, `react-redux_v5.x.x/flow_v0.54.x-v0.61.x` when the changes are made for `react-redux_v5.x.x/flow_v0.62.0-` To be honest I'm not sure what is the reason since this does not touch them. Maybe libdef matching functionality is a bit loose?

grlira:
I've tested this on a relatively complex component/container combination in our code base and it worked just fine :+1: 
Would be great to see this go in.

AndrewSouthpaw:
Seems fine to me then, we just need to get the merge conflicts resolved. Sorry for the slow movement.

GAntoine:
@villesau If you fix the conflicts I'll merge this in!

villesau:
@GAntoine @AndrewSouthpaw conflicts resolved. Also did some minor cleanup.

PR #1962, opened 2018-03-14
Allow any file type in first level definition folder

GAntoine:
No longer throws an error when there are non-test files and non-flow-version-range folders for first-level folders, ie: `definitions/npm/alasql_v1.x.x/`.

Also reverts the test file regex to only match and add test files.

### DIFF
--- a/cli/src/lib/libDefs.js
+++ b/cli/src/lib/libDefs.js
@@ -27,7 +27,7 @@ export type LibDef = {|
   testFilePaths: Array<string>,
 |};
 
-export const TEST_FILE_NAME_RE = /(^test_.*\.js$)/;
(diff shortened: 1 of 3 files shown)


AndrewSouthpaw:
Nice! I'm less familiar with this part of the code, it *looks* like this would also loosen checks for testing libdefs as well, which we don't want, yeah?

GAntoine:
Really it just loosens what's permitted in the first-level definition folder. Tests will still need to be names `test_<>.js`, and the files under flow version range folders are still required to be either libdefs or tests, and nothing else. Does that makes sense?

AndrewSouthpaw:
Seems reasonable. Looks like the build is failing?

GAntoine:
Bad travis build, fixed it!

PR #1969, opened 2018-03-16
Add typing for return type of underscore's _.times

jetpacmonkey:
There was a definition for `_.times`, but it didn't account for the fact that it can be used to build up an array with the return values of the iteratee. I tried to imitate the surrounding code style as much as possible, happy to make any additional changes as needed!

### DIFF
--- a/definitions/npm/underscore_v1.x.x/flow_v0.38.x-v0.49.x/test_underscore.js
+++ b/definitions/npm/underscore_v1.x.x/flow_v0.38.x-v0.49.x/test_underscore.js
@@ -83,6 +83,11 @@ var identityIsString: string = _.identity('foo');
 // $ExpectError
 var identityIsNotString: string = _.identity(42);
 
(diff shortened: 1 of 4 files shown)


jetpacmonkey:
It looks like some definitions I didn't touch are causing the build to fail?

GAntoine:
@jetpacmonkey Looking at the logs, the specs are indeed failing on your changes to underscore: `ERROR: underscore_v1.x.x/flow_v0.50.x-`

jetpacmonkey:
Hmm, that is the file that I changed, but the actual failures look like they're unrelated to my changes. I just ran the tests locally on master, and saw the same errors.

GAntoine:
Yeah looks like the current definition is broken for some of the newer flow versions.

jetpacmonkey:
So what's the path forward, then?

GAntoine:
I'll merge it in, even though it's failing, since the changes do look ok.

PR #1970, opened 2018-03-16
feat(redux-saga_v0.16): update delay signature

ADagen:
Redux-saga's `delay` allows to return a Promise that will resolve after ms milliseconds with val: https://redux-saga.js.org/docs/api/#delayms-val


ADagen:
What's wrong with Travis pipelines?

GAntoine:
Seems like the current definitions is broken for versions 62+

andrei-hameza:
As for me it makes sense to reuse definition for `delay` from here https://github.com/flowtype/flow-typed/blob/master/definitions/npm/redux-saga_v0.13.x/flow_v0.38.x-/redux-saga_v0.13.x.js#L119 It fits better.

AndrewSouthpaw:
Looks like the build is still failing, can you dig into it?

ADagen:
Yep, I'm going to investigate the case this evening)

GAntoine:
Closed in favour of #2502

Issue #1534, opened 2017-11-14
Improve Ramda's `#values` definition.


AndrewSouthpaw:
cc @LoganBarnett 

I'm working on improving `R.values`, which currently in flow only allows objects of the same value type.

```js
R.values(({ a: 1, b: 'foo' }: { a: number, b: string })) // boom
```

Based on the functionality, we should be returning an array of values with types equal to those of the object. The trouble with `$ElementType` is that it requires a key to be specified, otherwise it can't type check. As far as I can tell, there's no utility type that allows us to extract dynamically the types of an object. (This probably has to do with the inherent shortcomings of trying to type a dynamic language, see the discussion of `Object.values` with flow [here](https://github.com/facebook/flow/issues/2221)).

Returning `Array<mixed>` seems utterly useless, and in fact makes things worse since you can hardly do anything with a `mixed` type, and would be annoying to have to typecast as `Array<any>` every time.

AFAICT, the only reasonable solution right now is to return `Array<any>`, which makes me sad.

Does anyone see a better alternative?

LoganBarnett:
Yeah for what I can see `$Keys<T>` gives you a union of all of the key names for a `T`. `$ElementType` has to work with that union. It's almost like we want a `Literal` type which we can constrain by `$Keys<T>` and then we can feed that to `$PropertyType`. I would hope that would Just Work because we've resolved what `$PropertyType` wants: A literal string. We just did it further up the chain somewhere.

Would that work here in the case of `R.values`? I'm not sure - I'm kind of thinking out loud here with other things we've been running into with trying to harden the types of known objects as we perform highly abstract transformations on them.

Let's pretend for a moment that we could have any utility type we wanted with our magic utility type creating wand. What would we do to express `values`? We could somehow convert all of the property/element types into a tuple, but order isn't guaranteed. You'd have to disambiguate every entry, _however_ you could probably do it with a constraint. This might be where `$ElementType` indeed shines. I would expect the best we could do is get a union of all types that the collective properties of the object uses.

[...]

PR #1971, opened 2018-03-16
ramda: use `$Values` to represent the values function

LoganBarnett:
`values` in Ramda returns all of the values of an object. This is a natural fit for `$Values<T>`, which this PR changes the `values` definition to use.

There is a close cousin of `values` called `valuesIn`. It looks like the definition for `valuesIn` isn't even in the libdefs yet. I'm not terribly familiar with how differently we should type `valuesIn` so I'm leaving that out of this PR and it can be addressed in a future date.

This also begins using the new testing format per #1928.

@AndrewSouthpaw I'd totally forgotten about our discussion on #1534 but I think this is on the right track and not in conflict with anything we've talked about so far. Please correct me if I'm wrong though!

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
@@ -1705,7 +1705,7 @@ declare module ramda {
     o: O
   ): Array<[string, T]>;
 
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw:
(Sorry to be slow on this one.) Yep, totally right. `$Values` is a great fit. I'm pretty certain that utility type did not exist when that discussion took place. I'm going to close the other thread as well.

PR #1978, opened 2018-03-19
Add option to provide cache folder path in Environment property

tyger:
This options is quite useful in CI build tools like Travis to keep cache stashed between builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,11 @@ a short grace period during a libdef installation, but sometimes it is useful to
 do this update manually. Use this command if you want to download the most
 recent definitions into the cache for yourself.
 
+It is possible to specify a folder where cache will be stored, it could be useful
+for CI build tools such as Travis to allow it keep cache folder between builds.
+To specify cache folder define Environment property `FLOW_TYPED_CACHE_DIR`.
+By default `$(HOME)/.flow-typed` will be used.
+
 ## Active Maintenance Team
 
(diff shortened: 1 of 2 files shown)


diskshima:
It would be great to see this get integrated.
We're having problems with our CI as one build could get its flow-typed cache updated by another build potentially causing empty flow-typed files to be created in the former build.

GAntoine:
I'm holding off a little on this, as we don't set `env` variables anywhere else in the project, and I want to think on this before setting a precedent. It may make sense to add `env` variables for some of the existing flags, like `--libdefDir` or `--rootDir`.

AndrewSouthpaw:
Tangential to enabling env variables, is there a reason you don't just check in `.flowconfig` to your repo?

AndrewSouthpaw:
We're thinking that the best option is to accommodate CLI arguments such as `--cache_dir`, rather than env variables. There's no precedent for env vars while there is for CLI args, and CLI args are generally more discoverable through `--help`.

We like the idea of being able to specify a cache dir though, so would you mind reworking the PR to use CLI args?

tyger:
@AndrewSouthpaw  How do you see application of that CLI parameter? Provide it together with `install` and `update` commands?

GAntoine:
@tyger That's the idea, add it as a flag to the `install` and `update commands. You can look at how `--libdefDir` is implemented for some guidance. Let me know if you need a hand making the switch.

tyger:
Created new PR with `cli` commands: https://github.com/flowtype/flow-typed/pull/2016

AndrewSouthpaw:
Closing this one for now. :)

Issue #1963, opened 2018-03-14
Mismatch Reducer definitions in redux & redux-actions


sam-n-johnston:
I now get errors in my project in redux-actions:

```
This parameter is incompatible:
    30:   declare export type Reducer<S, A> = (state: S | void, action: A) => S;
                                                      ^^^^^^^^ undefined. This type is incompatible with. See lib: flow-typed/npm/redux_v3.x.x.js:30
    58: }, defaultState)
          ^^^^^^^^^^^^ object literal. See: app/reducers/appointmentsDurations.js:58
```

After digging into it, I believe it's due to #1958. The state's type in the reducer in [redux](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/redux_v3.x.x/flow_v0.55.x-/redux_v3.x.x.js#L27) is now `S | void`, but in [redux-actions](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/redux-actions_v2.x.x/flow_v0.39.x-/redux-actions_v2.x.x.js#L60), it's still just `S`. So `redux-action` returns a different type of reducer to `createStore` and breaks there I believe.

GAntoine:
Thanks for reporting this issue, would you be up to write a fix for `redux-actions`?

sam-n-johnston:
I tried to fix this, but got issues trying to fix the tests and didn't have much time. I can give it another go soon. Since `redux-actions` has to work with `redux`, I was wondering why `redux-actions` re-defines a type like Reducer instead of importing the one from the `redux` libdef?

GAntoine:
>  I was wondering why redux-actions re-defines a type like Reducer instead of importing the one from the redux libdef

Flow-typed doesn't yet have the ability to import definitions for other libdefs, although it is set as a target for version `3.0.0`.

PR #1980, opened 2018-03-19
Update redux-actions Reducer definition

dignite:
This is needed to match redux definitions changed in `v2.4.0`.
State can be `S` or `void`.

Resolves issue #1963


GAntoine:
@dignite Seems the change is causing the specs to fail, we can't merge this in until they pass.

dignite:
Thanks for the bump @GAntoine I hadn't noticed myself. Haven't had time to look into it yet

dignite:
I don't mind, I'm a bit too new at flow and didn't manage myself in my attempts. Will check out your solution!

PR #1991, opened 2018-03-22
feat(formik): improve formik types

sibelius:
I think FormikHOC is wrong

we need to type `withFormik` to make this work fine

right now I'm using FormikProps instead

something like this

```
type Values = { a: string }
type Props = {} & FormikProps<Values>
```

### DIFF
--- a/definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
+++ b/definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
@@ -1,5 +1,5 @@
 declare module "formik" {
-  import { ComponentType } from "react";
+  import type { ComponentType } from "react";
   
   declare export type FormikErrors<Values> = {
     [field: $Keys<Values>]: string
--- a/definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
+++ b/definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
@@ -1,0 +1,237 @@
+declare module "formik" {
+  declare export type FormikErrors<Values> = {
+    [field: $Keys<Values>]: string
+  };
+
+  declare export type FormikTouched<Values> = {
+    [field: $Keys<Values>]: boolean
+  };
+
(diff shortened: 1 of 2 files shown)


sibelius:
this makes sense for `withFormik`:

```
declare export var withFormik: (
    Component: React$ComponentType<Props>
  ): React$ComponentType<$Diff<Props, ...FormikProps<*>>
```

I've followed this https://flow.org/en/docs/react/hoc/

sibelius:
improved here

```jsx
declare export function withFormik<Props, Values>({
    mapPropsToValues: (props: Props) => Values,
    validationSchema: any, // TODO - improve this
  }): (ComponentType<Props>) => ComponentType<$Diff<Props, FormikProps<Values>>>;
```

How can I make sure this is working fine?

How can I check the type of a single variable for instance?

definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
```diff
@@ -0,0 +1,237 @@
...
+     *       const {setFieldValue, status, errors, dirty} = this.props;
+     *     }
+     *  }
+     *
+     *
+     * <MyFormikHoc/>
+    */
+   declare export type FormikHOC = {
+    dirty?: $PropertyType<FormikComputedProps<*>, 'dirty'>,
+    errors?: FormikErrors<Values>,
```
GAntoine: `Values` is undefined here, and is failing the specs.

definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
```diff
@@ -1,3 +1,6 @@
+import { ComponentType } from "react";
+import { FormikProps } from "formik";
```
saadq: I don't think this currently works, see discussion [here](https://github.com/flowtype/flow-typed/issues/1857).

sibelius: this is wrong I guess

sibelius:
@GAntoine could u review again

PR #1992, opened 2018-03-22
Cleanup runTests a bit

villesau:
This is handy if you are working on fixing libdefs for some specific flow version and don't want to be bothered about other versions while developing. `numberOfFlowVersions` already covers the main use case which is to run test suite against latest version of Flow though.

note that `?w=1` makes reviewing easier: https://github.com/flowtype/flow-typed/pull/1992/files?w=1

E: reverted last commit and now this contains only refactorings

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -471,9 +471,9 @@ async function removeTrashFromBinDir() {
 async function runTestGroup(
   repoDirPath: string,
   testGroup: TestGroup,
-  numberOfFlowVersions: number = 15,
-  errors = [],
+  orderedFlowVersions: Array<string>,
 ): Promise<Array<string>> {
+  const errors = [];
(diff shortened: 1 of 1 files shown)


villesau:
I think the refactorings before the argument implementation makes sense to get in anyways so if we end up not taking the argument, I can revert the last commit if other stuff looks OK.

AndrewSouthpaw:
I think it makes sense given the discussion, can you revert the last commit so we can look at the right diff?

villesau:
@AndrewSouthpaw done

AndrewSouthpaw:
To be honest, I haven't had the time to dive into the diff, but I trust your work and we'll fix it if something breaks. 😅

AndrewSouthpaw:
Oh, can you mention this in the CONTRIBUTING guide to help people? I remember wasting nearly an hour when first working on flow-typed just to figure out how to run tests, specifically just the ones for my library.

PR #1993, opened 2018-03-23
Improve return values for pretty-error

taybin:
render() returns a string, and most other functions return their own object to allow chaining.


GAntoine:
@taybin You'll need to fix the specs to match your changes, take a look at the Travis logs.

PR #1998, opened 2018-03-26
Add mockName type to jest mocked functions

grlira:
mockName was added to jest mocked functions as of jest v22 but seems to be missing from the flow type definitions.

### DIFF
--- a/cli/flow-typed/npm/jest_v22.x.x.js
+++ b/cli/flow-typed/npm/jest_v22.x.x.js
@@ -56,6 +56,11 @@ type JestMockFn<TArguments: $ReadOnlyArray<*>, TReturn> = {
     fn: (...args: TArguments) => TReturn,
   ): JestMockFn<TArguments, TReturn>,
   /**
+   * Accepts a string to use in test result output in place of "jest.fn()" to
+   * indicate which mock function is being referenced.
+   */
+  mockName(name: string): JestMockFn < TArguments, TReturn >,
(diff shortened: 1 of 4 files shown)


AndrewSouthpaw:
Can you rebase off the latest? I think the tests are failing due to you having an older version of the libdef.

grlira:
I checked and my commit is on top of 1ae5dd which is the latest according to github, so I'm not really sure of what's going on. Given that I updated some definitions is there something I should do beyond just changing the definition files themselves?

AndrewSouthpaw:
I figured it out, and introduced a fix in #2004. Once that gets merged, let's rebase this guy and make sure it all still checks out.

AndrewSouthpaw:
Okay, that's merged. Can you rebase?

grlira:
Thank you. I've rebased and added a couple tests for `mockName` now.
I'm just wondering... given that the flow libdef is also in `cli/flow-typed/npm/`, should that also be updated as part of this PR? Or do you guys have some process to update the cli flow-typed files independently?

AndrewSouthpaw:
Oh, nice, thanks for catching that, I forget it's also in `cli/flow-typed/npm`. Yes, the new method will need to be copied over there as well please. 😊

grlira:
Ok, that's added now :)

AndrewSouthpaw:
Great work, thanks for contributing!

PR #2000, opened 2018-03-26
Add libdef for deepmerge

Jarlakxen:
Library definition for [deepmerge](https://github.com/KyleAMathews/deepmerge).

### DIFF
--- a/definitions/npm/deepmerge_v2.1.x/flow_v0.57.3-/deepmerge_v2.1.x.js
+++ b/definitions/npm/deepmerge_v2.1.x/flow_v0.57.3-/deepmerge_v2.1.x.js
@@ -1,0 +1,12 @@
+declare module 'deepmerge' {
+  declare type Options = {
+    clone?: boolean,
+    arrayMerge?: (destination: any[], source: any[], options?: Options) => Array<any>
+  }
+
+  declare module.exports: {
+    <A, B>(a: A, b: B, options?: Options): A & B,
+    all<T>(objects: Array<$Shape<T>>, options?: Options): T
+  };
+
+}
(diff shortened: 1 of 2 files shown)


definitions/npm/deepmerge_v2.1.x/flow_v0.57.3-/deepmerge_v2.1.x.js
```diff
@@ -0,0 +1,12 @@
+declare module 'deepmerge' {
+  declare type Options = {
+    clone?: boolean,
+    arrayMerge?: (destination: any[], source: any[], options?: Options) => Array<any>
+  }
+
+  declare module.exports: {
+    <A, B>(a: A, b: B, options?: Options): A & B,
+    all<T>(objects: Array<$Shape<T>>, options?: Options): T
```
AndrewSouthpaw: This def requires the objects to be all the same shape, which does not reflect the docs.

definitions/npm/deepmerge_v2.1.x/test_deepmerge_v2.1.x.js
```diff
@@ -0,0 +1,8 @@
+const deepmerge = require("deepmerge");
+
+const m1: { a: string, b: string } = deepmerge({a: ''}, { b: '' })
+
+const m2: { a: string, b: string } = deepmerge.all([ {a: ''}, { b: '' } ])
+
+// $ExpectError
+const m3: { a: number, b: string } = deepmerge({a: ''}, { b: '' })
```
AndrewSouthpaw: Should probably demonstrate copy-over precedence, so something like:

```
deepmerge({ a: 1 }, { a: 'foo' }) // should be a string
```

Issue #1975, opened 2018-03-17
axios's ussing same type for request and response data 


punksta:
data and config.data types are same:
https://github.com/flowtype/flow-typed/blob/fa010af9306bd5f2835c7890638afb0cb107dcfe/definitions/npm/axios_v0.18.x/flow_v0.25.x-/axios_v0.18.x.js#L69-L77

I believe AxiosXHR should look like this.
```
declare class AxiosXHR<T, R> {
    config: AxiosXHRConfig<T>;
    data: R
```

Issue #1539, opened 2017-11-16
Issue with Axios


maxsalven:
It seems the libdef expects `response.data` to match the shape of submitted `data` in a post request? I think this is incorrect?

```
    axios({
      method: "post",
      url: "/user/12345",
      data: {
        firstName: "Fred",
        lastName: "Flintstone",
      },
    }).then(response => {
      console.log(response.data.wasSuccess);
    });
```

This raises a Flow error saying `wasSuccess` not found in object literal (whereas accessing `response.data.firstName` doesn't raise an error).

The user should be able to define the expected shape of `response.data`.

cc @timothystewart6 @haines

haines:
I think the problem is [here](https://github.com/flowtype/flow-typed/blob/d679eeac2f803286b3dc0cee4afcb01c0f3cc9bd/definitions/npm/axios_v0.17.x/flow_v0.25.x-/axios_v0.17.x.js#L84), where the input config has the same type parameter as the output promise.

cyorobert:
Same issue here. What I'd like to be able to do is something like this:

```javascript

type MyType = {
    foo : string,
    bar : number
}

type Error = {
     error : string
}

type Response = MyType & Error

function post() : Promise<Response> {
    return axios.post('/my-route')
                .then(response => response.data)
                .catch(() => ({error : "an error occurred"}))
}
```

PR #2006, opened 2018-03-28
Add response type to axios libdef

adambrgmn:
This PR adds a separate response type to axios calls, earlier the definitions
used the same infered type for config.data and response.data. This has been mentioned in #1975 and #1539.

But hopefully this wont break backwards compatability since I added the
original type (T) as default for the new response type (R). Also verified by tests.

As of now I have only updated the 0.18.x-version. But the same issue is present in earlier versions as well. If you think I should update the old ones as well I'd be happy to do so.

### DIFF
--- a/definitions/npm/axios_v0.18.x/test_axios_v0.18.x.js
+++ b/definitions/npm/axios_v0.18.x/test_axios_v0.18.x.js
@@ -4,6 +4,7 @@ import type {
   $AxiosXHR,
   CancelTokenSource,
   Cancel,
+  AxiosPromise,
 } from 'axios';
 (axios.get('foo'): Promise<*>);
 (axios.post('bar', {}, {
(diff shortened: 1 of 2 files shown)


definitions/npm/axios_v0.18.x/test_axios_v0.18.x.js
```diff
@@ -86,3 +87,14 @@ axios.all([
     // $ExpectError
     (a: string);
 })
+
+const promise1: AxiosPromise<{ foo: string }, { bar: string }> = axios({ url: '/', method: 'post', data: { foo: 'bar' }})
+promise1.then(({ data }) => {
+  // $ExpectError
+  data.foo;
```
AndrewSouthpaw: Add a check for `data.bar` is a string, ensure that carries through. :)

adambrgmn:
Thanks for the response (pun intended...)! I added a test to verify that "data.bar" is a string and all tests passes on my machine.

PR #2011, opened 2018-03-29
Added libdef for @nx-js/observer-util

eidheim:


definitions/npm/@nx-js/observer_util_v4.x.x/flow_v0.25.x-/observer_util_v4.x.x.js
```diff
@@ -0,0 +1,19 @@
+declare module '@nx-js/observer-util' {
```
GAntoine: Folder/file names should be `observer-util` with a dash.

GAntoine: @eidheim

PR #2013, opened 2018-03-30
Adding libdef for Winston Logging Package

gyrfalcon:
Added a libdef and test for the winston logging package.


gyrfalcon:
@GAntoine All comments addressed, thanks for the feedback!

PR #2014, opened 2018-03-31
Adding libdef for eventemitter3_v3.x.x

dszakallas:
Adds a libdef for eventemitter3 v3.

There were some breaking changes, outlined [here](https://github.com/primus/eventemitter3/releases/tag/3.0.0). You can also look at thet [ts definitions](https://github.com/primus/eventemitter3/blob/master/index.d.ts) for reference.


dszakallas:
I don't really understand what causes the syntax errors on line 44.

saadq:
You're missing a semicolon at line 40. So

```js
(a.listenerCount(): number)
(a.listeners('emit', true): boolean);
```

Is getting treated as

```js
(a.listenerCount(): number)(a.listeners('emit', true): boolean);
```

dszakallas:
thanks @saadq , my eye slipped over this one

PR #2016, opened 2018-04-01
Set cache directory with arguments of install and update commands

tyger:
Allows to specify path to cache directory with cli argument `cacheDir`. Useful for CI tools like travis.

### DIFF
--- a/cli/src/commands/update-cache.js
+++ b/cli/src/commands/update-cache.js
@@ -1,43 +1,44 @@
 // @flow
 
-import {updateCacheRepo} from '../lib/libDefs';
-
-import type {Argv} from 'yargs';
 import typeof Yargs from 'yargs';
+import {
+  _setCustomCacheDir as setCustomCacheDir,
+  ensureCacheRepo,
+} from '../lib/cacheRepoUtils';
+import {path} from '../lib/node';
 
 export const name = 'update-cache';
 export const description = 'Update the flow-typed definitions cache';
 
+export type Args = {
+  cacheDir?: string,
+};
 export function setup(yargs: Yargs) {
   return yargs.usage(`$0 ${name} - ${description}`).options({
-    debug: {
-      describe: 'Enable verbose messages for the update procedure',
-      alias: 'd',
-      type: 'boolean',
+    cacheDir: {
+      alias: 'c',
+      describe:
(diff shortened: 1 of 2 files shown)


tyger:
Any thoughts anybody?

GAntoine:
Totally forgot about this, I'll take a look at it soon!

cli/src/commands/update-cache.js
```diff
@@ -1,44 +1,46 @@
...
 
 export const name = 'update-cache';
 export const description = 'Update the flow-typed definitions cache';
 
+export type Args = {
+  cacheDir?: string,
+};
 export function setup(yargs: Yargs) {
   return yargs.usage(`$0 ${name} - ${description}`).options({
-    debug: {
```
GAntoine: Why did you choose to remove debug?

cli/src/commands/update-cache.js
```diff
@@ -1,44 +1,46 @@
...
-      verbose = process.stdout;
+    if (args.cacheDir) {
+      const cacheDir = path.resolve(args.cacheDir);
+      console.log('• Setting cache dir', cacheDir);
+      setCustomCacheDir(cacheDir);
     }
 
     console.log('Updating flow-typed definitions...');
-    await updateCacheRepo(verbose);
+    await ensureCacheRepo();
```
GAntoine: Do `updateCacheRepo` and `ensureCacheRepo` do the same thing?

tyger: Here is the thing: it used to call `updateCacheRepo` from `libDefs.js`, which in turn just called `ensureCacheRepo`.
But latest implementation of `install` command use newer set of methods from `npmLibDefs.js` and `cacheRepoUtils.js`, which as well contains new implementation of `ensureCacheRepo`. 
So to wrap up: yes, former `updateCacheRepo` and `ensureCacheRepo` does the same thing although old and new implementations of `ensureCacheRepo` have slight difference (not important though).

cli/src/commands/update-cache.js
tyger: It doesn't have much sense any more because of the reason described below. The new version `ensureCacheRepo` doesn't accept verbosity parameter where `debug` parameter used to go - it rather write all required information into log all the time.

PR #2017, opened 2018-04-02
add xterm libdef

saadq:
References

---

* TS Definition: https://github.com/xtermjs/xterm.js/blob/master/typings/xterm.d.ts
* TS Test File: https://github.com/xtermjs/xterm.js/blob/master/fixtures/typings-test/typings-test.ts


AndrewSouthpaw:
Thanks @saadq!

PR #2018, opened 2018-04-02
[react-redux] Add default export

Ashoat:
Avoiding any `export` declarations means that all declared modules are available as properties on the default export. The version of this libdef before the [revamp](https://github.com/flowtype/flow-typed/pull/1731) didn't have any `export` keywords, presumably for that reason.

Why not use named imports, you say? I have a specific case where I can't use named imports: the new module support in Node.js 9.3 can't handle named imports on commonjs modules. And since I'm not able to use the ESM version of `react-redux` because it relies on a named import from `react` (which doesn't have an ESM version), I have to use the commonjs version.

Regardless: as `react-redux` does indeed have a default export, the libdef should be declared accordingly.

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/react-redux_v5.x.x.js
@@ -3,12 +3,12 @@ import type { Dispatch, Store } from "redux";
 declare module "react-redux" {
   import type { ComponentType, ElementConfig } from 'react';
 
-  declare export class Provider<S, A> extends React$Component<{
+  declare class Provider<S, A> extends React$Component<{
     store: Store<S, A>,
     children?: any
   }> {}
 
-  declare export function createProvider(
+  declare function createProvider(
     storeKey?: string,
     subKey?: string
   ): Provider<*, *>;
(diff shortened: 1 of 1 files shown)


villesau:
What about explicitly defining default export for the module? Current libfdefs has private types which should not be needed to be exposed. Also, rect-redux actually does _not_ have default export: https://github.com/reactjs/react-redux/blob/master/src/index.js

Ashoat:
My apologies, you are totally correct. I assumed there was a default export because I was able to import a default export from what was actually a CommonJS module. But that's the CommonJS version, not the ESM version... which is the source of the trouble here.

Sadly it [doesn't seem like](https://medium.com/the-node-js-collection/an-update-on-es6-modules-in-node-js-42c958b890c) Node.js will ever support named imports from CommonJS modules:

> With CommonJS, on the other hand, the shape of a module is not known until after the code is evaluated. What this means is, without making significant changes to the ECMAScript language spec, it will not be possible to use Named Imports from a CommonJS module.

I guess this implies that we want to support ESM in Node.js, we're stuck with declaring default exports in all libdefs for the foreseeable future. At least until major libraries like React start distributing ESM versions, and the community standardizes on a way of a declaring ESM versions.

I'll update this PR to explicitly declare the default export.

AndrewSouthpaw:
LGTM. @villesau?

AndrewSouthpaw:
Also, build is still failing, but it seems like that's probably unrelated? Would be nice to get confirmation there, and figure out what the heck is causing `react-redux` libdef to be so angry.

Ashoat:
It appears that the build has been broken since #1731 due to test errors. The tests in the `react-redux_v5.x.x/flow_v0.54.x-v0.61.x` folder weren't updated (though the libdef was), so they have been failing since #1731.

There are additionally some failures in `react-redux_v5.x.x/flow_v0.62.0-` that don't appear to have been occurring when #1731 was merged. I'm not entirely sure what's going on there, as they are occurring on 0.62.0, which was tested by the CI (but didn't fail) at the time at merge.

villesau:
LGTM! The errors are unrelated and some sort of test suite glitch. Not sure why it matches also older versions of the libdef.

PR #2020, opened 2018-04-02
Allow `moment.utc(undefined)` calls

SBoudrias:
`moment.utc(undefined)` does the same as `moment.utc()`, so I feel it'd be convenient to not error on such a use case.

### DIFF
--- a/definitions/npm/moment_v2.x.x/flow_v0.28.x-/moment_v2.x.x.js
+++ b/definitions/npm/moment_v2.x.x/flow_v0.28.x-/moment_v2.x.x.js
@@ -1,26 +1,26 @@
 type moment$MomentOptions = {
-  y?: number|string,
-  year?: number|string,
-  years?: number|string,
-  M?: number|string,
-  month?: number|string,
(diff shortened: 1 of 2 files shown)


definitions/npm/moment_v2.3.x/test_moment-v2.js
```diff
@@ -1,25 +1,37 @@
 // @flow
-import moment from "moment";
+import moment from 'moment';
 
 // Parse
-const m3: moment = moment([123, 123]);
+let m: moment;
```
GAntoine: `m` is defined below, pick a different name like `m0`.

GAntoine: `m` is defined below, pick a different name like `m0`.

AndrewSouthpaw: Or use the new `describe` and `it` blocks! :)

GAntoine:
Closing due to age.

PR #2021, opened 2018-04-02
Firebase libdef update

meafmira:
Added actionCodeSettings type
Added parameter for sendEmailVerificationMethod
Changed onAuthStateChanged handler signature
Prettified


AndrewSouthpaw:
In the future, please keep prettification as a separate commit, because it's otherwise impossible to interpret these PRs.

Can you provide references to the docs / methods you specifically changed?

meafmira:
@AndrewSouthpaw sorry. Updated PR without prettier

PR #2027, opened 2018-04-03
Add new stub generation with guessed properties

KsawK:
New method of generating stubs, based on module properties. It's looking for one (can be changed with --maxDepth) layer of properties and checks their types. Mostly created for modules that exports object filled with functions (i.e. Lodash) or statics.
Ps. I am asking for patience, the first PR I made so I could make some mistakes of a novice

### DIFF
--- a/cli/src/lib/stubUtils.js
+++ b/cli/src/lib/stubUtils.js
@@ -79,85 +79,108 @@ function stubFor(moduleName: string, fileExt?: string): string {
   return moduleStub;
 }
 
-function getObjectToType(maxDepth?: number) {
-  let callstack: number = 0;
-  let maxCallstack: number = maxDepth || 1;
+const functionTemplate = '(%s) => any';
+const spaceByI = (i: number): string => '  '.repeat(i);
+const keyTypeTemplate = `  %s: %s,\n`;
 
-  const functionTemplate = '(%s) => any';
-  const spaceByI = (i: number): string => '  '.repeat(i);
+function objectToTypedTemplate(
+  obj: mixed,
+  currentDepth: number,
+  maxDepth: number,
+  functionHeader?: string,
+) {
+  const thisDepth = currentDepth + 1;
 
-  const keyTypeTemplate = `  %s: %s,\n`;
-  function objectToTypedTemplate(obj: mixed, functionHeader?: string) {
-    callstack++;
-    let output = '';
-    let entries = Object.entries(obj);
-    let formatedEntries = [];
-    if (functionHeader) {
-      formatedEntries.push(spaceByI(callstack + 1) + functionHeader);
-    }
-    for (let entrie of entries) {
-      formatedEntries.push(
-        format(
-          spaceByI(callstack) + keyTypeTemplate,
-          entrie[0],
-          objectToType(entrie[1], callstack > maxCallstack ? false : true),
-        ),
-      );
-    }
-    if (formatedEntries.length > 0) {
-      output = '{\n' + formatedEntries.join('') + spaceByI(callstack) + '}';
-    } else {
-      output = '{}';
-    }
-    callstack--;
-    return output;
+  let formatedEntries = [];
+  if (functionHeader) {
+    formatedEntries.push(spaceByI(thisDepth + 1) + functionHeader);
   }
 
-  function functionToType(fun: Function): string {
-    let output = '';
-    let raw = fun.toString();
-    let args = raw.slice(raw.indexOf('(') + 1, raw.indexOf(')'));
-    args = args.replace(/ /g, '');
-    if (args.length > 0) {
-      args = args
-        .split(',')
-        .map((el: string) => el + ': any')
-        .join(', ');
-    }
-    output = format(functionTemplate, args);
-    let functionEntries = Object.entries(fun);
-    if (functionEntries.length > 0) {
-      output = objectToTypedTemplate(
-        fun,
-        output.length > 0 ? output + ',\n' : undefined,
-      );
-    }
-    return output;
-  }
-  function objectToType(obj: mixed, deep: boolean = true): string {
-    let output = '';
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw:
It would be helpful to have more context around the motivation for this PR. What is the current problem you're facing, what's the current limitation and how does yours fix it?

KsawK:
PR is expanding generated stubs. Basic stub is generated with template:
```js
const moduleStubTemplate = `
declare module '%s' {
  declare module.exports: any;
}`.trim();
```
As I don't like write things that could be generated, my target was to create more accurate base with guessed properites from basic object returned by require(). I achived it by checking each properties by typeof and expanding other objects. Additional functions are expanded from Function.prototype.toString(). It saves time if somebody like to use small libraries which are exporting small number of functions. To avoid infinity loop on object reference there is maximal callstack of recursion.
Sample generated for 'colors': 
```
declare module 'colors' {
  declare module.exports: {
    themes: {},
    styles: {
      reset: any,
      bold: any,
      dim: any,
      italic: any,
      ... more
    },
    supportsColor: boolean,
    enabled: boolean,
    strip: (str: any) => any,
    ... more
  }
}
```

AndrewSouthpaw:
cc @GAntoine 
cc @villesau You should probably weigh in given our conversation about modules.

cli/src/lib/stubUtils.js
```diff
@@ -75,6 +79,89 @@ function stubFor(moduleName: string, fileExt?: string): string {
...
+  if (functionEntries.length > 0) {
+    output = objectToTypedTemplate(
+      fun,
+      output.length > 0 ? output + ',\n' : undefined,
+    );
+  }
+  return output;
+}
+
+const keyTypeTemplate = `  %s: %s,\n`;
```
GAntoine: This should be defined above the function that uses it, for readability,

cli/src/lib/stubUtils.js
```diff
@@ -75,6 +79,89 @@ function stubFor(moduleName: string, fileExt?: string): string {
...
+export function objectToType(obj: mixed, deep: boolean = true): string {
+  let output = '';
```
GAntoine: Entire function can be refactored as shown:

```
if (typeof obj === 'object') {
  return deep ? objectToTypedTemplate(obj) : 'any';
}
if (typeof obj === 'function' && deep) return functionToType(obj);
return typeof obj;
```

Haven't tested this chunk locally, so you may need to play around with the ternary return.

cli/src/lib/stubUtils.js
```diff
@@ -75,6 +79,89 @@ function stubFor(moduleName: string, fileExt?: string): string {
...
+  callstack--;
+  return output;
+}
+
+export function functionToType(fun: Function): string {
+  let output = '';
+  let raw = fun.toString();
+  let args = raw.slice(raw.indexOf('(') + 1, raw.indexOf(')'));
+  args = args.replace(/ /g, '');
+  if (args.length > 0) {
```
GAntoine: Try spacing out this code a little for readability.

cli/src/lib/stubUtils.js
```diff
@@ -75,6 +79,89 @@ function stubFor(moduleName: string, fileExt?: string): string {
...
 }
 
+const functionTemplate = '(%s) => any';
+const spaceByI = (i: number): string => '  '.repeat(i);
+
+let callstack = 0;
+let maxCallstack = 1;
+
+export function objectToTypedTemplate(obj: mixed, functionHeader?: string) {
+  callstack++;
```
GAntoine: Increment and return as a passed-in param.

cli/src/lib/stubUtils.js
KsawK: I agree that current form looks bad but i would refactor it to:
```js
if(deep){
  if(typeof obj === 'object'){
    return objectToTypedTemplate(obj);
  }else if(typeof obj === 'function'){
    return functionToType(obj)
  }
}else{
  if(typeof obj === 'object'){
    return 'any';
  }
}
return typeof obj;
```
It's more open for changes and easier to understand

GAntoine: Good point, checking `deep` first does make i easier to read. How about:

```
if(deep) {
  if(typeof obj === 'object') return objectToTypedTemplate(obj);
  if(typeof obj === 'function') return functionToType(obj);
}

if(typeof obj === 'object') return 'any';
return typeof obj;
```

No need for `else` statements when you return early (which is just good practice!)

KsawK: It's much better! I'll test tommorow my refactor and push it to github

KsawK:
@GAntoine  So i refactored code. Also i added handling error which fallback to normal method in case of uncatched error. I stayed with callstack as variable but closed it inside local scope. It's the clearest way while function is not self-recursed and give control if other recursion will go inside in the future.

GAntoine:
@KsawK I went ahead and refactored your refactor, just to clean things up a little more. I'd encourage you to give it a review, in case I broke something on the way.

One last think we'll need before merging this are some specs. Let me know if you're up to add them, and if not I'll add a few to cover `maxDepth`.

KsawK:
@GAntoine Thanks for clearing it. Sorry that's i'm adding now, after refactor but i added alternate function type handling which ignore entries and i changed old name in comment which i forgot while writing first version. Everything is working fine now and is probably styled properly. If you want to test, i recommend testing on "colors" which is dependency of flow-typed.

About specs, i'm not sure what you mean, can you explain?

GAntoine:
New changes look good. 👍 Never mind the specs, I though there were tests for `create-stub`, but it seems there aren't any at the moment. I'll go ahead and add some today, then we an merge this in.

KsawK:
Thanks for cooperation, i'm waiting for merge with excitement.

PR #2030, opened 2018-04-03
Add query-string_v6.x.x definitions

Zmetser:
No API changes compared to 5.1.x they just dropped support for old node versions.

Full changeset: https://github.com/sindresorhus/query-string/compare/v5.1.0...v6.0.0


Zmetser:
Thanks for the review @AndrewSouthpaw, sorry for missing the contribution guidelines for test. Great that they can be structured in a similar way to jest and the others.

definitions/npm/query-string_v5.1.x/flow_v0.32.x-/query-string_v5.1.x.js
```diff
@@ -11,9 +11,9 @@ declare module 'query-string' {
   |}
 
   declare module.exports: {
-    extract(str: string): string,
-    parse(str: string, opts?: ParseOptions): Object,
-    parseUrl(str: string, opts?: ParseOptions): { url: string, query: Object },
-    stringify(obj: Object, opts?: StringifyOptions): string,
+    extract(input: string): string,
+    parse(input: string, options?: ParserOptions): Object,
```
AndrewSouthpaw: We can type the result better. `{ [string]: string }`

definitions/npm/query-string_v5.1.x/flow_v0.32.x-/query-string_v5.1.x.js
```diff
@@ -11,9 +11,9 @@ declare module 'query-string' {
...
+    parseUrl(input: string, options?: ParserOptions): { url: string, query: Object },
+    stringify(obj: Object, options?: StringifyOptions): string,
```
AndrewSouthpaw: `obj: { [string]: mixed }` is probably our best bet.

definitions/npm/query-string_v5.1.x/flow_v0.32.x-/query-string_v5.1.x.js
```diff
@@ -11,9 +11,9 @@ declare module 'query-string' {
   |}
 
   declare module.exports: {
-    extract(str: string): string,
-    parse(str: string, opts?: ParseOptions): Object,
-    parseUrl(str: string, opts?: ParseOptions): { url: string, query: Object },
-    stringify(obj: Object, opts?: StringifyOptions): string,
+    extract(input: string): string,
+    parse(input: string, options?: ParserOptions): Object,
+    parseUrl(input: string, options?: ParserOptions): { url: string, query: Object },
```
AndrewSouthpaw: `query: { [string]: string }`

definitions/npm/query-string_v5.1.x/flow_v0.32.x-/query-string_v5.1.x.js
Zmetser: 👍  Good catch, although this should also be `{ [string]: mixed }` if I'm correct: https://github.com/sindresorhus/query-string/blob/master/index.js#L153

I'm making the changes.

Zmetser: Or would `{ [string]: string | Array<string> }` be better?

AndrewSouthpaw: You're right, it could be an array. So yes `{ [string]: string | Array<string> }` I think would be best.

heavenshell:
@AndrewSouthpaw 
Hi, after I updated this. I've got flow error at `parse()`.
e.g.
```
type T = { a?: string, b?: string }
// If location.search is '', a and b would be null
const { a, b }: T = parse(location.search)
```

IMO `parse()` can return `{}` if there is no query string at input.
https://github.com/sindresorhus/query-string/blob/master/index.js#L136

I think `{ [name?: string]: string | Array<string> }` is correct.
What do you think?

Regards.

Zmetser:
@heavenshell I think [your problem](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzAAcBDAJ1wFMwBeMACgEtsiBXfALjALOYHMAlFwDeYANrYSAW0pce-ALpz8vbHzAAfMAEEyZEgE8APPLUA+MAF9aF4ajBgylfKzLYwwy6i+p8BotQAKrQeYCQA-Mqq6lq6+samfGYANGAARpHcKvyaOnqGJtnmVqjAwGAAklBg8Bgk+Iw4AHRU5BgAFmCMuGAA5L2pJGHYACbpYAhwrDBjadTY0zCYOAShJKlpVlzBdKQUlPT9AkA) is that the values of that object are either a string or an array of strings. You should modify your code to prepare for that.

Here's an example for an Array:
```js
const { parse } = require("query-string");
parse('?a=1&b=2&b=3');
```
```
=> Object
    a: "1"
    b: ["2", "3"]
```

Also, I think flow can infer T for you, you don't have to explicitly set that.
If you do only need strings in your object, I think you have to [refine the type](https://flow.org/en/docs/lang/refinements/) for now on.

heavenshell:
@Zmetser Thank you for your advice!
I'll modify my code.

PR #2032, opened 2018-04-04
Some additions to mongoose@4.0.0

nodkz:
### New
- Added [descriminator](http://mongoosejs.com/docs/discriminators.html) method definitions.
- Add `_indexes` property to `Mongoose$SchemaField` it keeps info about indexes over particular field.

### Fixes
- Changed `declare export type` to `declare type` due eslint error.

### DIFF
--- a/definitions/npm/mongoose_v4.x.x/flow_v0.50.x-/mongoose_v4.x.x.js
+++ b/definitions/npm/mongoose_v4.x.x/flow_v0.50.x-/mongoose_v4.x.x.js
@@ -74,7 +74,8 @@ type SchemaOpts<Doc> = {
     | {
         createdAt?: string,
         updatedAt?: string
-      }
+      },
+  discriminatorKey?: string
 };
 
 type IndexFields = {
@@ -178,7 +179,8 @@ type Mongoose$SchemaField<Schema> = {
     description: ?string
   },
   enumValues?: ?(string[]),
-  schema?: Schema
+  schema?: Schema,
+  _index?: ?{ [optionName: string]: mixed }
 };
 
 declare class Mongoose$SchemaVirtualField {
@@ -261,6 +263,7 @@ declare class Mongoose$Document {
   static modelName: string;
   static schema: Mongoose$Schema<this>;
   static on(type: string, cb: Function): void;
+  static discriminator(name: string, schema: Mongoose$Schema<any>): Class<this>;
 
   collection: Mongoose$Collection;
   constructor(data?: $Shape<this>): this;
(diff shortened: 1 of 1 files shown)


AndrewSouthpaw:
You'll need to explain the motivation for these fixes, so when other people using mongoose can understand your decisions. :)

nodkz:
@AndrewSouthpaw fixes are very tiny. Described them top message.

definitions/npm/mongoose_v4.x.x/flow_v0.50.x-/mongoose_v4.x.x.js
```diff
@@ -492,15 +495,15 @@ declare class Mongoose$Connection {
...
-  declare export type MongooseSchemaField<Schema> = Mongoose$SchemaField<
+  declare type MongooseConnection = Mongoose$Connection;
+  declare type MongoId = MongoId;
+  declare type BSONObjectId = bson$ObjectId;
+  declare type ObjectId = bson$ObjectId;
+  declare type MongooseQuery<Result, Doc> = Mongoose$Query<Result, Doc>;
+  declare type MongooseDocument = Mongoose$Document;
+  declare type MongooseModel = typeof Mongoose$Document;
+  declare type MongooseSchema<Doc> = Mongoose$Schema<Doc>;
+  declare type MongooseSchemaField<Schema> = Mongoose$SchemaField<
```
AndrewSouthpaw: Explain more about this decision. By removing these types from exports, this would potentially break consumers. Why is it necessary to stop exporting them?

nodkz:
Hm, very strange. Cause thwy are working in my project. I removed them because get errors from the pre commit checks.

So i can return then back if you ask.

AndrewSouthpaw:
Pre commit checks in your own project? You should just ignore the flow-typed library in eslint if that's what's happening.

nodkz:
Nope in pre-commig flow-typed repo. I'll send you tomorrow screenshot. From mobile now.

nodkz:
@AndrewSouthpaw sorry for delay had a heavy sprint at work.

I return back `export` statement and when commit them I receive eslint error on pre-commit hook:
screen shot 2018-04-23 at 10 52 13

I removed `export` cause eslint checks did not allow me to made a commit.
So I made a commit with bypassing pre-hooks. 

Added after 10 minutes:
Looking on CI tests, eslint checks passed. But in dev machines `eslint` throws an error.

AndrewSouthpaw:
I was on vacation, sorry for the delay. 😄

I would suspect this is an issue with your eslint setup, since my eslinter never catches anything wrong with flow-typed. We can't approve changes based on your eslinting experience unless others report the same problem.

AndrewSouthpaw:
Ah, didn't actually check. Thanks! I'll keep an eye out for anyone else complaining about ESLint issues.

PR #2036, opened 2018-04-04
feat(jsonwebtoken): improve jsonwebtoken types

facundo-viale-olx:


definitions/npm/jsonwebtoken_v8.2.x/flow_v0.56.x-/jsonwebtoken_v8.2.x.js
```diff
@@ -0,0 +1,156 @@
+
```
GAntoine: Remove this whitespace

facundo-viale-olx:
I was not sure about putting all the new changes in current version ( 8.1.0) or just put them in the a new folder for the last version ( 8.2.0 ). I can merge the version into 8.x.x if you want or any other alternative =)

definitions/npm/jsonwebtoken_v8.2.x/flow_v0.56.x-/jsonwebtoken_v8.2.x.js
facundo-viale-olx: done

GAntoine:
@facundo-viale-olx One more thing, you'll need to copy over and teak the specs before I can get this merged in.

migueloller:
Something I'm surprised no one has noticed is that the types for `jwt$Callback` are wrong. Currently they are:

```js
declare type jwt$Callback = (tokenOrError: Error | string) => void;
```

But the function actually takes two arguments, `(error, token)`.

This could be represented with a union of the two signatures. One that takes an error and no token and another that takes null for the error and a string for the token.

facundo-viale-olx:
Done!

PR #2037, opened 2018-04-04
add puppeteer v1.2.x definitions

infinityfarm:
This adds definitions for [puppeteer](https://github.com/GoogleChrome/puppeteer)

### DIFF
--- a/definitions/npm/puppeteer_v1.2.x/flow_v0.58.x-/puppeteer_v1.2.x.js
+++ b/definitions/npm/puppeteer_v1.2.x/flow_v0.58.x-/puppeteer_v1.2.x.js
@@ -1,0 +1,1467 @@
+// @flow
+
+import ChildProcess from 'child_process';
+import EventEmitter from 'events';
+
+/** Keyboard provides an api for managing a virtual keyboard. */
+type $npm$puppeteer$Keyboard = {
+  /**
+   * Dispatches a keydown event.
+   * @param key Name of key to press, such as ArrowLeft.
+   * @param options Specifies a input text event.
+   */
+  down(key: string, options?: { text?: string }): Promise<void>,
+
+  /** Shortcut for `keyboard.down` and `keyboard.up`. */
+  press(key: string, options?: { delay?: number, text?: string }): Promise<void>,
+
+  /** Dispatches a `keypress` and `input` event. This does not send a `keydown` or keyup `event`. */
+  sendCharacter(char: string): Promise<void>,
+
+  /**
+   * Sends a keydown, keypress/input, and keyup event for each character in the text.
+   * @param text A text to type into a focused element.
+   * @param options Specifies the typing options.
+   */
+  type(text: string, options?: { delay?: number }): Promise<void>,
+
+  /**
+   * Dispatches a keyup event.
+   * @param key Name of key to release, such as ArrowLeft.
+   */
+  up(key: string): Promise<void>,
+};
+
+type $npm$puppeteer$MousePressOptions = {
+  /**
+   * left, right, or middle.
+   * @default left
+   */
+  button?: $npm$puppeteer$MouseButtons,
+
+  /**
+   * The number of clicks.
+   * @default 1
+   */
+  clickCount?: number,
+};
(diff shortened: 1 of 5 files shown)


definitions/npm/puppeteer_v1.2.x/flow_v0.58.x-/test_puppeteer_v1.2.x_connect.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import { connect } from 'puppeteer';
+import { describe, it } from 'flow-typed-test';
+
+describe('connect', () => {
```
AndrewSouthpaw: YAY you used the TDD blocks! 👏

definitions/npm/puppeteer_v1.2.x/flow_v0.58.x-/puppeteer_v1.2.x.js
```diff
@@ -0,0 +1,1437 @@
+// @flow
```
AndrewSouthpaw: This seems like you pulled it from somewhere? If so, where?

infinityfarm: @AndrewSouthpaw Yes, this was based on the [definitely typed](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/puppeteer) typescript definition. I've converted it by hand to flow

AndrewSouthpaw: Bravo. Can you link to the original in the top of the file, for reference by future contributors?

infinityfarm: @AndrewSouthpaw sure thing

saadq:
Would you be interested in changing these types to not be global? It would be a matter of of moving all the `type $npm$puppeteer$Whatever = ...` inside the `declare module` and changing it to `declare type Whatever = ...` instead. 

The `CONTRIBUTING.md` file seems to say:

> Right now we don't have a good way to write types inside declare module {} bodies that aren't exported

but this isn't really true anymore. If you use something like `declare type Foo = string` inside a `declare module` without an `export` keyword, it will be treated as a local type that won't be global.

Also, as a side note, doing imports outside of a `declare module` don't currently work. When you do

```js
import ChildProcess from 'child_process';
import EventEmitter from 'events';
```

Right now, both `ChildProcess` and `EventEmitter` will be given a type of `any`. You can only use `import type` or `import typeof`, and they both have to be done inside the `declare module`. It's described more [here](https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md#dont-import-types-from-other-libdefs).

AndrewSouthpaw:
Thanks for the additional review, @saadq!

infinityfarm:
@saadq Working on this now, can't figure out how to properly extend EventEmitter in the libdef. All my attempts thus far are resulting in an `any` being in my union type for `Page`, `Browser`, etc. Right now I'm thinking I'll remove the `EventEmitter` extension altogether if I can't fix it because at least then it will be mostly typed instead of a dangerous any cast. Let me know if you have any suggestions. From what I can tell in other lib defs that try to extend `EventEmitter` in the flow-typed repo (webdriverio), it's just an `any` type there.

hewsut:
Dumb question time from a guy (me) that gets confused about module exports easily. I have the latest puppeteer libdef installed from `flow-typed`. Why is Flow complaining about the `page` global?

```
Cannot resolve name page.

     13│ 	it('can log into the the app', async () => {
     14│ 		await login();
     15│
     16│ 		const pageUrl = page.url();
     17│
     18│ 		await expect(pageUrl).toMatch(`${appUrl}/dashboard`);
     19│ 	});
```

I tried to read through the libdef and I don't see where the global is declared... but it must be something I'm doing wrong. (And my poor attempt to append a `page` declaration on of any use failed.)

PR #2040, opened 2018-04-05
add cookie v0.3.x definitions

fabulator:
Definitions for cookie library - https://github.com/jshttp/cookie

### DIFF
--- a/definitions/npm/cookie_v0.3.x/flow_v0.58.x-/cookie_v0.3.x.js
+++ b/definitions/npm/cookie_v0.3.x/flow_v0.58.x-/cookie_v0.3.x.js
@@ -11,7 +11,7 @@ type SerializeOptions = {
 
 type ParseOptions = {
   decode?: ?((stringToDecode: string) => *),
-}
+};
 
 declare module 'cookie' {
   declare module.exports: {
--- a/definitions/npm/cookie_v0.3.x/flow_v0.58.x-/cookie_v0.3.x.js
+++ b/definitions/npm/cookie_v0.3.x/flow_v0.58.x-/cookie_v0.3.x.js
@@ -1,0 +1,22 @@
+type SerializeOptions = {
+  domain?: ?string,
+  encode?: ?((stringToDecode: string) => string),
+  expires?: ?Date,
+  httpOnly?: ?boolean,
+  maxAge?: ?number,
+  path?: ?string,
+  sameSite?: ?(boolean | 'lax' | 'strict'),
+  secure?: ?boolean,
+};
+
+type ParseOptions = {
+  decode?: ?((stringToDecode: string) => *),
+}
+
+declare module 'cookie' {
+  declare module.exports: {
+    serialize(name: string, val: string, options: ?SerializeOptions): string,
+    // eslint-disable-next-line flowtype/no-weak-types
+    parse(data: string, options: ?ParseOptions): Object,
+  };
+};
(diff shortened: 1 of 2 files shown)


definitions/npm/cookie_v0.3.x/flow_v0.58.x-/cookie_v0.3.x.js
```diff
@@ -0,0 +1,22 @@
+type SerializeOptions = {
+  domain?: ?string,
+  encode?: ?((stringToDecode: string) => string),
+  expires?: ?Date,
+  httpOnly?: ?boolean,
+  maxAge?: ?number,
+  path?: ?string,
+  sameSite?: ?(boolean | 'lax' | 'strict'),
+  secure?: ?boolean,
```
AndrewSouthpaw: The double question mark is unnecessary. They're optional props, but I'm assuming they're not maybe types, so it should be `domain?: string` for example.

definitions/npm/cookie_v0.3.x/flow_v0.58.x-/cookie_v0.3.x.js
```diff
@@ -0,0 +1,22 @@
...
+
+type ParseOptions = {
+  decode?: ?((stringToDecode: string) => *),
+};
+
+declare module 'cookie' {
+  declare module.exports: {
+    serialize(name: string, val: string, options: ?SerializeOptions): string,
+    // eslint-disable-next-line flowtype/no-weak-types
+    parse(data: string, options: ?ParseOptions): Object,
```
AndrewSouthpaw: `Object` is pretty useless. I'd go with someone like `{ [key: string]: string }` if possible.

definitions/npm/cookie_v0.3.x/flow_v0.58.x-/cookie_v0.3.x.js
fabulator: Library itself accept both options. There are checks:
```
if (options.path) {
// do something
}
```

Should I change it to ```domain?: string``` or leave it as it is?

definitions/npm/cookie_v0.3.x/flow_v0.58.x-/cookie_v0.3.x.js
fabulator: Type of value of object depend of decode options. I think it should be something like this:

```
type ParseOptions<ParserOutput> = {
  decode?: ?((stringToDecode: string) => ParserOutput),
};

declare module 'cookie' {
...
    parse<ParserOutput>(data: string, options: ?ParseOptions<ParserOutput>): { [key: string]: ParserOutput },
...
};
```

I'm not sure what is the best solution. If decode options is not defined, output is always string. But user have option to create own decoder with any response type he choose.

I think there are two options.
1. Set return type of decode option to string, which is logical because you are working with cookies. I will make definition clearer and simplier.
2. Use generic type as wrote in this comment.

AndrewSouthpaw: Unless there's a strong use case for it, I recommend keeping things simple. You could reference this comment in the libdef, in case it becomes necessary, the work doesn't need to be redone.

fabulator: Ok I updated my PR

PR #2041, opened 2018-04-05
Add react-apollo definitions

thymikee:
Based on https://github.com/apollographql/react-apollo/pull/1858 `react-apollo` package doesn't want to maintain flow types anymore, so I've decided to migrate them here.

For now I've added one basic test, but honestly there's a lot to test and add missing items (like new `ApolloConsumer` added in v2.1).


AndrewSouthpaw:
Nice work, thanks @thymikee.

PR #2043, opened 2018-04-05
Jasmine: Add `.and` for spies, `pending`, `DEFAULT_TIMEOUT_INTERVAL`

Mr-Wallet:
This exposes a bit more of Jasmine.

Unfortunately I haven't been able to run tests locally (Windows). Following the instructions in the README is kind of a non-starter because it immediately complains about an `ENOENT` related to the first definition. There's also a `CONTRIBUTING.md` in the definitions that throws a wrench in the works, which I had to locally delete to confirm structure.

I got a lot farther running `run_def_tests.cmd` but I hit a lot of errors, especially `ENOENT: no such file or directory, scandir 'C:\gitrepos\flow-typed\cli\.flow-bins-cache'`.

If anyone can lend advice, I'd much appreciate it, since I'd also like to submit some fixups for `lodash/fp` at some point.

### DIFF
--- a/definitions/npm/jasmine_v2.4.x/flow_v0.25.x-/jasmine_v2.4.x.js
+++ b/definitions/npm/jasmine_v2.4.x/flow_v0.25.x-/jasmine_v2.4.x.js
@@ -35,6 +35,7 @@ declare function fit(name: string, fn: Function, timeout?: number): void;
 declare function xit(name: string, fn: Function): void;
 
 declare function expect(value: mixed): JasmineExpectType;
+declare function pending(message?: string): void;
 declare function fail(err?: Error | string): void;
 
 // TODO handle return type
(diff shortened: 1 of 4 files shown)


GAntoine:
Hey @Mr-Wallet, sorry to hear you hit so many road blocks when testing. Fixing the `CONTRIBUTING.md` issue should be as easy as updating to the latest patch of `flow-typed`. As for the other issue, try creating the cache manually. There is an up-and-coming PR to allow you to set a custom cache-dir location, but that won't be available for a while.

GAntoine:
The latest version of flow-typed on npm supports extra file types. Try removing it globally and `npm install -g flow-typed@2.4.0`.

Mr-Wallet:
I'm not sure what you mean. Why would I try to confirm the file structure (`validate-defs`) from a global install of flow-typed?

GAntoine:
Most commands will check the file structure, including install (for some reason...). The latest version of flow-typed allows for more file types, and loosens the restrictions. Just upgrading should fix the issue you're facing.

PR #2045, opened 2018-04-06
Update redux-actions to match updated redux definitions

dashed:
Closes https://github.com/flowtype/flow-typed/issues/1963

Referred updated redux definitions are at: https://github.com/flowtype/flow-typed/pull/1958

-----

The semantics of the reducer returned by either `handleAction` or `handleActions` are actually different from the semantics of the reducer that are consumed by either `handleAction` or `handleActions`.

Reducers consumed by either `handleAction` or `handleActions` are assumed to be given the actual `State` type, since internally, `redux-actions` will perform the action type matching for us, and will always provide the expected state.

See the following tests:

- https://github.com/redux-utilities/redux-actions/blob/de4e8f13192d306fff9bb3929a140356d557def0/src/__tests__/handleAction-test.js#L45-L51

- https://github.com/redux-utilities/redux-actions/blob/de4e8f13192d306fff9bb3929a140356d557def0/src/__tests__/handleActions-test.js#L303-L308

This is why I've kept the `Reducer<S, A>` type as declared originally; and instead, define a new reducer type that is assumed by the `redux` library:

```
declare type ReduxReducer<S, A> = (state: S | void, action: A) => S;
```

### DIFF
--- a/definitions/npm/redux-actions_v2.x.x/flow_v0.39.x-/redux-actions_v2.x.x.js
+++ b/definitions/npm/redux-actions_v2.x.x/flow_v0.39.x-/redux-actions_v2.x.x.js
@@ -58,6 +58,7 @@ declare module "redux-actions" {
   declare function createActions(...identityActions: string[]): Object;
 
   declare type Reducer<S, A> = (state: S, action: A) => S;
+  declare type ReduxReducer<S, A> = (state: S | void, action: A) => S;
 
   declare type ReducerMap<S, A> =
     | { next: Reducer<S, A> }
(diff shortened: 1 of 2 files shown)


definitions/npm/redux-actions_v2.x.x/flow_v0.39.x-/redux-actions_v2.x.x.js
```diff
@@ -58,6 +58,7 @@ declare module "redux-actions" {
   declare function createActions(...identityActions: string[]): Object;
 
   declare type Reducer<S, A> = (state: S, action: A) => S;
+  declare type ReduxReducer<S, A> = (state: S | void, action: A) => S;
```
AndrewSouthpaw: Can you add some commentary on the difference between these two? It's not obvious, and would be helpful to others.

dashed: @AndrewSouthpaw I've added some commentary. Hopefully it's sufficient.

PR #2049, opened 2018-04-06
update puppeteer typedef and tests

infinityfarm:
@saadq This updates the typedef taking your comments into consideration and adds more extensive tests. Please let me know if you have any other comments/suggestions. cc @AndrewSouthpaw

### DIFF
--- a/definitions/npm/puppeteer_v1.2.x/flow_v0.58.x-/puppeteer_v1.2.x.js
+++ b/definitions/npm/puppeteer_v1.2.x/flow_v0.58.x-/puppeteer_v1.2.x.js
@@ -5,987 +5,1000 @@
  * https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/puppeteer
  */
 
-import ChildProcess from 'child_process';
-import EventEmitter from 'events';
+declare module 'puppeteer' {
+  import type { ChildProcess } from 'child_process';
 
-/** Keyboard provides an api for managing a virtual keyboard. */
-type $npm$puppeteer$Keyboard = {
+  declare class EventEmitter {
+    // deprecated
+    static listenerCount(emitter: EventEmitter, event: string): number;
+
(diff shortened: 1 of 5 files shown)


saadq:
Hey @infinityfarm, awesome work! So you mentioned in [this comment](https://github.com/flowtype/flow-typed/pull/2037#issuecomment-379282191) that you were having trouble extending the `EventEmitter` class, which I assume is why you created your own mock class. So as you probably noticed, you can only `import type` or `import typeof`, which only imports the type, so you can't use the type as a value. 

Instead, you will want to use Flow's private types/classes. Most of these are named like `module$Name`. in this case, you will want to extend `events$EventEmitter`.

Example:

```ts
declare module 'example' {
  declare class MyClass extends events$EventEmitter {

  }
}
```

There is a list of these Flow "private types" available [here](https://www.saltycrane.com/flow-type-cheat-sheet/latest/). The `events$EventEmitter` is in the [Node.js "private" types](https://www.saltycrane.com/flow-type-cheat-sheet/latest/#lib/node.js-private) section.

Other than that, everything looks really good!

infinityfarm:
@saadq I've updated it to use the private type the best I could. I had to create an `OverridableEventEmitter` class that extends the built in `events$EventEmitter` because using the `events$EventEmitter` unchanged resulted in the following error message:
screen shot 2018-04-07 at 12 05 44 am
If there's a better way to fix that please advise.

Thanks

saadq:
@infinityfarm Ah, yeah I remember having issues with the string literal vs string approach too... I had a StackOverflow post about it [here](https://stackoverflow.com/questions/49411515/limiting-the-type-of-eventname-in-a-class-that-extends-eventemitter-with-flow) that got answered.

The approach you did here looks fine to me too, I think this is ready for merging unless @GAntoine or anyone has any other comments.

GAntoine:
No comments, changes look good to me. 👍

Issue #1934, opened 2018-03-08
react-redux connect MapStateToProps state parameterization


Ashoat:
#1731 improved a lot of things! But it's worth noting that the past definition of `MapStateToProps` was parameterized on state, which make it possible to "flow" your Redux state types into your component prop types, without the need to explicitly type the return of `MapStateToProps`.

Wondering if there was an explicit reason that parameterization was dropped. Is there a bug in Flow or some sort of insufficiency that's preventing it from being typed that way in the revamp?

villesau:
@Ashoat  Thanks for reporting the issue! All I know is that it's been working previously, see this: https://github.com/facebook/flow/issues/5792#issuecomment-372346578 and [try flow link](https://flow.org/try/#0PTAEAEDMBsHsHcBQBLAtgB1gJwC6hwJ7oCmoA3qAMKwawB2xdOAKkcQDSgCi0xqjOAApZY6AM6gAvqEgjUoAORZiAQwDGOBQG4UtXKABUoFRIBKqjTLmLl6zToAmxNdBXL8bUABFkY9Cpw1AAsAHgBBAD5QAF5jOgJHZ1d3QhJQAHlUZBwfPwDgkOpaBiYo2IASH0hIQppMEpxOMgdff0CggC5vVvzQgwjJCMSXN1JU0gBZFXQAZRwA4mZYYVExENMZwS70gCMAK2ccMtAACjF5nGJt-cOAShiojcFh5NJiAA9MfUgAVzoNZD0UBqegMDS1VBdIr1ASsEghfqcLxbDI3DScJ7XA4aCInRCgUCoaZzBZLFbiLpTWYXRbLETidabCKIW5dE4g4oCKE0e7RKLQ+iwtghSrIaohHh8ATktZFCIYpmgABk3kEQ0QGvGoBlMXIshoJMuZPpYi65ywyDoAHNJDoRmIJEVQB9LnQHGYLDgAHQChohGVRMj40DKN3ELAne5BgkE5Q4H5YOigEItABuETIOCCvi96BNXv1qENtJlkhCwDT6oJkkQNcQWuLurIKi6dB+qB24dtGpBdHOhOJNONq11ZxpXWLvMDwbjCaT0ZjheLw4poHOCy9KmDNe7iF7-eodDBlwcut7x5OROppLpq1uJyKtx0tSPh2Ip+AQyAA)

[...]

PR #2057, opened 2018-04-09
[react-redux] Parameterize MapStateToProps on Redux state

Ashoat:
Context in #1934. This PR parameterizes the `MapStateToProps` function on Redux state so that the first parameter to that function doesn't have to be `Object`. This avoids forcing library users to either `any`-cast the Redux state or to do runtime assertions on it.

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/react-redux_v5.x.x.js
@@ -28,7 +28,7 @@ declare module "react-redux" {
   Com = React Component
   */
 
-  declare type MapStateToProps<SP: Object, RSP: Object> = (state: Object, props: SP) => RSP;
+  declare type MapStateToProps<RS: Object, SP: Object, RSP: Object> = (state: RS, props: SP) => RSP;
 
   declare type MapDispatchToProps<A, OP: Object, RDP: Object> = (dispatch: Dispatch<A>, ownProps: OP) => RDP;
 
@@ -50,8 +50,8 @@ declare module "react-redux" {
 
   declare type OmitDispatch<Component> = $Diff<Component, {dispatch: Dispatch<*>}>;
 
-  declare export function connect<Com: ComponentType<*>, DP: Object, RSP: Object>(
-    mapStateToProps: MapStateToProps<DP, RSP>,
+  declare export function connect<Com: ComponentType<*>, RS: Object, DP: Object, RSP: Object>(
+    mapStateToProps: MapStateToProps<RS, DP, RSP>,
     mapDispatchToProps?: null
   ): (component: Com) => ComponentType<$Diff<OmitDispatch<ElementConfig<Com>>, RSP> & DP>;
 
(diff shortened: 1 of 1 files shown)


GAntoine:
@Ashoat The specs need to be fixed before we can merge this in.

Ashoat:
Sorry what are the "specs"? You mean the failing tests? I don't have the bandwidth to address those right now, unfortunately. Context for why they are failing is [here](https://github.com/flowtype/flow-typed/pull/2018#issuecomment-379568272).

Up to you whether to merge this PR, but I'm pretty sure that the test failures are unrelated and that this change represents an improvement to the libdef.

villesau:
@Ashoat can you elaborate the issue still? Why parametarization was needed? I think I've been able to specify the type for state object just fine, like `Component = connect((state: {prop: number}) => state.prop)`. It didn't seem to affect any current tests, and I didn't see any additional test to point out the problem.

Ashoat:
@villesau, here is a [minimal repro](https://flow.org/try/#0C4TwDgpgBAIglgZzAQ2AYwBYEEAKBJAHiwD4oBeKACmTWDgHsA7ALiiwEpzSsBuAKFCRYiFOgxFWAbyiCIrACQBlAK4AjWQQTAATnEYBzUgF9SFeElSZchEvwHhoAWWRhFwVBAAq9HNvpgEAkUcVgB5VQArCFoAGigAJWCwyOjgUyotD2So2KgwPwDWYM4yUkScflkoUIBbOGBzUUwCAGF6GrAmCEY08ih5eAAzQdb2zsZu4DjJABMRSwxWRoWCACpiE34Z6IAbZG1oQeVGWgZGKAOaYHiIGeUADzbGCdpRmtYbq-k2jq6ezwca2IcRgIWqKVy5WyqWIlD4UCgNRcbg83l8-gQrGcrncwC8PgKgVBcXKwL47FYlDQYz+wFYPxKZQgXx+40mAMgBAGcGGBFq9WWYgIn1o8gAojsIDVJk9BnB9G9iMCEsFSAAyWA4Yh2AD0OqgngwiCgAHd6NoANYIPjUxhaRHI3H49EBMX3ZAdSV9SiZPFSKDIVhaXQGKBGRlUSTwgNBp0AOmQMT44f4l1oNzujyYL2AlCRONRBIxbo9YEl7F1+sNxpm9AgCCgjHowD4RxOdCYUFtOaC0NySXBOTScIR+ZReLRhMpvrkUEUEeCSc4UYRB2Aym05zT11uDyeObzjsLLoQFeTQA) if you'd like to play around to understand the problem more.

> It didn't seem to affect any current tests, and I didn't see any additional test to point out the problem. Without test case it's very hard to understand what it potentially fixed.

I agree with this philosophy; my first thought was to figure out why the tests were passing, and what would have to be added. But when I realized the tests were broken I decided not to bother. If you can get the tests fixed up I'll gladly add a test case for this :)

villesau:
@Ashoat Ok now I got it, thanks! Yeah I didn't come up with that kind of use case back when rewriting the libdefs so no wonder why I couldn't find benefits of paremeterizing it. Once I have time I could check out what could be done for test suite so that it wouldn't run unrelated tests.

AndrewSouthpaw:
I'm still not seeing what would be the benefit of writing out something like this?

```js
function connect<S: Object, SP: Object>(
  mapStateToProps: (state: S) => SP,
) {
  return reactReduxConnect(mapStateToProps);
}
```

Just genuinely curious. :)

Ashoat:
That’s just the minimal repro :)

Here is [how I actually use it](https://github.com/Ashoat/squadcal/blob/master/lib/utils/redux-utils.js). Basically, I have a wrapper around `connect` that combines Redux state with the `dispatch` function to create state-dependent action creators.

PR #2058, opened 2018-04-09
react-i18next libdef update

TLadd:
No longer require using a Translator type with the translate HOC.
Updated tests to use describe/it

Rather than doing
```
import { translate, type Translator, type TFunction } from 'react-i18next'
type Props = { s: string }
type PropsWithT = Props & { t: TFunction }
const Comp = ({ s, t }: PropsWithT) => { .... }
const translator: Translator<Props, PropsWithT> = translate()
const TranslatedComp = translator(Comp)
```

You now do
```
import { translate, type TFunction } from 'react-i18next'
type Props = { s: string,  t: TFunction }
const Comp = ({ s, t }: Props) => { .... }
const translator = translate()
const TranslatedComp = translator(Comp)
```

And TranslatedComp will still have the same type information (it knows that t is supplied to Comp through translate, but s still need to be passed). 

Took the same approach as the react-router-dom libdef uses to type the withRouter HOC.


gre:
Hi, this `{| ...TranslatorProps, ...P |}` seems to trigger a problem if a component has some non-optional props that are covered by defaultProp, it will says "Property `someProp` is missing in object type [1] but exists in `Props` [2]."

GAntoine:
@gre Can you open an issue linking to this PR?

TLadd:
Appears to be a limitation with typing HOC's in general. https://github.com/facebook/flow/issues/4644. Flow has no way of knowing what default props you have declared.

PR #2059, opened 2018-04-09
First draft of RxJS 6 support

radex:
Adds initial flow support for RxJS 6 — which gets rid of operators on `Observable` class and moves them to pure functions for efficiency.

This work is admittedly incomplete — it's just enough for my project to pass Flow, but it might be just barely enough for other projects, that won't move to RxJS 6 until there's Flow support.

Here's what's done:

- (almost) all operators are moved to the `rxjs/operators` export
- _some_ Observable creation functions are moved from Observable static methods to `rxjs/observable/xxx` exports

TODO (in future PRs if this one is allowed to be merged as-is):

- move the rest of Observable creation functions to the separate export
- fix some of the signatures (for operators that have multiple variants — a `xx & yy & zz` syntax must be used)
- tests
- audit for some operator name changes


radex:
> thanks for getting this started! one thing i noticed is that there are a lot of static methods, but they were removed weren't they?

Yeah — that's just a WIP artifact. I will move the rest of them to `rxjs/observable/xxxx`, since otherwise you wouldn't get Flow errors for using old syntax…

radex:
@matthewwithanm Moved the rest of the static methods out of Observable :)

matthewwithanm:
@nmote do you think this could be merged as-is or do you need tests first?

radex:
OK, will do!

radex:
@AndrewSouthpaw I copied & adjusted the tests from rxjs5 and fixed the failing tests :)

radex:
hey @AndrewSouthpaw, pinging :)

radex:
RxJS 6.0.0 was released, so it would be nice to have this merged :))

PR #2069, opened 2018-04-11
Export ApolloConsumer component with react-apollo

bwlt:
Add missing API
https://www.apollographql.com/docs/react/api/react-apollo.html#apollo-consumer

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -21,6 +21,10 @@ declare module 'react-apollo' {
     client: ApolloClient;
   }
 
+  declare export class ApolloConsumer extends React$Component<{}> {
+    render(): (client: ApolloClient) => React$Element<*>;
+  }
+
   declare export class ApolloProvider extends React$Component<ProviderProps> {
     childContextTypes: {
       client: ApolloClient,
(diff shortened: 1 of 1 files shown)


definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
```diff
@@ -21,6 +21,10 @@ declare module 'react-apollo' {
     client: ApolloClient;
   }
 
+  declare export class ApolloConsumer extends React$Component<{}> {
+    render(): (client: ApolloClient) => React$Element<*>;
```
AndrewSouthpaw: I *think* this should be `React$Node`, unless you have reasoning otherwise?

bwlt: Yes, a typo. But for now all this new definition is totally wrong :(
The ApolloConsumer render attribute should be a function that returns `React$Node`; and the render prop `(client: ApolloClient) => React$Node` should be defined on the `React$Component` first type argument.
Let me work on that

bwlt:
Can I work & push other commits on the new `Query` component and `Mutation` component?
At least we have some base code to be improved in the future

AndrewSouthpaw:
Sorry I'm not following, provide more words please. :)

bwlt:
I wanted to add definitions for `Query` and `Mutation` component, but I see that some definitions were merged through https://github.com/flowtype/flow-typed/pull/2086 .
I consider this pull request completed

AndrewSouthpaw:
I was on vacation, sorry for the delay. 😄

PR #2078, opened 2018-04-12
Handle defaultProps properly in react-i18next translate HOC

TLadd:
Addresses https://github.com/flowtype/flow-typed/issues/2076. The output of the HOC will no longer require props that are handled by the component's defaultProps.

Thanks to the flow docs for explicitly covering this issue: https://flow.org/en/docs/react/hoc/#toc-supporting-defaultprops-with-react-elementconfig.


TLadd:
Split into two commits here: https://github.com/flowtype/flow-typed/pull/2089

PR #2082, opened 2018-04-13
[algoliasearch] Fix async import() case

mrm007:
`algoliasearch` definition doesn't work when it's imported with an `await import()`. This fixes it.

### DIFF
--- a/definitions/npm/algoliasearch_v3.x.x/flow_v0.56.x-/algoliasearch_v3.x.x.js
+++ b/definitions/npm/algoliasearch_v3.x.x/flow_v0.56.x-/algoliasearch_v3.x.x.js
@@ -169,11 +169,13 @@ declare module 'algoliasearch' {
   declare type ClientOptions = $algoliasearch$ClientOptions;
   declare type Index = $algoliasearch$Index;
   declare type Settings = $algoliasearch$Settings;
-  declare module.exports: (
-    applicationID: string,
-    apiKey: string,
-    options?: ClientOptions
-  ) => Client;
+  declare module.exports: {
+    (
+      applicationID: string,
+      apiKey: string,
+      options?: ClientOptions
+    ): Client;
+  }
 }
 
 declare module 'algoliasearch/reactnative' {
@@ -181,11 +183,13 @@ declare module 'algoliasearch/reactnative' {
   declare type ClientOptions = $algoliasearch$ClientOptions;
   declare type Index = $algoliasearch$Index;
   declare type Settings = $algoliasearch$Settings;
-  declare module.exports: (
-    applicationID: string,
-    apiKey: string,
-    options?: ClientOptions
-  ) => Client;
+  declare module.exports: {
+    (
+      applicationID: string,
+      apiKey: string,
+      options?: ClientOptions
+    ): Client;
+  }
 }
 
 declare module 'algoliasearch/lite' {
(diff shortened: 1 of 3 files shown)


definitions/npm/algoliasearch_v3.x.x/test_algoliasearch-v3.x.x.js
```diff
@@ -1,8 +1,25 @@
 // @flow
-import algoliasearch from 'algoliasearch';
+import algoliasearchImport from 'algoliasearch';
```
AndrewSouthpaw: Can you use the new `describe` and `it` blocks for extra winning? More in the CONTRIBUTING.md file. :)

definitions/npm/algoliasearch_v3.x.x/flow_v0.56.x-/algoliasearch_v3.x.x.js
```diff
@@ -169,33 +169,39 @@ declare module 'algoliasearch' {
...
-    apiKey: string,
-    options?: ClientOptions
-  ) => Client;
+  declare module.exports: {
+    (
+      applicationID: string,
+      apiKey: string,
+      options?: ClientOptions
+    ): Client;
+  }
```
AndrewSouthpaw: I'd be curious to see, is there any documentation that guided your fix here? I think this would be useful to know for other libdefs if they're required asynchronously.

definitions/npm/algoliasearch_v3.x.x/test_algoliasearch-v3.x.x.js
```diff
@@ -1,8 +1,25 @@
...
+function syncImportTests() {
+  const searchIndex = algoliasearchImport('', '').initIndex('');
+  searchIndex.search('test').then(result => console.warn(result.hits));
+  // $ExpectError
+  searchIndex.search('test').nonPromiseProperty;
+}
 
-// $ExpectError
-searchIndex.search('test').nonPromiseProperty;
+function requireTests() {
```
AndrewSouthpaw: Is this function testing something different from the `syncImportTests` one?

definitions/npm/algoliasearch_v3.x.x/test_algoliasearch-v3.x.x.js
```diff
@@ -1,8 +1,25 @@
...
+  // $ExpectError
+  searchIndex.search('test').nonPromiseProperty;
+}
+
+async function asyncImportTests() {
+  const algoliasearch = await import('algoliasearch');
+  const searchIndex = algoliasearch('', '').initIndex('');
+  searchIndex.search('test').then(result => console.warn(result.hits));
+  // $ExpectError
+  searchIndex.search('test').nonPromiseProperty;
```
AndrewSouthpaw: No need to test this part multiple times... just check that async works.

definitions/npm/algoliasearch_v3.x.x/flow_v0.56.x-/algoliasearch_v3.x.x.js
mrm007: Not really, it was a bit of trial and error to get it working. I saw the syntax used somewhere, but I can't remember where.

definitions/npm/algoliasearch_v3.x.x/test_algoliasearch-v3.x.x.js
mrm007: Will do.

definitions/npm/algoliasearch_v3.x.x/test_algoliasearch-v3.x.x.js
mrm007: Nope, just the way the library is imported --  `import ...` vs `require(...)`.

mrm007:
Build is failing for some unrelated reason. What should I do @AndrewSouthpaw?

PR #2089, opened 2018-04-13
React i18next fix default props

TLadd:
Fixes https://github.com/flowtype/flow-typed/issues/2076

The output of the HOC will no longer require props that are handled by the component's defaultProps.

Thanks to the flow docs for explicitly covering this issue: https://flow.org/en/docs/react/hoc/#toc-supporting-defaultprops-with-react-elementconfig.


definitions/npm/react-i18next_v7.x.x/flow_v0.64.x-/react-i18next_v7.x.x.js
```diff
@@ -0,0 +1,92 @@
...
+  declare type Translator<P: {}, Component: React$ComponentType<P>> = (
+    WrappedComponent: Component
+  ) => React$ComponentType<
+    $Diff<React$ElementConfig<Component>, TranslatorPropsVoid>
+  >;
+
+  declare type TranslateOptions = $Shape<{
+    wait: boolean,
+    nsMode: "default" | "fallback",
+    bindi18n: false | string,
```
dpweberza: @GAntoine @TLadd 
The casing of this option is incorrect, it should be: bindI18n

https://github.com/i18next/react-i18next/blob/114f9e0e123ed59f404e7b7eec9546797c728639/src/context.js#L4

PR #2092, opened 2018-04-14
Add libdef for url-parse

PatriceVignola:
Added a libdef for the [url-parse package](https://www.npmjs.com/package/url-parse).


PatriceVignola:
I changed flow_all to flow_v0.59.x- (the smallest version that the tests ran against).

PR #2094, opened 2018-04-15
Modify the contributing guidelines testing steps

PatriceVignola:
Since `./run_def_tests.sh` is not working very well to run tests and
validate libdefs locally, I change the documentation to instead run the
command `flow-typed validate-defs definitions && flow-typed run-tests`.
This works like a charm.


GAntoine:
Much better! I might go in sometime and tweak it a little more, but this will hopefully reduce confusion for some.

PR #957, opened 2017-06-16
Use proper types for `addSnapshotSerializer`

eliperkins:
The current flow-typed definition will error on any proper usage of this function, since it needs to take in an object shaped as such, not a function `(input: Object): string`.

This PR uses the same types as Jest declares for these functions internally.

This function is [defined within Jest here](https://github.com/facebook/jest/blob/3c94c23786c223c17ee8f051be32cee549a2ff1a/packages/jest-jasmine2/src/jest-expect.js#L31), by assigning the value to the [function `addSerializer`, defined here](https://github.com/facebook/jest/blob/dec54a34ffbf37f859a252d05901e87615accb7b/packages/jest-snapshot/src/plugins.js#L25), which takes in a [`type Plugin`, defined here](https://github.com/facebook/jest/blob/dec54a34ffbf37f859a252d05901e87615accb7b/types/PrettyFormat.js#L43-L52).

You can see [how Jest tests this internally here](https://github.com/facebook/jest/blob/cd678bca8bc31fb1fa5b0387fcd240523b01d319/integration_tests/snapshot-serializers/__tests__/snapshot-test.js#L89-L92).

You can see more [documentation about `expect.addSnapshotSerializer(serializer)` in Jest's docs](https://facebook.github.io/jest/docs/expect.html#expectaddsnapshotserializerserializer).

PR #2096, opened 2018-04-15
Use proper types for addSnapshotSerializer

GAntoine:
Replaces #957


GAntoine:
@mikewoudenberg Does this make sense to you?

mikewoudenberg:
Awesome, looks like something we can work with!

PR #2097, opened 2018-04-16
Improve checks for connect in react-redux@5.x.x 

tf:
I found that, with the current definitions, the return values of `mapStateToProps`, `mapDispatchToProps` and `mergeProps` are not checked against the wrapped component as I would have expected.

## Example

Given a component that takes a string prop, trying to pass a number for that prop in `mapStateToProps` fails as expected:

```jsx
  type Props = {
    str: string
  };

  class Com extends React.Component<Props> {
    render() {
      return <div>{this.props.str}</div>;
    }
  }

  type State = {num: number, str: string};

  const mapStateToProps = (state: State) => {
    return { // error
      str: 123
    }
  };

  connect(mapStateToProps)(Com);
```
But soon as a property from state or a selector function is used instead of the number literal, no errors are detected anymore:

```jsx
  type Props = {
    str: string
  };

  class Com extends React.Component<Props> {
    render() {
      return <div>{this.props.str}</div>;
    }
  }

  type State = {num: number, str: string};

  const mapStateToProps = (state: State) => {
    return { // no error
      str: state.num
    }
  };

  connect(mapStateToProps)(Com);
```

Same basically holds for the other parameters of `connect`. There are also cases where unknown props are not detected correctly. 

As a result misspelled props or selector functions with incompatible return values go unnoticed.

I've created a branch with some examples that I'd expect to result in errors, but which pass except for simple cases:

https://github.com/flowtype/flow-typed/compare/master...tf:redux-selector-problem-2

I deliberately did not put in `$ExpectError` comments, so that the current error messages are visible. The Travis build is quite noisy since all of the tests for the `flow_v0.54.x-v0.61.x` and `flow-v0.62.x` variant are failing on `master`, but here's the relevant part:

https://travis-ci.org/tf/flow-typed/jobs/367131347#L1834

## Proposed Solution

Looking at the way `connect` is defined at the moment, the main problem IMO is that the return value of  `mapStateToProps` etc. is not directly related to the wrapped components prop types:

```js
  declare type MapStateToProps<S: Object, SP: Object, RSP: Object> = (state: S, props: SP) => RSP;

  declare export function connect<
    Com: ComponentType<*>,
    S: Object,
    DP: Object,
    RSP: Object,  // <== no direct relation to Com
    CP: $Diff<OmitDispatch<ElementConfig<Com>>, RSP>
    >(
    mapStateToProps: MapStateToProps<S, DP, RSP>,
    mapDispatchToProps?: null
  ): (component: Com) => ComponentType<CP & DP>;
```
At least when no `mergeProps` function is present, `mapStateToProps` and `mapDispatchToProps` should only be allowed to return valid props for the component. This PR therefore makes the following changes:

```js
  declare export function connect<
    P: Object, // <==
    Com: ComponentType<P>, // <== Introduce a type variable for the props of Com
    S: Object,
    DP: Object,
    RSP: $Shape<P>, // <== Allow only those props in state props
    CP: $Diff<OmitDispatch<ElementConfig<Com>>, RSP>
    >(
    mapStateToProps: MapStateToProps<S, DP, RSP>,
    mapDispatchToProps?: null
  ): (component: Com) => ComponentType<CP & DP>;
```

See also http://sitr.us/2015/05/31/advanced-features-in-flow.html#shapet.

If `mergeProps` is present, `mapStateToProps` and `mapDispatchToProps` are still allowed to return arbitrary data. Only the return value of `mergeProps` itself has to match the shape of the wrapped component's props.

The PR contains more valid and invalid examples.

The test failures either come from the `react-redux_v5.x.x/flow_v0.54.x-v0.61.x/` directory, which is not altered by this PR or Flow 0.62, which appears to have problem importing `ElementConfig` from `react`. For 0.63 and above tests pass. This problem also exists on `master`.

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/react-redux_v5.x.x.js
@@ -27,6 +27,7 @@ declare module "react-redux" {
   RMP = Returned merge props
   CP = Props for returned component
   Com = React Component
+  P = Props of Com
   */
 
   declare type MapStateToProps<S: Object, SP: Object, RSP: Object> = (state: S, props: SP) => RSP;
@@ -52,10 +53,11 @@ declare module "react-redux" {
   declare type OmitDispatch<Component> = $Diff<Component, {dispatch: Dispatch<*>}>;
 
   declare export function connect<
-    Com: ComponentType<*>,
+    P: Object,
+    Com: ComponentType<P>,
     S: Object,
     DP: Object,
-    RSP: Object,
+    RSP: $Shape<P>,
     CP: $Diff<OmitDispatch<ElementConfig<Com>>, RSP>
     >(
     mapStateToProps: MapStateToProps<S, DP, RSP>,
@@ -68,13 +70,14 @@ declare module "react-redux" {
   ): (component: Com) => ComponentType<OmitDispatch<ElementConfig<Com>>>;
 
   declare export function connect<
-    Com: ComponentType<*>,
+    P: Object,
+    Com: ComponentType<P>,
     A,
     S: Object,
     DP: Object,
     SP: Object,
-    RSP: Object,
-    RDP: Object,
+    RSP: $Shape<P>,
+    RDP: $Shape<P>,
     CP: $Diff<$Diff<ElementConfig<Com>, RSP>, RDP>
     >(
     mapStateToProps: MapStateToProps<S, SP, RSP>,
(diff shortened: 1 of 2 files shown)


tf:
I'll do some more real life testing in one of our projects. Just wanted to submit this early to get some feedback.

tf:
Sorry for not getting back earlier. I'll try to look into villesau's comments.

villesau:
@tf have you had time to check it out?

peter-leonov:
With the new Flow and fixed typings it's enough to just instantiate the component to get the (misleading) error: https://github.com/flow-typed/flow-typed/pull/3035/files#diff-a8524c9f8ed49a21b3ff85e6a1b27499R579

PR #2101, opened 2018-04-17
lodash/fp: Fix compact, unset, dissoc, dissocPath

Mr-Wallet:
lodash compact seems to have gotten fixed at some point since the fp definitions were added, this ports the same fix into fp. Namely: The input array is currently `Array<T>`, but it must be 
 `Array<?T>` for an output of `Array<T>`because the whole point of `compact` is to get rid of nil elements.

`unset/dissoc/dissocPath` returns a `boolean` in normal lodash, reflecting the success of the operation. However `fp` does not mutate the input, so instead it returns an `Object` which is a new object resulting from attempting the operation on the input object. This is simply a case of the original `fp` definition writer not noticing that its return type is different from normal lodash.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
@@ -1619,7 +1619,7 @@ declare module "lodash/fp" {
     // Array
     chunk<T>(size: number): (array: Array<T>) => Array<Array<T>>;
     chunk<T>(size: number, array: Array<T>): Array<Array<T>>;
-    compact<T, N: T>(array: Array<N>): Array<T>;
+    compact<T, N: ?T>(array?: ?$ReadOnlyArray<N>): Array<T>;
     concat<T, U, A: Array<T> | T, B: Array<U> | U>(
       base: A
     ): (elements: B) => Array<T | U>;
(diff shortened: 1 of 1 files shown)


definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
```diff
@@ -1619,7 +1619,7 @@ declare module "lodash/fp" {
     // Array
     chunk<T>(size: number): (array: Array<T>) => Array<Array<T>>;
     chunk<T>(size: number, array: Array<T>): Array<Array<T>>;
-    compact<T, N: T>(array: Array<N>): Array<T>;
+    compact<T, N: ?T>(array?: ?$ReadOnlyArray<N>): Array<T>;
```
GAntoine: Is `compact` callable without passing in an array?

Mr-Wallet: Virtually everything in `lodash/fp` can be called without providing the "expected" collection, and it seems like the definitions overlook this in general, a concern that I recently brought up [here](https://github.com/flowtype/flow-typed/pull/1262#issuecomment-382199299). In the case of `compact`, calling with no value or a nil value will return an empty array (which trivially qualifies as an `Array<T>`).

PR #2105, opened 2018-04-18
Support factory types for mapStateToProps and mapDispatchToProps

### LINKED ISSUE #1924: [react-redux] MapStateToProps and MapDispatchToProps missing alternate types
React Redux allows MapStateToProps to be:
- a function that returns:
  - an object
  - a function that itself returns an object

and allows MapDispatchToProps to be:

- an object
- a function that returns: 
  - an object
  - a function that itself returns an object

The previous version of the libdef, `flow_v0.54.x-v0.61.x/react-redux_v5.x.x.js`, included each of these types (with the exception of MapDispatchToProps' function that returns a function that returns an object). In the current version, `flow_v0.62.0-/react-redux_v5.x.x.js`, both MapStateToProps and MapDispatchToProps include only:

- a function that returns:
  - an object

Was there a reason for the alternate types' omission?

Also, shouldn't props/ownProps be declared optional?

grlira:
This fixes #1924 by introducing the factory type for MapStateToProps and MapDispatchToProps, at the expense of having to explicitly declare whether we're using the factory style or not.

I.E. this would require that  the type of mapStateToProps is explicitly declared like such

```javascript
const mapStateToPropsFactory: MapStateToPropsFactory<*, *, *> = () => {
    return (state, props) => {
// (...)
conect(mapStateToPropsFactory)(Component)
```
OR
```javascript
const mapStateToProps: MapStateToProps<*, *, *> = (state) => (
// (...)
conect(mapStateToProps)(Component)
```

Edit: changed the type of `mapStateToProps` and `mapStateToPropsFactory` to match the updates in [1fd1c00](https://github.com/flowtype/flow-typed/pull/2105/commits/1fd1c008534550acb1aaf7b649017d1a8f7c1f11)

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.62.0-/react-redux_v5.x.x.js
@@ -30,8 +30,12 @@ declare module "react-redux" {
   */
 
   declare type MapStateToProps<S: Object, SP: Object, RSP: Object> = (state: S, props: SP) => RSP;
+  declare type MapStateToPropsFactory<S, SP, RSP> = () => MapStateToProps<S, SP, RSP>;
+  declare type MapStateToPropsUnion<S, SP, RSP> = MapStateToPropsFactory<S, SP, RSP> | MapStateToProps<S, SP, RSP>;
 
   declare type MapDispatchToProps<A, OP: Object, RDP: Object> = (dispatch: Dispatch<A>, ownProps: OP) => RDP;
+  declare type MapDispatchToPropsFactory<A, OP, RDP> = () => MapDispatchToProps<A, OP, RDP>;
+  declare type MapDispatchToPropsUnion<A, OP, RDP> = MapDispatchToPropsFactory<A, OP, RDP> | MapDispatchToProps<A, OP, RDP>;
 
   declare type MergeProps<SP: Object, DP: Object, MP: Object, RMP: Object> = (
     stateProps: SP,
@@ -58,7 +62,7 @@ declare module "react-redux" {
     RSP: Object,
     CP: $Diff<OmitDispatch<ElementConfig<Com>>, RSP>
     >(
-    mapStateToProps: MapStateToProps<S, DP, RSP>,
+    mapStateToProps: MapStateToPropsUnion<S, DP, RSP>,
     mapDispatchToProps?: null
   ): (component: Com) => ComponentType<CP & DP>;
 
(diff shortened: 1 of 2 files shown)


lukeapage:
shouldn't `() => (*, *) => *` be wrapped up in an exported type name in the library, so you import that rather than the ugly long syntax?

villesau:
What happens if the annotation is missing? Have you considered of opening an issue to Flow instead? I don't think extra annotation should be needed, it looks like inference bug in Flow instead.

grlira:
@villesau if the annotation is missing then flow raises an error, because it can't disambiguate between `MapStateToProps` and `MapStateToPropsFactory`

I've raised a bug with Flow [#6212](https://github.com/facebook/flow/issues/6212)

AndrewSouthpaw:
My vote is 👎, mostly because it looks like it requires a lot of boilerplate that's fairly hard to reason about. It would seriously decrease the legibility. I'm all for supporting the factory types, and also I'd rather wait for flow to address the bug so types can be inferred.

lukeapage:
@AndrewSouthpaw @villesau do you know anyone on the flow team who can look at the issue? Otherwise we are stuck with maintaining our own types - we can't use flow-typed. And the flow repo has alot of bugs that don't get responded to or looked at - I don't feel confident the "bug" would ever be fixed.

I've pondered about investigating flow myself, but I'm not familiar with ML and there doesn't appear any documentation on developing flow that would help me get started..

AndrewSouthpaw:
> I've pondered about investigating flow myself, but I'm not familiar with ML and there doesn't appear any documentation on developing flow that would help me get started..

You, me, and everyone else. Facebook did a fine job of ensuring people don't hack on Flow that much by writing it in OCaml. 🙄 

I don't know anyone on the FB team unfortunately.

I hear you about the pain of maintaining your own types, and I'm not convinced that it would be the worth the pain to everyone else to adopt the extra boilerplate (which would be very hard to do in large codebases).

villesau:
This is very tricky case! We are a bit between the rock and the hard place with this. I completely understand the frustration about incompleteness on factory types. At the same time it will be very inconvenient to force people to explicitly type their connects, especially as it's not very idiomatic way of using Flow.

My view on this is very biased since I don't feel the pain of lack of factory types, and if this would be merged in, that would make using libdefs much harder for others. I wouldn't expect people to realize that there is an utility type to solve issue where Flow cannot infer the type. And the point about boilerplate is very good too.

I, for example, most likely wouldn't update to latest libdefs if this gets merged. It means that the situation would just be other way around. Would be very nice if we had contacts in Flow team.

PR #2118, opened 2018-04-20
Formik: Fixed ComputedProps fields access to read-only instead of write-only

meafmira:


GAntoine:
Wow, 2 PRs back-to-back for the same fix (#2117). Gonna have to close this one then, thanks for contributing!

PR #2122, opened 2018-04-22
Support to use string template for chalk

HsuTing:
[chalk v2.x.x](https://github.com/chalk/chalk#usage) can be used as `string template`.

For example:
```js
import chalk from 'chalk';

chalk`{red string template}`;
```

As a result,`declare module.exports` should be added to `declare module`.

### DIFF
--- a/definitions/npm/luxon_v0.2.x/flow_v0.32.0-/luxon_v0.2.x.js
+++ b/definitions/npm/luxon_v0.2.x/flow_v0.32.0-/luxon_v0.2.x.js
@@ -38,10 +38,10 @@ declare module "luxon" {
         localeCode?: ?string
       }
     ): string;
-    isValid: boolean;
-    name: string;
-    type: string;
-    universal: boolean;
+    +isValid: boolean;
+    +name: string;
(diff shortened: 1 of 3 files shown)


GAntoine:
@HsuTing You'll need to fix the specs for this change.

HsuTing:
@GAntoine I tried to fix, but I got error in local testing with [node cli/dist/cli.js validate-defs definitions && node cli/dist/cli.js run-tests](https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md#5-run-your-tests-with-flow-typed-validate-defs-definitions--flow-typed-run-tests-left-pad).

How should I do to run local testing?

HsuTing:
I copy [`chalk` flow file](https://github.com/chalk/chalk/blob/master/index.js.flow).
This work, but here is some problem about [`colorSupport `](https://github.com/chalk/chalk/blob/master/index.js.flow#L17-L22).
I had opened an issue in `chalk` to ask the type of [`colorSupport`](https://github.com/chalk/chalk/issues/267).

PR #2124, opened 2018-04-23
lodash and lodash/fp: get/getOr accept nil collections

Mr-Wallet:
Virtually all lodash functions can take `null`/`undefined` for their input collection; for example, `map` will never run its iterator function and simply return an empty array. However, it's not clear how much of this should be allowed by Flow. Allowing nil on absolutely everything may dramatically lower the overall usefulness of the type-checking because even though those _may_ work with nil inputs, it's very often the case that the programmer never _intended_ that as a possibility and would have preferred an error.

Despite that, there's one case which seems extremely useful to allow `null`/`undefined`, which is `get`/`getOr`. Perhaps the primary use of these functions over simply `accessing.things['directly']` is because it avoids `cannot read property "directly" of undefined` errors with lodash's built-in nil checks. If for whatever reason lodash cannot get the value desired, it will return `undefined`.

It would very convenient to be able to extend this safety up to the root level:
`fp.get(['optional', 'property'], root)`
Currently the above will pass Flow if `optional` is a primitive or not defined, but will fail Flow if `root` may be nil. This PR allows nil values for `root`. This avoids the extra verbosity of doing a separate nil check when `get` is already nil-checking from top to bottom.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
@@ -1126,7 +1126,7 @@ declare module "lodash" {
     functions(object?: ?Object): Array<string>;
     functionsIn(object?: ?Object): Array<string>;
     get(
-      object?: ?Object | ?$ReadOnlyArray<any>,
+      object?: ?Object | ?$ReadOnlyArray<any> | void | null,
       path?: ?$ReadOnlyArray<string> | string,
       defaultValue?: any
     ): any;
@@ -2764,8 +2764,8 @@ declare module "lodash/fp" {
     forOwnRight(iteratee: OIteratee<*>, object: Object): Object;
     functions(object: Object): Array<string>;
     functionsIn(object: Object): Array<string>;
-    get(path: $ReadOnlyArray<string> | string): (object: Object | $ReadOnlyArray<any>) => any;
-    get(path: $ReadOnlyArray<string> | string, object: Object | $ReadOnlyArray<any>): any;
+    get(path: $ReadOnlyArray<string> | string): (object: Object | $ReadOnlyArray<any> | void | null) => any;
+    get(path: $ReadOnlyArray<string> | string, object: Object | $ReadOnlyArray<any> | void | null): any;
     prop(path: Array<string> | string): (object: Object | Array<any>) => any;
     prop(path: Array<string> | string, object: Object | Array<any>): any;
     path(path: Array<string> | string): (object: Object | Array<any>) => any;
@@ -2775,15 +2775,15 @@ declare module "lodash/fp" {
     ): ((
       path: Array<string> | string
     ) => (object: Object | Array<any>) => any) &
-      ((path: Array<string> | string, object: Object | Array<any>) => any);
+      ((path: Array<string> | string, object: Object | $ReadOnlyArray<any> | void | null) => any);
     getOr(
       defaultValue: any,
       path: Array<string> | string
-    ): (object: Object | Array<any>) => any;
+    ): (object: Object | $ReadOnlyArray<any> | void | null) => any;
     getOr(
       defaultValue: any,
       path: Array<string> | string,
-      object: Object | Array<any>
+      object: Object | $ReadOnlyArray<any> | void | null
     ): any;
     propOr(
       defaultValue: any
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/test_lodash-fp-v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/test_lodash-fp-v4.x.x.js
@@ -157,6 +157,12 @@ get('2', [{ a: 'foo' }, { b: 'bar' }, { c: 'baz' }]);
 get('3', [[1, 2], [3, 4], [5, 6], [7, 8]]);
 get('3')([[1, 2], [3, 4], [5, 6], [7, 8]]);
 
+// Nil - it is safe to perform on nil root values, just like nil values along the "get" path
+get('thing', null);
+get('data', undefined);
+getOr('default', 'stuff', null);
+getOr(12345, 'info', undefined);
+
 // First argument must be string when looking for array items by index
 // $ExpectError number This type is incompatible with union: ?array type | string
 get(0, [1, 2, 3]);
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/test_lodash-v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/test_lodash-v4.x.x.js
@@ -16,6 +16,7 @@ import first from "lodash/first";
 import flatMap from "lodash/flatMap";
 import forEach from "lodash/forEach";
 import get from "lodash/get";
+import getOr from 'lodash/fp/getOr';
 import groupBy from "lodash/groupBy";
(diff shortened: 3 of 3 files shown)


Mr-Wallet:
I still haven't gotten any of that stuff working properly locally, see #2043 ... I think the README may be outdated.

AndrewSouthpaw:
I'm not following? Can you explain more?

Mr-Wallet:
Right off the bat, Step 5 of contributing says to use a command that doesn't work; I can't just be in the repo and write `flow-typed` and expect anything to happen. But that's fine, I can go to the `cmd/sh` files and see what the full process is. I can't test _only_ the definitions I changed with that, but so be it.

I get through the `yarn test` in `definitions` with 527 warnings, which I assume is okay, So then I try to `yarn install` in `cli` and it blows up running tests because it can't find `git` even though it's on my PATH. so I edit `getGitPath` to return a hard-coded string, I hit a number of errors, including `ENOENT: no such file or directory, scandir 'C:\gitrepos\flow-typed\cli\.flow-bins-cache'`. So cool, I create that directory, but then there's still 4 errors I don't know how to deal with:

```
> jest && npm run lint

 FAIL  src\commands\__tests__\runTests-test.js
  ● run-tests (command) › regression 1385 › returns error code if $ExpectError test doesn't fail

    expect(received).toEqual(expected)

    Expected value to equal:
      1
    Received:
      0

      21 |
      22 |     it("returns error code if $ExpectError test doesn't fail", async () => {
    > 23 |       expect(status).toEqual(1);
      24 |     });
      25 |
      26 |     it('console logs about unused suppression', async () => {

      at _callee2$ (src/commands/__tests__/runTests-test.js:23:30)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:296:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at invoke (node_modules/regenerator-runtime/runtime.js:152:20)
      at node_modules/regenerator-runtime/runtime.js:195:11
      at callInvokeWithMethodAndArg (node_modules/regenerator-runtime/runtime.js:194:16)
      at AsyncIterator.enqueue (node_modules/regenerator-runtime/runtime.js:217:13)
      at AsyncIterator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at Object.<anonymous>.runtime.async (node_modules/regenerator-runtime/runtime.js:241:14)
      at Object._callee2 (src/commands/__tests__/runTests-test.js:22:111)

  ● run-tests (command) › regression 1385 › console logs about unused suppression

    expect(collection)[.not].toContainEqual(value)

    Expected collection to be an array-like structure.
    Received: undefined

      28 |       const calls = ((console.log: any): JestMockFn<*, *>).mock.calls;
      29 |       const lastErrorMsg = calls[calls.length - 1][1];
    > 30 |       expect(lastErrorMsg).toContain(expectedError);
      31 |     });
      32 |   });
      33 | });

      at _callee3$ (src/commands/__tests__/runTests-test.js:30:36)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:296:22)
      at Generator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:62:40)
      at invoke (node_modules/regenerator-runtime/runtime.js:152:20)
      at node_modules/regenerator-runtime/runtime.js:195:11
      at callInvokeWithMethodAndArg (node_modules/regenerator-runtime/runtime.js:194:16)
      at AsyncIterator.enqueue (node_modules/regenerator-runtime/runtime.js:217:13)
      at AsyncIterator.prototype.(anonymous function) [as next] (node_modules/regenerator-runtime/runtime.js:114:21)
      at Object.<anonymous>.runtime.async (node_modules/regenerator-runtime/runtime.js:241:14)
      at Object._callee3 (src/commands/__tests__/runTests-test.js:26:135)

 PASS  src\lib\__tests__\libDefs-test.js
 PASS  src\lib\__tests__\semver-test.js
 PASS  src\commands\__tests__\search-test.js
 PASS  src\lib\__tests__\cacheRepoUtils-test.js
 PASS  src\lib\__tests__\codeSign-test.js
 PASS  src\lib\__tests__\flowVersion-test.js
 PASS  src\commands\__tests__\install-test.js (6.235s)
 FAIL  src\lib\npm\__tests__\npmLibDefs-test.js (20.72s)
  ● Console

    console.log src\lib\cacheRepoUtils.js:131

  ● npmLibDefs › findNpmLibDef › when no cached libDefs found › returns null

    Timeout - Async callback was not invoked within the 10000ms timeout specified by jest.setTimeout.

      at node_modules/jest-jasmine2/build/queue_runner.js:72:21
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:633:19)

  ● npmLibDefs › findNpmLibDef › when non-semver package provided › doesn't throw error

    Timeout - Async callback was not invoked within the 10000ms timeout specified by jest.setTimeout.

      at node_modules/jest-jasmine2/build/queue_runner.js:72:21
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:633:19)

Test Suites: 2 failed, 7 passed, 9 total
Tests:       4 failed, 95 passed, 99 total
Snapshots:   0 total
Time:        22.454s
Ran all test suites.
```

I just now realized that I can dig into your travis logs from here and the current problem looks easy to fix, but the point is that I don't see how someone new to this project can properly contribute without using your Travis CI as their first and only method for running tests. I can only hope that there's only one problem on this PR and my next push will pass...

Mr-Wallet:
I followed your steps and still got all the aforementioned warnings and errors. Nevertheless I pressed on and `run_def_tests.cmd` seems to have worked despite all the "failures".

Unfortunately the tests are failing because of what seems to be a bug in Flow. (Sadly I can't seem to find any github issues at the moment, but I know there's been a few related ones because I saw them back when I was working in my own project on what would eventually become this PR). Specifically, Flow often refuses to resolve to the correct version of a variadic (in this case, curried) function. There's 3 definitions, for 1-ary, 2-ary, and 3-ary calls to the curried function:
```
    getOr(
      defaultValue: any
    ): ((
      path: Array<string> | string
    ) => (object: Object | Array<any>) => any) &
      ((path: Array<string> | string, object: Object | $ReadOnlyArray<any> | void | null) => any);
    getOr(
      defaultValue: any,
      path: Array<string> | string
    ): (object: Object | $ReadOnlyArray<any> | void | null) => any;
    getOr(
      defaultValue: any,
      path: Array<string> | string,
      object: Object | $ReadOnlyArray<any> | void | null
    ): any;
```
and tests (e.g. `getOr(undefined, 'info', 12345);` are failing because 
```
 - Either no more than 1 argument is expected by function type [1].
 - Or no more than 2 arguments are expected by function type [2].
```
i.e. it never considers the 3-ary definition right under the other two. In my experience with a real project that had this sort of issue, re-ordering the definitions can fix the issue for the 3-ary case but it will just cause issues for another case. This may not be reconcilable in the current version of Flow, unless I've made a mistake, or there's some clever hack I don't know about.

Mr-Wallet:
OK, I think this should be good to go!

AndrewSouthpaw:
lol

PR #2130, opened 2018-04-26
Fix flow errors in pg_v7 for flow 0.70.0

eldog:
Using flow 0.70.0 I got errors with pg_v7.x.x so made these changes.

Thanks :grinning:

### DIFF
--- a/definitions/npm/pg_v7.x.x/flow_v0.28.x-/pg_v7.x.x.js
+++ b/definitions/npm/pg_v7.x.x/flow_v0.28.x-/pg_v7.x.x.js
@@ -212,12 +212,12 @@ declare module pg {
    * ToDo: should find a better way.
   */
   declare class Query extends Promise<ResultSet> {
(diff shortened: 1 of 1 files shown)


definitions/npm/pg_v7.x.x/flow_v0.28.x-/pg_v7.x.x.js
```diff
@@ -212,12 +212,12 @@ declare module pg {
    * ToDo: should find a better way.
   */
   declare class Query extends Promise<ResultSet> {
-    then<U>( onFulfill?: (value: ResultSet) => Promise<U> | U,
-      onReject?: (error: PG_ERROR) => Promise<U> | U
+    then<U>( null | onFulfill?: (value: ResultSet) => Promise<U> | U,
```
GAntoine: `onFulfill?` means you can just call `then()`, without the need to pass that option in. The `?` in ` null | onFulfill?` isn't valid syntax.

stekycz:
Already fixed in #2296

PR #2133, opened 2018-04-26
redux-exr

paska27:
Flow definitions for [redux-exr](https://github.com/paska27/redux-exr) lib


GAntoine:
Is this a WIP or is it ready to be merged in?

paska27:
it is ready.

the definitions are okay, but flow doesn't seem to understand them well.
some tests should fail.
i'm going to fix tests later on.

PR #2140, opened 2018-04-27
update to latest from react-navigation/1.x

roballsopp:
This update should simply get this repo synchronized with the flow types defined in [`react-navigation`](https://github.com/react-navigation/react-navigation/blob/master/flow/react-navigation.js). Most of the changes are from this recently approved pr: https://github.com/react-navigation/react-navigation/pull/4085. This fixes the following:

- The typedef for [`getParam`](https://reactnavigation.org/docs/navigation-prop.html#getparam-get-a-specific-param-value-with-a-fallback) is missing.

- The return type for `withNavigation` is incorrect per the [flow documentation](https://flow.org/en/docs/react/hoc/#toc-injecting-props-with-a-higher-order-component) for injecting props with an HOC. The injected prop should be removed from the returned component's prop types. I'm not sure why this doesn't seem to cause an issue in the example app, but flow insisted I need a `navigation` prop in my own app until I fixed this.

### DIFF
--- a/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
+++ b/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
@@ -113,27 +113,27 @@ declare module 'react-navigation' {
   |};
 
   declare export type NavigationReplaceAction = {|
-    type: 'Navigation/REPLACE',
-    key: string,
-    routeName: string,
-    params?: NavigationParams,
-    action?: NavigationNavigateAction,
+    +type: 'Navigation/REPLACE',
+    +key: string,
+    +routeName: string,
+    +params?: NavigationParams,
+    +action?: NavigationNavigateAction,
   |};
   declare export type NavigationPopAction = {|
-    type: 'Navigation/POP',
-    n?: number,
-    immediate?: boolean,
+    +type: 'Navigation/POP',
+    +n?: number,
+    +immediate?: boolean,
   |};
   declare export type NavigationPopToTopAction = {|
-    type: 'Navigation/POP_TO_TOP',
-    immediate?: boolean,
+    +type: 'Navigation/POP_TO_TOP',
+    +immediate?: boolean,
   |};
   declare export type NavigationPushAction = {|
-    type: 'Navigation/PUSH',
-    routeName: string,
-    params?: NavigationParams,
-    action?: NavigationNavigateAction,
-    key?: string,
+    +type: 'Navigation/PUSH',
+    +routeName: string,
+    +params?: NavigationParams,
+    +action?: NavigationNavigateAction,
+    +key?: string,
   |};
 
   declare export type NavigationAction =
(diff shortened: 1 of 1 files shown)


roballsopp:
This pr is literally a copy paste from that repo to this one, so it includes all the changes made to `react-navigation` since this repo was last updated, not just the ones made in the linked pr. I can make a different one into this repo with only the changes I made, but I figured I would go ahead and get everything in one go.

roballsopp:
@Ashoat doh, that could be why there are extra changes. this is from `react-navigation/master`. I opened a pr into [`react-navigation/1.x`](https://github.com/react-navigation/react-navigation/pull/4108) with the same changes. I'm happy to close this pr and open a new one with the correct base when thats merged, or just leave this open and ill push the changes on the correct base here.

Ashoat:
@roballsopp this PR is needed as well, but it needs to be for

roballsopp:
I ultimately want to update this libdef since this is the one everyone will actually use, but I'm following the [contributing guide](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/react-navigation_v1.x.x/CONTRIBUTING.md) under the `react-navigation_v1.x.x` folder of this repo, and it looks like they want to keep a copy of the libdef up to date in both `react-navigation` and `flow-typed` repos, with the `react-navigation` one being the source of truth. So I went there first, and as soon as its passing and merged, I'll bring it over here.

Ashoat:
What I don’t understand is: what is there to update on the react-navigation side? Presumably you sourced the update from there, so what are you updating exactly? Are you trying to replace the 1.x libdef with the master one?

roballsopp:
The `react-navigation` repo and this repo are both incorrect. I made the correction only to `react-navigation/master`, and then (with this pr) mistakenly tried to merge it into the existing libdef in `flow-typed`, which actually corresponds to the one in the `react-navigation/1.x` branch. The same corrections are also necessary in `react-navigation/1.x`, so now I'm trying to get my original corrections into `react-navigation/1.x`, and then finally into the existing libdef in `flow-typed`.

roballsopp:
I actually don't see any actual releases for `react-navigation` beyond 1.5.2. Do we actually want `react-navigation/master` to be in this repo yet? Maybe we just want the `react-navigation/1.x` branch into `flow-typed/react-navigation_v1.x.x`, and thats it?

Ashoat:
`react-navigation@next` actually already has `2.0.0-rc.7`, and the official 2.0 release should happen quite soon.

roballsopp:
ok this pr is now correct. this is the `react-navigation/1.x` branch going into the `flow-typed/react-navigation_v1.x.x`. there are still a few more changes here than I made in my pr into `react-navigation`, but I'm sure its just because this repo was already missing an additional commit or two. let me know if anything else looks off.

roballsopp:
I will also look into making another pr to create a `flow-typed/react-navigation_v2.x.x`.  I do see a [v2.0.0-rc.1](https://github.com/react-navigation/react-navigation/tree/v2.0.0-rc.1) tag in `react-navigation`, should that be the base for `flow-typed/react-navigation_v2.x.x`? I don't see an `2.0.0-rc.7`.

PR #2141, opened 2018-04-28
Update type to variant in <Typography>

wcandillon:


GAntoine:
@bugzpodder Could I get a review on this change?

PR #2143, opened 2018-04-28
partial def for leaflet module

turadg:
Partial definition for http://leafletjs.com/reference-1.2.0.html

### DIFF
--- a/definitions/npm/leaflet_v1.2.0/flow_all/leaflet_v1.2.0.js
+++ b/definitions/npm/leaflet_v1.2.0/flow_all/leaflet_v1.2.0.js
@@ -1,6 +1,3 @@
-// flow-typed signature: 9eac0d0d9e2129cc3d92d8f4f8c072db
-// flow-typed version: <<STUB>>/leaflet_v1.2.0/flow_v0.69.0
-
 declare module 'leaflet' {
 
   // http://leafletjs.com/reference-1.2.0.html#featuregroup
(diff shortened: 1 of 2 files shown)


definitions/npm/leaflet_v1.2.0/flow_all/leaflet_v1.2.0.js
```diff
@@ -0,0 +1,155 @@
+// flow-typed signature: 9eac0d0d9e2129cc3d92d8f4f8c072db
+// flow-typed version: <<STUB>>/leaflet_v1.2.0/flow_v0.69.0
```
GAntoine: Remove these 2 lines.

turadg:
Looking to merge as partial. I covered what we needed and want to share that for people to build on.

Anything you'd like to see in the code as caveat? (btw, feel free to edit in place.)

TrySound:
@turadg Did you see this one?
https://github.com/flowtype/flow-typed/pull/1074

turadg:
Shoot, nope! I searched Google and for files in master.

Why not merge that one? Seems like it would be valuable even as is. At the least to unify efforts to improve leaflet typing. :)

Incidentally, it looks like https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/leaflet/index.d.ts is very complete for v1.2. I tried https://github.com/joarwilk/flowgen but it has some bugs. The output can be manually fixed, but then there's the dependency on `geojson` which also is in DefinitelyTyped but not flow-typed.

TrySound:
If you'd like to help with that PR, ask Michael. I'm sure he will agree to give an access to his fork.

RumyantsevMichael:
Yes, I will

PR #2146, opened 2018-04-30
[react-apollo] Fix Query component typing

wzrdzl:
If case of error, result object doesn't contain `data` property, please see the original TypeScript typings: https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L97

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -206,7 +206,7 @@ declare module 'react-apollo' {
   declare export function cleanupApolloState(apolloState: any): void;
 
   declare type QueryRenderPropFunction<TData, TVariables> = ({
-    data: TData,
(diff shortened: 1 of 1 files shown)


wzrdzl:
@budde377 could you please review? I noticed that you submitted the first version of the Query component typing

budde377:
Sure! I'll get on it right away

budde377:
So according to the [documentation](https://www.apollographql.com/docs/react/api/react-apollo.html#query-render-prop) and [this](https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L89-L96) they never use `undefined` but rather returns an empty object. This is consistent with what I've experienced, so I believe that the current implementation is correct. Have you experienced something else?

budde377:
@wzrdzl I see. Well that was super frustrating..... Anyways, I believe the implementation is the final documentation, so I approve of your PR. If I was in a position to do so ;)

wzrdzl:
Thanks @budde377 !
@GAntoine can we merge this?

GAntoine:
Can do, thanks for the PR, and for the quick review @budde377.

PR #2147, opened 2018-04-30
Delete second parameter in removeAllListeners in socket.io-client

ArStah:


ArStah:
@GAntoine excuse me, but where in the docs you see a callback??
image
And socket.io uses EventEmitter for events system, so even in [docs of EventEmitter](https://nodejs.org/api/events.html#events_emitter_removealllisteners_eventname) there is only one optional argument... don't see any reason to have a callback in such function

GAntoine:
My bad, I was reading the wrong line. Feel free to ping me time, if I don't get back to you after a few days!

PR #2153, opened 2018-05-02
[react-helmet] Fix wrong typing of toComponent on attributes

budde377:
This PR fixes a bug in the annotations to React Helmet, where the `body` and `html` attributes have the same types as e.g. `title` and `meta` components.

### DIFF
--- a/definitions/npm/react-helmet_v5.x.x/flow_v0.53.x-/react-helmet_v5.x.x.js
+++ b/definitions/npm/react-helmet_v5.x.x/flow_v0.53.x-/react-helmet_v5.x.x.js
@@ -27,10 +27,15 @@ declare module 'react-helmet' {
     toComponent(): [React$Element<*>] | React$Element<*> | Array<Object>;
   }
 
+  declare interface AttributeTagMethods {
+    toString(): string;
+    toComponent(): {[string]: *};
+  }
+
   declare interface StateOnServer {
     base: TagMethods;
-    bodyAttributes: TagMethods,
-    htmlAttributes: TagMethods;
+    bodyAttributes: AttributeTagMethods,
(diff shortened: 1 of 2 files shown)


budde377:
@dquang-tran Can you review, plz?

definitions/npm/react-helmet_v5.x.x/flow_v0.53.x-/react-helmet_v5.x.x.js
```diff
@@ -27,10 +27,15 @@ declare module 'react-helmet' {
     toComponent(): [React$Element<*>] | React$Element<*> | Array<Object>;
   }
 
+  declare interface AttributeTagMethods {
+    toString(): string;
+    toComponent(): {[string]: *};
```
dquang-tran: So toComponent function's return type is an object with one key? Can you link to the corresponding part of the react-helmet code where this return type is specified?

budde377: Since `react-helmet` isn't typed at all, linking to where the return type is specified is a little hard. But as you can see [here](https://github.com/nfl/react-helmet/blob/master/src/HelmetUtils.js#L598), the cases of attributes for body and html are treated differently from the other methods by converting the tags to react props, i.e. not an array of react components.

The
```
{[string]: *}
```
is not an object with one key, but rather a map from string to `*` (see https://flow.org/en/docs/types/objects/#toc-objects-as-maps).

dquang-tran: Ah yes, that's right. Your solution looks good to me. Thanks for catching that!

PR #2158, opened 2018-05-02
Support to use string template for chalk 

HsuTing:
I changed branch in my repo. So, I pull request again, sorry.

#2122

### DIFF
--- a/definitions/npm/chalk_v2.x.x/flow_v0.25.x-/chalk_v2.x.x.js
+++ b/definitions/npm/chalk_v2.x.x/flow_v0.25.x-/chalk_v2.x.x.js
@@ -1,90 +1,90 @@
 // From: https://github.com/chalk/chalk/blob/master/index.js.flow
 
-type TemplateStringsArray = $ReadOnlyArray<string>;
-
-export type Level = $Values<{
-  None: 0,
-  Basic: 1,
-  Ansi256: 2,
-  TrueColor: 3
-}>;
-
-export type ChalkOptions = {|
-  enabled?: boolean,
-  level?: Level
-|};
-
-export type ColorSupport = boolean;
-
-export interface Chalk {
-  (...text: string[]): string,
-  (text: TemplateStringsArray, ...placeholders: string[]): string,
-  constructor(options?: ChalkOptions): Chalk,
-  enabled: boolean,
-  level: Level,
-  rgb(r: number, g: number, b: number): Chalk,
-  hsl(h: number, s: number, l: number): Chalk,
-  hsv(h: number, s: number, v: number): Chalk,
-  hwb(h: number, w: number, b: number): Chalk,
-  bgHex(color: string): Chalk,
-  bgKeyword(color: string): Chalk,
-  bgRgb(r: number, g: number, b: number): Chalk,
-  bgHsl(h: number, s: number, l: number): Chalk,
-  bgHsv(h: number, s: number, v: number): Chalk,
-  bgHwb(h: number, w: number, b: number): Chalk,
-  hex(color: string): Chalk,
-  keyword(color: string): Chalk,
-
-  +reset: Chalk,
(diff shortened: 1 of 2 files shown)


definitions/npm/chalk_v2.x.x/flow_v0.25.x-/chalk_v2.x.x.js
```diff
@@ -1,105 +1,90 @@
...
-  bgCyanBright: $npm$chalk$Chain,
-  bgWhite: $npm$chalk$Chain,
-  bgWhiteBright: $npm$chalk$Chain
-};
+export type ColorSupport = boolean;
 
-declare module "chalk" {
-  declare var enabled: boolean;
-  declare var supportsColor: boolean;
+export interface Chalk {
```
AndrewSouthpaw: Is there a reason you're exporting it?

definitions/npm/chalk_v2.x.x/flow_v0.25.x-/chalk_v2.x.x.js
```diff
@@ -1,105 +1,90 @@
...
+  +bgBlueBright: Chalk,
+  +bgMagentaBright: Chalk,
+  +bgCyanBright: Chalk,
+  +bgWhiteBrigh: Chalk,
+
+  supportsColor: ColorSupport
+};
+
+declare module "chalk" {
+  declare module.exports: Chalk
```
AndrewSouthpaw: It seems to me like what you want is to have both named exports and a default export, in which case you could accomplish this with using ES6 exports.

definitions/npm/chalk_v2.x.x/flow_v0.25.x-/chalk_v2.x.x.js
HsuTing: No. I just copy the file from [chalk](https://github.com/chalk/chalk/blob/master/index.js.flow).
If you think this should not export, I will remove this.

definitions/npm/chalk_v2.x.x/flow_v0.25.x-/chalk_v2.x.x.js
HsuTing: Should change anything?

I try:
```js
const chalk = require('chalk')`;
import { bold } from 'chalk';
```

This still work. Maybe you can give me a simple example?
I can know how to modify.

saadq: In my opinion what you have is correct, I was just explaining the reason why you need to mix `module.exports` and `export`.

saadq: Actually scratch that. Right now, all your types and interfaces are outside of the `declare module`. This will make them global, which is not what we want. You will want to move all the types to inside the `declare module`. For example:

```ts
declare module 'chalk' {
  declare type TemplateStringsArray = $ReadOnlyArray<string>;

  declare export type Level = $Values<{
    None: 0,
    Basic: 1,
    Ansi256: 2,
    TrueColor: 3
  }>;

  ...

  declare module.exports: Chalk;
}
```

HsuTing: OK, I know.
I did not have enough experience on `flow`.
I will try to modify.

Thank you!

definitions/npm/chalk_v2.x.x/flow_v0.25.x-/chalk_v2.x.x.js
AndrewSouthpaw: Ah, that's fine.

AndrewSouthpaw:
LGTM. If something breaks we can revert or roll out a fix. :)

Issue #2156, opened 2018-05-02
[testcafe] typedef adds 80-120s to Flow server startup


awilkins:
See [this issue](https://github.com/facebook/flow/issues/5858) on flow.

If typedefs are causing very long load times it can negatively affect use of Flow in workstation and CI situations.

Not sure what constructs in this typedef are causing the slow load times, but it's tiny next to my gargantuan 77klines-and-3.3MB-of-code generated AWS-SDK typedef, which adds something like 2s to load time, so it's not size.

Corollaries to this issue (that I don't expect immediate answers to but are worth thinking about)

* Does flow-typed have any kind of test metric that rates load time?
* If it can be determined what causes it, would a linter that detects these constructs be practical?

AndrewSouthpaw:
Seems like something strictly related to Flow and not flow-typed. @GAntoine, thoughts? I'd rather close this and let Flow sort it out, since it's kinda out of our hands.

villesau:
@AndrewSouthpaw agreed.

> If it can be determined what causes it, would a linter that detects these constructs be practical?

This is an interesting point. I'd rather raise an issue in Flow than preventing to use some patterns though.

awilkins:
Making some progress on this [here](https://github.com/facebook/flow/issues/5856).

TLDR : seems like the problem is multi-extension of interfaces with certain types as members. builtin and Core types seem to be OK, the DOM types are something that seem to reliably trigger it.

Workaround may be to declare interfaces that are simply unions of multiple interfaces as ... unions of those interfaces, rather than extensions.

Currently testing that theory by patching the testcafe typedef.

PR #2163, opened 2018-05-02
[testcafe] Replace multi-interface inheritance with intersection types

awilkins:
This fixes/workarounds #2156 (very slow initialization of flow server)

Without these changes testcafe defs add > 80s to startup time.

Other possible victims of this flow bug could be

* puppeteer
* data.either
* data.maybe
* sequelize


GAntoine:
Closing due to age, and since the current def isn't technically incorrect.

PR #2165, opened 2018-05-03
React-Navigation: Fixed issue related to export magic

mwalkerwells:
Context: https://github.com/facebook/flow/issues/3119
Bug: Flow would not recognize the module definition & would exhibit this behavior: https://github.com/flowtype/flow-typed/issues/2164

### DIFF
--- a/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
+++ b/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js
@@ -1,1095 +1,1166 @@
 // @flow
 
-declare module 'react-navigation' {
+// ---------------------------------------
+// Private Types
+// ---------------------------------------
+
+/**
+ * First, a bunch of things we would love to import but instead must
+ * reconstruct (mostly copy-pasted).
+ */
+
+// This is a bastardization of the true StyleObj type located in
+// react-native/Libraries/StyleSheet/StyleSheetTypes. We unfortunately can't
+// import that here, and it's too lengthy (and consequently too brittle) to
+// copy-paste here either.
+type $npm$ReactNavigation$StyleObj =
+  | null
+  | void
+  | number
+  | false
+  | ''
+  | $ReadOnlyArray<$npm$ReactNavigation$StyleObj>
+  | { [name: string]: any };
+type $npm$ReactNavigation$ViewStyleProp = $npm$ReactNavigation$StyleObj;
+type $npm$ReactNavigation$TextStyleProp = $npm$ReactNavigation$StyleObj;
+type $npm$ReactNavigation$AnimatedViewStyleProp = $npm$ReactNavigation$StyleObj;
+type $npm$ReactNavigation$AnimatedTextStyleProp = $npm$ReactNavigation$StyleObj;
+
+// This is copied from the Layout type in
+// react-native-tab-view/src/TabViewTypeDefinitions
+type $npm$ReactNavigation$TabViewLayout = {
+  height: number,
+  width: number,
+};
+
+// This is copied from react-native/Libraries/Image/ImageSource.js
+type $npm$ReactNavigation$ImageURISource = {
+  uri?: string,
+  bundle?: string,
+  method?: string,
+  headers?: Object,
+  body?: string,
+  cache?: 'default' | 'reload' | 'force-cache' | 'only-if-cached',
+  width?: number,
+  height?: number,
+  scale?: number,
+};
+type $npm$ReactNavigation$ImageSource = $npm$ReactNavigation$ImageURISource | number | Array<$npm$ReactNavigation$ImageURISource>;
+
+// This one is too large to copy. Actual definition is in
+// react-native/Libraries/Animated/src/nodes/AnimatedValue.js
+type $npm$ReactNavigation$AnimatedValue = Object;
+
+type $npm$ReactNavigation$HeaderForceInset = {
+  horizontal?: string,
+  vertical?: string,
+  left?: string,
+  right?: string,
+  top?: string,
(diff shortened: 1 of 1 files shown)


mwalkerwells:
@hansonw Do you have any insight here & do you think it could be related to hyperclick?

AndrewSouthpaw:
I'll wait on approval from @Ashoat before merging.

mwalkerwells:
@Ashoat Just submitted https://github.com/react-navigation/react-navigation/pull/4234, which implements your feedback.

mwalkerwells:
@Ashoat This diff has the same interface as the previous flow-typed library definition & passes all the flow-typed tests.

Why does the corresponding file in ``react-navigation`` have a different interface? https://github.com/react-navigation/react-navigation/blob/master/flow/react-navigation.js

Ashoat:
Your `react-navigation` PR is on the `master` branch, which corresponds to the 2.0 version of `react-navigation`. However, this PR is for the 1.0 version. #2210 introduces a libdef for the 2.0 version. If you want to update the 1.x version, update the `1.x` branch on `react-navigation`.

Please make sure to clean up both PRs and leave minimal changes necessary for the introduction of a default export. As far as I understand, both of your PRs should just add an extra chunk at the end of the libdef, and should not change the existing contents.

mwalkerwells:
@Ashoat Ah, gotcha regarding the versioning.

This diff seems larger than it really is...in order to use ``module.exports``, a decent number of lines must be modified. Because of this, I went ahead and addressed the fact that many private type definitions are incorrectly being exported (every type with an underscore).

This diff follows the guidelines for private types outlined here: https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md#prefix-global-variables-that-arent-really-meant-to-be-global

Any diff that appropriately implements the correct export interface is going to touch a lot of lines (i.e. all previous export references need to be removed, function annotations need to be moved to arrow functions, semicolons need to be commas, etc.)—this passes all tests & makes react-navigation more type-safe.

Because there are so many inter-dependencies among types, I moved them all out of the module declaration & explicitly exported each. This makes it much easier to reason about if writing the type interface for a new API.

Ashoat:
Wait, so you're saying `declare type` is exported? What's the distinction between `declare type` and `declare export type`? Besides that, is there actually any concrete benefit in reducing the total number of exports?

Ashoat:
Also, instead of putting everything in a `declare module.exports`, can't you just do `declare export default`, like #2018 does?

This PR really should broken up into three. As it stands is nigh-on impossible to review, since it changes pretty much every single line.

Can you turn this PR into a simple `declare export default` line, and then put your other proposed changes in separate PRs?

mwalkerwells:
https://github.com/flowtype/flow-typed/blob/master/CONTRIBUTING.md#prefix-global-variables-that-arent-really-meant-to-be-global
> Right now we don't have a good way to write types inside declare module {} bodies that aren't exported. This problem is being worked on, but in the meantime the best option is to just put a 
declaration outside the declare module {} and reference it.

The concrete benefit to not exporting private types is that end users will not be surprised in the future when it inevitably changes.

I'd use ``declare export default``, but it breaks hyperclick in Nuclide.

This is merely re-arranging the furniture in the same room without any new items—breaking this up into three diffs will take more time than I can justify (although I get why'd you request that at first glance). No types were removed & the exiting tests pass.

If you look at the file (not in diff view), I think you'll find the new module export is clean & easy to read: https://github.com/mwalkerwells/flow-typed/blob/93b6e2b24937e1e6dbf119891721f284a51ce376/definitions/npm/react-navigation_v1.x.x/flow_v0.60.x-/react-navigation_v1.x.x.js#L932-L1165

mwalkerwells:
@Ashoat Would you be willing to reconsider accepting this PR? I addressed your feedback (explicitly specifying an export) & followed flow-type guidelines to prefix non-exported.

Regarding using ``declare export default``...without a special build configuration, the project main file does not export an object with the key ``default``, which would be required.

This library definition uses ``module.exports`` just like the source file, which I'd argue is better. While I'd love to break this up into multiple diffs, I just don't have the time—prefixing types & adding an explicit export makes it look like a larger refactor than it really is.

I feel confident these changes are sound on the basis of the existing tests passing—& core library functionality is not at risk here.

Ashoat:
@mwalkerwells:

1. The expectation in the open source community (as well as in industry, which I'm sure you're familiar with) is that each PR should represent a small, indivisible change. By saying you don't have the time to split your PR into three, what you're really saying is that your time is worth more than your reviewers' time. We're all volunteers here.
2. Regarding pulling the non-exported types up into the global namespace, the main downside here is that error messages will now talk about `$npm$ReactNavigation$NavigationState` instead of `NavigationState`. I think this outweighs the benefit of "end users will not be surprised in the future when it inevitably changes". I don't expect too many end users are using the types prefixed with underscores anyways.
3. Changing the libdef to use the old CommonJS style instead of ESM style doesn't make sense to me. It's true that `react-navigation` uses CommonJS exports for compatibility with various web configs, but there's no reason the libdef can't be typed in the ESM style. The fact that this breaks your Nuclide setup sounds like a good reason to send Nuclide a PR, not a good reason to send this libdef a PR.

If you update this PR to be the addition of a `declare export default` statement, I'll gladly accept it. But as it stands I don't think this PR represents an improvement to the libdef.

mwalkerwells:
@Ashoat Definitely not saying that my time is worth more than your time—but it is true, I do not have more time to invest in this—just trying to make a helpful contribution.

A lot of lines are touched here, and while I get your point regarding making small indivisible changes, some changes are inherently large & cannot be broken up (like making non-exported types private).

Wish you made your second point earlier, because it actually is pretty convincing. Your third point does not resonate with me though. For these reasons, I'm abandoning these changes—thanks for your time.

PR #2166, opened 2018-05-03
[react-apollo] Improve usability of Query and Mutation components

budde377:
By exporting and using type arguments it becomes
very easy to use these components with e.g. apollo-codegen

So given a mutation with vars of type `MVars` and result of type `MData` or a query with data `QData` or vars `QVars`, we can annotate like this:

```jsx
<Mutation>
  {(update: MutationFunction<MVars>, result: MutationResult<MData>) => { ... }}
</Mutation>
<Query>
  {(props: QueryRenderProps <QData, QVars>) => { ... }}
</Query>

```

Needs tests!

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/test_react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/test_react-apollo_v2.x.x.js
@@ -1,9 +1,14 @@
 // @flow
 import * as React from 'react';
-import { it } from 'flow-typed-test';
+import { it, describe } from 'flow-typed-test';
 import {
+  Query,
+  Mutation,
   graphql,
   withApollo,
+  type MutationFunction,
+  type MutationResult,
+  type QueryRenderProps,
   type OperationComponent,
   type GraphqlQueryControls,
   type ChildProps,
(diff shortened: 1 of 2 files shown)


jonespen:
Nice! One nitpick (and probably not your fault, since it was like this before), but `QueryRenderProps.data` should always be there, but is nullable: https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L97 

So, something like 
```
data: ?(TData | {||})
```

instead? 

Also, `error` should be `error?: ApolloError`. If you want this in another PR, shout out ;)

budde377:
@jonespen, I agree! And the changes you request are actually already part of this PR, with the exception of `error`. But I'll add that now.

budde377:
@jonespen Do you mind doing a quick review to see if I got all the changes you requested?

jonespen:
Looks good! Btw, I'm not a maintainer or anything, so I guess I'm not able to do a proper review, but using this definition in my project now (together with types generated from apollo-codegen), works like a charm 👍

definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/test_react-apollo_v2.x.x.js
```diff
@@ -210,3 +214,49 @@ it('works with withApollo HOC', () => {
   (Manual: React.ComponentType<{}>);
 })
 
+describe('Query', () => {
+  it('works', () => {
```
AndrewSouthpaw: 👏👏👏

PR #2167, opened 2018-05-03
add Subscription component to react-apollo defs

mlc:
PR #2086 added types for react-apollo's `<Query>` and `<Mutation>` components, but not `<Subscription>`. This PR adds the `<Subscription>` component, and fixes a small typo in the definition of `<Query>`

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -236,12 +236,27 @@ declare module 'react-apollo' {
     pollInterval?: number,
     notifyOnNetworkStatusChange?: boolean,
     fetchPolicy?: FetchPolicy,
-    errorPolicty?: ErrorPolicy,
+    errorPolicy?: ErrorPolicy,
     ssr?: boolean,
     displayName?: string,
     delay?: boolean,
     context?: {[string]: any}
   }> {}
+
+  declare type SubscriptionResult<TData> = {
+    loading: boolean,
+    data?: TData,
+    error?: ApolloError,
+  }
+
+  declare type SubscriptionProps<TData> = {
+    subscription: DocumentNode,
+    variables?: { [string]: any },
+    shouldResubscribe?: boolean | (SubscriptionProps<TData>, SubscriptionProps<TData>) => boolean,
+    children: (result: SubscriptionResult<TData>) => React$Node,
+  }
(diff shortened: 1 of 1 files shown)


definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
```diff
@@ -234,13 +234,28 @@ declare module 'react-apollo' {
...
+  }
+
+  declare type SubscriptionProps<TData> = {
+    subscription: DocumentNode,
+    variables?: { [string]: any },
+    shouldResubscribe?: boolean | (SubscriptionProps<TData>, SubscriptionProps<TData>) => boolean,
+    children: (result: SubscriptionResult<TData>) => React$Node,
+  }
+
+  declare export class Subscription<TData> extends React$Component<SubscriptionProps<TData>> {}
```
AndrewSouthpaw: I'm pretty sure you don't need to export it explicitly (in the same way you don't have to `declare export type`), can you verify?

definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
```diff
@@ -236,13 +236,28 @@ declare module 'react-apollo' {
...
     fetchPolicy?: FetchPolicy,
-    errorPolicty?: ErrorPolicy,
+    errorPolicy?: ErrorPolicy,
     ssr?: boolean,
     displayName?: string,
     delay?: boolean,
     context?: {[string]: any}
   }> {}
 
+  declare type SubscriptionResult<TData> = {
```
budde377: I think this should be exported as well for higher usability.

definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
budde377: This is consistent with the way `Query` and `Mutation` are exported. And if I remove `export` from them I get the error like:

```
Cannot import Query because there is no Query export in react-apollo.
```

AndrewSouthpaw: Got it.

mlc:
sorry I was out and missed the conversation, but thanks @AndrewSouthpaw and @budde377 for getting this merged!

PR #2169, opened 2018-05-04
Make throttle-debounce function accept arguments

dmtrKovalenko:
Add generic type definition for ...args, that can be passed to throttle or debounce function callback


definitions/npm/throttle-debounce_v1.x.x/flow_v0.25.x-/throttle-debounce_v1.x.x.js
```diff
@@ -1,3 +1,6 @@
+// flow-typed signature: 2fd21db3d7d3c371b608c10f78bdc0a3
+// flow-typed version: da30fe6876/throttle-debounce_v1.x.x/flow_>=v0.25.x
```
GAntoine: Remove these lines at the top.

dmtrKovalenko:
@GAntoine done.

PR #2170, opened 2018-05-04
Add libdef knex 0.14

Jarlakxen:


GAntoine:
Could you point out any major changes from the previous version so I can focus a review on those?

facundo-viale-olx:
It's the same definition but I only added `  declare type Knex = Knex$Knex<any>;`

Jarlakxen:
It's the same definition but I only added `declare type Knex = Knex$Knex<any>;`

PR #2171, opened 2018-05-05
Fix problems with Formik types

TLadd:
* Correct FormikHandlers to match v0.11.x, rather than the 1.0.0 alpha
* Add missing validateForm to FormikActions
* Add some tests and use describe/it

The types for FormikHandlers looks to be based of the master Typescript typings, which is v1 alpha. Even so, they will need some corrections when we add the v1 flow-typed types. Changed it to match what's actually in v0.11.x (https://github.com/jaredpalmer/formik/blob/v0.11.11/src/formik.tsx)

### DIFF
--- a/definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
+++ b/definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
@@ -9,33 +9,58 @@ declare module "formik" {
     [field: $Keys<Values>]: boolean
   };
 
-  declare export type FormikActions<Values> = {
+  declare export interface FormikActions<Values> {
     /** Manually set top level status. */
-    setStatus: (status?: any) => void,
+    setStatus(status?: any): void;
     /**
      * Manually set top level error
      * @deprecated since 0.8.0
      */
-    setError: (e: any) => void,
+    setError(e: any): void;
     /** Manually set errors object */
-    setErrors: (errors: FormikErrors<Values>) => void,
+    setErrors(errors: FormikErrors<Values>): void;
     /** Manually set isSubmitting */
-    setSubmitting: (isSubmitting: boolean) => void,
+    setSubmitting(isSubmitting: boolean): void;
     /** Manually set touched object */
-    setTouched: (touched: FormikTouched<Values>) => void,
+    setTouched(touched: FormikTouched<Values>): void;
     /** Manually set values object  */
-    setValues: (values: Values) => void,
+    setValues(values: Values): void;
     /** Set value of form field directly */
-    setFieldValue: (field: $Keys<Values>, value: any, shouldValidate?: boolean) => void,
+    setFieldValue(
+      field: $Keys<Values>,
+      value: any,
+      shouldValidate?: boolean
+    ): void;
+    setFieldValue(field: string, value: any, shouldValidate?: boolean): void;
     /** Set error message of a form field directly */
-    setFieldError: (field: $Keys<Values>, message: string) => void,
+    setFieldError(field: $Keys<Values>, message: string): void;
+    setFieldError(field: string, message: string): void;
     /** Set whether field has been touched directly */
-    setFieldTouched: (field: $Keys<Values>, isTouched?: boolean, shouldValidate?: boolean) => void,
+    setFieldTouched(
+      field: $Keys<Values>,
+      isTouched?: boolean,
+      shouldValidate?: boolean
+    ): void;
+    setFieldTouched(
+      field: string,
+      isTouched?: boolean,
+      shouldValidate?: boolean
+    ): void;
+    /** Validate form values */
+    validateForm(values?: any): void;
     /** Reset form */
-    resetForm: (nextValues?: any) => void,
+    resetForm(nextValues?: any): void;
     /** Submit the form imperatively */
-    submitForm: () => void
-  };
+    submitForm(): void;
+    /** Set Formik state, careful! */
+    setFormikState(
+      f: (
+        prevState: $ReadOnly<FormikState<Values>>,
+        props: any
+      ) => $Shape<FormikState<Values>>,
+      callback?: () => any
+    ): void;
+  }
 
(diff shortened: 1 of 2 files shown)


definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
```diff
@@ -184,7 +199,7 @@ declare module "formik" {
       /** Value of the input */
       value: any,
       /* name of the input */
-      name?: string,
+      name?: string
```
cliedeman: Hello, I found another issue that we could maybe bundle into this PR?

https://github.com/jaredpalmer/formik/blob/v0.11.11/src/Field.tsx#L42 - According to the typescript definiation name is not optional.

Ciaran

TLadd: Sure, added the change in.

definitions/npm/formik_v0.11.x/test_formik.js
```diff
@@ -1,38 +1,120 @@
...
-  Field,
-  type FieldProps,
-  type FormikActions,
-  type FormikConfig
+import { Formik, Form, Field } from "formik";
+import type {
+  FieldProps,
+  FormikErrors,
+  FormikTouched,
+  FormikActions,
```
AndrewSouthpaw: Not used?

definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
```diff
@@ -9,33 +9,58 @@ declare module "formik" {
     [field: $Keys<Values>]: boolean
   };
 
-  declare export type FormikActions<Values> = {
+  declare export interface FormikActions<Values> {
```
AndrewSouthpaw: What motivated the switch to an interface? I'm not even seeing `FormikActions` being used?

definitions/npm/formik_v0.11.x/test_formik.js
TLadd: Updated to add tests for FormikActions. You end up using FormikActions some in userland when using Formik, so it makes sense to test these.

definitions/npm/formik_v0.11.x/flow_v0.53.x-/formik_v0.11.x.js
TLadd: Like the Typescript definition, there are some overloaded method signatures in FormikActions. As far as I know, flow supports that for interfaces, but doesn't for object types (https://github.com/facebook/flow/issues/1556#issuecomment-200051475)

AndrewSouthpaw: Makes sense; can you document that justification? Otherwise there might be some churn.

TLadd: Sure, added a comment explaining the use of interface

sibelius: I prefer type instead of interfaces

TLadd: I would use type if they supported overloaded method signatures. As it is, interface seems like the right choice.

stevehollaar:
I noticed something strange after including this update - it seems to slow down Flow initialization on a relatively small project from a few seconds to over a minute! (on both flow-bin: `0.71.0` and `0.72.0`). 
Before:
screen shot 2018-05-10 at 6 35 23 am
After updating this definition:
screen shot 2018-05-10 at 6 37 47 am

I wonder if this is related to https://github.com/flowtype/flow-typed/pull/2163.
@AndrewSouthpaw @TLadd @sibelius any chance someone can reproduce this performance regression?

TLadd:
Now that you mention it, the tests for this did take a lot longer to run than was typical

sibelius:
I think flow does not handle interfaces well

TLadd:
Well, I tried changing interface to type, and the tests still run slow, so I don't think that's it anyways. Using the previous version, the tests are fast. Will need to spend some time isolating what causes the slowdown.

sibelius:
can you provide a repository so we can help you figure it out, open an issue with this repo

TLadd:
I'm just running flow-typed run-tests formik_v0.11.x in the flow-typed repo.

TLadd:
Looks to be the use of SyntheticEvent in FormikHandlers that causes the slowdown.

If I change FormikHandlers to not include `SyntheticEvent<HTMLFormElement>` and `SyntheticEvent<any>` for handleSubmit and handleChange, respectively, the tests go back to running fast.
```
  declare export type FormikHandlers = {
    /** Form submit handler */
    handleSubmit: (e: any) => any,
    /** Classic React change handler, keyed by input name */
    handleChange: (e: any) => any,
    /** Classic React blur handler */
    handleBlur: (e: any) => void,
    /** Reset form event handler  */
    handleReset: () => void
  };
```

Not sure why this would be the case.

cliedeman:
A wild guess would because [EventTarget](https://github.com/facebook/flow/blob/v0.70.0/lib/dom.js#L154) used by [SyntheticEvent](https://github.com/facebook/flow/blob/v0.70.0/lib/react-dom.js#L113) is just a massive type.

sibelius:
we need an issue about it in flow repo

TLadd:
Redux form's libdef uses SyntheticEvent as well, and I don't see the same slowdown when running its tests. https://github.com/flowtype/flow-typed/blob/master/definitions/npm/redux-form_v5.x.x/flow_v0.53.x-/redux-form_v5.x.x.js#L16

TLadd:
Some issues that are probably related:
https://github.com/facebook/flow/issues/5856
https://github.com/facebook/flow/issues/5858 (this one actually does relate to interfaces)

PR #2172, opened 2018-05-05
Apollo Client Libdef (also add to react-apollo)

TLadd:
- [x] Add some more tests
- [x] Resolve interop with the current react-apollo libdef. 

It seems that since the react-apollo libdef is importing types from apollo-client (which doesn't have types), even if I include this apollo-client libdef in my project, types that react-apollo imports end up getting clobbered to any. Either need to 

1. stop importing those types in react-apollo libdef or 
2. actually use the types added here.

Actually using the types is definitely the endgame, but might be a large effort that could be split to another PR.

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -1,73 +1,762 @@
-// @flow
-import type {
-  ApolloClient,
-  DataProxy,
-  MutationQueryReducersMap,
-  ApolloQueryResult,
-  ApolloError,
-  FetchPolicy as FetchPolicyAC,
-  FetchMoreOptions as FetchMoreOptionsAC,
-  UpdateQueryOptions,
-  FetchMoreQueryOptions as FetchMoreQueryOptionsAC,
-  SubscribeToMoreOptions as SubscribeToMoreOptionsAC,
-  MutationUpdaterFn,
-} from 'apollo-client';
-import type { DocumentNode, VariableDefinitionNode } from 'graphql';
+declare module "react-apollo" {
+  /**
+   * Copied types from Apollo Client libdef
+   * Please update apollo-client libdef as well if updating these types
+   */
+  declare class ObservableQuery<T> extends Observable<ApolloQueryResult<T>> {
+    options: WatchQueryOptions;
+    queryId: string;
+    variables: { [key: string]: any };
+    isCurrentlyPolling: boolean;
+    shouldSubscribe: boolean;
+    isTornDown: boolean;
+    scheduler: QueryScheduler<any>;
+    queryManager: QueryManager<any>;
+    observers: Observer<ApolloQueryResult<T>>[];
+    subscriptionHandles: SubscriptionLINK[];
+    lastResult: ApolloQueryResult<T>;
+    lastError: ApolloError;
+    lastVariables: { [key: string]: any };
 
-declare module 'react-apollo' {
+    constructor(data: {
+      scheduler: QueryScheduler<any>,
+      options: WatchQueryOptions,
+      shouldSubscribe?: boolean
+    }): this;
 
-  declare export type NetworkStatus = 1 | 2 | 3 | 4 | 6 | 7 | 8;
+    result(): Promise<ApolloQueryResult<T>>;
+    currentResult(): ApolloCurrentResult<T>;
+    getLastResult(): ApolloQueryResult<T>;
+    getLastError(): ApolloError;
+    resetLastResults(): void;
+    refetch(variables?: any): Promise<ApolloQueryResult<T>>;
+    fetchMore(
+      fetchMoreOptions: FetchMoreQueryOptions<any> & FetchMoreOptions<any, any>
+    ): Promise<ApolloQueryResult<T>>;
+    subscribeToMore(options: SubscribeToMoreOptions<any, any>): () => void;
+    setOptions(
+      opts: ModifiableWatchQueryOptions
+    ): Promise<ApolloQueryResult<T>>;
+    setVariables(
+      variables: any,
+      tryFetch?: boolean,
+      fetchResults?: boolean
+    ): Promise<ApolloQueryResult<T>>;
+    updateQuery(
+      mapFn: (previousQueryResult: any, options: UpdateQueryOptions) => any
+    ): void;
+    stopPolling(): void;
+    startPolling(pollInterval: number): void;
+  }
 
-  declare export interface ProviderProps {
-    client: ApolloClient;
+  declare class QueryManager<TStore> {
+    scheduler: QueryScheduler<TStore>;
+    link: ApolloLink;
+    mutationStore: MutationStore;
+    queryStore: QueryStore;
+    dataStore: DataStore<TStore>;
+
+    constructor({
+      link: ApolloLink,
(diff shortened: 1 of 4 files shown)


GAntoine:
Are the checklist items TODOs, or are they done?

TLadd:
They are TODOs. Sorry for the confusion.

budde377:
@TLadd let me know when you're ready for a review and I'll gladly abide. Regarding the imports; considering the poor support for inter libdef dependencies (ref #16) I'm really starting to think that the imports should be removed and the types duplicated. I'm planning on removing all the imports from react-apollo and clean up duplicate declarations during the next week.

TLadd:
Yeah, from https://github.com/flowtype/flow-typed/issues/1857#issuecomment-371312890, it appears that even if the imports worked, the tests wouldn't. Just duplicating the types react-apollo needs is probably the path of least resistance at the moment. I think as long as we distinguish clearly in the file where the types are actually coming from (this section is actually react-apollo, this section is types from apollo-client, etc) maintenance won't be that bad.

Sorry this has been stagnating a bit; I'll have some time over the weekend to make a push and should get it to a mergeable state.

budde377:
Yes, that sounds like a good plan!

TLadd:
@budde377 @GAntoine think this is ready for a look. I copied the types from apollo-client into react-apollo as well. Sorry for the diff size.

I'm not sure how fun keeping the types in apollo-client and react-apollo in sync will be. I'm hoping it's not too bad, since it is just a matter of copy/pasting them over. If it gets to be too much of a pain, I'm not opposed to abandoning apollo-client and just maintain everything in react-apollo. At least in my projects, the only place I'm using apollo-client directly is to initialize the client (hence the tests checking ApolloClient's initialization). So not a huge deal, but I think they're worth having around if it's easy.

Added some tests focused around the Query and Mutation props I'm using.

Also some stylistic changes in the PR caused by prettier, just using the default config. I'm fine shifting the style some if there are complaints, but would prefer it be prettier-able.

Pulled both these libdefs into the project I'm using react-apollo in and seems to be behaving correctly.

Definitely still room for improvement in places, but hopefully a good step.

budde377:
Nice! I can prob. do a review monday afternoon (CET) :)

definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
```diff
@@ -1,74 +1,763 @@
...
+    data?: T,
+    extensions?: { [string]: any },
+    errors?: any[]
+  };
+  declare type GraphQLError = any;
+  declare type VariableDefinitionNode = any;
+  /** End From graphql */
+
+  declare export interface ApolloProviderProps<TCache> {
+    client: any; // ApolloClient<TCache>;
```
budde377: Any or apollo client?

TLadd: The trouble with using ApolloClient is that it's a class and it's nominally typed. So even though ApolloClient is the same in both the apollo-client and react-apollo libdefs, they are different types. We could maybe make it work with an interface in react-apollo, but didn't think it was worth it to diverge from what is defined in apollo-client.

budde377: I see... Let's leave it as is for now.

TLadd:
@GAntoine Looks like this is ready for you whenever you get the time. Thanks!

jedwards1211:
Example of the problem ([Try flow link](https://flow.org/try/#0C4TwDgpgBAIghsOUC8UDeAzA9lgXFAOwFcBbAIwgCcBfKAH3TruoCgWMiCBjYASywJQA5hGAAxHAAoAJgjj54iAJT4A-MXJV0LKFEqiilQbMQA6bFhbUgA)):

```js
@flow
type Data = {foo: number} | {||}

function getFoo(data: Data): ?number {
  return data.foo
}
```

```
4:   return data.foo
                 ^ Cannot get `data.foo` because property `foo` is missing in object type [1].
References:
3: function getFoo(data: Data): ?number {
                         ^ [1]
```

TLadd:
I believe this has always been the case (you still get an error if you select older versions of flow as well).

I'm typically checking that values are present when I'm accessing them, and it removes the type errors. This works for instance:
```
type Data = {foo: number} | {||}

function getFoo(data: Data): ?number {
  if (data.foo) {
  	return data.foo 
  }
  return null
}
```

Having said that, it may be reasonable to change data to just be `TData | void`. The `| {||}`, while correct, doesn't buy much.

jedwards1211:
@TLadd actually I read the react-apollo source, it never passes an undefined value.

I didn't realize flow has this obscure exception for checking if an object is exact empty or not.  Thanks for pointing that out.

TLadd:
@jedwards1211 There is actually still one case where undefined gets passed in: https://github.com/apollographql/react-apollo/issues/1977

Typescript actually types data as `TData | undefined` (https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L97) because they couldn't figure out a way to get the equivalent of {||}. It's possible that we should follow suit and just do `TData | void`; might make it a little more user friendly, since that error you were showing in that example can be a bit confusing.

jedwards1211:
Ah right.  Argh...

ganemone:
> @TLadd using `data: TData | {||} | void` no longer works with recent versions of Flow; it errors on any property you try to access on `data`, because no properties are present in `{||}`, and recent versions of Flow no longer allow any kind of untyped property access, even in existence checks. (Without casting to `any` or `Object` at least)
> 
> Are you sure Apollo can pass an empty object?
> 
> If so would you be okay with us changing this to `?$Shape<TData>`?

I think we can actually fix this using `$ObjMap` ([try flow link](https://flow.org/try/#0C4TwDgpgBAsghgawgeTMAlgewHZwDZQC8UAPAGoB8AFGQJREVQD8ZAUFB1K6JFAIoBXCACcQAJQjYAJiIAKwzGADOJACoAROMDiNiAb1YBIKVrgAuKADdM6KVAA+UACTIARgCt4YNZu0AaWEQUNCxcPApWAF92Tm5waFUAC3RsAHMlIigDDkNzLJjOKFcLJWBhFNSCqGjo1gAzAWwAYwwcKGAIUqowBWULQRFxSRlheUUVJIqlCnpsqHQ6qG7epQA6E21ZqoWlnvH101W4LcKOAHozqGxMKBEFYSUqjiacUqg80vK0zL3lA+0jqtXFVooULlApJhOrdhPcqi9sG9ilBPhUfit-nBAa4ANwxUFQcHXGH3R7PV7Ad4lMpo4i-JRMTGMuCM1wORwAcg5eMiQA))

I will submit a PR.

jedwards1211:
@ganemone I don't get it, the way you're using `$ObjMap` does the same thing as `$Shape` right?

jedwards1211:
@TLadd I don't think you were the one to introduce this, but can `MutationFunction` really resolve to `undefined`?  The return type is `Promise<void | FetchResult<...>>` which adds another pesky null check I really wish I could avoid.

TLadd:
It's the same in the typescript defs: https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L62

Not sure; would have to look through the source to see how/if this can happen.

jedwards1211:
No worries, just wanted to check if you knew off the top of your head, thanks!

PR #2173, opened 2018-05-05
upgrade run-tests command

XaveScor:
I replace 'github' package to '@octokit/rest' because 'github' is outdated.

Rewrited flow bins receive mechanism. In my version run-tests don't need to save zip archives. Script writes to disk only binary files.

Download functions moved to separate library.


AndrewSouthpaw:
cc @GAntoine 

@XaveScor is there a reason you closed this pr?

XaveScor:
I used the very old version of master. I recreate changes from actual version and create PR.

PR #2174, opened 2018-05-06
Export koa-bodyparser as a CommonJS module

KevinGrandon:
This matches the codebase here better: https://github.com/koajs/bodyparser/blob/master/index.js#L28

### DIFF
--- a/definitions/npm/koa-bodyparser_v4.x.x/flow_v0.56.x-/koa-bodyparser_v4.x.x.js
+++ b/definitions/npm/koa-bodyparser_v4.x.x/flow_v0.56.x-/koa-bodyparser_v4.x.x.js
@@ -21,5 +21,5 @@ declare module "koa-bodyparser" {
     onerror?: (err: Error, ctx: Context) => void
   |};
 
-  declare export default function bodyParser(opts?: Options): Middleware;
+  declare module.exports: (opts?: Options) => Middleware;
 }
--- a/definitions/npm/koa-bodyparser_v4.x.x/flow_v0.56.x-/test_koa-bodyparser_v4.x.x.js
+++ b/definitions/npm/koa-bodyparser_v4.x.x/flow_v0.56.x-/test_koa-bodyparser_v4.x.x.js
@@ -14,3 +14,7 @@ bodyParser({ strict: true });
 // $ExpectError (Must use valid types for options)
 bodyParser({ formLimit: 56 });
 bodyParser({ formLimit: "56kb" });
+
+// Should work when requiring koa-bodyparser
+var requiredBodyParser = require('koa-bodyparser');
+requiredBodyParser();


saadq:
Hey, I'm the one who did the initial contribution for `koa-bodyparser`, as well as `koa-send`, `koa-static`, and `koa-views`. Would you be interested in fixing the exports for the other three modules I mentioned as well?

As a side note, there should probably a mention in the `CONTRIBUTING.md` that modules meant to be used with Node should be CJS exports...

KevinGrandon:
Thanks for looking. I’d be happy to help update other libraries, but since this is my first PR here, I’d like to get this landed first and go from there.

AndrewSouthpaw:
> As a side note, there should probably a mention in the CONTRIBUTING.md that modules meant to be used with Node should be CJS exports...

Sounds like a great PR. 😉

KevinGrandon:
So folks using flow-typed can not use require() at all? It seems like the typedef should mimic what the library exports. If they export a CJS module, shouldn't we define the types as CJS compatible?

saadq:
Just my opinion, but I agree with @KevinGrandon that the exports should follow whatever the original module uses. If ES6 exports are used in the type definition, it enforces users to use `import` syntax. If a module is typed as CJS, both `require` and `import` will work.

I don't think client-side stuff is a problem to always use `import`, but for server-side Node stuff a lot of people just use `require` and don't bother transpiling `import` -> `require` so they will have issues. An example of this problem: https://github.com/chalk/chalk/issues/264

GAntoine:
> If a module is typed as CJS, both require and import will work.

Is that the case? From what I can tell, the keyword `import` isn't supported in CJS.

saadq:
Sorry, let me clarify. I meant that if a `flow-typed` libdef uses `module.exports`, the end-user can use either `require` syntax or `import` syntax. This is because `babel` will convert the CJS `module.exports` to an `export default` assuming you're using a plugin like `env`.

But yeah, it's true that you can do `require('modulename').default` to get a `flow-typed` libdef that uses ESM to work.

AndrewSouthpaw:
Wouldn't it be nice if we just had one way to put JS into modules? 😂 

Okay, that makes sense. We'll have the module type reflect the library. @saadq would you mind updating the contributing guide?

GAntoine:
They can require it, they'll just need to use the `requite('').default` syntax.

AndrewSouthpaw:
> But this is bundle system specific and would not work with a normal node module for typical required code.

Interesting. I don't work much with Node. Can you expand on what bundle system allows the `require('').default` syntax?

KevinGrandon:
> Interesting. I don't work much with Node. Can you expand on what bundle system allows the require('').default syntax?

There are various transpilers which provide the `.default` property, but this is not a native node property. Webpack, babel, typescript can all generate this depending on config, but not all bundlers do.

AndrewSouthpaw:
Got it. That seems like a compelling enough argument to follow the module pattern of the library. @GAntoine thoughts?

AndrewSouthpaw:
Just need sign-off from @GAntoine.

Can you also add/update documentation to the `CONTRIBUTING.md` in the correct place to discuss this decision?

GAntoine:
Yeah I'm not thrilled about this change, but considering the arguments above, this PR is the least painful of both approaches. I'll go ahead and get it merged, and any updates to the docs can be made in another PR.

PR #2178, opened 2018-05-07
Add react-transition-group definitions

el1t:
Add flow definitions for [react-transition-group](https://github.com/reactjs/react-transition-group)

Works fine, though according to the API some props could be a little more strict


AndrewSouthpaw:
Great work! Congrats on your first contribution, and thanks for following all the guidelines! 💪

PR #2179, opened 2018-05-07
Add stringify-object declaration

pascalduez:
See https://github.com/yeoman/stringify-object

### DIFF
--- a/definitions/npm/stringify-object_v3.x.x/flow_v0.50.x-/stringify-object_v3.x.x.js
+++ b/definitions/npm/stringify-object_v3.x.x/flow_v0.50.x-/stringify-object_v3.x.x.js
@@ -1,0 +1,19 @@
+declare module 'stringify-object' {
+  // Matches the `prop` parameter `mixed` type.
+  // Prevent errors when using computed properties.
+  declare type AnyObject = { [key: mixed]: mixed };
+
+  declare type Options = {|
+    indent?: string,
+    singleQuotes?: boolean,
+    filter?: (obj: AnyObject, prop: mixed) => boolean,
+    transform?: (obj: AnyObject, prop: mixed, originalResult: string) => string,
+    inlineCharacterLimit?: number,
+  |};
+
+  declare module.exports: (
+    input: AnyObject | Array<*>,
+    options?: Options,
+    pad?: string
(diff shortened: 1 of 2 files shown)


definitions/npm/stringify-object_v3.x.x/test_stringify-object_3.x.x.js
```diff
@@ -0,0 +1,111 @@
+// @flow
```
AndrewSouthpaw: Thanks for writing tests; please organize them with `describe` and `it` blocks.

definitions/npm/stringify-object_v3.x.x/flow_v0.50.x-/stringify-object_v3.x.x.js
```diff
@@ -0,0 +1,17 @@
+declare module 'stringify-object' {
+  declare type AnyObject = { [key: mixed]: mixed };
```
AndrewSouthpaw: Any reason to use the indexer property here? I think `{}` would suffice.

pascalduez: That's what I got first but the `obj[prop]` gives error.

```
Cannot access computed property using mixed [1].

      test_stringify-object_3.x.x.js:64:37
      64|   filter: (obj, prop) => typeof obj[prop] === 'string',
                                              ^^^^

   References:
      stringify-object_v3.x.x.js:7:30
       7|     filter?: (obj: {}, prop: mixed) => boolean,
                                       ^^^^^ [1]
```

Happy to get insights on this.

definitions/npm/stringify-object_v3.x.x/test_stringify-object_3.x.x.js
pascalduez: Done.

definitions/npm/stringify-object_v3.x.x/flow_v0.50.x-/stringify-object_v3.x.x.js
pascalduez: Done.

PR #2182, opened 2018-05-07
fix(sequelize): asupport beforeUpsert and afterUpsert hooks

jedwards1211:


jedwards1211:
If you look at it, it's a transient test failure, not my changes

PR #2184, opened 2018-05-08
[rethinkdb] Start libdef on rethink db

budde377:
Even though the future of rethinkdb might be dubious, this libdef really improves the usability of their excellent client lib.


definitions/npm/rethinkdb_v2.x.x/flow_v68.x-/test_rethinkdb_v2.x.x.js
```diff
@@ -0,0 +1 @@
+// TODO write some tests!
```
AndrewSouthpaw: so... do it. ;-) Write at least a few as a smoke screen, and remember to use describe/it blocks.

budde377:
@AndrewSouthpaw It has been done!

PR #2185, opened 2018-05-08
RxJS: Allow filtering maybe types with Boolean

matthewwithanm:
This uses the same logic as Array.

### DIFF
--- a/definitions/npm/rxjs_v5.0.x/flow_v0.34.x-/rxjs_v5.0.x.js
+++ b/definitions/npm/rxjs_v5.0.x/flow_v0.34.x-/rxjs_v5.0.x.js
@@ -438,6 +438,7 @@ declare class rxjs$Observable<+T> {
     scheduler?: rxjs$SchedulerClass
   ): rxjs$Observable<T>;
 
+  filter(predicate: typeof Boolean): rxjs$Observable<$NonMaybeType<T>>;
   filter(
     predicate: (value: T, index: number) => boolean,
     thisArg?: any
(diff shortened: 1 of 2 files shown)


definitions/npm/rxjs_v5.0.x/flow_v0.34.x-/rxjs_v5.0.x.js
```diff
@@ -438,6 +438,7 @@ declare class rxjs$Observable<+T> {
     scheduler?: rxjs$SchedulerClass
   ): rxjs$Observable<T>;
 
+  filter(predicate: typeof Boolean): rxjs$Observable<$NonMaybeType<T>>;
```
GAntoine: Why `typeof Boolean`? You can just use `boolean`.

matthewwithanm: This is for using `.filter(Boolean)` to convert `Observable<?T>` to `Observable<T>`, in the same way that Flow does for arrays.

GAntoine: I get that, I'm saying `typeof Boolean` should be changed to `boolean`. `typeof Boolean` isn't used in any other definition.

matthewwithanm: Sorry guess I wasn't clear. 😞 It is used in another definition, the lib type for `Array.filter()`. Using `boolean` wouldn't be the same as that would enable code like `obs.filter(true)` (which isn't valid) whereas this enables `obs.filter(Boolean)`, where `Boolean` isn't a type but rather a value: the JS function that converts a value to a boolean.

GAntoine: Ah gotcha!

PR #2187, opened 2018-05-08
Remove deprecated static property from axios

rgoldfinger:
Flow 62 removed support for static properties. The use of a static proprty in this typedef causes flow to error immediately. This fixes it by removing `static`.

screen shot 2018-05-08 at 1 34 06 pm


GAntoine:
You'll need to update the specs: `Cannot call `axios.CancelToken.source` because property `source` is missing in statics of `CancelToken` [1]`.

GAntoine:
Closing due to age.

PR #2188, opened 2018-05-08
Fixes for uuid: add v3, predefined namespaces

kumandroo:
First PR here, sorry if I do something wrong.

Flow is flagging valid uses like the following:
```
const uuidv5 = require('uuid/v5');

// ... using predefined DNS namespace (for domain names)
uuidv5('hello.example.com', uuidv5.DNS); // ⇨ 'fdda765f-fc57-5604-a269-52a7df8164ec'

// ... using predefined URL namespace (for, well, URLs)
uuidv5('http://example.com/hello', uuidv5.URL); // ⇨ '3bbcee75-cecc-5b56-8031-b6641c1ed1f1'
```

I added defs for DNS and URL as well as defs for v3.


GAntoine:
Looks great!

PR #2189, opened 2018-05-08
[react-apollo] Move default type variable

budde377:
By moving the type variable default definition the component becomes more usable.

Now we do not have to annotate the type of the variables, explicitly, in the child function, if they are not required.


budde377:
Plz rerun test

PR #2194, opened 2018-05-09
[react-apollo] Add defaults to type-variables on Subscription and Query

budde377:
Add type variable defaults to the `Subscription` and `Query` component props.


AndrewSouthpaw:
Thanks for more PRs @budde377. In the future, please CC a few contributors since I won't have much context around apollo.

mlc:
Thanks, @budde377! I can confirm that this fixes the problem I mentioned [here](https://github.com/flowtype/flow-typed/pull/2180#issuecomment-387868219).

AndrewSouthpaw:
I love all the work y'all are doing on apollo. @TLadd thanks for being on top of reviews, it's much better than us doing them since we know less about the library. 😄

PR #2207, opened 2018-05-14
fix ramda/memoizeWith type signature

nikoskalogridis:


AndrewSouthpaw:
You can also run tests with `./run_def_tests.sh ramda_v0.x.x`

nikoskalogridis:
@AndrewSouthpaw thanks for your comments. Actually I noticed that the tests in the master branch are already broken. The `startsWith`, `endsWith` tests have not been changed by that PR.

The reason for this PR is that the previous declarations using `...rest: Array<void>` are incorrectly matching and flow is confused. Removing them seems to handle all cases correctly.

definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
```diff
@@ -1845,14 +1845,13 @@ declare module ramda {
...
-  declare function memoizeWith<A, B>(
-    keyFn: (x: A) => string,
-    ...rest: Array<void>
-  ): (fn: (x: B) => B) => A;
-  declare function memoizeWith<A, B>(
-    keyFn: (x: A) => string,
-    fn: (x: B) => B
-  ): A;
+  declare function memoizeWith<A, B, C>(
+    keyFn: (...args: Array<A>) => C
```
AndrewSouthpaw: keyFn doesn't take an array of args AFAICT, can you point to docs otherwise?

definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
```diff
@@ -1845,14 +1845,13 @@ declare module ramda {
...
+  ): (...args: Array<A>) => (...args: Array<A>) => B;
+  declare function memoizeWith<A, B, C>(
+    keyFn: (...args: Array<A>) => C,
+    fn: (...args: Array<A>) => B
```
AndrewSouthpaw: does `keyFn` take the input to `fn` or the output from `fn`?

nikoskalogridis: It's takes the input parameters of the function when called and should return a unique key

definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
nikoskalogridis: It does not. It takes a number of arguments. Notice the ... rest operator

LoganBarnett:
@nikoskalogridis Yeah - the rest args thing that's sprinkled all throughout the Ramda defs is (I think) the original author's way of working around an earlier version of Flow that wasn't as compatible with JavaScript's ability to just leave args out of a function call. That, or they were mixing with things that provided more args that were strictly necessary when mixing with things like `Array.map` and `Array.forEach`. Either way, it's not something we need any longer, and we should really just make a pass to eliminate them in another pull request.

Re: argument types here, I think this is a good place to try capturing a generic type and then passing it through. Ramda's example of `memoizeWith` doesn't illustrate its functionality well. I cooked up another one [here](https://ramdajs.com/repl/?v=0.25.0#?let%20count%20%3D%200%0Aconst%20mult%20%3D%20R.memoizeWith%28%28a%2C%20b%29%20%3D%3E%20b%2C%20%28a%2C%20b%29%20%3D%3E%20%7B%0A%20%20%2B%2Bcount%0A%20%20return%20a%20%2A%20b%0A%7D%29%0A%0Amult%283%2C%205%29%20%2F%2F%2015%0Amult%283%2C%204%29%20%2F%2F%2012%0Amult%281%2C%205%29%20%2F%2F%2015%3F%21%20intentionally%20incorrect%0Acount%20%2F%2F%202%20%28skipped%20one%20with%20mult%281%2C%205%29%29). Maybe we can submit this as something to use for the original Ramda docs too.

Notably, the function signature of `keyFn` here is the same arg list used by `fn`. [This example](https://ramdajs.com/repl/?v=0.25.0#?let%20count%20%3D%200%0Aconst%20mult%20%3D%20R.memoizeWith%28%28a%2C%20b%29%20%3D%3E%20b%20%2B%20%27foo%27%2C%20%28a%2C%20b%29%20%3D%3E%20%7B%0A%20%20%2B%2Bcount%0A%20%20return%20a%20%2A%20b%0A%7D%29%0A%0Amult%283%2C%205%29%20%2F%2F%2015%0Amult%283%2C%204%29%20%2F%2F%2012%0Amult%281%2C%205%29%20%2F%2F%2015%3F%21%20intentionally%20incorrect%0Acount%20%2F%2F%202%20%28skipped%20one%20with%20mult%281%2C%205%29%29) shows that return types of `keyFn` and `fn` may differ. This will be much harder to capture as a type param. I tried seeing if the current use of rest args would strictly enforce that, but this [try](https://flow.org/try/#0FAEwpgxgNghgTmABAMwK4DsIBcCWB7dFPPAHgEEAaRAISoGEA+ACmEUQHp3EBJAZ0TKIIMAA5ZUCEIhj8YiDPkJ5k0uAHN+ACzAIA-K0QEwALkRMAdJZimycODACe5BgEpEAXgY0KBrAHc8UwsrGztHZzdPRDofF1MANzwcEGBgCAJeLGkAFTxqDzNrRHRUAFsAIx0qctNMuBx0NSoIU3LiKDAYdEivcrSMrJhcugKmIpKKqsQaxDqGpqFW9s7ujy8IYGRiMdzaHLw6F1TOHhVLcyLbeycyddFxSWkoKEMVLG1VDVmseuwoByo7xw-F4mjwqCgUmQMBwUHM-XQmX2dAA6nACGpRkUAGIYbCKaqmCaVOA9aabbZDPJUKmo9GNFxAA) doesn't show the errors I expect to see. We do have the ability on recent versions of Flow to extract a return type, but I don't think we have the inverse that extracts the parameters.

My initial thought is we go make another generator for things like this like we did for [partial](https://github.com/LoganBarnett/typedef-gen/blob/master/index.org). I'm not sure how many times making this choice is a good one though. We've run into situations where the observed behavior might be different than the intended behavior (or what we think is the intended behavior). Oftentimes this is due to defensive coding in an untyped JS world. In Flow we can make some assumptions. It's also perilous to build type defs off of undocumented behavior. At some point I'd like to start pulling in some of the Ramda maintainers and see what they think. I can be the one to start the conversation about this.

All that said, I don't want to hold up @nikoskalogridis and this change to the libdef leaves it better than it was before. I can't think of a quick way to make those args individually typed across both functions. So this PR LGTM.

LoganBarnett:
@nikoskalogridis #2247 (when merged) fixes the test issues with `startsWith` and `endsWith`. I'll have another PR shortly that handles the `...rest: Array<void>` stuff.

nikoskalogridis:
@AndrewSouthpaw  rebased with latest and tests pass. Sorry for the delay :-)

PR #2218, opened 2018-05-15
Move material-ui into @material-ui/core

mastertinner:


gabriel:
Looks like this needs another update, to conform to imports that are now like `import InputLabel from @material-ui/core/InputLabel`. 

I have a version copy that I am updating as I need it (but I have only fixed a few components, like CircularProgress, FormControl, and InputLabel):
https://gist.github.com/gabriel/4238a433cb7eb5645091a8ea1055ff3f

mastertinner:
Thanks, @gabriel! I wrote a couple of reviews there.

gabriel:
Oh nm, there are a few pull-requests already out for this, see https://github.com/flowtype/flow-typed

PR #2225, opened 2018-05-16
[react-transition-group] Replace an interface with an object type

maxmalov:
Looks like flow doesn't work well with interface and types intersections


maxmalov:
Still, a WIP, while I'm reproducing the error in tests from my project case

AndrewSouthpaw:
Go ahead and ping me when it's no longer a WIP.

maxmalov:
@AndrewSouthpaw added a test case to reproduce an issue

AndrewSouthpaw:
Looks good, thanks!

PR #2229, opened 2018-05-16
Minor fix for formik's `handleSubmit` function

EvHaus:
The `handleSubmit` function in formik isn't always going to be attached to `<form onSubmit/>` element. It might be attached to a button (ie. `<button onClick={handleSubmit} />` or some custom component or whatever else. So we can't assume it's always going to be an `HTMLFormElement` in the event object.

@TLadd Any objections to this change? I noticed you made some large changes to this libdef last.


TLadd:
It looks like the type in Formik (https://github.com/jaredpalmer/formik/blob/master/src/Formik.tsx#L269) is `(e: SyntheticEvent<any> | undefined)`. Would `(e: SyntheticEvent<any> | void) => any` work for your case?

PR #2231, opened 2018-05-16
exposed class ApolloError

dcgudeman:
I need access to the `ApolloError` class in order to properly type my code.

```javascript
// @flow
import React, { Component } from 'react';
import type { Node } from 'react';
import type { ApolloError } from 'react-apollo';

type Props = {
  children: Node,
  loading: boolean,
  error?: ApolloError,
};
```

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -377,7 +377,7 @@ declare module "react-apollo" {
     [queryName: string]: MutationQueryReducer<T>
   };
 
-  declare class ApolloError extends Error {
+  declare export class ApolloError extends Error {
     message: string;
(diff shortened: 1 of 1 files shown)


AndrewSouthpaw:
... by "apollo maintainers" I mean people who have been working on PRs to this libdef.

dcgudeman:
@thymikee @budde377

dcgudeman:
Great. @AndrewSouthpaw is that sufficient?

TLadd:
Since ApolloError is defined as a class export, one issue with this is that it would allow for some unsafe code to pass. For instance,

```
import { ApolloError } from 'react-apollo';

const error = new ApolloError({ errorMessage: 'error' })
```

We really would probably ideally want to expose ApolloError as an interface to be completely correct (since ApolloError is actually defined and exported out of apollo-client). I'm not 100% against this if it proves too cumbersome to type correctly, but would like to try and maintain correctness.

AndrewSouthpaw:
@TLadd It sounds like that can be a separate PR. I'm going to go ahead and merge it in, if there are significant objections we can do a revert.

PR #2234, opened 2018-05-18
[knex] Add some missing methods to knex definitions

tnayuki:


tnayuki:
@GAntoine Thanks, I did.

PR #2239, opened 2018-05-20
[apollo-client] Add extract and restore methods on ApolloClient

budde377:
Add `extract` and `restore` methods

See https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/ApolloClient.ts#L442

@TLadd


definitions/npm/apollo-client_v2.x.x/flow_v0.57.x-/apollo-client_v2.x.x.js
```diff
@@ -455,6 +455,8 @@ declare module "apollo-client" {
     reFetchObservableQueries(
       includeStandby?: boolean
     ): Promise<ApolloQueryResult<any>[]> | Promise<null>;
+    extract(optimistic?: boolean): TCacheShape;
```
TLadd: Could you add these two to react-apollo as well?

PR #2241, opened 2018-05-21
Add missing method definitions

alethes:


GAntoine:
@alethes Could you add some tests for those new methods?

GAntoine:
Closed as already present.

PR #2246, opened 2018-05-22
feat(formik): add Props generic to add missing props on FormikActions

sibelius:
we can access component Props on handleSubmit thorough FormikActions


GAntoine:
Seems one of the specs is failing, and expecting 2 arguments instead of just 1.

sibelius:
what is the right way to run only `formik` tests?

it is not easy to find this info on the docs

GAntoine:
`flow-typed run-tests formik`

sibelius:
any way to run only against latest flow version?

JCMais:
I've incorporated those changes on the other PR I've opened https://github.com/flowtype/flow-typed/pull/2262

sibelius:
@GAntoine check this instead https://github.com/flowtype/flow-typed/pull/2262

tks

PR #2256, opened 2018-05-24
added css as an export and wrote test for checking it

shrynx:
Fixes.
https://github.com/flowtype/flow-typed/pull/2190#issuecomment-391615925


Gpx:
I tried in local and it works for me

PR #2260, opened 2018-05-24
Ramda: Remove superfluous rest args

LoganBarnett:
From an earlier pass of Ramda's libdefs, many functions have a `...rest: Array<void>` at the end of the argument list. This may have been added as a means of accounting for composition with functions like `Array.forEach` which has "extra" arguments like the index and original array being operated on. It could have also been that earlier versions of Flow didn't play nice with spare arguments.

The original motivation isn't that important here. As a community of Ramda libdef consumers, we've run into frequent issues where Flow can't disambiguate which function signature to use because the `...rest: Array<void>` portion causes a conflict. Sometimes it means that we wind up with runtime errors because Flow assumed some bogus parameters we passed were good simply because they satisfied the `...rest: Array<void>` catch-all. The solution to these problems has been to simply remove the rest args from functions we run into as we encounter problems.

This PR simply removes all of the `...rest: Array<void>` params. The tests all passed locally with the only change being to move an `$ExpectError` down a line.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
@@ -316,7 +316,6 @@ declare module ramda {
     de: UnaryFn<D, E>,
     ef: UnaryFn<E, F>,
     fg: UnaryFn<F, G>,
-    ...rest: Array<void>
   ) => UnaryFn<A, G>) &
     (<A, B, C, D, E, F>(
       ab: UnaryFn<A, B>,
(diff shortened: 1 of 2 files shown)


definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
```diff
@@ -316,35 +316,30 @@ declare module ramda {
     de: UnaryFn<D, E>,
     ef: UnaryFn<E, F>,
     fg: UnaryFn<F, G>,
-    ...rest: Array<void>
-  ) => UnaryFn<A, G>) &
+      ) => UnaryFn<A, G>) &
```
AndrewSouthpaw: Looks like all the indentations here are off... 😬

LoganBarnett: I don't believe I indented them, although it's possible. I noticed there's a pre-commit linter application across anything that's changed. I'll try making the adjustment and see if it's me or the linter.

LoganBarnett: It was me!

I think.

AndrewSouthpaw:
Sorry, looks like there are merge conflicts now @LoganBarnett :)

LoganBarnett:
Fixed!

PR #2265, opened 2018-05-25
Update of ioredis definitions 

tchiss:
Hi,
This is an update of ioredis definitions. I removed some any types and added the real type inspired from the existing redis definitions.


definitions/npm/ioredis_v3.x.x/flow_v0.46.x-/ioredis_v3.x.x.js
```diff
@@ -12,6 +12,9 @@
     const redis = new Redis();
  =============================================== */
 
+import {callback} from "../../awaiting_v2.x.x/flow_v0.33.x-/awaiting_v2.x.x";
```
AndrewSouthpaw: You can't import other type definitions, see the contributing guide.

tchiss: fixed :-)

PR #2280, opened 2018-05-28
Update react-navigation_v1.x.x from the lastest of react-navigation repo

vpodolyan:
This PR updates type definitions for React Navigation 1.x.x


AndrewSouthpaw:
Build needs to be fixed. You can run tests locally too to reduce the feedback loop time.

vpodolyan:
@AndrewSouthpaw fixed

PR #2286, opened 2018-05-29
Fix function declaration used with `$el.data("key")`

PikachuEXE:
According to https://github.com/flowtype/flow-typed/issues/2160#issuecomment-386303024
The `data(key: string, _: void): any;` is more specific than `data(key: string, value: any): JQuery`
But flow is using `data(key: string, value: any): JQuery` anyway since it's on top

This commit reverse their positions to allow `data(key: string, _: void): any;` being used for expression like `$el.data("key")`


PikachuEXE:
@AaronAsAChimp Please see #2299

PR #2288, opened 2018-05-29
Add correct typings for CustomPicker HoC

harjis:
I used [this](https://github.com/facebook/flow/issues/4759#issuecomment-327713650) as a guidance in writing the Hoc type annotations.

[This](https://github.com/casesandberg/react-color/blob/befb042d425e89d79c24c6611aa5bff9f2ab06a9/src/helpers/color.js#L30-L37) is what is injected by `CustomPicker`

### DIFF
--- a/definitions/npm/react-color_v2.x.x/flow_v0.56.x-/react-color_v2.x.x.js
+++ b/definitions/npm/react-color_v2.x.x/flow_v0.56.x-/react-color_v2.x.x.js
@@ -172,12 +172,14 @@ declare module "react-color/lib/components/common" {
     HexColor,
     RGBColor,
     HSLColor,
+    HSVColor,
     ColorChangeHandler
   } from "react-color";
 
   declare type PartialColorResult = {|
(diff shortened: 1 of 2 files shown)


definitions/npm/react-color_v2.x.x/flow_v0.56.x-/react-color_v2.x.x.js
```diff
@@ -131,16 +139,13 @@ declare module "react-color" {
   ) => void;
 
   declare export type InjectedColorProps = {
-    hex?: string,
-    hsl?: HSLColor,
-    rgb?: RGBColor,
-    onChange?: ColorWrapChangeHandler
-  };
-
-  declare export type ExportedColorProps = {
```
harjis: Removing `ExportedColorProps` is a breaking change since someone might have imported it and used it in their application. Since I removed the need for it I decided to remove the type annotation too. If this is seen as not good change I can add it back.

PR #2291, opened 2018-05-29
fix(sequelize): fix association create method types

jedwards1211:
Typing an ORM is hard 😕 
The association create method types incorrectly relied on the type of the attributes once created rather than the type of the attributes required to create a model.

There is still a problem with the `belongsToMany` methods: if you try to pass additional options for the `through` table, it complains that the primary key attributes are missing, because it doesn't have a way to omit them from the through table's initial attributes type.  Fixing this would require a breaking change to the flow types...which is an example of why typedefs should be released in their own packages with their own versioning instead of tracking the version of the package here.

### DIFF
--- a/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/sequelize_v4.x.x.js
+++ b/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/sequelize_v4.x.x.js
@@ -101,13 +101,13 @@ declare module "sequelize" {
    * @see  http://docs.sequelizejs.com/en/latest/api/associations/belongs-to/
    * @see  Model
   */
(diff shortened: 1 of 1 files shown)


jedwards1211:
thanks!

jedwards1211:
Hi @GAntoine,
There are some fixes I need to make that require breaking changes to the parameters of some types.
Is there any way to do that without causing people who are using the current types to get the new ones automatically?  It would be a nasty surprise.

GAntoine:
They won't get the new ones unless they run `update`, or haven't checked in their definitions.

jedwards1211:
@GAntoine that's the thing though, I think people will run `update` to get improvements to any typedefs for any other lib they're using, and accidentally pull in the new typedefs for sequelize...

PR #2292, opened 2018-05-29
Add opentok 2.6.2 defs

pronebird:
ps: I define types inside of `declare module`, although the guidelines state to move them outside and prefix. If that's still the case let me know and I'll prefix types but I think that works very well with recent Flow.


pronebird:
@GAntoine alright, let me take a look.

pronebird:
Updated my PR. I am not entirely sure what to test in such simple annotations (e.g. no generics etc..), so I've added quite a lot of tests. If you feel like it's over the top, please advise and I'll refine tests.

pronebird:
🕺

PR #2298, opened 2018-05-30
lodash: isNaN should accept any like other isX functions

lukeapage:
I changed it to be like the other is functions.. I'm guessing it was missed at some point?
Otherwise at least boolean and Class<any> are missing


lukeapage:
@GAntoine isNative looks ok to me - it specifies a number of types that return false and then for `any` it returns a boolean

PR #2306, opened 2018-06-01
[react-router-dom] Prop types should be exact

wcandillon:


oriSomething:
Hey, why did update `<NavLink>` and `<Link>` to exact type?

You can clearly see [here](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/NavLink.js#L20) and [here](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/NavLink.js#L46) and in [`<Link>`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L58) that it's not exact and the props are chained to [`<a>`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L75).

PR #2308, opened 2018-06-01
Firebase v5 definitions

meafmira:


wcandillon:
@meafmira Thanks for contributing this. Out of curiosity, was this type definition build manually or was it generated from somewhere?

I did notice that `currentUser` should be:
-  currentUser: $npm$firebase$auth$User | null;

wcandillon:
Also onSnapshot should allow async callbacks (that return a promise). 
=> void
should become
=> void | Promise<void>

This is true in a lot of places where a callback is allowed as a parameter.

meafmira:
@wcandillon It's based on firebase_v4 but with some fixes. For example changed 'firebase/app' exports. I'll fix callbacks

wcandillon:
yes thanks for fixing the /app export :)

meafmira:
@wcandillon are you sure about `onSnapshot` signature? I checked this https://firebase.google.com/docs/reference/js/firebase.firestore.Query#onSnapshot and see it should return unsubscribe function not a Promise

meafmira:
@wcandillon Updated `onSnapshot` and `currentUser` signatures

wcandillon:
Yes because the user might use `await` in his callback.

Thanks a lot for the changes, you are the man 🙌🏻

meafmira:
@wcandillon Oh I didn't understand what do you mean at first.
I've updated onSnapshot callbacks signatures and fixed all tests

wcandillon:
@meafmira Thank you for your work. I just reviewed and tested this type definition in my project and it works for me 👍🏻

GAntoine:
LGTM based on @wcandillon approval 👍

PR #2311, opened 2018-06-01
Create firebase_functions_v1.x.x.js

wcandillon:


wcandillon:
@GAntoine I understand that you don't support dependencies at the moment but it would still be incredibly useful to the community to have these type definitions. Maybe you can run the tests against all type definitions or add them without test? I understand that it's less than ideal but these should still be useful to the community.

GAntoine:
Closing in favour of a new PR.

PR #2312, opened 2018-06-01
Add opentracing v0.14.x definitions

shuhei:
Adding definitions and tests for `opentracing` v0.14.x. The definitions are based on `.d.ts` files from the package. v0.14.x has been the latest version since [23 Apr 2017](https://github.com/opentracing/opentracing-javascript/commit/eb768cb00c7d52b2b980947733c832e14096ee80).

The source code of the package is at https://github.com/opentracing/opentracing-javascript. I didn't include `private` and `protected` class fields because they are not meant to be used by package consumers.


shuhei:
Thanks for reviewing and merging :-)

PR #2313, opened 2018-06-02
lodash: isString should accept any like other isX methods

lukeapage:
The specific thing missing for me was $ReadOnlyArray but this seems more consistent and correct.


lukeapage:
@GAntoine do you know why this and the other one like it are stalled? I did some research and it isn't possible to make these better - tbh until flow has support we may as well have
```
isX(any): boolean;
```
https://github.com/facebook/flow/issues/34#issuecomment-394291134

GAntoine:
@villesau What's your opinion on this change?

villesau:
LGTM. Back in the days I just didn't want to touch these when refining others as I wasn't sure if they have some purpose.

GAntoine:
You heard the man, approved!

PR #2323, opened 2018-06-04
Create icons_v1.x.x

wcandillon:


wcandillon:
@GAntoine Could you help me with this failing test? I need to resolve another type `@material-ui/core`. I'm surprised because in a previous type definition that refered to "express", the module was resolved properly.

GAntoine:
> I'm surprised because in a previous type definition that referred to "express"

I actually went ahead and changed that definition after merging, when I noticed the reference. I just didn't catch it during the review, and the flow version range was locked at `flow_v0.53.x`, instead of `flow_v0.53.x-`, which caused none of the tests to run, and therefore reported no failures.

At the moment, you can reference one library from another, though work is being done to improve that. You'll need to remove the reference and replace with custom objects, or `any`/`mixed` types.

wcandillon:
Thanks for looking into it. Same comment as for #2311

GAntoine:
Closing as v4 for for icons exists

PR #2325, opened 2018-06-04
Add metrics v0.1.x definitions

shuhei:
Adding definitions and tests for [metrics](https://github.com/mikejihbe/metrics) v0.1.x. The definitions are based on the package's `index.d.ts` file and manually adjusted.


definitions/npm/metrics_v0.1.x/flow_v0.25.x-/metrics_v0.1.x.js
```diff
@@ -0,0 +1,204 @@
+import events from 'events';
```
GAntoine: FYI, this will import `events` as `any`.

PR #2331, opened 2018-06-06
reselect: improve return type of createStructuredSelector

NilSet:


jedwards1211:
Sigh, this caused issues for me back when the selectors were callable objects, now this change is causing issues for me again like this:

`Cannot call "foo" because callable signature is missing in object type [1]`

NilSet:
Can you give me a small repro?

jedwards1211:
I think it's actually just a bad error message coming from flow, because if I cast one of my arguments in the selector call to `any`, the error goes away.

PR #2336, opened 2018-06-06
fix(sequelize): add missing fields to Model class

jedwards1211:


jedwards1211:
https://github.com/sequelize/sequelize/blob/master/lib/model.js#L41

PR #2340, opened 2018-06-08
Fix type signature of react-router-dom

kogai:
Fixed lack of several annotations of react-router-dom which for using exact type.


kogai:
I think #2342 would be more reasonable than this, therefore 'll close it.

PR #2342, opened 2018-06-08
relax react-router-dom Link constraints

caoilte:
Small PR to relax over-strict type checking on router-react-dom.

react-router-dom `<Link>` accepts any user properties and passes them through to the generated `<a>`. Also `<NavLink>` extends `<Link>`. 

for reference 
https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md#others

code 
https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L75
https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/NavLink.js#L46

### DIFF
--- a/definitions/npm/react-router-dom_v4.x.x/flow_v0.63.x-/react-router-dom_v4.x.x.js
+++ b/definitions/npm/react-router-dom_v4.x.x/flow_v0.63.x-/react-router-dom_v4.x.x.js
@@ -14,14 +14,14 @@ declare module "react-router-dom" {
     children?: React$Node
   |}> {}
 
-  declare export class Link extends React$Component<{|
+  declare export class Link extends React$Component<{
     className?: string,
     to: string | LocationShape,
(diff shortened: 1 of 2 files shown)


caoilte:
Hi @wcandillon and @GAntoine. This is a tiny PR to correct a change you introduced yesterday to react-router-dom that was a little bit too strict. I've linked to the docs so you can see how.

Thanks for your awesome work getting this library updated.

GAntoine:
LGTM, waiting on @wcandillon for approval.

thabemmz:
Would be good to also update the tests to confirm this does not go wrong again in the future. See https://github.com/flowtype/flow-typed/pull/2343 for an example.

caoilte:
Thanks for the suggestion @thabemmz. I'll do that now.

PR #2343, opened 2018-06-08
[react-router-dom] Make <Link> component of ReactRouterDOM not exact + update tests

thabemmz:
## Context
In https://github.com/flowtype/flow-typed/pull/2306, the `<Link>` component has been made exact. However, in the official React Router DOM documentation (https://reacttraining.com/react-router/web/api/Link/others), it is explicitly stated that the `<Link>` component accepts all attributes for the `<a>` element in HTML.

Instead of defining all `<a>` properties explicitly in the flow-type, I chose to make the `<Link>` component not exact. I do this to make the `<Link>` flow-type independent of the HTML-spec for  `<a>`, since some of the HTML-specced properties are dynamic (`data-` attributes for example). I think this is the most pragmatic way.

## What has been done?
* Make `Link` property definition not exact
* Updated tests for the `Link` component to verify the `<a>` attributes can be passed


GAntoine:
I believe this is already being addressed in #2342.

PR #2349, opened 2018-06-09
[react-router]/[react-router-dom] add missing props to Redirect and Route components

theres:
Some of properties of `Redirect` and `Route` components are missing in definitions. That was for longer time, but last change (#2306) that makes `react-router-dom` components exact has made it visible and cause compatibility issue. This PR aim to fix contract of [Redirect](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Redirect.md) and [Route](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/Route.md).


definitions/npm/react-router_v4.x.x/flow_v0.63.x/react-router_v4.x.x.js
```diff
@@ -0,0 +1,130 @@
+declare module "react-router" {
```
GAntoine: Change folder name to `flow_v0.63.x-` (notice the `-`).

theres:
@GAntoine directory name is fixed & I've rebase onto master

PR #2355, opened 2018-06-12
Handle uninstalling of scoped packages

bugzpodder:
scoped packages (eg @material-ui) needs to include the scope and the package name in order to do lookups.


GAntoine:
Nice addition! Could you add a spec to `install-test.js` to cover this case?

bugzpodder:
i looked at it and its a bit hard, but i'll give it a try later.

cli/src/commands/install.js
```diff
@@ -296,10 +297,11 @@ async function installNpmLibDefs({
         // If a libdef is already installed for some dependency, we need to
         // uninstall it before installing the new (potentially updated) ver
         const libDef = npmLibDef.libDef;
-        //const toInstall = libDefsToInstall.has(libDef.name);
-        //console.log(`Found ${libDef.name} already installed. Uninstall? ${toInstall != null ? 'yes' : 'no'}`);
-        if (libDefsToInstall.has(libDef.name)) {
-          libDefsToUninstall.set(libDef.name, fullFilePath);
+        const scopedPkgName = getScopedPackageName(libDef);
+        //const toInstall = libDefsToInstall.has(scopedPkgName);
+        //console.log(`Found ${scopedPkgName} already installed. Uninstall? ${toInstall != null ? 'yes' : 'no'}`);
```
AndrewSouthpaw: Delete cruft.

PR #2368, opened 2018-06-14
lodash_v4.x.x - Change concat to use $ReadOnlyArray

Pephers:
Fix lodash_v4.x.x type definition for `concat`.


lukeapage:
Not sure this makes sense/is required? what code change requires this ?

KenanY:
I think this is a step towards resolving https://github.com/flowtype/flow-typed/issues/1099.

lukeapage:
This change is a spread argument.. the array is essentially created by language sugar, so I cannot see why this is needed. That’s why I didn’t make this change when changing others on this line.

Pephers:
As @lukeapage said it turns out this isn't needed. Some other type definition changes actually solved the issue I had with concat. I'll close the PR.

PR #2370, opened 2018-06-15
lodash: allow number as argument to get, getOr

### LINKED ISSUE #2128: lodash and lodash/fp: get() coerces argument with toKey so numbers should be OK
example:

```js
import { get } from 'lodash';

const a = { '1': 1, '2': 2 };

// flow type error but works because lodash uses toKey to coerce to string
get(a, 1);

// no flow type error
get(a, '1');
```

cc @mvanlonden

beaugunderson:
Fixes #2128.

lodash coerces arguments to `get` and `getOr` with `toKey`, meaning that numbers are valid argumnets to `get` and `getOr`.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
@@ -1127,7 +1127,7 @@ declare module "lodash" {
     functionsIn(object?: ?Object): Array<string>;
     get(
       object?: ?Object | ?$ReadOnlyArray<any> | void | null,
-      path?: ?$ReadOnlyArray<string> | string,
+      path?: ?$ReadOnlyArray<string | number> | string | number,
       defaultValue?: any
     ): any;
     has(object: Object, path: Array<string> | string): boolean;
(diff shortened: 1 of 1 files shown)


beaugunderson:
hmm ugh; found a corner case?

```
Could not decide which case to select. Since case 1 [1] may work but if it doesn't case 2 [2] looks promising too. To
fix add a type annotation to inferred union of array element types (alternatively, provide an annotation to summarize
the array element type) [3].

     static/scripts/state/diagnosticView/selectors.js
        3│ import { diagnosticViewEntityKey } from './api';
        4│ import { getPatientDetail } from 'state/patient/selectors';
        5│
 [3]    6│ const getDiagnosticViewEntities = get(['entities', diagnosticViewEntityKey]);
        7│
        8│ export const getDiagnosticViews = compose(values, getDiagnosticViewEntities);
        9│

     flow-typed/npm/lodash_v4.x.x.js
 [1] 2767│     get(path: $ReadOnlyArray<string | number> | string | number): (object: Object | $ReadOnlyArray<any> | void | null) => any;
 [2] 2768│     get(path: $ReadOnlyArray<string | number> | string | number, object: Object | $ReadOnlyArray<any> | void | null): any;
```

guidance appreciated...

AndrewSouthpaw:
I believe the `object: ... | void | null` is what's causing the trouble, because that can be difficult for Flow to correctly interpret the branch.

Do you get this error with the original libdef?

PR #2382, opened 2018-06-17
Update luxon typings

alizhdanov:
1. Missing `keepLocaTime` on `setZone` method [source code](https://github.com/moment/luxon/blob/master/src/datetime.js#L1135)

2. `unit` on `setZone` method is optional, because it has default value set to `miliseconds` [source code](https://github.com/moment/luxon/blob/master/src/datetime.js#L1585)


GAntoine:
Seems there's a failing test, you'll need to fix it before we can merge this in.

PR #2383, opened 2018-06-17
Fix ElementType in react-bootstrap

richardsamuels:
react-bootstrap uses the elementType proptype from this package: https://github.com/react-bootstrap/prop-types-extra/blob/master/src/elementType.js, which declares elementType as string or a React component. 

As written, the libdef will only accept strings, so this should fix that.


definitions/npm/react-bootstrap_v0.32.x/flow_v0.53.x-/react-bootstrap_v0.32.x.js
```diff
@@ -1,8 +1,7 @@
 declare module "react-bootstrap" {
-  import type { Node, Element, Component, Ref } from 'react';
+  import type { Node, Element, ElementType, Component, Ref } from 'react';
```
GAntoine: Why import `ElementType` when you've switched all of them to `React$ElementType`?

GAntoine:
Closing since stale.

PR #2408, opened 2018-06-22
Update express_v4.x.x defs for compatibility with flow_v0.75.0

### LINKED ISSUE #2407: Deprecated $call syntax in express_v4.16.x.js using flow 0.75.0
After upgrading to flow 0.75.0 the express definition file reports the following error but I cannot find any documentation on what changed:
```
Warning ----------------------------------------------------------------------- interfaces.flow/express_v4.16.x.js:200:3

Deprecated $call syntax. Use callable property syntax instead. (`deprecated-call-syntax`)

   200|   $call: (
          ^^^^^
```

antongolub:
https://github.com/facebook/flow/issues/3084 is closed (flow v0.55) so `$call` workaround may be thrown away.

fixes #2407


antongolub:
[Got build failure on travis](https://travis-ci.org/flowtype/flow-typed/jobs/395382495), but can not reproduce it locally (node v8.11.3, flow v0.63.0, OSX 10.13.5).
```bash
ERROR: express_v4.16.x/flow_v0.63.x-
 * express_v4.16.x/flow_v0.63.x- (flow-v0.63.0): Unexpected Flow errors(139):
   Segmentation fault
   
   null
```
As I see [some other builds](https://travis-ci.org/flowtype/flow-typed/jobs/395140844) fails too, so I've just excluded flow-v0.63.0 to pass the test.  Could anybody clarify this issue?

fxfactorial:
is this related to this error as well: 

```

Cannot call express because a callable signature is missing in module express [1].

 [1]  2│ import * as express from 'express';
..
  this.application = express();
```

antongolub:
@fxfactorial, this libdef fix is only related to flow_v0.75.0 update (legacy `$call`able is now deprecated).

You're trying to use ES6 default export, but module declared as CJS. Unfortunately, there's no way to support both notations at the same time.
```javascript
Found both declare module.exports and declare export in the same module. Modules can only have 1 since they are either
an ES module xor they are a CommonJS module
```
Moreover, the flow does not allow you to particularly override definitions, so the only way:
```javascript
import express from 'express'
```

antongolub:
@doberkofler, I could not find the cause of test failure on version 0.63. And I dont know whether it was correct to exclude this version from the test suite. Still looking for a review and approve from @GAntoine

GAntoine:
The segmentation fault is unrelated, if your other specs passed I'll merge this in when its ready.

GAntoine:
@antongolub You can rebase on the latest version of master, and the specs should pass again. I'd recommend no longer splitting at `0.63.0`, since the issue has been addressed.

antongolub:
Done, but travis is still trying to run test on 0.63. It looks like `flow-typed/cli/dist` wasn't rebuilt.

antongolub:
@GAntoine, could you restart the last build again?

antongolub:
@GAntoine, now everything is fine. `rebase and merge`?

PR #2410, opened 2018-06-22
Added missing type for Glyphicon

pwfcurry:


GAntoine:
Could you add a quick test for this addition?

PR #2413, opened 2018-06-22
Add generics to the set and update functions in the lodash

ej9x:


GAntoine:
Could you add a few specs to test the new generics?

GAntoine:
Closing since stale.

PR #2414, opened 2018-06-23
New libdef: js-joda 1.8.x

mmcgahan:
Weak tests, but this is a large lib and I'm eager to share types across a few projects at Meetup sooner rather than later.


mmcgahan:
Not sure what's causing the segmentation fault, but the build passes in Travis for our fork: https://travis-ci.com/meetup/flow-typed/jobs/130954073 (not sure if you can access that build log but it's all green)

PR #2419, opened 2018-06-24
Change $call to callable property syntax

mastertinner:
@wzrdzl 
@thymikee 
@rasmusprentow 
@mic
@bwlt 
@TLadd
@budde377


GAntoine:
Is this relevant to all versions of flow, or just 0.75+?

mastertinner:
Thanks, @GAntoine. I did as you suggested.

GAntoine:
You'll need to the cap the other version at `-0.74.x`, you're getting a disjoint flow version error. Let me know if you need some guidance in doing that. Also could you rebase both PRs on master? That will help fix one of the failing specs.

PR #2420, opened 2018-06-25
Use callable property syntax instead of $call

mastertinner:
This fixes a new error introduced with flow 0.75:
Deprecated $call syntax. Use callable property syntax instead.


GAntoine:
Same question as the other PR.

mastertinner:
Thanks, @GAntoine. I did as you suggested.

mastertinner:
@GAntoine, it looks like it's all green now :)

PR #2421, opened 2018-06-25
Upate Knex def for 0.75

facundo-viale-olx:


facundo-viale-olx:
I moved the tests.

GAntoine:
@facundo-viale-olx One last thing, I'll need you to fix the names of some of the files. All files should end with the full definition name-version combo (knex_v0.15.x.js). Also the connecting caracters should be `_`, and not `-`, as they could cause issues. So files like `test_knex-v0.14.js` should be named `test_knex_v0.14.js`.

facundo-viale-olx:
done

PR #2425, opened 2018-06-25
react-redux: mapDispatchToProps can return an object

### LINKED ISSUE #2284: [react-redux] mapDispatchToProps incorrectly maps action type to component
when using `mapDispatchToProps` as Object it forces component properties to return redux action objects.... 
`react-redux@5.x.x`

```js
// actions.js
export const LOAD_PRODUCT = {
  REQUEST: "product/LOAD_PRODUCT_REQUEST",
};

export const loadProductAction = (id: number) => ({
  type: LOAD_PRODUCT.REQUEST,
  id
});


// container.js
const mapStateToProps = state => ({
});

const mapDispatchToProps = {
  loadData: loadProductAction
};

export const ProductInfo = connect(mapStateToProps, mapDispatchToProps)(
  ProductView
);


// component.js
type Props = {
  loadData: number => void,
};

export class ProductView extends React.Component<Props> { 
  render() { ... }
}


```

flow check returns 
```
undefined [1] is incompatible with object literal [2] in the return value.

     src/shared/features/product/product-info/component.js
     23│ };
     24│
     25│ type Props = {
 [1] 26│   loadData: number => void,
     27│   product: TProduct
     28│ };
     29│

     src/shared/features/product/actions.js
 [2]  6│ export const loadProductAction = (id: number) => ({
      7│   type: LOAD_PRODUCT.REQUEST,
      8│   id
      9│ });
```

which assumes that component props function `loadData` should return `Action` type which is incorrect, since it's wrapped with dispatch and does not return anything 

changing `mapStateToProps` 

```
const mapDispatchToProps = (dispatch: any => void) => ({
   loadData: id => dispatch(loadProductAction(id))
});
```
fixes flow error, but I would like to use object variant for some reason. I believe it used to work previously

Anujan:
This fixes #2284

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.63.0-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.63.0-/react-redux_v5.x.x.js
@@ -33,6 +33,8 @@ declare module "react-redux" {
 
   declare type MapDispatchToProps<A, OP: Object, RDP: Object> = (dispatch: Dispatch<A>, ownProps: OP) => RDP;
 
+  declare type MapObjDispatchToProps<RDP: Object> = RDP;
+
   declare type MergeProps<SP: Object, DP: Object, MP: Object, RMP: Object> = (
     stateProps: SP,
     dispatchProps: DP,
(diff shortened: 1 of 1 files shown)


AndrewSouthpaw:
There are test errors, you'll need to fix that first.

AndrewSouthpaw:
cc @villesau

PR #2428, opened 2018-06-26
fix: ramda path should support string and number

foray1010:
according to https://ramdajs.com/docs/#path
```
Idx = String | Int
```
the path should be string or integer


foray1010:
reopen until fix errors

PR #2429, opened 2018-06-26
[react-intl] proper flowtypes for injectIntl using React.ElementConfig

bobrovnik:
Main reason for this PR is that current implementation of injectIntl completely ignores flowtype of wrapped component.

Starting from flowtype@.0.63.0 we have React.ElementConfig<...> method for writing proper types for Higher Order Components (HOCs) https://flow.org/en/docs/react/types/#toc-react-elementconfig. 

Note: this PR introduces no changes to existing code, but changes the way how injectIntl works for flowtype => 0.63 version


definitions/npm/react-intl_v2.x.x/flow_v0.63.x-/react-intl_v2.x.x.js
```diff
@@ -0,0 +1,259 @@
+/**
+ * Original implementation of this file by @marudor at https://github.com/marudor/flowInterfaces
+ * Copied here based on intention to merge with flow-typed expressed here:
+ * https://github.com/marudor/flowInterfaces/issues/6
+ */
+// Mostly from https://github.com/yahoo/react-intl/wiki/API#react-intl-api
+
+import type { Element, ChildrenArray } from "react";
```
pascalduez: Should be moved inside the module declaration, otherwise those will resolve to `any`.

definitions/npm/react-intl_v2.x.x/flow_v0.63.x-/react-intl_v2.x.x.js
```diff
@@ -0,0 +1,259 @@
...
+
+type $npm$ReactIntl$LocaleData = {
```
AndrewSouthpaw: These should all be wrapped inside the module generally.

bobrovnik: Done. 
I think this part I just copy pasted from prev implementation.

definitions/npm/react-intl_v2.x.x/flow_v0.63.x-/react-intl_v2.x.x.js
bobrovnik: Moved it inside the module declaration. Thanks for the hint

definitions/npm/react-intl_v2.x.x/flow_v0.63.x-/react-intl_v2.x.x.js
AndrewSouthpaw: Makes sense, I appreciate you taking the extra time to improve the libdef!

PR #2431, opened 2018-06-26
mongoose 5.x.x libdef

nodkz:


definitions/npm/mongoose_v5.x.x/test_mongoose-v5.js
```diff
@@ -0,0 +1,136 @@
+// @flow
```
AndrewSouthpaw: Test should be in the same directory under the flow version.

nodkz: Done. 
Also, moved tests in mongoose@4

PR #2442, opened 2018-06-27
remove type for DayOfWeek.valueOf() to prevent error

mmcgahan:
Flow does not allow classes that inherit from `Object` to implement `static valueOf(arg)`. Unfortunately, that's what `js-joda` does. It might not be safe, but that's the way it is. Consumers will have to add `$FlowFixMe` around any usage of `valueOf()` in their code for the time being.


AndrewSouthpaw:
Can you describe what's going wrong currently that motivated the PR? I don't usually see `$ExpectError` inside a libdef so not sure what it's doing for consumers.

PR #2443, opened 2018-06-28
[react-redux] Support static properties hoisting in connect

### LINKED ISSUE #830: react-redux: connect(...) doesn't retain statics' types
As of react-redux@3.1.0, redux hoists non-react statics to the wrapper class. 

Is there some way to maintain the API of the wrapped class through the connect() call in addition to exporting a React.Component with similar Props?

maximeg:
react-redux connect does hoist not react statics as per https://github.com/reduxjs/react-redux/blob/master/src/components/connectAdvanced.js#L288

So I added support for that in typing as I needed it.

After digging, this fixes #830.

Cheers.

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.63.0-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.63.0-/react-redux_v5.x.x.js
@@ -27,6 +27,7 @@ declare module "react-redux" {
   RMP = Returned merge props
   CP = Props for returned component
   Com = React Component
+  ST = Static properties of Com
   */
 
   declare type MapStateToProps<S: Object, SP: Object, RSP: Object> = (state: S, props: SP) => RSP;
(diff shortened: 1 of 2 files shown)


villesau:
Nice, this is great @maximeg !

SBoudrias:
Awesome! Thanks a lot @maximeg

cloudyster:
This change broke some of our type checks, here are some snippet's (sorry, cant' give more details at the moment):
```
     src/components/app/ProtectedRoute.jsx
     32│         return (
     33│           <div>
     34│             <NavBar />
 [1] 35│             <OrganisationSelect org={org} />
     36│             {element}
     37│           </div>
     38│         );

 [2] 86│     CP: $Diff<$Diff<ElementConfig<Com>, RSP>, RDP>,

```
and 
```
Cannot create SomeInputs element because:
 • Either property formValueNumber is missing in props [1] but exists in props [2].
 • Or property searchSomeMode is missing in props [2] but exists in props [1].
 • Or ST [3] is not a React component.

     src/components/admin/SomeAdmin.jsx
     156│                     {messageRestrictedForSupport}
     157│                   </div>
     158│                 </div>
 [2] 159│                 <SomeInputs formValueNumber={''} formValueComment={''} />
     160│                 <div className="row">
     161│                   <div className="col-12">
     162│                     <hr />

 [3]  64│     ST: {[_: $Keys<Com>]: any}

     src/components/admin/SomeSearchFormModal.jsx
 [1]  44│                 <SomeInputs searchSomeMode={true} />

```

Benjamin-Dobell:
This commit seems to have broken chaining with other libraries/libdefs that expect a `ComponentType` e.g.

```
  declare module.exports: <WrappedComponent: React$ComponentType<*>>(
  ) => (
    Component: WrappedComponent
  ) => React$ComponentType<$Diff<React$ElementConfig<WrappedComponent>, {somePropToOmit: number}>>;
```

When you proceed to render you get the dreaded:

```
Cannot create SomeComponent element because props [1] is incompatible with empty [2].
```

If I roil this back, everything works as expected. I suspect this is what @cloudyster ran into also.

Benjamin-Dobell:
Pretty sure this is actually the root cause of what was "fixed" in https://github.com/flow-typed/flow-typed/pull/2629

By the way, @villesau, I'm pretty sure that's only passing now because you've replaced actual type references (e.g. `React$ComponentType`) with `any` i.e. `import {ComponentType} from "react";`

The latter of which being a result of the ongoing dependency saga (#16 and co).

villesau:
@Benjamin-Dobell React has builtin libdefs in Flow which makes it fine to import within module declaration.

Benjamin-Dobell:
I'm aware React has in-built defs within Flow, but to use them shouldn't they be accessed as `React$ComponentType` etc. without any explicit imports?

Benjamin-Dobell:
@villesau Whoops, you're quite right! I really ought to have tested that at least once before saying something. As you've rightly pointed out, I should probably go update my libdefs!

Also, after writing my own definitions (PR submitted) for __material-ui__'s higher order components, specifically `withStyles`, I'm no longer running into issues with this commit. Previously I had to roll back this commit (or use explicit types through-out my HOC chain) in order to make things work. So ah, yeah... sorry for bothering you!

villesau:
@Benjamin-Dobell no problem! This conversation actually led me to learn something too! :) I've been wondering how to define react classes without `React$Component` by importing component type `import type { Component } from 'react'`. And here is my finding: `declare export var MyComponent: Class<Component<{someProp: number}>>` <- no need to extend!

PR #2446, opened 2018-06-28
Fix and add test case for named imports from winston

KevinGrandon:
Without this error, the following error occurs:

> [flow] Cannot import `createLogger` because there is no `createLogger` export in `winston`. Did you mean `import createLogger from "..."`?


GAntoine:
You'll need to fix this specs before we can merge this in.

KevinGrandon:
I'm not really clear at this point how to fix the tests, sorry =( If anyone can provide some guidance, I'm happy to pick this back up. I do think it's still an issue, but will close for now until I have more time to look into tests.

PR #2450, opened 2018-06-29
[ramda] add types for chain function

honzabrecka:
https://ramdajs.com/docs/#chain


GAntoine:
Could you add a test to cover this addition?

honzabrecka:
Tests added.

PR #2451, opened 2018-06-29
redux-saga - declare takeLeading

stropho:


stropho:
I am quite confused why it has failed. Can anybody help?

PR #2262, opened 2018-05-24
Improve withFormik config type definitions

JCMais:
Change signature to match the one at https://github.com/jaredpalmer/formik/blob/v0.11.11/README.md#withformikoptions

PR #2452, opened 2018-06-29
Formik improvements

bdrobinson:
The formik typedefs were good but it's very verbose having to explicitly label the render prop args with `FormikActions<Values>` etc. Instead, this PR makes them automatically inferred by `initialState`. This makes things more friendly for newcomers and work better out of the box.

I've made `setFieldValue` much more powerful – it can now verify that the value being set is the correct type. I also deleted some duplicate methods definitions as I thought they were undermining the type checking.

If people do need to explicitly set the type of `Values` rather than inferring it from `initialState` (although I imagine the latter will suit the majority of usecases), I've added a test that demonstrates how to do that.

I've increased the required version of flow to 0.59 from 0.53 – I tried running the tests on the repo as it was before my changes, and they required flow 0.59 anyway (I think they use the `$ReadOnly` helper) so anyone who tried using these with 0.53-0.58 probably would have got an error anyway.

I've tried to split my changes into small commits to make it easier to see what changes I've made.

Let me know if you have any questions. We've been using these new typedefs internally at work and they seem to have worked well so far.


GAntoine:
@sibelius Could I get your eyes on this PR?

sibelius:
using $Call is a nice tricky

this looks good, and works fine for us

I'd like to bring the changes of this other PR as well https://github.com/flow-typed/flow-typed/pull/2262

GAntoine:
Awesome, thanks for chiming in so quickly! I'll get this merged in.

Issue #2461, opened 2018-07-03
Axios should not care about request/response payload types


LoganBarnett:
I stumbled across this due to what I think is a regression in #1975.

The basic gist is that the Axios libdef is assuming a few things:

1. Any payload passed as part of a `POST` or `PUT` will be the same data type that comes back in the response ([demonstrated here](https://github.com/LoganBarnett/axios-flow-req-res-types)).
2. We can trust the server to always give us the data type we desire.

For 1, this might be the case a lot of the time but it's not something we can assure is always the case. There are no rules that stipulate that data returned must be of the same type that is provided. I think #1975 addressed this, but it seems the request/response types are now in lockstep again.

As a corollary to 1, how can we assume the server is going to yield the correct type value? There are validation/deserialization solutions for Flow currently, and we should be leaning on those to ensure soundness in our communications code. We can fix 1 by separating the types, but we can fix 1 and 2 together by simply making the payloads `mixed` on both the request and the response type. I don't think we can safely say anything else about our data being exchanged. Perhaps we can do some kind of trick where we ensure the response payload can't have functions in it since we know it will be the result of `JSON.parse`. This is covered in https://github.com/facebook/flow/issues/4183 but `mixed` would make for a fine first pass.

Thoughts?

Issue #2405, opened 2018-06-21
Flow 0.75.0 reports 'Deprecated $call syntax' error with Axios


benjreinhart:
Hey there! I'm getting the following error regarding the `$call` (lowercase c) syntax. I'm not familiar with this syntax and there's nothing in the flow docs about this. The only mention of it I could find was an old [github issue](https://github.com/facebook/flow/issues/3781#issuecomment-299068112) in the flow repo and it wasn't helpful.

This started happening when I upgraded flow from 0.74.0 to 0.75.0.

* flow: 0.75.0
* yarn: 1.7.0
* typed/npm/axios_v0.18.x.js

```
> yarn flow
yarn run v1.7.0
$ ./node_modules/.bin/flow status
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ flow-typed/npm/axios_v0.18.x.js:102:5

Deprecated $call syntax. Use callable property syntax instead. (deprecated-call-syntax)

      99│   declare type AxiosPromise<T,R = T> = Promise<AxiosXHR<T,R>>;
     100│   declare class Axios {
     101│     constructor<T,R>(config?: AxiosXHRConfigBase<T,R>): void;
     102│     $call: <T,R>(
     103│       config: AxiosXHRConfig<T,R> | string,
     104│       config?: AxiosXHRConfig<T,R>
     105│     ) => AxiosPromise<T,R>;
```

eqyiel:
@madole you can read a bit more about it here: https://github.com/facebook/flow/releases/tag/v0.75.0

AFAICT, `s/$call/[[call]]/g`, but it seems like that doesn't work.

madole:
I did try that as well without any luck. Ended up hacking the type, not even sure it's doing the same thing now :(

ryota-murakami:
thanks all!

i'm facing too this error and also even for packages other than axios.
so is there a way(i guess .flowconfig setting) to suppress `deprecated-call-syntax` Error?

it is useful until reflect fix to type definition file.

dashed:
[...]

PR #2464, opened 2018-07-04
remove type param from axios such that responses and requests are all now mixed

LoganBarnett:
`eslint` was complaining so I had to bypass the check - there's something it doesn't like about the `[[call]]` syntax, even though this is now preferred to `$call`. This might be related to #2405. I did not change the `[[call]]` syntax, so I'm not sure this worked prior.

This PR removes the polymorphic types for `axios`. This both decouples the request and response payloads (a possible regression of #1975), and improves soundness by moving the validation/deserialization to the consumer (or a blind `any` cast, which is really what was going on before).

Addresses #2461.

### DIFF
--- a/definitions/npm/axios_v0.11.x/flow_v0.25.x-v0.74.x/axios_v0.11.x.js
+++ b/definitions/npm/axios_v0.11.x/flow_v0.25.x-v0.74.x/axios_v0.11.x.js
@@ -1,5 +1,5 @@
 declare module "axios" {
-  declare interface AxiosXHRConfigBase<T> {
+  declare interface AxiosXHRConfigBase {
     headers?: Object;
     maxContentLength?: number;
     params?: Object;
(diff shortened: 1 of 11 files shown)


AndrewSouthpaw:
It seems to me that setting the return types as `mixed` will require frustrating `any`-casting to actually interact with the return time. If you do something like

```js
axios.get('/foo').then(res => { console.log(res.foo) })
```

does that throw an error? Can you write a test and make sure that's possible?

LoganBarnett:
@AndrewSouthpaw being pedantic here: The response itself does not have `foo`, so the code in your example would not and should not type check.

Perhaps you meant:
``` javascript
axios.get('/foo').then(res => { console.log(res.data.foo) })
```

In this case it will fail to type check as well (rightfully so). You aren't strictly required to cast to `any` here. You could check to see if `foo` was present and what type it was before attempting to use it. These examples are doing the exact same thing in Flow's eyes:

``` javascript
type Foo = { foo: string }
(res: $AxiosXHR<Foo>) => console.log(res.data.foo.toUpperCase()
```

``` javascript
type Foo = { foo: string }
(res) => console.log(((res.data: any): Foo).foo.toUpperCase()
```

In both cases, you're saying "Flow just trust me on this one" and we all know what kind of lie that is to Flow and ourselves ;) If we treat `data` as `mixed`, it puts the burden on us to ensure that whatever `data` is, we have to do our due diligence first before we do anything meaningful with it. Granted that for complex data types, this can be a lot of work. To address that, there are some type safe validation libraries out there such as [validated](https://github.com/andreypopp/validated) and (shameless plug) [flow-degen](https://github.com/loganbarnett/flow-degen). There are others out there as well.

AndrewSouthpaw:
RE: `foo` on `res`, yes, you're quite right, I meant `res.data.foo`.

So, at *most*, as a consumer of axios I'd want to do something like 

```js
type Foo = { foo: string }
(res: $AxiosXHR<Foo>) => console.log(res.data.foo.toUpperCase()
```

If I had to do

```js
(res: $AxiosXHR<Foo>) => {
  if (res.data.foo) {
    ...
  }
}
```

or anything else like it using a validation library, I would get very, very sad. Yes, it's not technically type safe, and also I'm okay with that; it's something I like about Javascript, that's it's not *actually* a typed language and sometimes I can choose to not worry about it without having to validate everything. If I'm actually concerned about it I can pass it through the validation, but generally I'd rather not worry about it.

I think that's a reflection of how I like to code and what purpose I want Flow to serve for me. Everyone is different. So, my vote is to not merge this PR. I'm also happy to be outvoted and just keep my own older version locally. @GAntoine @villesau @adityavohra7. Feel free to CC other people e.g. maintainers of this libdef if you want.

AndrewSouthpaw:
Man, what I'd do for a forceful recast that wasn't `((foo: any): MyType)`. Ugh. It just looks so awful, and the lack of a forceful recast (to me) makes stuff like `mixed` pretty much useless.

Yeah curious to hear input from others.

RE: aside: The CI tests are ... kinda crap. Restarted.

villesau:
I'm with @LoganBarnett here. Soundness should be priority, and that's the reason we even use tools like Flow :)

Although, I totally understand @AndrewSouthpaw s point, and checks would probably make annoy me too. But I have news for you: Flow have supported [optional chaining](https://github.com/tc39/proposal-optional-chaining) now [for a while ](https://github.com/facebook/flow/releases/tag/v0.75.0) which makes this kind of checks a lot easier :)

AndrewSouthpaw:
@villesau A couple thoughts about optional chaining... it's only stage 1 right now, so offering it as a solution doesn't seem fair to the community. The bigger issue in my opinion is that it only solves part of the problem. My example wasn't complete. Here's what I'm actually afraid of needing to do:

```js
type Foo = { a: string, b: number, c: SomeOtherComplexType }
axios.get('/foo').then((res: $AxiosXHR<mixed>) => {
  if (typeof res.data.a === 'string' && typeof res.data.b === 'number' /* && argh wtf c */) {
    ...
  }
})
```

Not only is that annoying to write out (or maintain a bunch of type-checking helpers), basically my fear is that we'd be forcing developers to perform a lot of runtime checks in order to provide type safety. It's potentially (probably?) totally unfounded, but I'm having concerns about performance impact on slow phones, and also the crummy dev experience of just having to actually perform type checking on every blasted AJAX request in my codebase.

I'm also definitely willing to be overruled here in the face of a majority vote. I'd love to get some other opinions to weigh in, and I can shamefully override my `axios` libdef and live dangerously if people decide my opinion is not the winning one. 😉

villesau:
> it's only stage 1 right now, so offering it as a solution doesn't seem fair to the community

Well.. JSX is not part of Javascript standard and neither is Flow it self so you will always need a transpiler for your Flow code. And flow clearly seems to go optional chaining path as they have implemented the support for it. At least they have strong belief in optional chaining, I don't think that kind of features are implemented just for fun. There are only very few experimental Javascript features Flow supports optional chaining being one of them.

But your example indeed looks very annoying. Didn't think of more complex objects in previous message. Do we have any good examples of other similar cases?

AndrewSouthpaw:
> Well.. JSX is not part of Javascript standard ... one of them.

Fair point. I think we draw the line in different places, and I do see what you're saying.

As far as complex cases, I could point to any number of them within our codebase. We have plenty server calls and they often return complex objects of different types.

PR #2467, opened 2018-07-04
Fix Hoc support in react navigaton

wcandillon:


Ashoat:
Hey @wcandillon, can you give the [CONTRIBUTING](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-navigation_v2.x.x/CONTRIBUTING.md) file a quick read? In the future please make sure to send any PRs for this libdef to the `react-navigation` repo first.

Can you send me a code snippet of what was breaking without this change? I suspect there's a way to simplify the type parameters here, but want to make sure it doesn't break whatever you're using.

PR #2475, opened 2018-07-05
add typings for empty / isEmpty / notEmpty / isNotEmpty in chaijs assert

aliceriot:
Was just adding these to my project and realized they weren't included, so I thought I'd go ahead and try opening a PR. The typings are pretty straightforward I think, lmk if there's anything else I need to add.

Thanks!


GAntoine:
Nice! Could you also add them to the `chai_v3.5.x` definition?

PR #2489, opened 2018-07-07
metalsmith_v2.x.x: Fix File type to be less strict

jirutka:
File object is not always backed by a file on disk, it can be also created by a plugin; such object typically don't have properties stats and mode.

### DIFF
--- a/definitions/npm/metalsmith_v2.x.x/flow_v0.75.x-/metalsmith_v2.x.x.js
+++ b/definitions/npm/metalsmith_v2.x.x/flow_v0.75.x-/metalsmith_v2.x.x.js
@@ -20,12 +20,23 @@ declare module 'metalsmith' {
 
   declare type File = {
     contents: Contents,
-    mode: string,
-    stats: Stats,
+    mode?: string,
+    stats?: Stats,
     [string]: any,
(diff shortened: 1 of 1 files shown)


definitions/npm/metalsmith_v2.x.x/flow_v0.75.x-/metalsmith_v2.x.x.js
```diff
@@ -71,11 +82,11 @@ declare module 'metalsmith' {
     run(files: FilesMap, plugins: Array<Plugin>, cb: Callback<FilesMap>): void;
     run(files: FilesMap, plugins: Array<Plugin>): FilesMap;
```
GAntoine: Do all these methods need versions compatible with `RFile`? If so, then this pattern isn't very efficient.

jirutka: No, only in methods that creates and returns _file_ (i.e. `readFile` and `read`) is guaranteed that the _file_ contains `mode` and `stats` (and no extra properties). `RFile` is de facto subtype of `File`.

PR #2444, opened 2018-06-28
lodash: when using _.get, keys can be numbers (for the array case)

lukeapage:

PR #2436, opened 2018-06-27
Port lodash 4.x.x libdefs to lodash-es

tadaskay:

Issue #2488, opened 2018-07-07
Support modularized NPM packages


josemigallas:
When I require [lodash](https://www.npmjs.com/search?q=keywords:lodash-modularized) in my projects, I rarely want to use the whole library but I just install the modularized methods like `lodash.shuffle`, `lodash.clonedeep`...

However in this case flow-typed can't find definitions for those packages so either I have to:

- `npm install` the whole library so that `flow-typed` can install the definitions automagically, or
- explicitly `flow-typed install lodash@...` manually every time, which is bad

I actually wouldn't mind having the whole definition file in my project, since it's gitignored but I would like `flow-typed` to somehow recognize a modularized package is being used. This is of course not directly related to lodash but to any library that supports this modularization.

AndrewSouthpaw:
Definitely would be a Nice To Have. Folks are welcome to take a stab at this one, though I think it'll be challenging.

Out of curiosity, what other packages do this style of modularization?

josemigallas:
Well actually I haven't seen that anywhere else so far! Not even underscore.. If I did take a look at this, could it be a solution to "simply" add all modularized packages under definitions/? The after after all just NPM packages.

AndrewSouthpaw:
I'm gonna go with "no" on this solution, because it would be a maintenance nightmare, unless there was supporting code to automatically copy the types into the modularized folders.

Brianzchen:
@AndrewSouthpaw If we had dependent definitions would you be in favour of this and then we can add more definitions for the community?

```
declare module 'lodash.shuffle' {
  import typeof { shuffle } from 'lodash';

  declare module.exports: typeof shuffle;
}
```

Brianzchen:
^ Follow example Pr above if anyone wants to add subset packages

PR #2490, opened 2018-07-08
Join `lodash` and `lodash-es` libdefs

antongolub:
It seems to be difficult to support several libdebs in future. `lodash-es` is a custom build with exactly the same contracts. But `get` method in `lodash`:
```javascript
get(
      object?: ?Object | ?$ReadOnlyArray<any> | void | null,
      path?: ?$ReadOnlyArray<string | number> | string | number,
      defaultValue?: any
    ): any;
```
and `lodash-es` are different:
```javascript
  declare export function get(
    object?: ?Object | ?$ReadOnlyArray<any> | void | null,
    path?: ?$ReadOnlyArray<string> | string,
    defaultValue?: any
  ): any;
```

Let's reuse `lodash` export:
```javascript
declare module "lodash-es" {
  declare module.exports: $Exports<"lodash">;
}
```
@tadaskay, @AndrewSouthpaw what do you think about?

#2444 #2436


tadaskay:
Looks like the way to go. But will flowtyped pull definitions properly when it is defined in such a way (provided project has `lodash-es` dependency but not `lodash`)?

antongolub:
@tadaskay I'm afraid will not.
@GAntoine, @AndrewSouthpaw is it possible to define [lodash-modularized packs](https://www.npmjs.com/search?q=keywords:lodash-modularized) and avoid copy-paste?

AndrewSouthpaw:
Not sure about the modularized packs, but I think it wouldn't be the right way to go, because most people (based on no evidence whatsoever) use the general lodash package rather than the modularized ones. 

I don't think symlinking it would do what we wanted either...

I'm at a loss for how to avoid copy-pasting. @GAntoine any thoughts?

GAntoine:
AFAIK copy-pasting is the only solution at the moment, and I don't see an easy way to avoid doing it.

> Also I don't think we should entirely delete lodash-es because there are probably people who currently depend on it, and it will unfairly break them.

Agreed.

> It should be possible to add something to a script somewhere that auto copies the files for us, but I don't have immediate thoughts about where that script would live or be executed.

We could write a script that runs on master builds (on travis) and copies the files over, kinda like what we have/are working on in the experimental branch. Though I'm just throwing out ideas.

antongolub:
Ok. Let's do some practical steps.
1) Fix `lodash-es.get` for now: #2495 
2) Add an issue about modularized packs: #2488
3) Close this PR.

PR #2498, opened 2018-07-12
react-apollo: Fix $call deprecated error on flow 0.75-

TLadd:
Remove $call where it was used. $call was being used as a workaround for a union
of objects and a function (https://github.com/facebook/flow/issues/1948#issuecomment-338440545)

A sealed object also resolves this issue and removes the deprecation error


TLadd:
@budde377 @wzrdzl @thymikee @rasmusprentow @mic @bwlt Removing $call to work with flow 0.75-

GAntoine:
@TLadd This LGTM, I'll wait until the others you've pinged see this before merging it in.

TLadd:
Thanks all for the quick reviews/merge!

PR #2503, opened 2018-07-13
Improve scripts for faster testing feedback loop; improve docs.

AndrewSouthpaw:
- Breaks apart scripts to set up / test the cli from the tests to run libdef tests.
- Testing libdefs are now way faster as a consequence.
- Improves docs in `CONTRIBUTING.md` for new users.


CONTRIBUTING.md
```diff
@@ -16,7 +17,27 @@ to existing libdefs.
...
+$ ./build_and_test_cli.sh
+```
+
+to build the CLI. Once that's finished, you can repeatedly run
+
+```
+$ ./quick_run_def_tests.sh
+```
+
+to run tests for any libdefs that have changes. (Caveat: the tests are only run for the most recent flow version of the libdef, so if you make changes to an earlier flow version of the libdef, the tests won't run against them, and honestly you should just update your flow anyway and stop working on older versions 😉).
```
GAntoine: Would remove `and honestly you should just update your flow anyway and stop working on older versions 😉`.

CONTRIBUTING.md
```diff
@@ -16,7 +17,27 @@ to existing libdefs.
 * [Writing tests](#writing-tests)
   * [Use `describe` and `it` blocks to limit scope](#use-describe-and-it-blocks-to-limit-scope)
 
-## Contributing to the definitions repository
+## Running libdef tests
+
+We write tests for libdefs. When you make changes to a libdef, you should make sure the tests still pass or add new tests for new functionality.
```
GAntoine: `you should make sure the tests still pass or add new tests for new functionality ` -> `verify that the tests still pass, and add tests to cover new functionality `.

AndrewSouthpaw: Thanks. #firstPassesAtWords

CONTRIBUTING.md
AndrewSouthpaw: √

PR #2506, opened 2018-07-16
type(formik): improve withFormik hoc type hinting

JCMais:
This makes the formik definition more close to the TS one at their repo.

Also changes `React.*` to use `React$*` since the first does not work correctly because imports insides type definitions do not work correctly.

This also types the `withFormik` hoc correctly, to take into consideration the following:

> Note: If foo does not exist in Props you will get an error! $Diff<{}, { foo: number }> will be an error. To work around this use a union with void, see: $Diff<{}, { foo: number | void }>. An optional prop will not completely remove foo. $Diff<{ foo: number }, { foo?: number }> is { foo?: number } instead of {}.

https://flow.org/en/docs/react/hoc/#toc-injecting-props-with-a-higher-order-component

I have a question tho, I've created a `TypeOrVoid` type helper, but have not exported it, is it going to pollute the global scope? Or if they are not exported they are scoped to the file?

Asking that because I saw for example on `styled-components` type definitions, where they declare the types using something like `$npm$styledComponents$blah`

### DIFF
--- a/definitions/npm/formik_v0.11.x/flow_v0.59.x-/formik_v0.11.x.js
+++ b/definitions/npm/formik_v0.11.x/flow_v0.59.x-/formik_v0.11.x.js
@@ -1,6 +1,4 @@
 declare module "formik" {
-  import type { ComponentType } from "react";
-
   declare export type FormikErrors<Values> = {
     [field: $Keys<Values>]: ?string
   };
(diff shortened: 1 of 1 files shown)


sibelius:
it won't pollute as this is inside declare module

only pollute if this is outside declare module

PR #2509, opened 2018-07-17
fix generics for bluebird Promise.all

stekycz:
From flow version 0.76.0 resolution of subtyping polymorphic types is more strict. See https://github.com/facebook/flow/releases


GAntoine:
Closed in favour of #2605

PR #2510, opened 2018-07-17
react-apollo: fix stopPolling()

oskarkook:
`stopPolling()` does not take an argument. For reference:
https://www.apollographql.com/docs/react/essentials/queries.html#render-prop
https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/core/ObservableQuery.ts#L500


AndrewSouthpaw:
Thanks for the contribution!

PR #2511, opened 2018-07-18
Improvements to Firebase v5 definitions

nfarina:
This builds on #2308 and adds the following changes:

  - Adds the `app.firestore()` function to get the associated Firestore instance for an initialized App.
  - Fixes argument types for the callback passed to `onSnapshot()` to allow `QuerySnapshot`.
  - Changes the `data` parameter to be typed as `Object` instead of `{}`, allowing access to keys (instead of assuming that data is an empty object).

/cc @meafmira @meafmira


meafmira:
LGTM

meafmira:
thx @nfarina

PR #2513, opened 2018-07-18
Add libdef NewRelic 4.x.x

facundo-viale-olx:


definitions/npm/newrelic_v4.x.x/flow_v0.53.x-/newrelic_v4.x.x.js
```diff
@@ -0,0 +1,71 @@
+// @flow strict
+/* eslint-disable flowtype/no-weak-types */
+
+// @see https://docs.newrelic.com/docs/agents/nodejs-agent/supported-features/nodejs-agent-api
+declare module 'newrelic' {
+
+  declare interface TransactionHandle {
```
AndrewSouthpaw: Generally we've been finding interfaces to be less performant. Do you definitely need it to be an interface?

definitions/npm/newrelic_v4.x.x/test_newrelic_v4.x.x.js
```diff
@@ -0,0 +1,43 @@
+// @flow
```
AndrewSouthpaw: Put the test under the versioned flow directory.

Also, use describe/it blocks.

definitions/npm/newrelic_v4.x.x/flow_v0.53.x-/newrelic_v4.x.x.js
facundo-viale-olx: Oh! I wasn't aware of that. I'm moving this to a type.

definitions/npm/newrelic_v4.x.x/test_newrelic_v4.x.x.js
facundo-viale-olx: Done

AndrewSouthpaw:
Cool, looking better. Did you want to change the `interface` over to types as well?

facundo-viale-olx:
I switched `NewRelicAPI` to a type. I don't know how can i turn this into a type:

```
  declare interface Instrument {
    (opts: { moduleName: string, onRequire: () => void, onError?: (err: Error) => void }): void;
    (moduleName: string, onRequire: () => void, onError?: (err: Error) => void): void;
  }
```

AndrewSouthpaw:
... can you not just change it to a type? Looks like it'd be fine. Essentially if you're never doing `implements Instrument` that's a pretty clear indication it doesn't need to be an interface.

facundo-viale-olx:
Sorry my bad. I thought defining a constructor was only in a `ìnterface`, maybe I never saw a `type` defining a constructor. 
Now I know the main difference is the `implements`.

PR #2520, opened 2018-07-19
add the ability to refine types from Ramda's filter

LoganBarnett:
We can't really do `reject` here since it's an inversion, and I'm not aware of any special `$Not<T>` just yet.

Currently, you can't do this in Ramda with Flow:
``` javascript
import { filter, map } from 'ramda'
const ns = filter(n => n != null, [1, 2, null])
map(n => n.toString(), ns) // n can be null so you can't do .toString()
```

There's a small concession to make it work: You have to indicate that you actually want to do a refinement. This is because refinement predicates are considered a different type than ordinary predicates.
``` javascript
import { type RefineFilter, filter, map } from 'ramda'
const notNull = (x: mixed): bool %checks  => n !=null
const ns = (filter: RefineFilter)(notNull, [1, 2, null])
map(n => n.toString(), ns) // works
```

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
@@ -407,13 +407,25 @@ declare module ramda {
     ) => UnaryFn<A, C>) &
     (<A, B>(ab: UnaryFn<A, B>) => UnaryFn<A, B>);
 
-  declare type Filter = (<K, V, T: Array<V> | { [key: K]: V }>(
-    fn: UnaryPredicateFn<V>,
-    xs: T
-  ) => T) &
-    (<K, V, T: Array<V> | { [key: K]: V }>(
-      fn: UnaryPredicateFn<V>
-    ) => (xs: T) => T);
+  // This kind of filter allows us to do type refinement on the result, but we
+  // still need Filter so that non-refining predicates still pass a type check.
+  declare type RefineFilter =
+    & (<K, V, P: $Pred<1>, T: Array<V> | { [key: K]: V }>(
+      fn: P,
+      xs: T
+    ) => Array<$Refine<V, P, 1>>)
+    & (<K, V, P: $Pred<1>, T: Array<V> | { [key: K]: V }>(
+      fn: P
+    ) => (xs: T) => Array<$Refine<V, P, 1>>)
+
+  declare type Filter =
+    & (<K, V, T: Array<V> | { [key: K]: V }>(
+      fn: UnaryPredicateFn<V>,
+      xs: T
+    ) => T)
+    & (<K, V, T: Array<V> | { [key: K]: V }>(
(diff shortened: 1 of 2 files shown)


definitions/npm/ramda_v0.x.x/flow_v0.62.x-/test_ramda_v0.x.x_list.js
```diff
@@ -147,12 +148,34 @@ const str: string = "hello world";
...
+
+    it('filters objects by passing the value to the predicate', () => {
+      const s: { [key: string]: string } = _.filter(x => x === "2", {
+        a: "2",
+        b: "3"
+      });
+    })
+
+    it('refines the element type using the predicate', () => {
+      const notNull = (x): bool %checks => x != null
```
AndrewSouthpaw: Check the reverse, that it errors when `RefineFilter` is not applied.

definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
```diff
@@ -407,13 +407,25 @@ declare module ramda {
...
-    fn: UnaryPredicateFn<V>,
-    xs: T
-  ) => T) &
-    (<K, V, T: Array<V> | { [key: K]: V }>(
-      fn: UnaryPredicateFn<V>
-    ) => (xs: T) => T);
+  // This kind of filter allows us to do type refinement on the result, but we
+  // still need Filter so that non-refining predicates still pass a type check.
+  declare type RefineFilter =
+    & (<K, V, P: $Pred<1>, T: Array<V> | { [key: K]: V }>(
```
AndrewSouthpaw: Nicely spotted on the `$Pred` and `$Refine`, didn't know those existed.

definitions/npm/ramda_v0.x.x/flow_v0.62.x-/test_ramda_v0.x.x_list.js
LoganBarnett: Good catch! Done.

definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
LoganBarnett: Thanks! It's still officially undocumented per https://github.com/facebook/flow/issues/2464 - but [this comment](
https://github.com/facebook/flow/issues/2464#issuecomment-251995029) links to the comment block where there's kind of enough to go off of. To be honest, it took me quite a long time to understand how it was intended to be applied. But hey, we have a living example now!

AndrewSouthpaw:
Still have some travis errors tho, can you look into them?

LoganBarnett:
@AndrewSouthpaw based on the error it looked like a glitch in the run. I did a rebase which ran it again and now it passes ¯\\_(ツ)_/¯

PR #2525, opened 2018-07-22
Key can be a actionCreator also

### LINKED ISSUE #2524: redux-actions handleActions: Key can be actionCreator
```js
 declare function handleActions<State, Action>(
    reducers: {
      [key: string]: Reducer<State, Action> | ReducerMap<State, Action>
    },
    defaultState?: State
  ): ReduxReducer<State, Action>;
```

Over here, The input key can also be ActionCreator; according to https://redux-actions.js.org/introduction/tutorial
```js
const decrement = createAction('DECREMENT');

const reducer = handleActions(
  {
    [increment]: state => ({ ...state, counter: state.counter + 1 }),
    [decrement]: state => ({ ...state, counter: state.counter - 1 })
  },
  defaultState
);
```

https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/redux-actions_v2.x.x/flow_v0.39.x-/redux-actions_v2.x.x.js#L111

Anenth:
fixes #2524

### DIFF
--- a/definitions/npm/redux-actions_v2.x.x/flow_v0.39.x-/redux-actions_v2.x.x.js
+++ b/definitions/npm/redux-actions_v2.x.x/flow_v0.39.x-/redux-actions_v2.x.x.js
@@ -108,7 +108,7 @@ declare module "redux-actions" {
 
   declare function handleActions<State, Action>(
     reducers: {
-      [key: string]: Reducer<State, Action> | ReducerMap<State, Action>
+      [key: string| Function]: Reducer<State, Action> | ReducerMap<State, Action>
     },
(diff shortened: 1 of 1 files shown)


Anenth:
@GAntoine Need some help running the tests.

I removed some `$ExpectError` from `test_handleActions.js`, expecting a test failure after executing the `npm run test` from the `definitions` folder but everything passed.

villesau:
I wouldn't necessarily merge this. This allows you to pass _any_ function to key. That's not how the library is defined to use. More type safe way is to call actionCreatorFunction.toString() in the key.

PR #2526, opened 2018-07-22
firebase_v5.x.x: make auth provider ids static and add GetOptions

eddedd88:


eddedd88:
The AuthProvider types are actually missing more stuff, please give me a few minutes to update the providers with the latest definitions.

eddedd88:
Ok, I updated them to what firebase has in their docs: https://firebase.google.com/docs/reference/js/firebase.auth

PR #2530, opened 2018-07-24
fix type coverage on react-redux's connect

LoganBarnett:
When I upgraded to `react-redux` to 5, we had a drop in type coverage on `connect`. From some experimentation I found that it's resulting from the type import within the libdef. The types imported from redux aren't overly complex and I understand them to be relatively stable, so I've copied them in directly.

I've tested this on my internal repo and found the coverage is applied again. I don't know how to express that as a test in the `flow-typed` suite, however. I tried adding `flow-strict` to the test files as well as `[strict]` settings to `cli/.flow-config`, but still no joy. I'd love any other suggestions anyone can think of.

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.63.0-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.63.0-/react-redux_v5.x.x.js
@@ -1,7 +1,17 @@
-import type { Dispatch, Store } from "redux";
-
 declare module "react-redux" {
   import type { ComponentType, ElementConfig } from 'react';
+
+  // These types are copied directly from the redux libdef. Importing them in
+  // this libdef causes a loss in type coverage.
+  declare type DispatchAPI<A> = (action: A) => A;
+  declare type Dispatch<A: { type: $Subtype<string> }> = DispatchAPI<A>;
+  declare type Reducer<S, A> = (state: S | void, action: A) => S;
+  declare type Store<S, A, D = Dispatch<A>> = {
+    dispatch: D;
+    getState(): S;
+    subscribe(listener: () => void): () => void;
+    replaceReducer(nextReducer: Reducer<S, A>): void
+  };
 
   declare export class Provider<S, A> extends React$Component<{
     store: Store<S, A>,
@@ -57,7 +67,7 @@ declare module "react-redux" {
     S: Object,
     SP: Object,
     RSP: Object,
-    CP: $Diff<OmitDispatch<ElementConfig<Com>>, RSP>,
+    CP: $Diff<ElementConfig<Com>, RSP>,
     ST: {[_: $Keys<Com>]: any}
     >(
     mapStateToProps: MapStateToProps<S, SP, RSP>,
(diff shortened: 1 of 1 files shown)


definitions/npm/react-redux_v5.x.x/flow_v0.63.0-/react-redux_v5.x.x.js
```diff
@@ -57,7 +67,7 @@ declare module "react-redux" {
     S: Object,
     SP: Object,
     RSP: Object,
-    CP: $Diff<OmitDispatch<ElementConfig<Com>>, RSP>,
+    CP: $Diff<ElementConfig<Com>, RSP>,
```
villesau: Why OmitDispatch was removed? Usually react-redux passes it to component so it's not required to pass it explicitly.

villesau: @LoganBarnett ping :)

villesau: @LoganBarnett I tried the version in our codebase and look like it relies that connect passes the dispatch to component in this case.

villesau: @LoganBarnett I guess correct type in omit for dispatch is actually `{dispatch?: Dispatch<*>}`, so this might fix the problem: `declare type OmitDispatch<Component> = $Diff<Component, {dispatch?: Dispatch<*>}>;`

LoganBarnett: @villesau got it - okay I've made that change. How is your codebase encountering this issue? I'd like to capture it in a test if possible.

Also, if you know of ways to capture type coverage itself as a test I'm all ears. So far it feels like I'm spitballing just to get these to add coverage and any changes in the future could just remove them without failing anything.

villesau:
Thanks @LoganBarnett, great catch!

pronebird:
This PR breaks redux-thunk actions for us.  

We used to type `Action` as either being `{ type: ACTION_NAME }` or a `Function`. It's not possible anymore since action is now forced to be `{ type: $Subtype<string> }`.

villesau:
@pronebird try this instead of changes in this PR:

```
  declare type DispatchAPI<A> = Function;
  declare type Dispatch<A: { type: $Subtype<string> }> = DispatchAPI<A>;
  declare type Reducer<S, A> = (state: S | void, action: A) => S;
  declare type Store<S, A, D = Dispatch<A>> = {
    // rewrite MiddlewareAPI members in order to get nicer error messages (intersections produce long messages)
    dispatch: D;
    getState(): S;
    subscribe(listener: () => void): () => void;
    replaceReducer(nextReducer: Reducer<S, A>): void
  };
```

This is what we use at Smartly.io currently.

Note that this PR follows how vanilla redux interface looks like.

Redux-thunk is very problematic from type system perspective as middlewares (thunks included) alters the interface at runtime. See this issue for for further details: https://github.com/flow-typed/flow-typed/issues/574

LoganBarnett:
@villesau do we need to pull in your Smartly defs or is that an override specific to your middlewares?

villesau:
@LoganBarnett that's just an override to make thunks work without too much of a hassle. Note that it's not similar to Redux interface _without_ thunks. I don't have solution which would fit for all: Plain redux does not support function, but thunk middleware adds support for it.

PR #2534, opened 2018-07-26
react-modal: Fix signature of props.onRequestClose

kemuridama:
This PR makes it possible for `props.onRequestClose()` to receive an `Event` object.

`props.onRequestClose()` will be called by `requestClose()` at [here](https://github.com/reactjs/react-modal/blob/master/src/components/ModalPortal.js#L287) and it will be called by handlers of both `onClick` and `onKeyDown`.
Therefore, `props.onRequestClose()` will receive either `SyntheticMouseEvent` or `SyntheticKeyboardEvent`.

### DIFF
--- a/definitions/npm/react-modal_v3.1.x/flow_v0.54.1-/react-modal_v3.1.x.js
+++ b/definitions/npm/react-modal_v3.1.x/flow_v0.54.1-/react-modal_v3.1.x.js
@@ -33,7 +33,7 @@ declare module 'react-modal' {
     },
     appElement?: HTMLElement | string | null,
     onAfterOpen?: () => void | Promise<void>,
-    onRequestClose?: () => void,
+    onRequestClose?: (SyntheticMouseEvent<>) => void,
     aria?: {
       [key: string]: string
     },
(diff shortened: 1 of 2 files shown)


definitions/npm/react-modal_v3.1.x/flow_v0.54.1-/react-modal_v3.1.x.js
```diff
@@ -33,7 +33,7 @@ declare module 'react-modal' {
     },
     appElement?: HTMLElement | string | null,
     onAfterOpen?: () => void | Promise<void>,
-    onRequestClose?: () => void,
+    onRequestClose?: (SyntheticMouseEvent<> | SyntheticKeyboardEvent<>) => void,
```
GAntoine: Does it make sense to generalize this to `SyntheticEvent<>`?

kemuridama: I thought it may as well define concrete type when this handler has clear type to handle...😛
But it's easier to use generic event type for handling events.

Thank you for your advice, I modified it.

PR #2536, opened 2018-07-26
Update jest_v22.x.x.js

Whoaa512:
Fix the `toHaveProperty` input args per https://jestjs.io/docs/en/expect#tohavepropertykeypath-value


AndrewSouthpaw:
Thanks for the first-time contribution!

PR #2538, opened 2018-07-27
directory-tree: change Array to ReadOnlyArray

HsuTing:


AndrewSouthpaw:
Ah, I see it now. I think you don't want to do it for the function though, because otherwise in the exclude you'd need to specify the array as read only, which is silly when you're just passing it in as function params. Thoughts?

AndrewSouthpaw:
Other way: keep `children: $ReadOnlyArray` but change back `exclude: Array`... essentially I'd want to be able to do 

```js
tree('/foo', { exclude: [/bar/] })
```

and not have it blow up. My understanding is that setting `exclude` to be `$ReadOnlyArray`, the above would throw an error.

Does that make sense?

AndrewSouthpaw:
Oh, nice, that's what I was looking for. I was expecting

```js
  it('use exclude array', () => {
    const exclude: Array<RegExp> = [/some_path_to_exclude/];
    exclude.push(/test/);                                   

    const useExcludeArray: dirTreeType = dirTree(__dirname, {
      exclude,   // <-- I thought this would throw an error
    });                                                     
  });     
```

Since that's fine, I'm good with it.

AndrewSouthpaw:
In that case go ahead and add the change to `children` back in as well.

directory-tree__change_array_to_readonlyarray_by_hsuting_ _pull_request__2538_ _flow-typed_flow-typed

HsuTing:
I did 😁.

PR #2542, opened 2018-07-29
Added pipeK definition for Ramda

edahlseng:


definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
```diff
@@ -427,7 +427,7 @@ declare module ramda {
       fn: UnaryPredicateFn<V>,
     ) => (xs: T) => T)
 
-  declare class Monad<T> {
+  declare interface Monad<T> {
```
AndrewSouthpaw: Don't use interfaces when it can be avoided, there's some evidence it slows down flow. Any reason this was changed?

edahlseng:
@GAntoine, I can add tests. Will try to get to it this week, but it may bleed over into next week.

definitions/npm/ramda_v0.x.x/flow_v0.62.x-/ramda_v0.x.x.js
edahlseng: This was changed in order to utilize structural typing instead of nominal typing, so that anything that supports the `chain` method can be used as a monad.

AndrewSouthpaw: Sure... as far as I can tell though, it's not actually being used right now? It would work whether you used it as a `class` or and `interface`?

edahlseng: In the `PipeK` definition that I added, all of the arguments are functions that return an object that implements `Monad<T>`, which is where this interface is primarily used.

AndrewSouthpaw: Oh, got it. That makes sense, thanks for explaining. As much as I use Ramda, there's still a bunch I don't know much about, including monads and pipeK.

AndrewSouthpaw:
Ping?

edahlseng:
Ah, thanks for bumping me on this @AndrewSouthpaw. I have time this evening and will push tests then!

edahlseng:
Tests added. Thanks for the patience!

PR #2546, opened 2018-07-30
Add definition for Faker.js

joeybaker:
Creates a new definition for [faker.js](https://github.com/Marak/faker.js)

### DIFF
--- a/definitions/npm/faker_v4.x.x/flow_v0.25.x-/faker_v4.x.x.js
+++ b/definitions/npm/faker_v4.x.x/flow_v0.25.x-/faker_v4.x.x.js
@@ -1,0 +1,264 @@
+declare module "faker" {
+  declare module.exports: {
+    address: {
+      zipCode: (localeFormat?: string) => string,
+      city: (mustacheTemplate?: string) => string,
+      cityPrefix: () => string,
+      citySuffix: () => string,
+      streetName: () => string,
+      streetAddress: (useFullAddress?: boolean) => string,
+      streetSuffix: () => string,
+      streetPrefix: () => string,
+      secondaryAddress: () => string,
+      county: () => string,
+      country: () => string,
+      countryCode: () => string,
+      state: (useAbbr?: boolean) => string,
+      stateAbbr: () => string,
+      latitude: () => string,
+      longitude: () => string
+    },
+    commerce: {
+      color: () => string,
+      department: () => string,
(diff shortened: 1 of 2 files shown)


definitions/npm/faker_v4.x.x/flow_v0.25.x-/faker_v4.x.x.js
```diff
@@ -0,0 +1,261 @@
...
+      paragraph: (sentenceCount?: number) => string,
+      paragraphs: (paragraphCount?: ?number, separator?: string) => string,
+      text: (times?: number) => string,
+      lines: (lineCount?: number) => string
+    },
+    name: {
+      firstName: (gender?: mixed) => string,
+      lastName: (gender?: mixed) => string,
+      findName: (
+        firstName: string,
```
JelteF: this should be `firstName?: ?string,`, because it's optional

joeybaker: Huh. The docs indicate it's required: https://marak.github.io/faker.js/faker.name.html#-static-findName__anchor, but the code shows that its not: https://github.com/Marak/faker.js/blob/3a4bb358614c1e1f5d73f4df45c13a1a7aa013d7/lib/name.js#L68-L77

Thanks! Will change.

joeybaker: Done.

definitions/npm/faker_v4.x.x/test_faker_v4.x.x.js
```diff
@@ -0,0 +1,311 @@
+import faker from "faker";
```
AndrewSouthpaw: Move the test file into the versioned flow directory.

definitions/npm/faker_v4.x.x/test_faker_v4.x.x.js
```diff
@@ -0,0 +1,311 @@
...
+
+faker.address.zipCode();
```
AndrewSouthpaw: Use `it/describe` blocks for more winning.

joeybaker:
@AndrewSouthpaw thanks for the review! I moved the file and added a few `describe` blocks. I'm not sure there's much benefit to breaking down the tests more.

PR #2547, opened 2018-07-31
Add rsuite v3 libdef

simonguo:


definitions/npm/rsuite_v3.x.x/flow_v0.53.x-/rsuite_v3.x.x.js
```diff
@@ -0,0 +1,1738 @@
...
+  import type {
+    Node,
+    Element,
+    ElementType,
+    Component,
+    ChildrenArray,
+    ComponentType,
+    ElementRef,
+    Ref
+  } from "react";
```
AndrewSouthpaw: This won't work, check the contributing guidelines for alternatives.

simonguo: Has been fixed.

definitions/npm/rsuite_v3.x.x/flow_v0.53.x-/rsuite_v3.x.x.js
```diff
@@ -1,16 +1,7 @@
...
-    ElementType,
-    Component,
-    ChildrenArray,
-    ComponentType,
-    ElementRef,
-    Ref
-  } from "react";
-
-  declare class moment$Moment {}
+  import type { ChildrenArray } from "react";
```
AndrewSouthpaw: Doesn't work, check contributing guidelines.

simonguo: The documentation states: "Do not import types from other libdefs" ， So I changed to private React types.  

But there is no `React$ChildrenArray` in the private type. I would love to know how react-bootstrap works： 
https://github.com/flow-typed/flow-typed/blob/b3b854c87f74b2fbe48d18e262c142f9d890727b/definitions/npm/react-bootstrap_v0.32.x/flow_v0.53.x-/react-bootstrap_v0.32.x.js#L2

AndrewSouthpaw: It doesn't, I think a lot of libdefs did this practice before we realized that importing types from other libdefs will `any` them.

I'd either copy the definition of `ChildrenArray`, or just create the type as `any` locally so we're more explicitly aware of what we're doing.

Make sense?

simonguo: I understand.
I have remove other libdefs, can you check if there are any problems?

AndrewSouthpaw:
Sorry, just been busy... looking now.

PR #2551, opened 2018-07-31
fix(lodash_v4.x.x): fix bad types for omitBy/pickBy

jedwards1211:
Fixes problems like
```
Cannot call `pickBy` because undefined [1] is incompatible with object type [2] in type argument `T`.
```

Somebody constrainted `T: void | null` for the case where the input object is not defined, but that means it won't accept a predicate that takes a defined object.


jedwards1211:
done, assuming the tests pass

AndrewSouthpaw:
Nope. Instructions for running tests are in the contributing doc.

jedwards1211:
@AndrewSouthpaw okay, it's fixed now, looks like the CI had unrelated transient failures...

GAntoine:
., @AndrewSouthpaw you're good to move forward w

PR #2563, opened 2018-08-02
fix(react-apollo): remove | void from data type

jedwards1211:
react-apollo will never pass undefined (https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L377)


TLadd:
There is actually still a case where undefined is passed as data: https://github.com/apollographql/react-apollo/issues/1977

If you treat data as never being undefined, you'll end up getting runtime exceptions if your query fails on its first call. This line is why: https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L395 `{}.data`

jedwards1211:
Typing stuff gets to be such a god-awful mess sometimes...

PR #2565, opened 2018-08-03
Added client to MutationResult

jonespen:
According to https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L17

cc @budde377 

Edit: Hm, tests failed for node8, but can't see why. Is there any way to trigger another build?


definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/test_react-apollo_v2.x.x.js
```diff
@@ -525,6 +525,7 @@ describe("<Mutation />", () => {
           }
           const d1: Res = data;
           const s: string = data.res;
+          client;
```
jonespen: this is kind of awkward, but seems to do the trick ;)

PR #2566, opened 2018-08-03
[material-ui] Improve CircularProgress definition

jmarceli:
According to the official [MUI docs](https://material-ui.com/api/circular-progress/), CircularProgress component should have a different set of allowed values for `color` and `mode` props.


jmarceli:
Rewritten and reopened as [PR 2567](https://github.com/flow-typed/flow-typed/pull/2567)

PR #2567, opened 2018-08-03
[material-ui] Improve CircularProgress definition

jmarceli:
According to the official [MUI docs](https://material-ui.com/api/circular-progress/), CircularProgress component should have a different set of allowed values for `color` prop and may have an optional `variant` prop.
I'm not sure about `mode` which seems to be a legacy prop. Maybe it should be removed from a type definition?


GAntoine:
@jmarceli This needs a rebase, I just merged in another change to material-ui/core

jmarceli:
Fixed (I hope).

PR #2570, opened 2018-08-03
Add libdef for js-yaml

jirutka:
This libdef is based on [TS type definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/js-yaml/index.d.ts).

I'm not sure where to use `mixed` and where `any` (I know the difference between them, but I don't know js-yaml library so well).


GAntoine:
This LGTM. For reference, `any` can flow into other types, but `mixed` cannot, since it's not a subtype of all types, like `any`. There's a SO question on it [here](https://stackoverflow.com/a/30023184/1179430).

Issue #2561, opened 2018-08-02
[Winston 3]: Shorthand log methods optionally take additional parameters


carmanchris31:
https://github.com/flow-typed/flow-typed/blob/14afab0c3b6b21a809ae2f06f757a5b168660b0e/definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js#L51

Minor issue, but since flow is very strict... winston log methods take additional parameters. Specifically, `logger[level](...args)` is converted to `logger.log(level, ...args)` here:
https://github.com/winstonjs/winston/blob/6d01c78/lib/winston/create-logger.js#L62

And, in turn, this optionally supports string interpolation and custom meta as documented here:
https://github.com/winstonjs/winston/blob/6d01c78/lib/winston/logger.js#L108-L110

and demonstrated here:
https://github.com/winstonjs/winston#string-interpolation

Leaving this here in case there are any takers. I may be able to submit something later if not.

EDIT:

TypeScript definitions for reference:
https://github.com/winstonjs/winston/blob/master/index.d.ts#L53-L71

I received errors on doing local `yarn install` inside definitions & cli and `./build_and_test_cli.sh` unfortunately so there seems to be somewhat of a barrier of entry for contributing

PR #2572, opened 2018-08-03
[Winston] Add extra parameters to log methods

carmanchris31:
For https://github.com/flow-typed/flow-typed/issues/2561

Practically lifted from https://github.com/winstonjs/winston/blob/master/index.d.ts#L51-L71 but could use some additional eyes/verification.


GAntoine:
@carmanchris31 You'll need to fixup the specs to match this change.

carmanchris31:
@GAntoine Turns out the TS definitions were incorrect. Through investigating the winston source code and some local tests I was able to pull it back to a couple minimal changes that pass the existing specs.

Will rebase and squash commits if desired.

PR #2573, opened 2018-08-03
fix: react-apollo: don't constrain ignored return type of user-passed…

jedwards1211:
… functions

@TLadd please don't ever stipulate `void` for the return type of a user-passed function, *except, perhaps, if it's necessary to ensure that the function completes synchronously.*

I have a lot of invalid errors in my code because of this.

It's perfectly reasonable to pass an async function for many callbacks, and constraining the return type to `void` in those cases causes a pointless error when an async function is passed (since async functions return a `Promise`.

It also causes annoying errors when passing an arrow function that calls something else which happens to return non-void.
I also often call `apollo-magic-refetch` in my `update` functions.  It returns a `Promise` which obviously is ignored by `Mutation`, but let's not have this cause a flow error please.

If whatever calls the user-defined function doesn't care about its result or synchronous completion, the return type should be `any`.

The use of `| void` also seems suspiciously prevalent in these types.  Are you sure that in all of those cases the value can be `undefined` but not `null`?

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -60,7 +60,7 @@ declare module "react-apollo" {
       link: ApolloLink,
       queryDeduplication?: boolean,
       store: DataStore<TStore>,
-      onBroadcast?: () => void,
+      onBroadcast?: () => any,
       ssrMode?: boolean
     }): this;
 
@@ -186,7 +186,7 @@ declare module "react-apollo" {
       document: DocumentNode,
       variables: any,
       updateQueries: { [queryId: string]: QueryWithUpdater },
-      update: ((proxy: DataProxy, mutationResult: Object) => void) | void,
+      update: ((proxy: DataProxy, mutationResult: Object) => any) | void,
       optimisticResponse: Object | Function | void
     }): void;
     markMutationResult(mutation: {
@@ -195,7 +195,7 @@ declare module "react-apollo" {
       document: DocumentNode,
       variables: any,
       updateQueries: { [queryId: string]: QueryWithUpdater },
-      update: ((proxy: DataProxy, mutationResult: Object) => void) | void
+      update: ((proxy: DataProxy, mutationResult: Object) => any) | void
     }): void;
     markMutationComplete({
       mutationId: string,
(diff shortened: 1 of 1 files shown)


TLadd:
The entire libdef is a port from typescript (which apollo-client and react-apollo are written in), where the return type for all these functions are typed as void. For instance: https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L68

jedwards1211:
I get it, but there's no reason to assume Apollo's defs are perfect.  I don't even know if TS vehicle the same way as flow in this regard.

TLadd:
For your use case, I would generally favor using `mixed` vs `any`. Some of the ones you are changing are not really user-passed functions though. 

subscribeToMore, for instance, is something react-apollo passes back to you via the children render prop function. By changing its return type to any, it allows for doing something like 

```
<Query ... >
  {({ subscribeToMore }) => {
    subscribeToMore({ ... }).then(() => "Whoops")
  }}
</Query>
```

which should error. Leaving that as void seems better to me. A lot of the types you're changing also are internal details that I wouldn't think a consumer of react-apollo would interact with directly. QueryManager, DataStore, ZenObservableSubscriber, Observable, Transaction, CacheWatchCallback, are all pretty deep into apollo-client/apollo-cache/apollo-link, so I'm not sure if there's any value in changing them.

The three places I do think it makes sense to change from void -> mixed are 
1. onCompleted of Mutation
2. onError of Mutation
3. MutationUpdaterFn

as they are props supplied by a user and it is fine if they return a value even though nothing is done with that value. If you do still want to change the others, please post use-cases where you have type errors that you don't think are valid.

jedwards1211:
What you say about `subscribeToMore` makes sense, are you saying I actually changed that or just using it as an example?  I couldn't find anywhere that I changed the return type of `subscribeToMore`, and actually it looks like it returns `() => void` (un unsubscribe function) rather than `void`.

If you do see places I changed the return type of a non-user-provided function, please do let me know which specific ones and I'll fix it.  It looks like I even missed things like ZenObservableObserver.

I agree QueryManager, DataStore, ZenObservableSubscriber etc. are not likely to be used by developers, but what would we gain by keeping them the way they are?  On the off chance that someone does use them, the void return types could end up wasting their time, which would be prevented if we go ahead and change them.

TLadd:
Yeah, it's quite probably true that it would make sense to just replace some of those internal types with any to reduce the complexity of the libdef; there are certainly some that I don't think provide any real-world value. It is important to maintain types for things like the client and cache, since those get used directly by users via props (update, ApolloConsumer, etc), but I'm sure a lot of the Observable stuff could go.

jedwards1211:
Yeah, some might not even be public API...though having ported the defs for Sequelize, I can say, there can easily be gray areas...numerous fields that weren't publicly documented in Sequelize have come in handy

GAntoine:
Merging this is based on the conversation above and TLadd's approval

PR #2578, opened 2018-08-04
Use /bin/sh instead of /bin/bash for better portability

jirutka:
All three scripts are POSIX-sh compliant, so it's unnecessary to require
specifically bash, they works on any POSIX shell. Not every unix system
comes with bash preinstalled and even less install bash executable into
/bin. /bin/sh is defined by POSIX, so more portable.

### DIFF
--- a/build_and_test_cli.sh
+++ b/build_and_test_cli.sh
@@ -1,5 +1,5 @@
-#!/bin/bash
-set -o errexit
+#!/bin/sh
+set -e
 
 cd definitions
 yarn install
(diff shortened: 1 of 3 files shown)


GAntoine:
Possibly dumb question, but is this windows shell friendly?

jirutka:
What do you mean by Windows shell? PowerShell and cmd.exe are not POSIX shells, so no. However, you have `*.cmd` scripts in the repo for Windows.

GAntoine:
And that's why I don't ask questions late at night. 😅

Issue #2478, opened 2018-07-05
js-cookie getJSON type


jmarceli:
Why `js-cookie.getJSON` flow type is `mixed`?
With **lodash pick** I get:
```
Cannot call pick because mixed [1] is incompatible with object type [2].
```

Does **getJSON** return type shouldn't be `object` or `any`?
I'd be happy to get some opinions about that.

AndrewSouthpaw:
You could probably add generics, rather than `mixed`, so require add a type for the output of `getJSON`.

AndrewSouthpaw:
Nope, nevermind. I opened an issue with Flow: https://github.com/facebook/flow/issues/6597. In the meantime, pretty much your only other options is to just explicitly make those return types `any` so you can do whatever you want with it.

AndrewSouthpaw:
Go ahead and make a PR for it; having them typed as `mixed` is probably useless and just requires all consumes to cast first to `any` so it's just a frustrating experience.

AndrewSouthpaw:
You could also try borrowing from the `axios` libdef, that might do what you want.

PR #2582, opened 2018-08-06
[js-cookie]: change getJSON type to Object

jmarceli:
It seems that `Object` is a correct return type for js-cookie `.getJSON` method as it will always try to return a parsed JSON object [source](https://github.com/js-cookie/js-cookie/blob/master/src/js.cookie.js#L147).
While getting a string with `.getJSON` [might be possible](https://github.com/js-cookie/js-cookie/blob/master/src/js.cookie.js#L128) I don't think it is a valid return type as for non JSON cookies `.get` method should be used.

This PR is related to the issue #2478


GAntoine:
It seems the specs are failing with flow v0.78, you'll need to address that before I can merge this in.

PR #2584, opened 2018-08-06
antd: add more typing

bwlt:


bwlt:
The API of this module is very vast. Cover all the API is too huge, so I will add more typing every time I need something more  😅

PR #2590, opened 2018-08-07
fix(react-apollo): export ApolloClient

jedwards1211:
since it's not compatible with the one from the apollo-client typedefs, we might as well use this one wherever we're using react-apollo...


jedwards1211:
@AndrewSouthpaw @GAntoine hey guys, Flow 0.70 seems to be segfaulting in CI often.  Can we just disable that version?

PR #2592, opened 2018-08-08
fix(ramda): allow pipeK as inferred type

naterious:


retyui:
This method `pipeK()` will `Deprecated since v0.26.0`

retyui:
@AndrewSouthpaw
I think you can close this PR )

AndrewSouthpaw:
Well, it doesn't make it any worse, so I'm fine with it in this case. 

@LoganBarnett anything to add?

LoganBarnett:
@AndrewSouthpaw I don't see any harm in adding it either.

Re: Deprecation I did notice that we have a few functions in the libdef that have been deprecated _and removed_ from Ramda. We should consider splitting the Ramda versions. That said, the matrix of files we have to support with just a "head" version of Ramda is getting cumbersome. I'll write up a ticket for it and elaborate more there.

AndrewSouthpaw:
K, LGTM, but conflicts will need to be resolved before continuing.

AndrewSouthpaw:

PR #2593, opened 2018-08-08
feature(memory-cache): add basic memory cache definition

naterious:


GAntoine:
Hi @naterious, you'll need to add some specs to this definition before we can accept it. You can take a look at some of the other small libraries for examples of specs.

PR #2595, opened 2018-08-08
fix(fluture): fix type definitions for fluture

naterious:


GAntoine:
Could you add some specs to the test file to cover your changes?

GAntoine:
Closing as fluture has change quite a bit, and new versions are available.

PR #2598, opened 2018-08-08
Move creation of .flow-bins-cache to earlier in test cycle

AlbinTheander:
Before this commit, the folder was created in runTestGroup (if it did
not exist before). However, it was accessed before that, making the
tests crash if the folder was not manually created.

This commit moves the creation to the first thing in the test run.

This is only a problem when you check out and run the build the first time, but it took me a while to figure out what was wrong.

I'm using OSX, maybe the folder creation works differently on other platforms.

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -492,10 +492,6 @@ async function runTestGroup(
     );
   }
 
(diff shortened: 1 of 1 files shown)


jirutka:
> I'm using OSX, maybe the folder creation works differently on other platforms.

No, it's broken even on Linux.

PR #2602, opened 2018-08-09
Update @storybook/react_v3.x.x libdef

Vages:
Downloading the libdefs made my current project crash because it lacked two things:

1. The exported `forceReRender` function
2. A first argument to the `.add`-method of the Story interface.

I've added these.

### DIFF
--- a/definitions/npm/@storybook/react_v3.x.x/flow_v0.28.x-/react_v3.x.x.js
+++ b/definitions/npm/@storybook/react_v3.x.x/flow_v0.28.x-/react_v3.x.x.js
@@ -1,13 +1,14 @@
 type NodeModule = typeof module;
 
 declare module '@storybook/react' {
-  declare type Renderable = React$Element<any>;
-  declare type RenderFunction = () => Renderable;
+  declare type Renderable = React$Node;
+  declare type RenderFunction = (context: Context) => Renderable;
+  declare type Context = { kind: string, story: string };
 
   declare type StoryDecorator = (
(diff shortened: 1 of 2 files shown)


Vages:
I see that there's some error in Travis when it is run on Node v8. I'd like to fix this, but I don't have any clue about where to start. Does anyone have any tips?

image

GAntoine:
@Vages You can ignore the segfault, I'll be taking care of it later this afternoon.

definitions/npm/@storybook/react_v3.x.x/flow_v0.28.x-/react_v3.x.x.js
```diff
@@ -1,13 +1,14 @@
 type NodeModule = typeof module;
 
 declare module '@storybook/react' {
-  declare type Renderable = React$Element<any>;
-  declare type RenderFunction = () => Renderable;
+  declare type Renderable = React$Node;
```
pascalduez: `React$Node` is only available from Flow `>=0.53.0`, so this breaks all previous versions. 
The libdef should be cloned and tweaked for later versions:
`definitions/npm/@storybook/react_v3.x.x/flow_v0.53.x-/react_v3.x.x.js `

pascalduez: I would say to revert it for the `flow_v0.28.x-v0.52.x` and can stay as is for the `flow_v0.53.x`.

PR #2603, opened 2018-08-09
Make CookiesSetOptions with defaults optional

TomPridham:
the properties on `CookiesSetOptions` that come with defaults should be optional

### DIFF
--- a/definitions/npm/koa_v2.0.x/flow_v0.47.x-/koa_v2.0.x.js
+++ b/definitions/npm/koa_v2.0.x/flow_v0.47.x-/koa_v2.0.x.js
@@ -188,15 +188,15 @@ declare module 'koa' {
   };
   // https://github.com/pillarjs/cookies
   declare type CookiesSetOptions = {
-    maxAge: number, // milliseconds from Date.now() for expiry
-    expires: Date, //cookie's expiration date (expires at the end of session by default).
-    path: string, //  the path of the cookie (/ by default).
     domain: string, // domain of the cookie (no default).
-    secure: boolean, // false by default for HTTP, true by default for HTTPS
-    httpOnly: boolean, //  a boolean indicating whether the cookie is only to be sent over HTTP(S),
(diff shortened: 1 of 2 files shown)


GAntoine:
You'll need to fix the issue this change introduces:

```
Cannot extend `net$Server` [1] with `Server` because in property `listen`:
    - Either number [2] is incompatible with undefined [3] in the first argument.
    - Or string [4] is incompatible with undefined [3] in the first argument.
    - Or object type [5] is incompatible with undefined [3] in the first argument.
   
      koa_v2.0.x.js:8:3
              v----------------------------------------
         8|   declare class Server extends net$Server {
         9|     listen(port: number, hostname?: string, backlog?: number, callback?: Function): Server,
        10|     listen(path: string, callback?: Function): Server,
        11|     listen(handle: Object, callback?: Function): Server,
        12|     close(callback?: Function): Server,
        13|     maxHeadersCount: number,
        14|     setTimeout(msecs: number, callback: Function): Server,
        15|     timeout: number,
        16|   }
              ^
   
   References:
      koa_v2.0.x.js:8:32
         8|   declare class Server extends net$Server {
                                           ^^^^^^^^^^ [1]
      koa_v2.0.x.js:9:18
         9|     listen(port: number, hostname?: string, backlog?: number, callback?: Function): Server,
                             ^^^^^^ [2]
      <BUILTINS>/node.js:1374:17
      1374|   listen(port?: number, hostname?: string, backlog?: number, callback?: Function): net$Server;
                            ^^^^^^ [3]
      koa_v2.0.x.js:10:18
        10|     listen(path: string, callback?: Function): Server,
                             ^^^^^^ [4]
      koa_v2.0.x.js:11:20
        11|     listen(handle: Object, callback?: Function): Server,
                               ^^^^^^ [5]
```

TomPridham:
i kept getting permission errors for some versions of flow when trying to run the tests, but i think that should fix it.

PR #2604, opened 2018-08-09
Fix concatMap optional parameters

hally9k:
Apologies if this is incorrect but I thought I'd have a go at it myself rather than drop another issue into the mix.

Makes concatMap's resultSelector param optional as per http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-concatMap

### DIFF
--- a/definitions/npm/rxjs_v6.x.x/flow_v0.34.x-/rxjs_v6.x.x.js
+++ b/definitions/npm/rxjs_v6.x.x/flow_v0.34.x-/rxjs_v6.x.x.js
@@ -898,7 +898,7 @@ declare module "rxjs/operators" {
     ): rxjs$Observable<T> => rxjs$Observable<U>;
     concatMap<+T, U, V>(
       f: (value: T, index: number) => rxjs$ObservableInput<U>,
-      resultSelector: (
+      resultSelector?: (
         outerValue: T,
         innerValue: U,
         outerIndex: number,


hally9k:
Based on previous commits to this file, maybe you might know who might be a good reviewer for this? @rgbkrk

GAntoine:
> Apologies if this is incorrect but I thought I'd have a go at it myself rather than drop another issue into the mix.

❤️

GAntoine:
Could you add a spec to cover the change? I've set this on hold until @rgbkrk or someone else finds some time to review it.

rgbkrk:
This is correct, matches the rxjs implementation.

hally9k:
Unfortunately I seem to be hitting a wall with the spec, first timer here.
If I pull `flow-typed/flow-typed` and run `./build_and_test_cli.sh` with node 9.11.2 It runs with a minor complaint but passes the tests.
But if I pull my fork `hally9k/flow-typed` and run `./build_and_test_cli.sh` with node 9.11.2 it borks out and fails on:

```
  ● run-tests (command) › regression 1385 › console logs about unused suppression

    expect(number).toContain(value)

    Expected number:
      15
    To contain value:
      "Unused suppression"
```

GAntoine:
Are there any other changes to your fork that aren't covered by this PR? If that's not the case, I'd nuke the fork, re-fork from the latest version of master and open another PR with this change.

PR #2605, opened 2018-08-09
Fix Bluebird.all for flow 0.76+

dlam:
Similar to https://github.com/flow-typed/flow-typed/pull/2509, but will be actively trying to get the travis checks to pass.

This solves an issue with stricter rules on generics introduced in flow 0.76.

Bigger question: Is there any reason why the arg for Bluebird.all is typed as `T | $Promisable<T>` instead of `$Promisable<T>` since `$Promisable` is already a type alias for `Promise<T> | T`?


dlam:
It actually kind of looks like the build is failing on master or the tests are somewhat flaky? Is this expected or should I just retry?

GAntoine:
Ignore the failing specs for now, I'll take care of them later today. This LGTM, and does fix bluebird for 76+.

> Bigger question: Is there any reason why the arg for Bluebird.all is typed as T | $Promisable<T> instead of $Promisable<T> since $Promisable is already a type alias for Promise<T> | T?

Might make sense to open an issue and ping whoever wrote the libdef, or made that change.

PR #2606, opened 2018-08-09
react-redux: Make leaked types less likely to conflict with user types

matthewwithanm:
These are leaking global `Store` and `Dispatch` types currently and ("Store" at least) is a pretty common type name in the Nuclide code base. As a result, if we forget to import our `Store` type, we get this libdef's `any`-typed Store. This makes the names a little more esoteric so that they're less likely to conflict.

Originally, I had tried to just move the imports into the declaration block but that caused errors because redux couldn't be found.


matthewwithanm:
Oh, looks like this makes the tests fail. What's the recommended way to deal with this without leaking globals? Or should we just make sufficiently esoteric aliases and leak those?

villesau:
@matthewwithanm Can you pull the latest version of react-redux? It should also solve this problem.

villesau:
I'm gonna close this now as #2530 should have tackled the same problem. Feel free to make a new PR if you don't think that's the case. Thanks for the contribution anyways!

PR #2607, opened 2018-08-10
Add flag property "ordered" to chai

s-diez:
http://www.chaijs.com/api/bdd/#method_ordered


GAntoine:
@s-diez You'll need to add a spec to cover the addition.

s-diez:
@GAntoine Thanks for the reminder. I added a check for the new property.

PR #2620, opened 2018-08-13
Fixes history.replace definition

pastelsky:
According to the official docs, the second parameter is optional. https://github.com/ReactTraining/history#navigation


pastelsky:
@AndrewSouthpaw Done!

PR #2621, opened 2018-08-14
material-ui export props proposal

futuranna:
I would like to export prop definitions of components as well. 
I used Grid as an example.

Any one interested?


AndrewSouthpaw:
LGTM. I'd recommend you CC some other contributors and get their input, just git blame and see who's given to this libdef.

Issue #2623, opened 2018-08-14
Changes in `pickBy` method of `lodash` definition to allow numbers are breacking


tyger:
Changes that are introduced by this PR https://github.com/flow-typed/flow-typed/pull/2613 are actually breaking changes. As makes object `pickBy` specific to object with keys of union type `string|number`.
So it makes `flow` to complain about the next code:
```javascript
type InterestFull = {
    id: string,
    name: string,
}
type DictionaryStr<T> = { [string]: T };
let interests = {
   'one': { id: 'one', name: '' }
}
pickBy(interests, interest => !!interest.name)
```
```
Cannot call pickBy because string [1] is incompatible with number [2] in the indexer property's key of type argument T.

     public/scripts/xxx.js
      457│         let interestsWithDetails = pickBy(interests, interest => !!interest.name);

     flow-typed/npm/lodash_v4.x.x.js
 [2] 1221│     pickBy<A, T: { [id: string | number]: A }>(

     public/flowTypes.js
 [1]   13│ type DictionaryStr<T> = { [string]: T };
```

tyger:
#2624 That PR introduces more correct way to solve the same task.

tyger:
#2624 PR has been merged. Issue should be fixed after next release of this package.

PR #2624, opened 2018-08-14
Fix `pickBy` and `omitBy` method definitions from `lodash` so that they correctly accept objects with `string` and `number` typed ids.

tyger:
In response to the next issue #2623

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -679,7 +679,7 @@ export function setup(yargs: Yargs) {
 }
 
 export async function run(argv: Args): Promise<number> {
-  if (!(await fs.exists(BIN_DIR))) {
+  if (!await fs.exists(BIN_DIR)) {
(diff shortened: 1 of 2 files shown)


davidgomes:
Great find. I can confirm that my PR #2613 was indeed correct and this is the right way to do it.

cli/src/commands/runTests.js
```diff
@@ -679,7 +679,7 @@ export function setup(yargs: Yargs) {
 }
 
 export async function run(argv: Args): Promise<number> {
-  if (!(await fs.exists(BIN_DIR))) {
+  if (!await fs.exists(BIN_DIR)) {
```
tyger: This change because of `eslint` complaint about extra parenthesis.

tyger:
@davidgomes  Thank you.

villesau:
Thanks for the contributions @tyger and @davidgomes !

Issue #2518, opened 2018-07-19
Regression: commit c5d8a breaks react-redux typedefs when using higher order components


arekkas:
Commit c5d8a135ce136817172c7d31e36225b223c09e17 changed something in the way higher order components work and seems to be breaking how combining multiple higher order components work(ed).

The error I'm experiencing is the following:

```
Cannot create `App` element because props [1] is incompatible with empty [2].

   src/index.js:15:7
   15|       <App />
             ^^^^^^^ [1]

References:
   flow-typed/npm/react-redux_v5.x.x.js:63:5
   63|     CP: $Diff<OmitDispatch<ElementConfig<Com>>, RSP>,
           ^^ [2]
```

This is achieved by simply combining two HOCs:

```
export default withRouter(connect(mapStateToProps)(App));
```

I created a minimalistic app based on CRA which can be found here: https://github.com/arekkas/flow-redux

To reproduce the error run

```
$ git checkout ...
$ npm i
$ npm run flow
```

This works fine when reverting to commit https://github.com/flow-typed/flow-typed/commit/cf33ff8762a0f78053aeaf6c390d1f675fefcf08#diff-7439332cbe96ef2bfd6a520ff5bc33ea

I'm also pinging the authors of https://github.com/flow-typed/flow-typed/commit/c5d8a135ce136817172c7d31e36225b223c09e17 as they probably know best what's going on: @maximeg @villesau

villesau:
HAH! I think I fixed it, and it actually wasn't react-redux which failed in this case :) https://github.com/flow-typed/flow-typed/issues/2518

PR #2629, opened 2018-08-15
fix react-router-dom and react-redux type compatibility

villesau:
Fixes https://github.com/flow-typed/flow-typed/issues/2518

Should not have other side effects, apparently it matters wether you use flow "private" internal types or import from react.

Verified the compatibility with this repo: https://github.com/arekkas/flow-redux

It's a bit tricky to provide good test for this, the whole compatibility error is a bit weird.


villesau:
hmm looks like flow 0.70.0 segfaults. @GAntoine thoughts?

GAntoine:
Rebase on master, that'll should fix it.

villesau:
Thanks! I Thought this was a fresh issue.

PR #2634, opened 2018-08-16
Add $ReadOnlyArray to Relay's Variable union type

sgwilym:
This allows types generated by relay-compiler to be used with this libdef.


sgwilym:
Okay I have to come clean here and admit I don't know what I'm doing. This change seemed to fix the failing flow tests in our repo, and I don't see anything like the errors here. Could someone with some more knowledge help me out?

PR #2637, opened 2018-08-16
Add missing definition for rxjsv6

sergio-toro:
Adds missing definition for **fromPromise** `import { fromPromise } from 'rxjs';`


GAntoine:
Could you add a spec that covers the addition?

GAntoine:
pink @sergio-toro

PR #2641, opened 2018-08-17
[lodash 4.x] 'keyBy': do not return map of maybe types

### LINKED ISSUE #924: [lodash v4.x.x] keyBy returns  a map of maybe type values
lodash `keyBy` returns  a [map of maybe type](https://github.com/flowtype/flow-typed/blob/master/definitions/npm/lodash_v4.x.x/flow_v0.47.x-/lodash_v4.x.x.js#L200) values which makes it incompatible with maps that expect no undefined values.

There seems to be no workaround for this other than making all maps in my code contain maybe types or short circuiting return types of `keyBy` with `any`. Either of those are obviously undesired.

So my question is if the return type in signature is correct?
Since `keyBy` iterates over the values of array/collection of `T` it should not be possible to end up with undefined values unless `T` is a maybe type itself.

For reference, [TypeScript definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/index.d.ts#L8179) of `keyBy` also returns a dictionary of `T` and not `?T`.

Current Flow definition:
```
// flow-typed signature: b611e91eea833723f66a55f458ee8e27
// flow-typed version: 13ceae43c9/lodash_v4.x.x/flow_>=v0.47.x
/* ... */
declare class Lodash {
/* ... */
    keyBy<T, V>(array: ?Array<T>, iteratee?: ValueOnlyIteratee<T>): {[key: V]: ?T};
    keyBy<V, A, I, T: {[id: I]: A}>(object: T, iteratee?: ValueOnlyIteratee<A>): {[key: V]: ?A};
/* ... */
}
```

theres:
As stated in #924 and in comment in #681, current type definition of lodash `keyBy` is invalid. `fp` version is already valid.

Resolve  #924

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.28.x-v0.37.x/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.28.x-v0.37.x/lodash_v4.x.x.js
@@ -373,7 +373,7 @@ declare module "lodash" {
       path: ((value: any) => Array<string> | string) | Array<string> | string,
       ...args?: Array<any>
     ): Array<any>;
-    keyBy<T, V>(array: ?Array<T>, iteratee?: Iteratee2<T, V>): { [key: V]: ?T };
+    keyBy<T, V>(array: ?Array<T>, iteratee?: Iteratee2<T, V>): { [key: V]: T };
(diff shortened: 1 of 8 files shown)


villesau:
Thanks!

PR #2642, opened 2018-08-17
[redux-mock-store] update Store type

theres:
After changes in #2530 it appears that Store type generated by previous definition of redux-mock-store is invalid. This PR makes it compatible again.


villesau:
Thanks @theres !

PR #2645, opened 2018-08-18
Fix and improve storybook/react definitions

### LINKED ISSUE #2609: Storybook `addDecorator` was broken in #2602
#2602

```
Error ---------------------------- src/components/Common/Tooltip/index.stories.js:17:46
Cannot call `story` because function [1] requires another argument.
   src/components/Common/Tooltip/index.stories.js:17:46
   17|   .addDecorator(story => <div style={style}>{story()}</div>)
                                                    ^^^^^^^
References:
   flow-typed/npm/@storybook/react_v3.x.x.js:8:33
    8|   declare type RenderFunction = (context: Context) => Renderable;
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ [1]
```

@Vages maybe make `context` as optional?

```js
declare type RenderFunction = (context?: Context) => Renderable;
```

pascalduez:
- Following of #2602 
- Should fix #2609 
- Some improvements


villesau:
I think storybook-host definitions starts to fail with these: https://github.com/flow-typed/flow-typed/tree/master/definitions/npm/storybook-host_v4.x.x/flow_v0.54.1-

Also, in overall it's better to prefer `import type { Element } from 'react'` over React$Element.

pascalduez:
> I think storybook-host definitions starts to fail with these

Then they will be updated as well.
The current state of all Storybook and plugs defs is quite poor, and recent updates broke several of them. So I spent quite some time researching and testing.

> Also, in overall it's better to prefer import type { Element } from 'react' over React$Element.

Really? I've seen several PRs here reverting it from import to global. 
I'm fine with both, but we should maybe most clearly state on the desired form.

villesau:
@pascalduez Sorry, meant actually `ElementType`, not `Element` to be precise. I have seen some differences on compatibility when using flow internal types vs imported types. And to be honest, I've not cracked the reason why they behave differently in some cases. Here is an example where they behaved differently: https://github.com/flow-typed/flow-typed/pull/2629

When using flow builtin types, it's possible to import them (and React is builtin). When using other libdefs, it's unfortunately not yet supported (it's a CI & CLI issue, introducing type-dependencies to flow-typed is pretty big task).

Personally I would recommend to import React types when possible.
Here is more about it: https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md

Vages:
Was about to finally sit down and fix this myself, but I think you have done a much better job than I would. Looks great. 👍 

Have you found any authoritative documentation on the Storybook API? I could not find any official docs, and their examples are few and far between.

pascalduez:
@Vages 

> Have you found any authoritative, extensive documentation on the Storybook API?

There's not indeed! So parsing the source code to verify (a bit tedious since its heavily modularized) and also a lot of testing (unit, manual) was required.

pascalduez:
@villesau 

> I have seen some differences on compatibility when using flow internal types vs imported types. And to be honest, I've not cracked the reason why they behave differently in some cases.

Once again, I'm not against using the imported types. But the lack of tangible material and explanation is a bit dubious. The types are in only one place after all. 
`declare export type ElementType = React$ElementType;` 
Anyway, can this be achieved in a later PR? So I tackle all plugins at once.

> Sorry, meant actually ElementType, not Element to be precise.

That's what I used first, but is actually not working as expected and a bit misleading actually. 
So `Element` is what we need in  this case. 

```
/**
 * Type of a React element. React elements are commonly created using JSX
 * literals, which desugar to React.createElement calls (see below).
 */
```

https://github.com/facebook/flow/blob/master/lib/react.js#L156-L178

villesau:
@pascalduez Yeah I totally understand that it sounds dubious. And like said, its not clear to me either why there might be differences as clearly there shouldn't be. The main point still is that if something can be imported, it should be. This can be merged without doing the change though. Will you look into `storybook-host` definition still?

AndrewSouthpaw:
Seems like @villesau has this one, leaving it to him. :-)

PR #2646, opened 2018-08-19
Fix and improve storybook/addon-a11y definitions

pascalduez:
Following of #2645


pascalduez:
@AndrewSouthpaw There is a test file actually, and already well populated.

pascalduez:
@AndrewSouthpaw Sort of requires #2645 to get along. Also #2647 while your at it ;)

PR #2648, opened 2018-08-20
Restore moment defintion for older flow versions

pascalduez:
In #2126 the moment definition was renamed and the minimum Flow version bumped for no reason, no specific new feature was used. This has the consequence of making the moment definition unavailable for Flow lower than 0.63 users. See https://github.com/flow-typed/flow-typed/pull/2126#issuecomment-414279605

This rename restore the previous version.

It was tested in a Flow 0.52 env with no error.

Is there a  way for the `run-tests` command to target a specific Flow version? (Older than the last 15 versions).


cobrabr:
I'm getting the following Flow error since this PR was merged, while trying to use `fn`:

```
import moment, { fn as momentProto } from 'moment';
                 ^^ Named import from module `moment`. This module only has a default export. Did you mean `import fn from ...`?
```

Prior to this (this morning, even), that code was working properly.

GAntoine:
@cobrabr What flow version are you on?

cobrabr:
@GAntoine -- we're using Flow 0.61 currently. Taking a look at the PR now.

pascalduez:
> we're using Flow 0.61 currently. Taking a look at the PR now.

This means prior to this MR you had no types at all :) Because the minimum version to install it was `0.63``. So `moment`  was resolving to `any`. 
Now you have types, and `moment` does not export a `fn` function. 

There's no changes in the defs here, just a rename, so it can't be the issue.

cobrabr:
That makes sense. It's strange because the code _does_ work. It's a test case, and it runs successfully...

cobrabr:
Here's the doc on `fn`: https://momentjs.com/docs/#/parsing/

pascalduez:
> It's strange because the code does work. It's a test case, and it runs successfully...

Yup, it just not properly declared as an export in the definition here.

GAntoine:
@pascalduez What's your verdict on #2648? It just copy-pastes the current definitions into `flow_v0.25.x-v0.62.x`.

pascalduez:
@GAntoine Why not, but basically, this shouldn't be necessary, as there's no "bug" actually. I've scanned very quickly the definition and couldn't spot something fancy introduced in newer Flow versions. We need to run tests to be sure. 

> Is there a way for the run-tests command to target a specific Flow version? (Older than the last 15 versions).

GAntoine:
Good enough for me, I'll close the other PR, and we'll call that matter settled.

PR #2661, opened 2018-08-23
Add location object support to history_v4.x.x push and replace functions

### LINKED ISSUE #2660: [history v4.x.x] history.push & history.replace should accept object as first argument
According to the [docs](https://github.com/ReactTraining/history#navigation), the first argument of `history.push`, `history.replace`, etc can be a location object.

```
// If you prefer, use a single location-like object to specify both
// the URL and state. This is equivalent to the example above.
history.push({
  pathname: "/home",
  search: "?the=query",
  state: { some: "state" }
})
```

mike1808:
Fix #2660 and add more tests

I used interfaces so that I can overload methods.


villesau:
Nice, thanks @mike1808 !

PR #2665, opened 2018-08-24
fix missing type for PG query submittable

stekycz:
This approach is used for pg-cursor or pg-query-stream.


villesau:
@stekycz would mind adding tests for the changes you made?

stekycz:
@villesau Done, thanks

PR #2666, opened 2018-08-24
Add `Card` & `Step` component definitions to `antd`

reyronald:
- [Card](https://ant.design/components/card/#Card), [source code 1](https://github.com/ant-design/ant-design/blob/master/components/card/index.tsx)
- [Steps](https://ant.design/components/steps/#Steps), [source code 1](https://github.com/ant-design/ant-design/blob/master/components/steps/index.tsx), [source code 2](https://github.com/react-component/steps/blob/9e62045c58ce7a9edcfd281db912fa11e28ac920/src/Steps.jsx)
- [Step](https://ant.design/components/steps/#Step), [source code](https://github.com/react-component/steps/blob/9e62045c58ce7a9edcfd281db912fa11e28ac920/src/Step.jsx)

### DIFF
--- a/definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
+++ b/definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
@@ -23,22 +23,22 @@ declare module 'antd' {
   
   declare export class Grid extends React$Component<{
     className?: string,
-    style?: CSSStyleDeclaration
+    style?: $Shape<CSSStyleDeclaration>,
   }> {}
 
   declare export class Meta extends React$Component<{
     avatar?: React$Node,
     className?: string,
     description?: React$Node,
-    style?: CSSStyleDeclaration
+    style?: $Shape<CSSStyleDeclaration>,
     title?: React$Node,
   }> {}
 
   declare export class Card extends React$Component<{
     actions?: Array<React$Node>,
     activeTabKey?: string,
-    headStyle?: CSSStyleDeclaration,
-    bodyStyle?: CSSStyleDeclaration,
+    headStyle?: $Shape<CSSStyleDeclaration>,
+    bodyStyle?: $Shape<CSSStyleDeclaration>,
     bordered?: boolean,
     cover?: React$Node,
     defaultActiveTabKey?: string,
@@ -238,7 +238,7 @@ declare module 'antd' {
           }
         ) => React$Node),
     size?: 'default' | 'small',
-    status: 'wait' | 'process' | 'finish' | 'error',
+    status?: 'wait' | 'process' | 'finish' | 'error',
   }> {
     static Step: typeof Step;
   }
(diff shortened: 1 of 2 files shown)


reyronald:
@GAntoine Sure! I just pushed them, is that sufficient?

By the way, I had to change the `CSSStyleDeclaration` type used to annotate the `style?` props to `$Shape<CSSStyleDeclaration>` to get the expected behavior and clear the following error:

```
 * antd_v3.x.x/flow_v0.25.x- (flow-v0.78.0): Unexpected Flow errors(2):
   Error --------------------------------------------- test_antd_v3.x.x.js:71:9
   Cannot create `Card.Grid` element because object literal [1] is incompatible
    with `CSSStyleDeclaration` [2] in property
   `style`.
      test_antd_v3.x.x.js:71:9
                  v---------------------------------------------------------
      71|         <Card.Grid style={{ width: '25%', textAlign: 'center' }}>
      72|           Content
      73|         </Card.Grid>
                  -----------^

   References:
      test_antd_v3.x.x.js:71:27
      71|         <Card.Grid style={{ width: '25%', textAlign: 'center' }}>
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ [1]
      antd_v3.x.x.js:26:13
      26|     style?: CSSStyleDeclaration,
                      ^^^^^^^^^^^^^^^^^^^ [2]
   Found 1 error
```

However, I'm not sure why that was happening, do you have any clues?

GAntoine:
Yup, the new specs look good. It's more about keeping future changes consistent with the current definition, then making sure that new definitions are correct.

No idea why the error was coming up. Maybe @villesau could shed some light? Either way, unless he has some reservation about the change, this LGTM.

reyronald:
@GAntoine I totally get it, in fact I didn't add specs originally because I was a little bit pressed on time at the moment. The tests I added are not using ALL of the props, but I think it should be fine :)

About the error, yeah, I found it odd, I'm just curious to know what causes it and why using $Shape works :S

villesau:
@reyronald @GAntoine  CSSStyleDeclaration is a class, and you pass in an object. That's why flow fails. Here is the flow declaration for CSSStyleDeclaration https://github.com/facebook/flow/blob/41b0eab99cdc5199421f7cccad9e0c4950f8b2f9/lib/cssom.js#L71

You can try CSSStyleDeclaration in console to see how it behaves.

Why $Shape fixes the issue? No idea. It's either bug or a feature :)

reyronald:
@villesau I see! I guess I was thrown out by how it is defined here: https://github.com/facebook/flow/blob/41b0eab99cdc5199421f7cccad9e0c4950f8b2f9/lib/dom.js#L1563, which I assumed is what Flow picks up when defining the style as an object in JSX, but apparently there's more to it than that.

And about $Shape, judging by its behavior I guess your assumption is right. Maybe I just found an undocumented feature 😆 

Thanks!

PR #2670, opened 2018-08-26
Update FormattedMessage in react-intl_v2.x.x.js

nextbook:
`FormattedMessage` can be configured to return strings by using the technique described [here](https://github.com/yahoo/react-intl/issues/999#issuecomment-335799491). This change enables you have the correct Flow type using this technique, as in the following example:

```jsx
<FormattedMessage id="id" defaultMessage="Request pricing">
  {(msg: string) => <a href={`mailto:email@example.com?subject=${msg}`}>{msg}</a>}
</FormattedMessage>
```


GAntoine:
@nextbook Could you add a spec to cover the addition?

nextbook:
@GAntoine consider it done!

PR #2674, opened 2018-08-27
[react-test-renderer] Export ReactTestRenderer type

skidding:
This enables importing the ReactTestRenderer type directly.

```js
import type { ReactTestRenderer } from 'react-test-renderer';
```

Useful when building libraries that receive a TestRenderer instance as an argument.


skidding:
FYI `src/commands/__tests__/runTests-test.js` timed out on Node 6.

PR #2675, opened 2018-08-27
feat: update ramda dissocPath to work with generic type

mattgstevens:
the previous definition assumed that all the values would be the same type, which will not work for any other type


GAntoine:
@mattgstevens Could you add a spec to cover this change?

PR #2685, opened 2018-08-29
Undo restore of old moment flow version range

GAntoine:
References #2648


cobrabr:
LGTM, @GAntoine, but I'm not sure whether that will solve the problem I mentioned on #2648.

GAntoine:
Then we'll need to fix the definitions for `flow_v0.25.x-v0.62.x`, as something in there is breaking.

cobrabr:
I guess the first step would be to merge this one. Once this is in, I can test again and let you know right away.

PR #2616, opened 2018-08-12
Add various RxJS statics

wbyoung:
The non-static versions of these have been deprecated. See:

- [`combineLatest`](https://rxjs-dev.firebaseapp.com/api/operators/combineLatest)
- [`zip`](https://rxjs-dev.firebaseapp.com/api/operators/zip)
- [`forkJoin`](https://rxjs-dev.firebaseapp.com/api/operators/forkJoin)
- [`race`](https://rxjs-dev.firebaseapp.com/api/operators/race)

PR #2694, opened 2018-09-04
[RxJS v6] add some missing defs & fix broken defs

berlysia:
# targets
## static
- [defer](https://rxjs-dev.firebaseapp.com/api/index/function/defer) (add)
- [combineLatest](https://rxjs-dev.firebaseapp.com/api/index/function/combineLatest) (add, rel. #2616)
- [pipe](https://rxjs-dev.firebaseapp.com/api/index/function/pipe) (improvement: allow to accept unlimited arguments)

## operators
- [withLatestFrom](https://rxjs-dev.firebaseapp.com/api/operators/withLatestFrom) (fix)
- [zip](https://rxjs-dev.firebaseapp.com/api/operators/zip) (fix)
- [combineLatest](https://rxjs-dev.firebaseapp.com/api/operators/combineLatest) (fix)
  - These operators are broken in case of multiple arguments, and now fixed in this PR
- [map](https://rxjs-dev.firebaseapp.com/api/operators/map) (add test)

### DIFF
--- a/definitions/npm/rxjs_v6.x.x/flow_v0.34.x-/rxjs_v6.x.x.js
+++ b/definitions/npm/rxjs_v6.x.x/flow_v0.34.x-/rxjs_v6.x.x.js
@@ -1252,8 +1252,8 @@ declare module "rxjs/operators" {
   };
 }
 
-interface rxjs$Operators$withLatestFrom {
-  <+T, A, B, C, D, E, F, G, H>(
+type rxjs$Operators$withLatestFrom = {
+  <T, A, B, C, D, E, F, G, H>(
     a: rxjs$Observable<A>,
     b: rxjs$Observable<B>,
     c: rxjs$Observable<C>,
@@ -1264,7 +1264,7 @@ interface rxjs$Operators$withLatestFrom {
     resultSelector: (t: T, a: A, b: B, c: C, d: D, e: E, f: F, g: G) => H
   ): rxjs$Observable<T> => rxjs$Observable<H>;
 
-  <+T, A, B, C, D, E, F, G>(
+  <T, A, B, C, D, E, F, G>(
     a: rxjs$Observable<A>,
     b: rxjs$Observable<B>,
     c: rxjs$Observable<C>,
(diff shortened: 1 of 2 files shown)


definitions/npm/rxjs_v6.x.x/flow_v0.34.x-/rxjs_v6.x.x.js
```diff
@@ -1423,119 +1238,146 @@ declare module "rxjs/operators" {
...
-    withLatestFrom<+T, A>(a: rxjs$Observable<A>, _: void): rxjs$Observable<T> => rxjs$Observable<[T, A]>;
+    // deprecated in favor of static combineLatest
+    // see: https://rxjs-dev.firebaseapp.com/api/operators/combineLatest
+    combineLatest: rxjs$Operators$combineLatest;
 
     refCount<T>(): rxjs$ConnectableObservable<T> => rxjs$Observable<T>;
   };
 }
 
+interface rxjs$Operators$withLatestFrom {
```
AndrewSouthpaw: interfaces have been known to slow down libdefs. Can you confirm that it's necessary for this to be an `interface` and not something else?

berlysia: Is it enough to rewrite `interface` with `type` ? Also works after applying [my rewriting patch](https://gist.github.com/berlysia/b07685aca114ab3ea8164cb8f43b14b8) .

I've already tried some patterns of implementation but every of those were not work.

Tried & failed patterns:
- in the manner of overloading, same as static [forkJoin](https://github.com/flow-typed/flow-typed/blob/7deafd3517fd8934f288f1f14393c748c5588825/definitions/npm/rxjs_v6.x.x/flow_v0.34.x-/rxjs_v6.x.x.js#L446)/[zip](https://github.com/flow-typed/flow-typed/blob/7deafd3517fd8934f288f1f14393c748c5588825/definitions/npm/rxjs_v6.x.x/flow_v0.34.x-/rxjs_v6.x.x.js#L719) operator
    - reordered upside down(decreasing in number of arguments)
    - pairwised with presence / absence of resultSelector (also inc/dec in number of arguments)

- in the manner of intersection, same as static [combineLatest](https://github.com/flow-typed/flow-typed/blob/7deafd3517fd8934f288f1f14393c748c5588825/definitions/npm/rxjs_v6.x.x/flow_v0.34.x-/rxjs_v6.x.x.js#L319) operator
    - (same as above)

villesau:
Nice, thanks!

PR #2698, opened 2018-09-04
react-router-dom: add support for chaining withRouther with other HOCs

eddedd88:
`withRouter` is getting rid of the types if is chained to another HOC. I ran into this using `withRouter` with material-ui `withStyles`. For example this currently works:
```
const MyComponent = (props: { a: string }) => {}

withRouter(
  withStyles({})(MyComponent)
)

<MyComponent />
```

I used the same solution used in material-ui `withStyles` definitions that allows chaining the HOC functions.


villesau:
Thanks for the contribution @eddedd88!

PR #2702, opened 2018-09-06
[react-navigation] Update libdef from upstream

MortalKastor:
Includes react-navigation/react-navigation#4917 & react-navigation/react-navigation#4836


villesau:
Thanks @MortalKastor !

PR #2703, opened 2018-09-06
[ignore] Fix `add` param array type

reyronald:
A developer should be able to do:

```js
const paths: string[] = ['.abc/*', '!.abc/d/']
ign.add(paths)
```

However, the type annotation `Array<(string | Ignore)>` doesn't allow it,
use `$ReadOnlyArray` instead so that Flow doesn't complain. [Check the
documentation for `$ReadOnlyArray` for an explanation](https://flow.org/en/docs/types/arrays/#toc-readonlyarray).

/cc @kaelzhang


definitions/npm/ignore_v4.x.x/test_ignore.js
```diff
@@ -32,3 +32,7 @@ ign.createFilter()(['foo']);
 (ign.ignores('.abc/a.js'): boolean);
 // $ExpectError
 ign.ignores(['foo']);
+
+const paths: string[] = ['.abc/*', '!.abc/d/']
+ign.add(paths)
+ign.filter(paths)
```
AndrewSouthpaw: Oops, missed on thing... Please put the test file into the same flow versioned folder as the implementation. That's our structure moving forward.

reyronald: Like that?

reyronald:
Done! :)

PR #2704, opened 2018-09-06
react-apollo: More concise type for QueryRenderProps

meafmira:
So flow can infer `data` type from `loading` and `error` states

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -899,35 +899,52 @@ declare module "react-apollo" {
 
   declare export function cleanupApolloState(apolloState: any): void;
 
-  declare export type QueryRenderProps<
-    TData = any,
-    TVariables = OperationVariables
-  > = {
-    data: TData | {||} | void,
-    loading: boolean,
-    error?: ApolloError,
+  declare export type RenderPropsUtils<TData, TVariables> = {|
     variables: TVariables,
     networkStatus: NetworkStatus,
     refetch: (variables?: TVariables) => Promise<ApolloQueryResult<TData>>,
     fetchMore: ((
       options: FetchMoreOptions<TData, TVariables> &
-        FetchMoreQueryOptions<TVariables>
+        FetchMoreQueryOptions<TVariables>,
     ) => Promise<ApolloQueryResult<TData>>) &
       (<TData2, TVariables2>(
         options: { query: DocumentNode } & FetchMoreQueryOptions<TVariables2> &
-          FetchMoreOptions<TData2, TVariables2>
+          FetchMoreOptions<TData2, TVariables2>,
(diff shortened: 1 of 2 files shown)


meafmira:
cc @budde377

budde377:
Let me just elaborate a bit on the problem here. Logically we know that:
```
¬error ∧ ¬loading ⇒ data      (1)
```
This is what many users relies on, and what you want to express with flow. What you are expressing however is this:
```
¬error ∧ ¬loading ⇔ data      (2)
```
But since 
```
data ⇒ ¬error ∧ ¬loading 
```
which can be written as 
```
error ∨ loading  ⇒ ¬data 
```
is not true, as per my example above, `(2)` is not true either.

arahansen:
I would love to see these types match what apollo outputs for their `.d.ts` which is:

```
export interface QueryResult<TData = any, TVariables = OperationVariables> extends ObservableQueryFields<TData, TVariables> {
    client: ApolloClient<any>;
    data: TData | undefined;
    error?: ApolloError;
    loading: boolean;
    networkStatus: NetworkStatus;
}
```

their types currently omit the `{||}` use case. Is it reasonable to do the same here?

budde377:
You and me both @arahansen! But unfortunately the concrete implementation defaults to the empty object (see https://github.com/apollographql/react-apollo/blob/master/src/Query.tsx#L370). An extremely poor design choice imo.

ganemone:
I think we can solve this by using `$ObjMap`. ([try flow link](https://flow.org/try/#0C4TwDgpgBAsghgawgeTMAlgewHZwDZQC8UAPAGoB8AFGQJREVQD8ZAUFB1K6JFAIoBXCACcQAJQjYAJiIAKwzGADOJACoAROMDiNiAb1YBIKVrgAuKADdM6KVAA+UACTIARgCt4YNZu0AaWEQUNCxcPApWAF92Tm5waFUAC3RsAHMlIigDDkNzLJjOKFcLJWBhFNSCqGjo1gAzAWwAYwwcKGAIUqowBWULQRFxSRlheUUVJIqlCnpsqHQ6qG7epQA6E21ZqoWlnvH101W4LcKOAHozqGxMKBEFYSUqjiacUqg80vK0zL3lA+0jqtXFVooULlApJhOrdhPcqi9sG9ilBPhUfit-nBAa4ANwxUFQcHXGH3R7PV7Ad4lMpo4i-JRMTGMuCM1wORwAcg5eMiQA))

There are basically three cases we need to handle with `props.data`.
1. null
2. empty object
3. object with props

For the purposes of types, we can consider this to be equivalent to null OR an object with optional properties corresponding to the generic types. I will submit a PR at least for reference and to further the discussion.

matthargett:
any update on this PR getting merged?

jedwards1211:
@budde377 I love your formal logic analysis 😆

jedwards1211:
Correct me if I'm wrong, but in the case of a refetch, won't `data` be fully defined while `loading`?  If you want to show the existing data while the new data is loading, you have to do the full blown type checks anyway.

PR #2706, opened 2018-09-06
signInWithEmailAndPassword should return Promise<UserCredential>

alaindresse:
signInWithEmailAndPassword should return Promise<$npm$firebase$auth$UserCredential>
see https://firebase.google.com/docs/reference/js/firebase.auth.Auth#signInWithEmailAndPassword


villesau:
Thanks for the contribution @alaindresse !

PR #2707, opened 2018-09-06
object-assign: use native `Object$Assign` type

### LINKED ISSUE #2705: Should `object-assign` be typed as `Object$Assign`?
Flow has a native type for the special function `Object.assign`: [it is
called `Object$Assign`][core]. As the `object-assign` npm package
exports a ponyfill for `Object.assign`, it seems reasonable that its
type should be `Object$Assign`.

A simple benefit of this is that the code

```js
Object.assign = require("object-assign");
```

would type-check. Currently, it does not.

A more impressive benefit of this is that the code

```js
const objectAssign = require("object-assign");
const x = {a: 1};
const y: {a: number, b: string} = objectAssign({}, x, {b: "hi"});
```

would type-check, while replacing `a: 1` with (say) `a: "one"` would
not. In other words, this offers substantially more type safety, instead
of just returning `Object`.

A temporary downside is that Flow’s definition of `Object$Assign` is
currently missing some edge cases (facebook/flow#6856,
facebook/flow#6857). The current `flow-typed` definition for
`object-assign` handles some (but not all) of those edge cases
correctly.

It seems to me that `object-assign` should be typed as `Object$Assign`
for consistency. But I understand that this is a breaking change, albeit
on edge cases and due to bugs in Flow itself.

Would a PR containing essentially

```diff
--- a/definitions/npm/object-assign_v4.x.x/flow_v0.25.x-/object-assign_v4.x.x.js
+++ b/definitions/npm/object-assign_v4.x.x/flow_v0.25.x-/object-assign_v4.x.x.js
@@ -1,3 +1,3 @@
 declare module "object-assign" {
-  declare module.exports: (target: any, ...sources: Array<any>) => Object;
+  declare module.exports: Object$Assign;
 }
```

plus some updates to tests be welcomed?

[core]: https://github.com/facebook/flow/blob/2ffa7c1bacc513371e9aede0223dd6ea409e9911/lib/core.js#L44

wchargin:
Summary:
Resolves #2705. See that issue for rationale and discussion.

Test Plan:
Running `flow-typed run-tests object-assign` passes.

wchargin-branch: object-assign-native

### DIFF
--- a/definitions/npm/object-assign_v4.x.x/flow_v0.25.x-/object-assign_v4.x.x.js
+++ b/definitions/npm/object-assign_v4.x.x/flow_v0.25.x-/object-assign_v4.x.x.js
@@ -1,3 +1,3 @@
 declare module "object-assign" {
-  declare module.exports: (target: any, ...sources: Array<any>) => Object;
+  declare module.exports: Object$Assign;
 }
--- a/definitions/npm/object-assign_v4.x.x/test_object-assign_v4.x.x.js
+++ b/definitions/npm/object-assign_v4.x.x/test_object-assign_v4.x.x.js
@@ -1,9 +1,36 @@
 // @flow
 
-import objectAssign from 'object-assign';
+import objectAssign from "object-assign";
 
-const result1: Object = objectAssign({}, null, 2, {}, []);
-const result2: Object = objectAssign(null);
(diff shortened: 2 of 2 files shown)


villesau:
Looks good, thanks for the contribution @wchargin!

PR #2713, opened 2018-09-07
element-ready: Create libdefs

reyronald:
Create library definitions for the [`element-ready`](http://npm.im/element-ready) package by @sindresorhus ([see source code here](https://github.com/sindresorhus/element-ready/blob/master/index.js))

The return type of the exported function is actually a `PCancelable` from [`p-cancelable`](http://npm.im/p-cancelable), but since there are no libdefs for that package yet I typed it with a Promise and added the TODO comment. I will submit another PR for those libdefs later and then come back to resolve it. If the maintainer's don't mind, this can be merged as it is now.

Reference: https://github.com/sindresorhus/element-ready/pull/14

### DIFF
--- a/definitions/npm/element-ready_v3.x.x/flow_v0.25.x-/test_element-ready.js
+++ b/definitions/npm/element-ready_v3.x.x/flow_v0.25.x-/test_element-ready.js
@@ -1,0 +1,43 @@
+// @flow
+
+import elementReady, {type Options} from 'element-ready';
+
+elementReady('#unicorn');
+elementReady('#unicorn', null);
+elementReady('#unicorn', {});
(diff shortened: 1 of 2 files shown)


definitions/npm/element-ready_v3.x.x/flow_v0.25.x-/test_element-ready.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+
+import elementReady, {type Options} from 'element-ready';
```
AndrewSouthpaw: Can you group using `describe` / `it` blocks to make it easier for future maintainers?

AndrewSouthpaw:

PR #2714, opened 2018-09-07
[p-cancelable]: Create libdefs

reyronald:
Create libdefs for [p-cancelable](http://npm.im/p-cancelable), [see source code here.](https://github.com/sindresorhus/p-cancelable/blob/master/index.js)

There's one thing I have my doubts on. `PCancelable.fn` pushes a new argument to the received function like this:

```js
	static fn(userFn) {
		return (...args) => {
			return new PCancelable((resolve, reject, onCancel) => {
				args.push(onCancel);
				userFn(...args).then(resolve, reject);
			});
		};
	}
```
[Source](http://github.com/sindresorhus/p-cancelable/blob/54b12bba45d3b38180f7a490bc2bc24e384612ac/index.js#L15-L22)

Since rest params should be the last argument of a function, it is not possible to annotate that function like this:

```js
static fn<T, U>(
  userFn: (...input: T[], onCancel: (handler: () => void) => void) => Promise<U>
  //       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  //       This is not possible.    
): (...input: T[]) => PCancelable<U>;
```

So my question is, is there any elegant way to achieve the above without the multiple overloads technique?

Side note: PR #2713 is waiting for this to be merged.


AndrewSouthpaw:
> So my question is, is there any elegant way to achieve the above without the multiple overloads technique?

Not that is immediately obvious.

AndrewSouthpaw:
Still failing tests :-)

reyronald:
Yes @AndrewSouthpaw , thank you for the reminder! I'll take a look this week :)

AndrewSouthpaw:
Hmm... I dunno. I'd say you can rewrite the libdef to not include 0.71 using folder names, give that a try and see if the tests pass that way?

reyronald:
Hey @AndrewSouthpaw, I ended up moving the definitions to Flow v0.72.0 and above only. Didn't want to have to maintain duplicated files for two ranges of versions just to exclude v0.71.0 :S.

Build is passing now, let me know if anything else is needed.

AndrewSouthpaw:
Seems reasonable. Oh, Flow. You're so bizarre sometimes.

PR #2720, opened 2018-09-11
lodash 'last' must return maybe - empty arrays return undefined

deecewan:
Currently, calling `_.last` returns `T`, where the argument is `Array<T>`. The return in that case will actually be `void` - the [lodash source](https://github.com/lodash/lodash/blob/4.17.10/lodash.js#L7609) explicitly makes that the case.

Previously false negative:

```js
const arr: Array<number> = [];
const res = _.last(arr);
(res: number); // passes
```

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
@@ -302,7 +302,7 @@ declare module "lodash" {
     ): Array<T>;
     join<T>(array: Array<T>, separator?: ?string): string;
     join<T>(array: void | null, separator?: ?string): '';
-    last<T>(array: ?$ReadOnlyArray<T>): T;
+    last<T>(array: ?$ReadOnlyArray<T>): ?T;
     lastIndexOf<T>(array: Array<T>, value?: ?T, fromIndex?: ?number): number;
     lastIndexOf<T>(array: void | null, value?: ?T, fromIndex?: ?number): -1;
     nth<T>(array: T[], n?: ?number): T;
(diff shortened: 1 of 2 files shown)


deecewan:
@AndrewSouthpaw should that really be a consideration, though? I want a type system so that it catches dumb bugs, such as trying to get the last item in an array with no items. The reason for this PR is because we had such a bug come up.

I get that it impedes on ease-of-use - I also don't think that is a function of a type system to provide. Not sure the general consensus on that tho. I would imagine that the 'source-of-truth' types repo should be *correct*, and if people would like to opt-out of some correctness, that should be done on a case-by-case basis.

cc @marudor @davidgomes I'm not sure who's actually worked on this, but maybe either or you can give an indication? otherwise I'll just make this change in our local libdef.

marudor:
I would be on @deecewan side. Lets make it correct.
Yes it might be "a problem" in a few cases.
But technically you rarely know that you 100% have a non empty Array.

davidgomes:
I disagree with this change. First of all, it's inconsistent with the array index access operator `[i]` (which returns `T`). Secondly, it forces a `undefined` check when people are 100% sure that the element will exist (it's not that rare in my opinion) which is a potential performance issue (alternatively, people can use `$ExpectError` which is annoying).

As a side note, this is typed as `null | undefined | T` I believe, and `_.last` returns `undefined | T`.

villesau:
> First of all, it's inconsistent with the array index access operator [i] (which returns T)

This is very strong argument IMO, although I understand both parties. It's very tricky to enforce null-safety on JS.

marudor:
I might have been to quick.
The [i] argument makes sense.
_.last and _.first are basically convienience functions for index access. So I guess you would expect the same behaviour with both.

deecewan:
@davidgomes I understand that argument. As devil's advocate, using `// $ExpectError` is significantly less annoying than having an undefined access error in production.

The array index argument makes a lot of sense. Going to close.

jedwards1211:
I definitely wish Flow provided some way to opt-in to stricter correctness checks for cases like this and array index access.

villesau:
@jedwards1211 there's been some talk about this on Flow Discord channel, and they were already almost about to merge this: https://github.com/facebook/flow/pull/6823

jedwards1211:
cool, thanks for pointing that out, sounds like it will eventually be possible to have strict libdefs for lodash!

Issue #2125, opened 2018-04-24
lodash/fp: compose([a, b, c]) is identical to compose(a, b, c)


beaugunderson:
But we get a type error with the latter and not the former.

Example:

```js
import { compose } from 'lodash/fp';

const fn = (x) => x;

// flow error
compose(fn, fn, fn)(5);

// no flow error
compose([fn, fn, fn])(5);
```

cc @mvanlonden

AndrewSouthpaw:
Fixed by #3454

gilmoreorless:
Can this issue be re-opened? It was fixed by #3454, but that was then reverted in #3475 due to causing other bugs.

All that means that this issue is still a bug in the most recent type definitions.

jedwards1211:
Seems Flow can't quite flowRight :trollface:

PR #2726, opened 2018-09-12
Use spread func param def in lodash compose functions

mvanlonden:
Resolves https://github.com/flow-typed/flow-typed/issues/2125


AndrewSouthpaw:
Fixed by #3454

PR #2730, opened 2018-09-13
Correct typings for `zip` in lodash-es

sgrowe:
[`zip` in lodash](https://lodash.com/docs/4.17.10#zip) returns an array as long as the longest input array. This means that if one of the arrays is shorter than another the output from `zip` will contain `undefined`s, but the types do not currently take this into account. As a result the following code type checks with "no errors" but will [throw an exception at runtime](https://runkit.com/5b0ea901ccf3ca0012f6bf05/5b9a3da26526760012332dd5):

```
zip(['A'], ['A', 'B']).map(([x, y]) => x.toLowerCase() + y.toLowerCase())
```


AndrewSouthpaw:
Hmm build is still failing but seems to be broken from before. I'll go ahead and merge it.

PR #2731, opened 2018-09-14
Antd types extra

ramsesz:
add some more more antd types


achepukov:
👍 🥇

definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
```diff
@@ -1,52 +1,64 @@
+import {Component, Node} from 'react';
```
villesau: These should be moved inside module declaration: If they are on top of file they returns `any`

ramsesz: ok, I have moved them now

definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
```diff
@@ -58,96 +82,207 @@ declare module 'antd' {
...
+    getFieldsError(names?: Array<string>): mixed,
+    getFieldsValue(fieldNames?: Array<string>): mixed,
+    getFieldValue(fieldName: string): mixed,
+    isFieldsTouched(names?: Array<string>): boolean,
+    isFieldTouched(name: string): boolean,
+    isFieldValidating(name: string): boolean,
+    resetFields(names?: Array<string>): void,
+    setFields(obj: Object): void,
+    setFieldsValue(obj: Object): void,
+    validateFields(callback: ValidateCallback): mixed,
```
pkyeck: @ramsesz Is this the way to do same method, different attributes? Asking because it's not working for me - only takes the last `validateFields()` and won't stop complaining that `options` is missing when I only pass in a callback :(

PR #2739, opened 2018-09-20
[normalizr] Use $ReadOnlyArray

c960657:
I believe normalizr does not mutate the input values, so it should be safe to use  `$ReadOnlyArray` here.

Disclaimer: I am a flow typing newbie, so please review carefully :-)


villesau:
In order to merge this PR we expect the contributor to have at least 6 years of prior Flow experience.

villesau:
Just kidding, looks good!

PR #2741, opened 2018-09-20
Fix commander

HsuTing:
- `arguments` in `outputHelp` should be a function
  - [api docs](http://tj.github.io/commander.js/#Command.prototype.outputHelp)
- `command.parse` can return `argv`, `rawArgv` and `options`.
  - [api docs](http://tj.github.io/commander.js/#Command.prototype.parse)

### DIFF
--- a/definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
+++ b/definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
@@ -243,7 +243,7 @@ declare module "commander" {
      *
      * @api public
      */
-    outputHelp(): void;
+    outputHelp((defaultHelp: string) => string): void;
 
     /**
      * Output help information and exit.
--- a/definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
+++ b/definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
@@ -159,8 +159,7 @@ declare module "commander" {
      * @return {Command} for chaining
      * @api public
      */
-    parse(argv: Array<string>): {
-      ...this,
+    parse(argv: Array<string>): this & {
       args: Array<string>,
       rawArgs: Array<string>,
       [string]: any,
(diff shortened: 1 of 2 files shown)


definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
```diff
@@ -159,7 +159,12 @@ declare module "commander" {
      * @return {Command} for chaining
      * @api public
      */
-    parse(argv: Array<string>): this;
+    parse(argv: Array<string>): {
+      ...this,
+      args: Array<string>,
+      rawArgs: Array<string>,
+      [string]: any,
+    };
```
HsuTing: I am not sure that this is the best way to fix it.
If you have a good idea to modify this, tell me please.

definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
```diff
@@ -243,7 +248,7 @@ declare module "commander" {
      *
      * @api public
      */
-    outputHelp(): void;
+    outputHelp((defaultHelp: string) => string): void;
```
villesau: based on docs, this is optional, you can use e.g `outputHelp(cb?: ?(defaultHelp: string) => string): void;`

definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
villesau: Other way could be `this & { args: Array<string>, ... }`. I'm not sure which is better.

definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
HsuTing: OK, I will fix this.

definitions/npm/commander_v2.x.x/flow_v0.28.x-/commander_v2.x.x.js
HsuTing: I like your answer. `this & { args: Array<string>, ... }` make this more understandable.
I will change.

villesau:
Great, thanks!

PR #2743, opened 2018-09-20
add getAll method

eleith:
method documented here: 

https://cloud.google.com/nodejs/docs/reference/firestore/0.15.x/Firestore#getAll


villesau:
Thanks!

PR #2749, opened 2018-09-22
Refactor validation error handling

villesau:
Before the CLI has passed all the validation errors all over the place, and by pushing into this array accumulated the errors. I changed the cli to rather just simply throw in case of (validation) error. This reduces cross-dependencies and variables passed to different functions, and we get rid of mutable array which ties everything together and context variable. In overall most of the function signatures became much simpler.

What this means in practice? Before you might have got multiple errors at the same time, but now the execution stops when ever error occurs. Usability wise this is IMO very minor downside compared to upside of cleaner code. Same behaviour can be achieved with exception approach as well later if we find it useful: We can catch errors, store them and re-throw accumulated errors. I didn't think this is wort effort at this point though. E: Ended up to collect getNpmLibDefs errors and throw them all at once.

Why I did this? Well, cleaner code is always cleaner code. This is also works as a bit of ground work towards dependency support in runTests.

I also removed some dead code.


cli/src/lib/libDefs.js
```diff
@@ -431,49 +409,6 @@ function writeVerbose(stream, msg, writeNewline = true) {
   }
 }
 
-const NAMESPACE_DIR_EXCEPTIONS = [
-  '.cli-metadata.json',
-  'node_modules',
-  'package.json',
-  '.eslintrc',
-];
-async function getLibDefNamespaces(defsDirPath: string) {
```
GAntoine: Why where these methods removed?

cli/src/lib/libDefs.js
```diff
@@ -483,25 +418,10 @@ export async function getLocalLibDefs(validationErrs?: VErrors) {
...
 export async function getCacheLibDefs(
   verbose?: VerboseOutput = process.stdout,
-  validationErrs?: VErrors,
 ) {
   await ensureCacheRepo(verbose);
   await verifyCLIVersion(path.join(CACHE_REPO_DIR, 'definitions'));
-  return getLibDefs(CACHE_REPO_DEFS_DIR, validationErrs);
-}
-
-export async function getCacheLibDefVersion(libDef: LibDef) {
```
GAntoine: Same with this one. Are they unused?

cli/src/lib/npm/npmLibDefs.js
```diff
@@ -470,65 +419,79 @@ export async function getInstalledNpmLibDefs(
   return installedLibDefs;
 }
 
+/**
+ * Retrieve single libdef.
+ */
+async function getItem(
```
GAntoine: Could this be a little more specific, like `getSingleLibdef`?

cli/src/lib/libDefs.js
villesau: `getLocalLibDefs` is not used anywhere, which was only place using `getLibDefNamespaces `

cli/flow-typed/npm/jest_v23.x.x.js
```diff
@@ -958,7 +1102,13 @@ declare var expect: {
...
   /** Matches any received string that contains the exact expected string. */
   stringContaining(value: string): string,
-  stringMatching(value: string | RegExp): string
+  stringMatching(value: string | RegExp): string,
+  not: {
+    arrayContaining: (value: $ReadOnlyArray<mixed>) => Array<mixed>,
+    objectContaining: (value: {}) => Object,
+    stringContaining: (value: string) => string,
+    stringMatching: (value: string | RegExp) => string,
+  },
```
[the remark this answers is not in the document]
villesau: I was working on tests so thought that up to date jest libdefs would make it easier :) Definitely would've been easier to have separate PR for updating libdefs. And to highlight just in case, these are CLI libdefs, not the ones users downloads

cli/src/lib/libDefs.js
villesau: Not used anywhere.

cli/src/lib/npm/npmLibDefs.js
villesau: god point!

Issue #2746, opened 2018-09-21
Proposal: Drop support (stop testing) for 0.53.0 and earlier libdefs


villesau:
Flow has slightly different way of handling suppression comments before and after flow 0.53.0. This causes the cli to have code that generates two two .flowccofigs and run tests a bit differently: 

https://github.com/flow-typed/flow-typed/blob/master/cli/src/commands/runTests.js#L443
https://github.com/flow-typed/flow-typed/blob/master/cli/src/commands/runTests.js#L543

By dropping this we could reduce complexity of test script. This does not mean that we couldn't still be able to have older libdefs available, but only that we don't test them anymore. Now we test only against last 15 versions. Only case when we currently test agains 0.53.0 and earlier is when upper bound of libdef is around that version. 0.53.0 flow version is over one year old anyways and should not be used anymore.

AndrewSouthpaw:
I'm fine with dropping support for testing before 0.53.

pascalduez:
`0.53` is not "that" old, a bit more that one year. But I understand the need to move forward and lighten the tests.  
FWIW, we are stuck with `0.52` at work :cry: , because we didn't have the resources to migrate our large code-base to `0.53` when it was released, and now the task is daunting.  

Would it be possible to manually run tests for a specific version? 
`runTests --version=0.52` or something?

villesau:
If we merge the PR, there wouldn't be such option :/ The problem is that older Flows does not accept `include_warnings=true` in .flowconfig, but newer versions require it in order to print warnings. This was the original PR to introduce that logic: https://github.com/flow-typed/flow-typed/pull/1407

That's pretty annoying detail when doing further refactorings: If we continue to support <0.53.0 we need to split the test run into older and newer which limits the flexibility. I'd like to get the `runTests` in as pleasant condition as possible so that it would be easier to implement cross-dependencies between libdefs in future.

PR #2750, opened 2018-09-22
Don't run tests for older Flows

villesau:
This helps to refactor runTests further. 0.53.0 flow version is so old that we don't need to support older versions than that.

fixes: https://github.com/flow-typed/flow-typed/issues/2746


pascalduez:
`0.53` is not "that" old, a bit more that one year. But I understnd the need to move forward and lighten the tests.  
FWIW, we are stuck with `0.52` at work :cry: , because we didn't have the ressources to migrate our large codebase to `0.53` when it was released, and now the task is daunting.  

Would it be possible to manually run tests for a specific version? 
`runTests --version=0.52` or something?

PR #2751, opened 2018-09-23
Allow 8 args to open-ended redux-saga calls

dhui:
Existing libdefs only allow 6 args to redux-saga calls. This PR extends that to 8.


dhui:
@villesau Thanks for the pointer!
It looks like v0.68.0 is the version of flow where the existing libdefs are broken. I've updated the version folder accordingly and tests are passing now.

villesau:
Nice!

PR #2754, opened 2018-09-23
Update jest-enzyme definitions

pascalduez:
Update jest-enzyme definitions to the latest versions.  
Keep removed matchers for retro-compat.


villesau:
Nice!

PR #2756, opened 2018-09-23
Add a11y-dialog definitions

pascalduez:
See https://github.com/edenspiekermann/a11y-dialog

### DIFF
--- a/definitions/npm/a11y-dialog_v5.x.x/test_a11y-dialog_v5.x.x.js
+++ b/definitions/npm/a11y-dialog_v5.x.x/test_a11y-dialog_v5.x.x.js
@@ -1,0 +1,167 @@
+// @flow
+
+import { describe, it } from "flow-typed-test";
+import A11yDialog from "a11y-dialog";
+
+let el = document.getElementById("my-accessible-dialog");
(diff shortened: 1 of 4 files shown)


definitions/npm/a11y-dialog_v5.x.x/test_a11y-dialog_v5.x.x.js
```diff
@@ -0,0 +1,179 @@
...
+    dialog.off("show", true);
+  });
+
+  it("should error on missing parameters", () => {
+    // $ExpectError
+    dialog.off("show");
+  });
+});
+
+// describe("The `dialog` property", () => {
```
villesau: Why this is commented out? I would rather remove it if it's not needed, or if it's something like "should work but can't make it work because of X", i would comment that there too.

pascalduez: I need to give it another try. Open to suggestions. 
`dialog.dialog` can be either `HTMLDialogElement | HTMLElement`. 
Here I want to test only the `HTMLDialogElement` case, but I can't make Flow understand the condition.

PR #2757, opened 2018-09-24
Add tip for automatic updates to README.md

soulofmischief:
This is the way I'm updating my flow types automatically with yarn, and I thought it could be useful to others.


soulofmischief:
I'm going to close this, because it seems there is a gotcha when using this method for only one script... I use it for multiple so I didn't catch it before writing this PR. I'm going to just write up a longer explanation on a separate repo and throw up a PR with a link later.

Issue #2755, opened 2018-09-23
Mongoose v5 connect() type may be wrong


laudeon:
Hello,

I wanted to point the fact that the mongoose connect() function is typed as if it accepts two arguments (uri and ?options) and returns void. 

however, according to the Mongoose documentation this function can accept a third argument (a callback) and returns a Promise: [see here](https://mongoosejs.com/docs/connections.html#callback).

Is there a specific reason to that unfortunate situation where I cannot catch any connection errors using the returned promise without having Flow telling me that connect() returns void? If not, any interest in a PR?

laudeon:
Thanks! Sounds good :+1:

PR #2760, opened 2018-09-24
added callback and promise support to mongoose connect def #2755

laudeon:
Hello, I've added callback and promise support to the connect mongoose function and the following tests. Not sure about the Promise<any> though. Any thoughts?


laudeon:
> «Promise» resolves to `this` if connection succeeded
[from the doc](https://mongoosejs.com/docs/api.html#mongoose_Mongoose-connect)

Do you agree that means the callback takes `this` (I imagine `this` should refer to the mongoose Connection instance) as an argument? Thus, I may say that I have some doubts about the `void` here.

If it's the case and the promise resolves to `this`, then, would it be better to use directly `Mongoose$Connection` instead of `any`?

villesau:
@laudeon I think that in this case you can actually define it as `this` instead! You can also smoke test the assumption by testing it liker: `mongoose.connect('mongodb://host:port/dbname').then(data => (data: Mongoose$Connection))`

laudeon:
The `this` keyword did not pass the tests. I used Mongoose$Connection. Thanks for the help! I did not know before about the type casting :).

PR #2761, opened 2018-09-25
fix: styled-component withTheme hoc def

theshortcut:
This PR simplifies and corrects the `withTheme` hoc type for styled-components.


GAntoine:
Looks like this change failed the specs, could you take a look at them when you get some time?

theshortcut:
closing. efforts should go towards styled-components v4

PR #2767, opened 2018-09-27
Update moment js to add defineLocale function

MarcoPolo:
same type as updateLocale (https://momentjs.com/docs/#/customization/)

### DIFF
--- a/definitions/npm/moment_v2.3.x/flow_v0.25.x-/moment_v2.3.x.js
+++ b/definitions/npm/moment_v2.3.x/flow_v0.25.x-/moment_v2.3.x.js
@@ -300,6 +300,7 @@ declare class moment$Moment {
   static isDate(obj: any): boolean,
   static locale(locale: string, localeData?: Object): string,
   static updateLocale(locale: string, localeData?: ?Object): void,
+  static defineLocale(locale: string, localeData?: ?Object): void,
   static locale(locales: Array<string>): string,
   locale(locale: string, customization?: Object | null): moment$Moment,
   locale(): string,


GAntoine:
Addition looks good, could you add a quick spec to cover it?

valscion:
This method seems to have been added to `moment` only in v2.8.0: https://github.com/moment/moment/blob/2.8.0/CHANGELOG.md#280-see-changelog

I see there's already two different flow types for `moment` here. How should we proceed in adding new types? These are currently here:

* `moment_v2.3.x/flow_v0.25.x-`
* `moment_v2.x.x/flow_v0.63.x-`

This PR seems to be the reason why we now have two versions: https://github.com/flow-typed/flow-typed/pull/2648

GAntoine:
> I see there's already two different flow types for moment here. How should we proceed in adding new types?

You can either create a moment typedef under `moment_v2.8.x`, which will cover version 2.8 exclusively, or update the definition for `moment_v2.x.x`, and let those with lower moment version deal with the addition... Yeah not a great situation all around.

>  I wanted to add new types to moment package and figured I'd tackle the open issues and pull requests related to moment while I'm at it

Always appreciated!

valscion:
> or update the definition for `moment_v2.x.x`, and let those with lower moment version deal with the addition... Yeah not a great situation all around.

Yeah I think I'll go with this approach. It's better that Flow doesn't complain about valid use cases for modern versions, I think, than being extra careful of not adding types for old versions.

I'm not sure why there are `v2.3.x` and `v2.x.x` versions, either. I'll see if I can figure that out and simplify the situation so that we wouldn't have to maintain two, almost similar, type definitions.

GAntoine:
> I'll see if I can figure that out and simplify the situation so that we wouldn't have to maintain two, almost similar, type definitions.

Might make sense to check the git logs and see who built the 2.3.x version, and ping them. They might have a reason for doing it.

valscion:
> Might make sense to check the git logs and see who built the 2.3.x version, and ping them. They might have a reason for doing it.

I did, see https://github.com/flow-typed/flow-typed/pull/2783 for my adventure.

The original reason was "to follow semver". Apart from the one new addition, that rule hasn't held very strictly for moment so I went with less strict rules.

valscion:
Ok well, I didn't ping them, because the original discussion was heated and I didn't want to derail #2783 for an unwanted argument. I'll happily go through my ideas in a more synchronous channel than GitHub issues if you want to have more insight

GAntoine:
Haha no worries. From what I understand what you're proposing makes sense, so I'm not gonna put you through the ringer lol. You're welcome to join out slack channel (when I can figure out how to invite/auto-invite new people...)

valscion:
The code from this PR was merged in #2780 so this can be closed.

PR #2769, opened 2018-09-29
Update ramda_v0.x.x.js

iKethavel:
Updated method `assocPath` parameter `key`, because this parameter should also support number type
```
const state = {
  controllers: [
    { tracks: [] },
    { tracks: [ 'song1' ] }
  ]
}

const index = 1
const data = ['newSongs1']

R.assocPath(
  ['controllers', index, 'tracks'],
  data,
  state
)
```


GAntoine:
Closing as already fixed in newer version

PR #2773, opened 2018-09-30
Add colors definitions

pascalduez:
See https://github.com/Marak/colors.js   

For the `String.prototype` API I'm looking for a way to extends builtin `String` class. But I guess it's not yet possible, right?


villesau:
I don't think so. Not at least conditionally: The type should be applied only when the package is imported.

PR #2774, opened 2018-09-30
Add a Prettier config file

pascalduez:
Hi,

I noticed quite some inconsistencies regarding formatting on the definitions. 
I use Prettier when writing them, but with bare defaults in the doubt.

This is a proposal, just a few config keys. Open to bikeshedding. 

Just a first step, but already useful.

We could go even further and add a script alias, a precommit-hook check, if needed.

@villesau @AndrewSouthpaw @GAntoine

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+---
+singleQuote: true
+trailingComma: "es5"
+parser: "flow"


pascalduez:
> If we prettify in precommit hook 

I always find this a bit intrusive, and as you said could be cumbersome with tests.
Maybe just run a `prettier check` and warn.
Or actually, just adding that config file without any process, would already help, for those using it while writing.

pascalduez:
A good example about why this would be helpful :) #2804

villesau:
@marudor placing it with prettier is easy, but it might break existing tests if we run it for the whole definitions folder.

AndrewSouthpaw:
I... think I have strong feelings about enforcing prettier on libdefs. There's some, such as Ramda, where we need to do formatting that makes it easier to read but prettier would pitch a fit, and I *really* don't like what prettier ends up doing.

Is there a way to enable it on a per libdef basis, or exclude certain libdefs, or...?

We don't use prettier at work so my knowledge of the tool is limited.

villesau:
I'm not biggest fan of prettier either :) The concept is great but output often is much uglier than I would expect from code. But I have conditioned to it as we use it at work, and I believe the value of having standardized output especially in case of flow-typed where there are 650 contributors is pretty big. The only concern from my side is how it affects to error locations.

langri-sha:
@AndrewSouthpaw Prettier provides a [`// prettier-ignore`](https://prettier.io/docs/en/ignore.html#javascript) escape hatch to ignore the next node, if someone wants to preserve formatting.

villesau:
Should we run prettier for all the existing libdefs?

PR #2776, opened 2018-10-02
Needed change for flow 0.82.0

doberkofler:


GAntoine:
Looks like this change broke the specs. Also, if this is indeed needed for 0.82.x, might make sense to create a new folder called `flow_v0.82.x-`, cap the current one at `flow_v0.58.x-v0.81.x`

doberkofler:
Correct, but in my understanding the change would also work in older versions.

GAntoine:
Ok cool, then it's just the specs that need to be addressed.

GAntoine:
Closing as already complete.

PR #2779, opened 2018-10-02
chore(antd): remove repetition

ferhatelmas:


pascalduez:
Tests should be fixed by #2784

PR #2789, opened 2018-10-04
Limit recompose's libdef to versions before v0.27.0

dhui:
In [v0.27.0](https://github.com/acdlite/recompose/releases/tag/0.27.0), recompose started shipping it's own libdefs

This change prevents clobbering of the recompose libdef.


pascalduez:
Which means users of versions lower than `0.26` won't be able to get the definitions anymore. 
Does range works for packages as well? `recompose_v0.x.x-v0.26.x`

dhui:
AFAIK, version ranges cannot be specified for packages.
Search for `pkgVersion` in the code base.

You can see the pkg SemVer being applied here: https://github.com/flow-typed/flow-typed/blob/acdbea3b8c53cd8595f1e0078e9d072282331d7f/cli/src/lib/npm/npmLibDefs.js#L221

GAntoine:
> AFAIK, version ranges cannot be specified for packages.

Correct.

IMO, I'd rather let those on version lower then `0.26` retain access to these definitions, then delete them for the convenience of those using the new built-in types. They can work around the issue when using the CLI, or manually delete/git ignore these definitions.

dhui:
@GAntoine 
It's pretty annoying to have to manually manage the libdef file. e.g. every time I update flow types, I need to delete the recompose libdef afterwards.

How do you feel about having a libdef for every version of recompose before `v0.27.x`?
We can do this with symlinks or file copies. It's quite simple to generate the files/symlinks (see below) but any maintenance would be more annoying. Copying vs symlinks each have their maintenance pros and cons (e.g. updating a few libdefs vs many).

```bash
# lists all recompose versions
$ git tag | grep '^v0\..' | sed -E -e 's/v0\.([0-9]+)\..*/v0.\1.x/' | sort -u
v0.10.x
v0.11.x
v0.12.x
v0.13.x
v0.14.x
v0.15.x
v0.16.x
v0.17.x
v0.18.x
v0.19.x
v0.2.x
v0.20.x
v0.21.x
v0.22.x
v0.23.x
v0.24.x
v0.25.x
v0.26.x
v0.28.x
v0.29.x
v0.3.x
v0.30.x
v0.4.x
v0.5.x
v0.6.x
v0.7.x
v0.8.x
v0.9.x
$
```

GAntoine:
> How do you feel about having a libdef for every version of recompose before v0.27.x?

Oh boy, now that would be a lot of duplicated code! After thinking on this, I'm ok moving this PR forward as-is. Anyone on a version below 0.26 is welcome to open an issue, and we'll copy over `0.26.x` to their specific version of recompose.

dhui:
@GAntoine Thanks for reconsidering!

PR #2786, opened 2018-10-03
Add react-native-simple-store definitions

amhinson:
Add support for react-native-simple-store: https://github.com/jasonmerino/react-native-simple-store

PR #2805, opened 2018-10-09
Add missing 'delete' method in react-native-simple-store

gre:
ctx https://github.com/flow-typed/flow-typed/pull/2786

switched to another synthax because `declare function delete` would not pass eslint

### DIFF
--- a/definitions/npm/react-native-simple-store_v1.x.x/flow_v0.25.x-/react-native-simple-store_v1.x.x.js
+++ b/definitions/npm/react-native-simple-store_v1.x.x/flow_v0.25.x-/react-native-simple-store_v1.x.x.js
@@ -2,9 +2,14 @@ declare module 'react-native-simple-store' {
   declare type Key = string | Array<Array<any>>;
   declare type Value = string | boolean | Object | Array<any>;
 
-  declare function get(key: Key): Promise<?any>;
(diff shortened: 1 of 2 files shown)


definitions/npm/react-native-simple-store_v1.x.x/flow_v0.25.x-/react-native-simple-store_v1.x.x.js
```diff
@@ -2,9 +2,14 @@ declare module 'react-native-simple-store' {
   declare type Key = string | Array<Array<any>>;
   declare type Value = string | boolean | Object | Array<any>;
 
-  declare function get(key: Key): Promise<?any>;
-  declare function save(key: Key, value: Value): Promise<?any>;
-  declare function update(key: string, value: Value): Promise<?any>;
-  declare function keys(): Promise<?any>;
-  declare function push(key: string, value: Value): Promise<?any>;
+  declare type SimpleStore = {
+    get: (key: Key)=> Promise<?any>,
```
pascalduez: `?any` is weird, `any` alone is enough, since it bascially means no typing, so anything will pass.

gre: good point

PR #2807, opened 2018-10-09
Support underscore with flow 0.75

almost:
I just replaced `$call` with `[[call]]`


pascalduez:
Well this was already fixed in #3067, so after a rebase this only fixes the comment.

Thanks anyway!

PR #2808, opened 2018-10-09
Add file-saver definitions

pascalduez:
See https://github.com/eligrey/FileSaver.js

### DIFF
--- a/definitions/npm/file-saver_v2.x.x/test_file-saver_v2.x.x.js
+++ b/definitions/npm/file-saver_v2.x.x/test_file-saver_v2.x.x.js
@@ -1,0 +1,58 @@
+// @flow
+
+import { describe, it } from 'flow-typed-test';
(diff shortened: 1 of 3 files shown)


definitions/npm/file-saver_v2.x.x/test_file-saver_v2.x.x.js
```diff
@@ -0,0 +1,58 @@
+// @flow
+
+import { describe, it } from "flow-typed-test";
+import saveAs from "file-saver";
+
+let blob = new Blob([""]);
```
AndrewSouthpaw: Can you switch the `"` to `'`? Clearly relevant to #2774 ...

pascalduez: Done. See the config file would be useful ;)

PR #2811, opened 2018-10-10
added parse-decimal-number to definitions

FermiDirak:
added parse-decimal-number to definitions

### DIFF
--- a/definitions/npm/parse-decimal-number_v1.x.x/flow_v0.25.x-/parse-decimal-number_v1.x.x.js
+++ b/definitions/npm/parse-decimal-number_v1.x.x/flow_v0.25.x-/parse-decimal-number_v1.x.x.js
@@ -1,3 +1,9 @@
 declare module 'parse-decimal-number' {
-  declare module.exports: (value: string, inOptions?: any, enforceGroupSize?: boolean) => number;
+  declare module.exports: {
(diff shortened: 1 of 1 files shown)


definitions/npm/parse-decimal-number_v1.x.x/flow_v0.25.x-/parse-decimal-number_v1.x.x.js
```diff
@@ -0,0 +1,3 @@
+declare module 'parse-decimal-number' {
+  declare module.exports: (value: string, inOptions?: any, enforceGroupSize?: boolean) => number;
```
pascalduez: `inOptions` can be either `string | Array<string> | { thousands: string, decimal: string }`.

pascalduez: The return value can be either `number | NaN`.

pascalduez: Missing `parseDecimalNumber.factorySettings`, `parseDecimalNumber.setOptions`, `parseDecimalNumber.withOptions` signatures.

GAntoine:
@FermiDirak You'll need to add specs for the new def.

PR #2812, opened 2018-10-10
[React-Apollo] Add context to QueryOpts

nathantalewis:
It is possible to pass `context` to query options as seen in the docs:
https://www.apollographql.com/docs/react/api/react-apollo.html#graphql-config-options-context

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -781,7 +781,8 @@ declare module "react-apollo" {
     fetchPolicy?: FetchPolicy,
     pollInterval?: number,
     skip?: boolean,
-    errorPolicy?: ErrorPolicy
+    errorPolicy?: ErrorPolicy,
+    context?: { [key: string]: any }
   |};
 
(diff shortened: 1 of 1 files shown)


villesau:
@nathantalewis looks like the tests are failing. Would you mind fixing them too?

nathantalewis:
I just gave a quick scan through the failures and I can't see the relation with the change I made. I don't consider myself well versed on the React-Apollo types. I only looked far enough into things to add the missing property that was causing problems in the project I was working on and decided I would offer that improvement back to the community. I think you will need someone more familiar with the React-Apollo project to resolve the failing specs.

I won't be offended if you decide to close this PR until someone able to fix the specs can look into the failures. Although, someone more familiar with React-Apollo might want to give a general review of the React-Apollo types to bring all of them in line with what is documented on the Apollo website rather than this one-off addition that I did.

villesau:
Thanks @nathantalewis! Could you rebase against master to see if it's still happening? There is a chance that it's related to recent flow updates and someone else might have already fixed the problem.

nathantalewis:
Rebased. Let's see how it goes. 🤞

villesau:
hmm.. meh. Unfortunately Travis is having bad time at the moment. Most of the builds are currently failing like this one. Maybe some resource quota stuff, dunno. I'll come back to this tomorrow and see if it still fails unexpectedly.

Issue #2782, opened 2018-10-02
react-select for flow 0.53+


IanVS:
It seems like react-select was [limited](https://github.com/flow-typed/flow-typed/commit/19506e57e6a19b813ec13cbab4244b3ab7f938cb) to flow 0.52, and now that I've upgraded to 0.53 I've lost all my flow typing for react-select.  What would it take to get a definition set available for newer versions of flow?  Thanks!

AndrewSouthpaw:
You're welcome to copy the libdef and create a new version, and see how the tests work out.

PR #2817, opened 2018-10-10
Add react-select definitions for Flow >= 0.53

pascalduez:
Refs #2782 

Copy the 0.52 definitions plus several improvements and cleans.


pieplu:
Since this update, I disocver than the definition is uncomplet.

The lib export more than just the default Select Class : https://github.com/JedWatson/react-select/blob/v1.x/src/index.js

Issue #1385, opened 2017-10-20
// $ExpectError tests can never fail


aaronjensen:
```js
// $ExpectError
const x: number = 3;
```

That should fail, but doesn't.

villesau:
This commit from flow changed the behaviour: https://github.com/facebook/flow/commit/c62d5ebac2ef7881192b760b023ce54bb57140e6
And there is an issue open about it: https://github.com/facebook/flow/issues/5009
Seems rather serious!

aaronjensen:
Agreed, quite serious. I'd hate to be the one to try and reenable it on this repo. I'd guess the only way to reenable it now would be to grep for warning in the output of flow since it won't exit w/ a non-zero status code :/

GAntoine:
Flow recently made unused suppression comments warnings instead of errors. Warnings are not included by default. (You need `include_warnings=true` in the `.flowconfig`). I'll look into how this affects running specs on travis.

villesau:
Made a naive fix for the problem: https://github.com/flowtype/flow-typed/pull/1407 at least it should point to right direction.

villesau:
@GAntoine it might make sense to run the test suite for all the PRs open at the moment now when the fix is merged.

GAntoine:
Yeah that makes sense, though they'll need to rebase first. I'll go through the active ones and let people know.

julienw:
@gantoine this is still not working: a lot of tests are not failing even though $ExpectError fails to catch an error... See the tests for react-redux_v5 as an example: they should clearly fail but currently they don't.

julienw:
Ok, actually this works, but because in travis tests are only run on differences, we haven't seen these errors.

villesau:
yeah would not be fair to fail tests on non-related libdefs.

PR #2820, opened 2018-10-11
[react-apollo] Extract QueryComponentProps

jevakallio:
As a user of `react-apollo`, it's a common use case to want to create your own higher-order `Query` component to specify default options, additional behaviours, etc.

In order to type that higher order component, we need to export the Props type of the `react-apollo` `Query` component, so that a consumer of the library can use it to define their HoC props type by extending it.

This commit extracts and exports a generic `QueryComponentProps<TData, TVariables>` type. It has no functional changes.

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -933,7 +933,7 @@ declare module "react-apollo" {
     QueryRenderProps<TData, TVariables>
   ) => Node;
 
-  declare export class Query<TData, TVariables> extends React$Component<{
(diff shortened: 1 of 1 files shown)


jevakallio:
The failing Travis test does not seem to be related to this change -- it's a failing test for regression https://github.com/flow-typed/flow-typed/issues/1385.

```
run-tests (command) › regression 1385 › returns error code if $ExpectError test doesn't fail
```

villesau:
@jevakallio Thanks for the contribution! I rerun the test suite and looks like there are bunch of valid errors too.

matthargett:
ping on this? flow is failing for all the teams I work with that use react-apollo.

villesau:
@matthargett feel free to fork and reopen the PR with fixed test cases. I'm afraid that the original author is no longer interested in this.

villesau:
@jevakallio any progress? :)

PR #2821, opened 2018-10-11
[winston] add transport level and silent properties

### LINKED ISSUE #2814: [winston 3.x.x] Setting level of a winston transport will fail
If I try setting re-setting the level of a [winston](https://github.com/winstonjs/winston) transport like their README reccomends:

```js
const transports = {
  console: new winston.transports.Console({ level: 'warn' }),
  file: new winston.transports.File({ filename: 'combined.log', level: 'error' })
};

const logger = winston.createLogger({
  transports: [
    transports.console,
    transports.file
  ]
});

logger.info('Will not be logged in either transport!');
transports.console.level = 'info';
transports.file.level = 'info';
logger.info('Will be logged in both transports!');
```

flow will complain with:

```
Cannot assign logLevel to transports.console.level because property level is missing in $winstonConsoleTransport [1].

     src/cli/index.js
     68│         }
     69│
     70│         if (logLevel) {
     71│           transports.console.level = logLevel;
     72│         }
     73│
     74│         return logLevel || process.env.LOG_LEVEL || 'error';

     src/logger.js
 [1] 22│   console: new winston.transports.Console({ level }),
```

runarberg:
closes #2814

### DIFF
--- a/definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
+++ b/definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
@@ -24,7 +24,10 @@ declare type $winstonFileTransportConfig<T: $winstonLevels> = {
   level?: $Keys<T>
 };
 
-declare class $winstonTransport {}
+declare class $winstonTransport {
+  level?: string;
+  silent?: boolean;
+}
 
 declare class $winstonFileTransport<T> extends $winstonTransport {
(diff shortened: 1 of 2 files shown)


villesau:
Thanks @runarberg! btw, I made an issue of your second comment: https://github.com/flow-typed/flow-typed/issues/2833 Feel free to improve the libdefs, there are plenty of examples on how to expose also named imports! :)

PR #2804, opened 2018-10-08
add array contains

eleith:
add https://firebase.google.com/docs/firestore/query-data/queries#array_membership (array contains) query operator

PR #2822, opened 2018-10-11
Add array-contains option for queries

nfarina:
See: https://firebase.google.com/docs/firestore/query-data/queries


pascalduez:
Dupe of #2804

nfarina:
Thanks @pascalduez, closing in favor of #2804.

PR #2824, opened 2018-10-11
Add typeof to material-ui handleEvent

KevinGrandon:
Without this fix, the following error occurs in flow 0.83:

image

### DIFF
--- a/definitions/npm/@material-ui/core_v1.x.x/flow_v0.58.x-/core_v1.x.x.js
+++ b/definitions/npm/@material-ui/core_v1.x.x/flow_v0.58.x-/core_v1.x.x.js
@@ -163,7 +163,7 @@ declare module "@material-ui/core/ButtonBase/createRippleHandler" {
     eventName: string,
     action: string,
     cb: ?Function
-  ) => handleEvent;
+  ) => typeof handleEvent;
 }
 
 declare module "@material-ui/core/ButtonBase" {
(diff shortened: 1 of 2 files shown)


villesau:
@KevinGrandon Thanks for the contribution! I believe the fix is probably right, but looks like the tests are not passing:

```
 * @material-ui/core_v1.x.x/flow_v0.58.x- (flow-v0.83.0): Unexpected Flow errors(2):
   Error ----------------------------------------------------------------------------------------- test_core_v1.x.x.js:42:1
   
   Cannot extend property `Component` [1] with `TestComponent` because in the return value of property `render`:
    - Either undefined [2] is incompatible with null [3].
    - Or property `@@iterator` is missing in undefined [2] but exists in `$Iterable` [4].
   
      test_core_v1.x.x.js:42:1
          v--------------------------------------------
      42| class TestComponent extends React.Component<{
      43|   ...$Exact<WithStyles>,
      44|   requiredProp: string
      45| }> {
      46|   render() {
      47|     const root = this.props.classes.root
      48|   }
      49| }
          ^
   
   References:
      test_core_v1.x.x.js:42:29
                                      v----------------
      42| class TestComponent extends React.Component<{
      43|   ...$Exact<WithStyles>,
      44|   requiredProp: string
      45| }> {
          -^ [1]
      test_core_v1.x.x.js:46:11
      46|   render() {
                    ^ [2]
      <BUILTINS>/react.js:14:5
      14|   | null
              ^^^^ [3]
      <BUILTINS>/react.js:20:5
      20|   | Iterable<?React$Node>;
              ^^^^^^^^^^^^^^^^^^^^^ [4]
   
```

To fix the errors, please refer to:
https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md

KevinGrandon:
> @KevinGrandon Thanks for the contribution! I believe the fix is probably right, but looks like the tests are not passing:

Thanks for the review! I believe these are new errors that are also existing in master today, I've gone ahead and pushed a new commit with a change that fixes the tests.

PR #2825, opened 2018-10-12
forEach shouldn't require a boolean return

nfarina:
TypeScript defs had the same bug (now fixed): https://github.com/firebase/firebase-js-sdk/issues/555

Documentation to back up optional boolean: https://firebase.google.com/docs/reference/js/firebase.database.DataSnapshot#forEach


villesau:
Nice!

PR #2826, opened 2018-10-12
Add the missing List.Item.Meta in antd types

piuccio:
Documented here: https://ant.design/components/list/#List.Item.Meta


villesau:
Thanks! :+1:

PR #2830, opened 2018-10-12
[winston_v3.x.x] Add 'metadata' type to winstonFormatSubModule

pinkstrings:
[Source](https://github.com/winstonjs/logform/blob/master/metadata.js)


villesau:
Thanks!

PR #2831, opened 2018-10-12
Export all of the OpenTok types

pronebird:
For mocking (and not only) purposes it's beneficial to have all of the types available for `import`


pronebird:
Just rebased that on top of master.

villesau:
Sounds reasonable! Although in overall it's good to expose as little as possible and strive to have the exposed types to do the job as well as possible.

PR #2832, opened 2018-10-12
Make Archive.url optional in OpenTok typings

pronebird:
Based on the docs the `url` field can be null:

>url  | String | The download URL of the available MP4 file. This is only set for an archive with the status set to "available"; **for other archives, (including archives with the status "uploaded") this property is set to null**. The download URL is obfuscated, and the file is only available from the URL for 10 minutes. To generate a new URL, call the OpenTok.getArchive() or OpenTok.listArchives() method.

Reference: https://tokbox.com/developer/sdks/node/reference/Archive.html


villesau:
Thanks!

PR #2834, opened 2018-10-12
Unbreak backbone for flow 0.82+

zgotsch:
Flow 0.82 removed the ability to use functions as type annotations. This broke the backbone config, so I added some `typeof`s. Also, there was and instance where a function was shadowing a property (`defaults` on `Backbone.Model`) which Flow no longer allows. This was converted to a callable object.

Since the new typedef uses the type-spread syntax, it can only be used from Flow 0.42 onwards.


zgotsch:
Seems like the old backbone test was wrong.

villesau:
Nice, thanks for the contribution @zgotsch!

definitions/npm/backbone_v1.x.x/flow_v0.42.x-/backbone_v1.x.x.js
```diff
@@ -0,0 +1,322 @@
+// flow-typed signature: ba506076507fa389f4f828349e657080
```
pascalduez: Should be removed.

villesau: Good point, I made a PR to prevent this from happening altogether: https://github.com/flow-typed/flow-typed/pull/2835

PR #2835, opened 2018-10-13
Prevent generated comments

villesau:
Fixes, and prevents having generated comments again: https://github.com/flow-typed/flow-typed/pull/2834/files#r224958797

Also:
- Introduces new rule [no-warning-comments](https://eslint.org/docs/rules/no-warning-comments)
  - Although the comments are not warnings, this rule succeeds preventing generated comments
  - I made a feature request to Eslint to make the rule more generic as warnings are not only comments that someone (e.g we) want to disallow, please +1 it: https://github.com/eslint/eslint/issues/10976
- updates related packages
- fixes some issues where eslint parser fails to parse the file (although works in Flow..)
  - in classes `[[call]]` should be changed to `static` unnamed method and in interface `[[call]]` is not needed
- Fixes some broken libdef that was revealed by eslint

And one controversial change
- removes eslint rule that complains about weak types. There was 1600+ warnings about it and I don't think that it's providing any value. In libdefs weak types might be very valid: Many libdefs takes `any` or `Object` in by definition. E: actually now this just silences it when running the suite on CI. people who have editor integrations still sees the warnings.


definitions/npm/execa_v0.10.x/flow_v0.75.x-/execa_v0.10.x.js
```diff
@@ -79,8 +79,8 @@ declare module 'execa' {
   }
 
   declare interface Execa {
-    [[call]](file: string, args?: $ReadOnlyArray<string>, options?: Options): ThenableChildProcess;
-    [[call]](file: string, options?: Options): ThenableChildProcess;
+    (file: string, args?: $ReadOnlyArray<string>, options?: Options): ThenableChildProcess;
+    (file: string, options?: Options): ThenableChildProcess;
```
villesau: This is valid syntax, but something in our babel/eslint stack fails to parse it for some reason. Both syntaxes behave identically.

definitions/npm/react-redux_v4.x.x/flow_v0.53.x-/test_connect.js
```diff
@@ -186,4 +186,4 @@ const CC10 = connector10(C3);
 //
 // ConnectedComponent
 //
-(CC1.WrappedComponent: C1);
+(CC1.WrappedComponent: React.ComponentType<React.ElementProps<typeof C1>>);
```
villesau: essentially just `typeof` fix.

pascalduez:
Nice!

> removes eslint rule that complains about weak types. There was 1600+ warnings about it and I don't think that it's providing any value. In libdefs weak types might be very valid: Many libdefs takes any or Object in by definition.

I understand that having that much warnings on the repo doesn't feels right.  
But I find this rule quite useful while writing definitions, it's has an incentive effect to search for better types when possible. At least for me :)

pascalduez:
Agree on shell outputs, we don't even see them after some time. I was more referring to inline lint, through editors integration. 

> I've built mental model to avoid weak types.

Same here, but surely not the case fo everyone. 

Anyway, just sharing feedback here.

villesau:
Came up with better solution: Only hide warnings when testing libdefs using `--quiet`: https://eslint.org/docs/user-guide/command-line-interface

definitions/.eslintrc
```diff
@@ -4,6 +4,7 @@
     "flowtype"
   ],
   "rules": {
-    "flowtype/no-weak-types": 1
+    "flowtype/no-weak-types": 1,
+    "no-warning-comments":  [2, { "terms": ["flow-typed signature", "flow-typed version"], "location": "start"}]
```
GAntoine: 👍

PR #2837, opened 2018-10-14
Add multimatch definitions

langri-sha:
Add library definitions for `multimatch`.

See:
- https://github.com/sindresorhus/multimatch
- https://www.npmjs.com/package/multimatch

### DIFF
--- a/definitions/npm/multimatch_v2.x.x/flow_v0.37.x-/multimatch_v2.x.x.js
+++ b/definitions/npm/multimatch_v2.x.x/flow_v0.37.x-/multimatch_v2.x.x.js
@@ -1,0 +1,23 @@
+declare module "multimatch" {
+  declare type Options = {|
+    debug?: boolean,
+    dot?: boolean,
+    flipNegate?: boolean,
+    matchBase?: boolean,
+    nobrace?: boolean,
+    nocase?: boolean,
+    nocommnet?: boolean,
+    noext?: boolean,
+    noglobstar?: boolean,
+    nonegate?: boolean,
+    nonull?: boolean
+  |};
+
+  declare function multimatch(
+    paths: string | string[],
+    patterns: string | string[],
+    options?: Options
+  ): string[];
+
+  declare module.exports: typeof multimatch;
+}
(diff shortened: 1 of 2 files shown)


definitions/npm/multimatch_v2.x.x/flow_v0.37.x-/multimatch_v2.x.x.js
```diff
@@ -0,0 +1,23 @@
...
+    flipNegate: boolean,
+    matchBase: boolean,
+    nobrace: boolean,
+    nocase: boolean,
+    nocommnet: boolean,
+    noext: boolean,
+    noglobstar: boolean,
+    nonegate: boolean,
+    nonull: boolean
+  };
```
villesau: When you define options, do you really have to define all of the properties? If not, i'd define the object properties as optional: `nonull?: boolean`. I also think that it would make sense to make the Options exact type(`{|prop: boolan, ...etc|}`: It probably doesn't make any sense to pass unknown props.

definitions/npm/multimatch_v2.x.x/flow_v0.37.x-/multimatch_v2.x.x.js
```diff
@@ -0,0 +1,23 @@
...
+
+  declare function minimatch(
```
pascalduez: `multimatch`?

definitions/npm/multimatch_v2.x.x/flow_v0.37.x-/multimatch_v2.x.x.js
langri-sha: :+1: makes sense, that's exactly how the options object is treated. Will make amends :raised_hands:!

definitions/npm/multimatch_v2.x.x/flow_v0.37.x-/multimatch_v2.x.x.js
langri-sha: Ugh nice catch... I pivoted from `minimatch` to `multimatch` and left this behind.

PR #2839, opened 2018-10-14
Add typedefs for next.js v7

mdowds:
Based on the v4 definitions, with new definitions for the new features added since then.

This is my first PR so let me know if there's anything I've missed!


ussuritiger:
@mdowds thanks for the typedefs for the Next.js 7. Although after this update I am getting an error while importing `getConfig`. So
```
import getConfig from 'next/config';
```
results into
```
Cannot resolve module `next/config`.

   1| import getConfig from 'next/config';
                            ^^^^^^^^^^^^^
```
Any ideas about this?

PR #2841, opened 2018-10-15
Add punycode v2.1.x definitions

pascalduez:
See https://github.com/bestiejs/punycode.js


pascalduez:
@GAntoine Done, thanks!

PR #2844, opened 2018-10-15
fix(httpServer) Fixing issue when importing type Server from http whe…

theboyWhoCriedWoolf:
…n including express types.

### DIFF
--- a/definitions/npm/express_v4.16.x/flow_v0.32.x-/express_v4.16.x.js
+++ b/definitions/npm/express_v4.16.x/flow_v0.32.x-/express_v4.16.x.js
@@ -1,4 +1,4 @@
-import type { Server as HttpNodeServer } from "http";
+import * as http from 'http';
 import type { Socket } from "net";
 
 declare type express$RouterOptions = {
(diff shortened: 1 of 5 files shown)


theboyWhoCriedWoolf:
This PR is to solve the issue when importing type {Server} from 'http'. Noe using both flow-typed files everything works.

definitions/npm/express_v4.16.x/flow_v0.32.x-/express_v4.16.x.js
```diff
@@ -1,4 +1,4 @@
-import type { Server } from "http";
+import * as http from "http";
 import type { Socket } from "net";
```
pascalduez: Please move all those imports inside the module declaration. Otherwise they resolve to `any`.

theboyWhoCriedWoolf: Hello @pascalduez,

Not sure I understand what you are request. Please could you elaborate.

Thanks

pascalduez: Any type import, such as `import type { Socket } from "net";` declared **outside** the module declaration `declare module "express" {}` will resolve to `any`. That's apparently what you wanted to fix.  
From what I can deduce, importing not the types but the modules itself works?

theboyWhoCriedWoolf: This is correct.

Yeah, I tried adding it within the module itself however this broke all the tests. Unless I can make it accessible only to the `express$Application` class which requires the dependency?

theboyWhoCriedWoolf: @pascalduez, I have tried different variations and the original implementation seems to be the only successful one.

The issue was not getting `any`, it was that without declaring http in this way, when using this module in an application with http, importing `type { Server } from 'http'` fails. It seems to be overriding the namespace somehow.

I am open to suggestions?

Thanks

pascalduez: Maybe the overriding was due to the fact the import was "global" here in the definition.   
All those imports in lib defs are a bit of a black box, and at this point I can't assert anything without giving it a try... Sorry if I've been confusing you.

theboyWhoCriedWoolf: No problem. Please let me know if there is a solution you can think of that I can bake in as it would be really helpful in my project as at the moment I am using a workaround.

Thanks

AndrewSouthpaw:
Thanks for the review @pascalduez

PR #2848, opened 2018-10-17
fix(ramda): move object `map` declaration before array ones

JalilArfaoui:
Since Flow 0.71, it seems that Flow tries to validate `map` usage on Objects with the Array definition.
Problem exits only when using curried version.
This, fixes the issue simply by moving Object definition before Array definition


villesau:
@GAntoine I think yarn.lock changes are fine (although not related to this PR in particular): Latest yarn has introduced integrity hash to yarn.lock file.

JalilArfaoui:
@GAntoine @villesau I reverted the changes on yarn.lock. 

Sorry, I though that `integrity` properties (automatically added at `yarn install`) couldn't hurt.

villesau:
Thanks for the contribution @JalilArfaoui!

PR #2849, opened 2018-10-17
Update jest defs for dom testing library

giuband:
The custom jest matchers for dom-testing-library have been updated in the last few months, here I'm updating their definitions.
Reference: https://github.com/gnapse/jest-dom#custom-matchers


GAntoine:
Are these all still valid for version 0.23.x, or were they introduced on a later patch?

giuband:
They would be valid for any jest version, as they are introduced by the external dependency `jest-dom`. These matchers have been added progressively during the last few months, but their definitions in this file have not been updated.

PR #2854, opened 2018-10-18
[react-apollo] Update type for QueryRenderProps

ganemone:
This improves the types for QueryRenderProps with respect to `props.data`. It correctly handles `props.data` being `null`, an empty object, or `TData`.

See https://github.com/flow-typed/flow-typed/pull/2172#issuecomment-409831959 for more context

### DIFF
--- a/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
+++ b/definitions/npm/react-apollo_v2.x.x/flow_v0.58.x-/react-apollo_v2.x.x.js
@@ -1,5 +1,7 @@
 declare module "react-apollo" {
   import type { ComponentType, Element, Node } from 'react';
+
+  declare type MakeOptional = <V>(V) => ?V;
   /**
    * Copied types from Apollo Client libdef
    * Please update apollo-client libdef as well if updating these types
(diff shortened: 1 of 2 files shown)


ganemone:
cc @wzrdzl @thymikee @rasmusprentow

GAntoine:
Code-wise this LGTM, I'll wait to see if the people you pinged will weigh in before merging this.

GAntoine:
Thanks everyone for contributions and/or feedback!

jedwards1211:
> It correctly handles props.data being null

@ganemone  just wanted to point out that `void` only accepts `undefined`, not `null` (not that it's causing me problems in this case)

```
1: const foo: void = null
                     ^ Cannot assign `null` to `foo` because null [1] is incompatible with undefined [2].
```

jedwards1211:
If Apollo ever ends up getting rid of the empty object behavior, would you guys have any opposition to changing this to just `TData | void`?

jedwards1211:
Wish I had weighed in on this before merge, but my fault that I didn't.
It can get pesky to refactor the following code to add the null checks that are now necessary:
```js
        updateAfterDelete={(cache: DataProxy, {data: rawData}: FetchResult<DestroyDeviceGroupMutationData>) => {
          const data = normalizeData(rawData)
          if (!data) return
          const {destroyDeviceGroup: {deviceGroupId}} = data
```

PR #2858, opened 2018-10-19
Add markdown-table definitions

langri-sha:
Add library definitions for `markdown-table`.

See:
- https://github.com/wooorm/markdown-table
- https://www.npmjs.com/package/markdown-table


villesau:
@langri-sha It depends a bit. E.g Lodash accepts falsy values by the spec so it's natural that the libdefs also accepts them. If there is no clear spec about it, it depends on the case I think. If you think that below would catch more bugs, it might be better. If it does not, then it probably does not matter. I don't have exact answer or guideline to provide for this unfortunately.

langri-sha:
Thanks! I went for the `Options` type listed below, because I think it would provide a better experience for `markdown-table` dependents.

PR #2859, opened 2018-10-19
[@babe/register] Add configFile option

langri-sha:
Adds the `{ configFile }` option.

See https://babeljs.io/docs/en/next/options#configfile.


villesau:
Nice!

PR #2860, opened 2018-10-19
prettier: fix type of `resolveConfig.sync`

wchargin:
Summary:
As the name suggests, the function does not return a promise.

Test Plan:
Regression tests added; they fail before this change and pass after it.
Also, changing the asynchronous version to _not_ return a promise causes
the new tests to fail.

wchargin-branch: prettier-fix-resolveconfig-sync


wchargin:
Thanks for the quick reviews and merges!

PR #2863, opened 2018-10-20
[react-navigation] Type NavigationScreenProp.getParam

Ashoat:
Thanks to @Klathmon [here](https://github.com/react-navigation/react-navigation/pull/5074)!

### DIFF
--- a/definitions/npm/react-navigation_v2.x.x/flow_v0.60.x-/react-navigation_v2.x.x.js
+++ b/definitions/npm/react-navigation_v2.x.x/flow_v0.60.x-/react-navigation_v2.x.x.js
@@ -533,7 +533,16 @@ declare module 'react-navigation' {
       eventName: string,
       callback: NavigationEventCallback
     ) => NavigationEventSubscription,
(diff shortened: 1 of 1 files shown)


definitions/npm/react-navigation_v2.x.x/flow_v0.60.x-/react-navigation_v2.x.x.js
```diff
@@ -533,7 +533,16 @@ declare module 'react-navigation' {
       eventName: string,
       callback: NavigationEventCallback
     ) => NavigationEventSubscription,
-    getParam: (paramName: string, fallback?: any) => any,
```
pascalduez: Hi, 
maybe add a test?  
Also because, if I'm not mistaken, calling function/methods with generics is not available before Flow 0.72.

villesau: @pascalduez the function type parameter is just not exposed outside prior that.

PR #2865, opened 2018-10-22
Add io-ts libdef

orlandoc01:
Really excited about the work gcanti has done for Typescript in his [io-ts](https://github.com/gcanti/io-ts) library and want to have the same thing for flow. I've added a libdef I've used for this library in case anyone else wants to use this with flow.


villesau:
Nice, thanks @orlandoc01 !

PR #2869, opened 2018-10-22
sequelize: improved coverage when using sequelize.define

eidheim:


villesau:
Hi @eidheim. Looks like the tests are failing, see: https://travis-ci.org/flow-typed/flow-typed/jobs/444737459

PR #2875, opened 2018-10-23
fix(moment): allow maybe string types in moment instantiation

arahansen:
Moment allows for an empty constructor, which was accounted for. But the types did not account for passing a variable that may have an undefined/null value.


arahansen:
tests are failing, but they appear to be unrelated to my changes? Happy to debug further if someone could help point me to how my changes might affect these tests.

PR #2877, opened 2018-10-23
[@storybook/addon-links] Fix linkTo return type

pascalduez:
Following of storybooks/storybook/issues/4104

See https://github.com/storybooks/storybook/blob/release/3.4/addons/links/src/preview.js#L8


villesau:
Nice! :)

PR #2878, opened 2018-10-24
Add compose export of react-apollo

### LINKED ISSUE #2580: "compose" method from react-apollo is not included inside flow-typed
Hi flow-typed community!

I saw that  "compose" method inside react-apollo library is not declare so flow show "no export" error.

Is it not included for any reason?

Thanks so much!

KevinGrandon:
The react-apollo library exports a compose helper which is useful for building applications leveraging react-apollo. This PR adds definitions to the compose methods to avoid flow errors.

https://www.apollographql.com/docs/react/basics/setup#compose

Fixes #2580


villesau:
Thanks @KevinGrandon !

PR #2885, opened 2018-10-24
Improve type definitions for next.js and next-routes Router events

benhjames:
[next.js](https://github.com/zeit/next.js) now uses a property on `Router` called `events` to manage callbacks.

I've updated the definitions in `next.js` and `next-routes` accordingly.

`events` now uses overloads too, so you can be sure that you have a valid event name! 🎉


villesau:
Nice!

PR #2889, opened 2018-10-24
[enzyme] Updated definition - Wrapper.setProps

eoingroat:
Updated def to match docs + implementation (https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/setProps.md)


villesau:
Nice, thanks @eoingroat !

PR #2892, opened 2018-10-24
Mongoose: Add strictQuery, update strict with 'throw'

pronebird:
Both `strictQuery` and `strict` support `throw` these days.

1. It's partially outlined in docs: https://mongoosejs.com/docs/guide.html#strictQuery
2. In code: https://github.com/Automattic/mongoose/blob/master/lib/cast.js#L234


villesau:
Looks good, thanks for pointing out the docs & code!

PR #2907, opened 2018-10-25
add prettier to unify on style

eleith:
while eslint-prettier is installed, it doesn't seem like the rules are being applied as expected as the firebase file has a number of conflicts. 

i've been attempting to keep the firebase definitions updated, but i've had to suppress prettier from running because it was trying to fix a number of conflicts within the file.

by adding a prettier config with what i believe is the intent for this package, i hope to avoid the manual suppression and have prettier now fix these conflicts and make editing easier going forward.

my goal is not to have an opinion on what options are in prettier, but rather ensure whatever those opinions are from the team are consistent with the definitions that i update.

if i'm on the wrong path, please point me in the right direction. my current hypothesis is that a config is needed and hence this PR.

### DIFF
--- a/definitions/npm/firebase_v5.x.x/flow_v0.34.x-/firebase_v5.x.x.js
+++ b/definitions/npm/firebase_v5.x.x/flow_v0.34.x-/firebase_v5.x.x.js
@@ -85,9 +85,11 @@ declare type $npm$firebase$auth$Auth$Persistence = {
   +LOCAL: 'local',
   +SESSION: 'session',
   +NONE: 'none',
-}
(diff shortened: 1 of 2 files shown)


eleith:
for the firebase file in particular, it ensures consistency with semicolons and line lengths. there are many conflicts within the file, so it was unclear which way to go (with or with out semicolons) so i went with what seems to be general defaults i see in other projects

eleith:
there are no new definition rules being applied, this PR is pure formatting only

pascalduez:
See #2774

Issue #2903, opened 2018-10-25
Momentjs  >2.14.0 construct support `{}` and `[]`


gusdecool:
Momentjs starting from [v2.14.0](https://momentjs.com/docs/#/parsing/now/) support construct using empty object and empty array.
Looks like flow typed haven't updated to support this type. https://github.com/flow-typed/flow-typed/blob/77d9eda30fb387061a41ac90bd967f24f47cd6fd/definitions/npm/moment_v2.x.x/flow_v0.25.x-/moment_v2.x.x.js#L123-L131

We will need to update the allowed parameters to also support these types.

villesau:
@gusdecool would you be willing to submit a pull request? here are instructions about contributing the libdefs: https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md

gusdecool:
Sure, i will find time to make the Pull Request 🙂

gusdecool:
Pull request added 🙏

PR #2913, opened 2018-10-27
issue-2903 momentjs support empty array and object

gusdecool:
issue: https://github.com/flow-typed/flow-typed/issues/2903

Allow momentjs constructed using empty array and object. No additional test cased added as we already have the test case here https://github.com/flow-typed/flow-typed/blob/67180766627b6494c5ed464a9aefb3eeb6b9fb68/definitions/npm/moment_v2.x.x/test_moment-v2.js#L9-L10

Btw i found issue committing the PR with GPG activated and got error error .

```shell
error: gpg failed to sign the data
fatal: failed to write commit object
```

Probably because of the pre-hooks commit rule. So this PR is created using commit file that directly updated from GitHub interface.

### DIFF
--- a/definitions/npm/moment_v2.x.x/flow_v0.25.x-/moment_v2.x.x.js
+++ b/definitions/npm/moment_v2.x.x/flow_v0.25.x-/moment_v2.x.x.js
@@ -129,7 +129,11 @@ declare class moment$Moment {
       | string
       | null
       | void
+      | []
+      | {}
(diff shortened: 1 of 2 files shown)


gusdecool:
i just removed the test cased that is not related anymore with latest update of momentjs 🙏

villesau:
@gusdecool looks like the branch is conflicting. Could you rebase?

PR #2918, opened 2018-10-28
[luxon] fix zone options

NShahri:
`keepCalendarTime` is depreciated in `DateTime.setZone` and `DateTime.toUtc` and will be replaced by `keepLocalTime`.
- Docs: https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-setZone
- https://github.com/moment/luxon/blob/master/src/datetime.js#L1144
- https://github.com/moment/luxon/blob/master/src/datetime.js#L1121

`keepCalendarTime` is still supported by `luxon`:
- https://github.com/moment/luxon/blob/master/src/datetime.js#L1152

Changes:
- Add tests to cover deprecated `keepCalendarTime` and new `keepLocalTime`
- Update SetZoneOptions to support both fields


villesau:
Thanks @NShahri ! :+1:

PR #2920, opened 2018-10-29
Enable to export Component as named-export

kogai:
It seems OK to treat `react-copy-to-clipboard` as it publishes own component as named-export
https://github.com/nkbt/react-copy-to-clipboard/blob/master/src/index.js


villesau:
Nice, thanks @kogai!

PR #2927, opened 2018-10-31
source and target props do not need to be related

maxaggedon:
See this issue https://github.com/mridgway/hoist-non-react-statics/issues/47


AndrewSouthpaw:
Tests pass locally, thanks for the review @pascalduez

PR #2928, opened 2018-10-31
Mongoose populate

maxaggedon:
set the correct return type for `Mongoose$Document.popluate()` (https://github.com/Automattic/mongoose/blob/4545d44c56a1fe62fa21d783b70bb3a8ff1734cd/lib/document.js#L2483)

### DIFF
--- a/definitions/npm/mongoose_v4.x.x/flow_v0.50.x-/test_mongoose-v4.js
+++ b/definitions/npm/mongoose_v4.x.x/flow_v0.50.x-/test_mongoose-v4.js
@@ -120,16 +120,28 @@ export const UserSchema: Mongoose$Schema<Mongoose$Document> = new Schema(
       set: (v: string) => v.toLowerCase().trim(),
       required: true
     },
-    name: String
+    name: String,
+    admin: {
+      type: mongoose.Schema.Types.ObjectId,
+      ref: "Admin"
+    }
   },
   {
     timestamps: {
       createdAt: "created_at",
       updatedAt: "updated_at"
     },
-    collection: "admin"
+    collection: "user"
   }
 );
+
+const User = mongoose.model("User", UserSchema);
+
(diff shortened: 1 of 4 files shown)


villesau:
Hi @maxaggedon! Looks like the libdef is failing on 0.85.0. The reason is that 0.85.0 is much stricter with type annotations, see: https://medium.com/flow-type/asking-for-required-annotations-64d4f9c1edf8

Would you be willing to fix those issues too?

maxaggedon:
Hi @villesau ! I did the same fixes for v5 ;)

definitions/npm/mongoose_v4.x.x/flow_v0.50.x-/test_mongoose-v4.js
```diff
@@ -2,7 +2,7 @@
 
 import mongoose, { Schema } from "mongoose";
 
-export const AdminSchema = new Schema(
+export const AdminSchema: Mongoose$Schema<AdminDoc> = new Schema(
```
villesau: I think that in this case you want to use type parameters like `export const AdminSchema = new Schema<Mongoose$Schema<AdminDoc>>(...` ?

villesau: Flow added support for these few versions back.

maxaggedon: I just tried it like `new Schema<AdminDoc>(...)` but I get a `Missing type annotation for 'Mongoose$Schema'` again. Did not figure out why...

mi6gan:
Callback-version of Document#populate is not purposed to return anything at all. So, imo, it would be better and cleaner to add a separate definition to callback-less version, like this one:
```
populate(path: string): this;
```

PR #2932, opened 2018-11-01
Next.js 7.x.x router push and replace not correct type

alizhdanov:
`Router` components `push` and `replace` have URL as a string, but it can be either string or URL object. The same one as `Link` is using.

`push` and `replace` are using property `change` which accept object or string. [Source code](https://github.com/zeit/next.js/blob/canary/packages/next-server/lib/router/router.js#L146)

`Link` is using `formatUrls` which has the same logic - [Surce code](https://github.com/zeit/next.js/blob/canary/packages/next-server/lib/link.js#L47)

### DIFF
--- a/definitions/npm/next_v7.x.x/flow_v0.53.x-/next_v7.x.x.js
+++ b/definitions/npm/next_v7.x.x/flow_v0.53.x-/next_v7.x.x.js
@@ -65,21 +65,21 @@ declare module "next/head" {
   declare module.exports: Class<React$Component<any, any>>;
 }
 
-declare module "next/link" {
-  declare export type URLObject = {
-    +href?: string,
-    +protocol?: string,
-    +slashes?: boolean,
-    +auth?: string,
-    +hostname?: string,
-    +port?: string | number,
-    +host?: string,
-    +pathname?: string,
-    +search?: string,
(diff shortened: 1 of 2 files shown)


definitions/npm/next_v7.x.x/flow_v0.53.x-/next_v7.x.x.js
```diff
@@ -65,21 +65,21 @@ declare module "next/head" {
...
-    +auth?: string,
-    +hostname?: string,
-    +port?: string | number,
-    +host?: string,
-    +pathname?: string,
-    +search?: string,
-    +query?: Object,
-    +hash?: string
-  };
+declare type URLObject = {
```
AndrewSouthpaw: Can you talk about why you moved it outside the module definition? Generally we try to keep them contained.

AndrewSouthpaw: Nope that seems reasonable thanks for explaining.

AndrewSouthpaw:
It looks like there are errors in the Travis build, can you look into those?

alizhdanov:
@AndrewSouthpaw done

Issue #2880, opened 2018-10-24
Update styled-components to version 4.x


kusmierz:
Is there any major breaking change between 3 and 4 version? Could we just copy 3.x to 4.x?

pascalduez:
See #3228

PR #2933, opened 2018-11-01
Add type-defs for styled-components v4

omninonsense:
I'd like someone to thoroughly review this!

## To-do:
 - [x] Add types for `styled-components/native` (**help wanted**)
 - [x] Add `KeyFrames` to `Interpolation`
 - [x] Add `CSSRules`(`Interpolation[]`) to `Interpolation`
 - [x] Improve tests for `css` (and maybed `keyframes`?); add comments too.
 - [x] Add tests for `ref`s
 - [x] Add tests for stuff that doesn't exist (ie `styled.derp`/`styled('derp')`)
 - [x] Set lower bound to flow 0.75!
 - [x] Add support for `attrs` API

These are currently _impossible_ (to the best of my knowledge) to express these safely and without breaking the types for the "main" functionality of the library, due to Flow's lack of support for some core React features (`React.forwardRef`) this library relies on:

 - [ ] Add support for `withCompnent` API
 - [ ] Add support for `as` prop API
 - [ ] Add support for custom props on built-in components (ie `<MySpan myProp=... />`)

Please let me know if you can think of anything else to add to the list.

---

Also, not sure if it's wise to simply merge this (if the PR is good enough), and then hope someone else contributes with whatever I missed?


omninonsense:
This should hopefully address #2880

omninonsense:
Also, the justification for the horribly misleading `StyledElementType<T> = T` type is to make it work with forward refs (v4 now uses `ref` instead of `innerRef`). I forgot to add tests for that! I'll add them later or tomorrow.

omninonsense:
Travis is back! :tada:

AndrewSouthpaw:
I suggest you tag a few people who have worked on this libdef already. :-)

omninonsense:
> I suggest you tag a few people who have worked on this libdef already.

@AndrewSouthpaw Good idea!

Hi @dashed, and @shrynx! Would you two like to take a look at this too, please? :cat2:

zaguiini:
Hi! I'm trying to use it but with no success. Successfully installed as custom definitions but I'm getting errors.

How should I use it, @omninonsense?

image

omninonsense:
Hi @zaguiini!

So, the way you're trying to call the function won't work with this libdef (I don't remember if this works for v3). The main reason is that we lie about the type that's returned by `styled` in order to compensate for the fact Flow currently doesn't have a good way of correctly typing `React.forwardRef()`s (see facebook/flow#6103), which is something styled-components started using in v4 instead of `innerRef`. 

What we essentially do is:

 - for built-ins like `styled.div`, we say that it returns a string (`"div"`), but we hide/wrap this lie in `StyledElementType<T>`
- for components we say that we're returning that same component type

This is not ideal, but it means that we have working refs, and prop validation is preserved for the non built-in components.

Once forward ref support lands in flow, these can be improved a lot, to even add custom props accepted by the styled component (which the wrapped component ignores, etc), and maybe even properly typing the functions inside the template literals. Although, I think flow doesn't really understand template literals expand to `fn(strings: string[], ...interpolations : Interpolation[])`, so I didn't bother too much with having `Interpolation` accept a prop parameter in there.

omninonsense:
Also, I never answered you question. In the example you provided, I'd just do:

~~~jsx
type Props = {
  a: boolean;
}

const CardContainer = styled.div`
  /* ... */
  background-color: ${(props: WithTheme<Props>) => props.theme.light.white};
  /* ... */
`
~~~

I usually have a `theme.js` file with something like this (this is where the `WithTheme<>` mentioned above comes from):

~~~jsx
export const theme = {
  /* ... */
}

export type Theme = typeof theme
export type WithTheme<P : {} = {}> = P & {| theme: Theme |}
~~~

dashed:
> Hi @dashed, and @shrynx! Would you two like to take a look at this too, please? 🐈


@omninonsense At the moment, I don't have a project currently using styled-components v4 + flow types; and unfortunately,  I won't have the bandwidth to review this PR anytime soon.

Awesome job though! 👍

maamounapprise:
you would need to add for macro too

AndrewSouthpaw:
If we don't get a PR from people with more context within a week, I'm okay with merging it in (after reviewing with little context) and letting to be "better than nothing," and people can continue adding fixes as needed. @omninonsense if this stagnates for too long just ping again and I'll merge it in.

iDuuck:
I have a problem when using the `as` prop for the new styled-components version. I have a case, where I style the `react-router-dom`'s own `Link`. After declaring it, I want to have a link to Instagram, but not with `Link`, but with a `a` tag.

```jsx
<ListLink as="a" target="_blank" href="https://instagram.com/fintory">
  Instagram
</ListLink>
```

But now, I am getting an error, that I am not supplying the right Props:

```
Cannot create ListLink element because property to is missing in props [1] but exists in object type [2].

     client/components/Footer/index.js
     128│                   <ListLink as="a" target="_blank" href="https://www.facebook.com/fintory.agency/">
     129│                     Facebook
     130│                   </ListLink>
 [1] 131│                   <ListLink as="a" target="_blank" href="https://instagram.com/fintory">
     132│                     Instagram
     133│                   </ListLink>
     134│                   <ListLink as="a" target="_blank" href="https://twitter.com/fintory">
     135│                     Twitter
     136│                   </ListLink>

     flow-typed/npm/react-router-dom_v4.x.x.js
 [2]  20│   declare export class Link extends React$Component<{
      21│     className?: string,
      22│     to: string | LocationShape,
      23│     replace?: boolean,
      24│     children?: React$Node,
      25│   }> {}
```

zaguiini:
> I have a problem when using the `as` prop for the new styled-components version. I have a case, where I style the `react-router-dom`'s own `Link`. After declaring it, I want to have a link to Instagram, but not with `Link`, but with a `a` tag.
> 
> ```js-jsx
> <ListLink as="a" target="_blank" href="https://instagram.com/fintory">
>   Instagram
> </ListLink>
> ```
> 
> But now, I am getting an error, that I am not supplying the right Props:
> 
> ```
> Cannot create ListLink element because property to is missing in props [1] but exists in object type [2].
> 
>      client/components/Footer/index.js
>      128│                   <ListLink as="a" target="_blank" href="https://www.facebook.com/fintory.agency/">
>      129│                     Facebook
>      130│                   </ListLink>
>  [1] 131│                   <ListLink as="a" target="_blank" href="https://instagram.com/fintory">
>      132│                     Instagram
>      133│                   </ListLink>
>      134│                   <ListLink as="a" target="_blank" href="https://twitter.com/fintory">
>      135│                     Twitter
>      136│                   </ListLink>
> 
>      flow-typed/npm/react-router-dom_v4.x.x.js
>  [2]  20│   declare export class Link extends React$Component<{
>       21│     className?: string,
>       22│     to: string | LocationShape,
>       23│     replace?: boolean,
>       24│     children?: React$Node,
>       25│   }> {}
> ```

You forgot to add the `to` prop. Make it optional if you don't want to pass it.

omninonsense:
@iDuuck Try doing what @zaguiini suggested. I don't think this is related to `styled-components` typings directly (could be wrong). I wasn't aware of the `as` prop mechanism in `styled-components`.

zaguiini:
> @iDuuck Try doing what @zaguiini suggested. I don't think this is related to `styled-components` typings directly (could be wrong). I wasn't aware of the `as` prop mechanism in `styled-components`.

There is indeed one "as" mechanism: https://www.styled-components.com/docs/basics#extending-styles

omninonsense:
I decided to give up on `withComponent`, `attrs` APIs and the `as` prop API _for now_; there is no way to add them without breaking more important stuff (or adding them at all).

I'll look into adding the react native stuff later today or tomorrow.

omninonsense:
I think this is ready-ish for review. @AndrewSouthpaw, at least until we get better support for better typing from [@facebook/flow](https://github.com/facebook/flow)

SpainTrain:
For tracking React feature support in flow:

- https://github.com/facebook/react/issues/12553
- https://github.com/facebook/flow/issues/6103

omninonsense:
> Generally looks good. Unless there's a strong need, I'd collapse the separate test files into one and scope with `describe/it` blocks.

Will do! I only split them up because I assumed it would make them easier to read/understand. :smiley:

omninonsense:
@AndrewSouthpaw I used a few (random) other libdefs to see how to structure it. Is it okay like this?

omninonsense:
Had a random epiphany that `attrs` _can_ be supported.

omninonsense:
Forgot about `styled-components/native`. :eyes: 

Sorry about the noise!

omninonsense:
Added `$Supertype`, the reason is explained here: https://github.com/flow-typed/flow-typed/pull/2994#issuecomment-444673119

omninonsense:
Hi @theshortcut!

> Is it possible to add support for `displayName`, `defaultProps` etc. using the string literal approach this PR is using?

I am not sure; if I had to take a guess, I'd say no. :disappointed: 

But, I think it is possible with [Flow ≥ 0.89](), where they introduced [`React.Config<Props, DefaultProps>`](https://flow.org/en/docs/react/types/#toc-react-config) and [`React.AbstractComponent<Config, Instance>`](https://flow.org/en/docs/react/types/#toc-react-abstractcomponent), which allows us to be _a lot_ more expressive. I haven't given them a closer look yet, but from a quick glance it seems like it gives us _almost_ as much versatility/expressiveness as TypeScript has!

So, I think I will leave these as-is for the current version range, since I don't know how to improve them for versions < 0.89, and I'll make another libdef for versions ≥ 0.89 when I get the time (unless someone wants do it :grinning:).

---

@AndrewSouthpaw should I make a separate PR for Flow ≥ 0.89, or should I just put it in here? I think it might be better to put it in this PR, but I don't feel strongly about it.

theshortcut:
I've added this to a relatively large application and it is working pretty well. I did have to mark all our uses of static methods on styled components with a `$FlowFixMe` for now. The one really weird thing is that the `attrs` support seems to be a little temperamental. Example failure:

```
Cannot call styled.header.attrs with function bound to attributes because function [1] is not an object.

     59│   `};
     60│ `;
     61│
 [1] 62│ export const PageHeader = styled.header.attrs(({ heroImageUrl }) => ({
     63│   style: {
     64│     backgroundImage: `url(${heroImageUrl}?w=1&h=500&fit=crop)`,
     65│   },
     66│ }))`
     67│   align-items: center;
     68│   background-position: center;
     69│   background-size: cover;
```

but other simpler uses like this seem to be ok:

```
const SocialLink = styled(Box).attrs(() => ({ is: 'a' }))`
```

omninonsense:
Aaah, I didn't know `attrs` accepted a function, too. Thanks for figuring this out! The typedef you added looks good to me, I'll commit it sometime this week when I add the 0.89 typedefs, too.

iDuuck:
Just to chime in on this one here: I am currently working with `styled-bootstrap-grid` which isn't typed. I used this pull request as a starting base to type the library and also my current project.

This is what I came up with (for now – for `styled-bootstrap-grid`):

```js
// @flow
import { type StyledComponentType, type StyledElementType } from 'styled-components'

declare module 'styled-bootstrap-grid' {
  declare type Media = {
    xs: (declarations: Array<*>) => string,
    sm: (declarations: Array<*>) => string,
    md: (declarations: Array<*>) => string,
    lg: (declarations: Array<*>) => string,
    xl: (declarations: Array<*>) => string,
  }

  declare module.exports: {
    GridThemeProvider: StyledComponentType<StyledElementType<'div'>>,
    Container: StyledComponentType<StyledElementType<'div'>>,
    Col: StyledComponentType<StyledElementType<'div'>>,
    Row: StyledComponentType<StyledElementType<'div'>>,
    media: Media,
  }
}
```

And this is what I am using in the "Stylesheets" of my application:

```js
export const StyledRow = styled(Row)`
  width: ${(props: { columnCount: number }) => 80 * props.columnCount}vw;
  justify-content: center;

  ${media.lg`
    width: auto;
  `};
`
```

But when I am using those declarations, I am getting a red template literal, a red variable declaration, and a red "Component" when asking for the coverage of the file:

image

omninonsense:
Hi @iDuuck!

So, this line of code doesn't work the way it "should" work (not sure if this is a bug, or intentional):

~~~jsx
import { type StyledComponentType, type StyledElementType } from 'styled-components'
~~~

The `StyledComponentType` and `StyledElementType` types are `any`. To fix it, you have to move it _inside_ the `declare module …` block. See [**How do I build type definitions that depend on other definitions?**](https://github.com/flow-typed/flow-typed/wiki/FAQs#how-do-i-build-type-definitions-that-depend-on-other-definitions "FAQs · flow-typed/flow-typed Wiki").

I think we've all fallen victim to this peculiarity at some point. :)

iDuuck:
It's 2019 and this issue is still a thing?

... just joking. Wanted to ask, when this will be merged into the master (@AndrewSouthpaw)?

Thanks for the heads up, @omninonsense. Helped a lot!

AndrewSouthpaw:
I'm just waiting for @omninonsense to say it's ready to go. TBH none of us have the context required to deeply review this libdef, so just waiting on the community say-so.

omninonsense:
Sorry for the radio silence; I didn't forget about this PR! I am just a little bit swamped with stuff going on. I'll try and get it done this weekend.

egeriis:
@omninonsense Anything we can do to help this get wrapped?

omninonsense:
There is a bug I tried to fix when using the `attrs`, the return type is always `any` for some reason. I am not sure why this happens. There is no good way to work around this apart from manually annotating your types. Sorry! I _think_ this might be a limitation of flow, at least for this version range.

---

Also, the `attrs` API is "ridiculously" strict in some cases. So, when doing

~~~jsx
const StyledComp = styled(Comp).attrs({a: 3})`
  // ...
`
~~~

it will complain inexact types. It can be worked around by explicitly adding an annotation to the partial props, but it's a bit ugly.

~~~jsx
const StyledComp = styled(Comp).attrs({a: 3}: {a: number})`
  // ...
`
~~~

omninonsense:
@AndrewSouthpaw This is ready to be merged, I think. I don't think it will get any better until we port them to ≥ 0.89

omninonsense:
Oof! Those are v2 and v3 test failing for `styled-components/native`. :confused:

iDuuck:
Don't know, if I am still a `flow` noob, or I am right, but this is what I am currently getting for `flow coverage`:

image

So, it basically tells me, that this is not covered. You have an idea, why this is so, @omninonsense? `$ flow` isn't giving me an error, but the coverage seems low for me.

omninonsense:
@iDuuck The typedefs for the current Flow version don't bother with adding support for typing the props function, because Flow doesn't fully support template literals anyway. The types would be visibly more complicated, and we'd get absolutely nothing for it. So, I decided to not add it.

In other words, given `f`:

~~~jsx
const f = (strings: string[], ...interpolations: Interpolation[]) => {
 // …
}
~~~

Flow does not understand that doing this:

~~~jsx
f `2 + 2 = ${2+2}`
~~~

is the same as doing this:

~~~jsx
f(["2 + 2 = ", ""], 2 + 2)
~~~

Type-checking will only work for the second/explicit version, not for the template literal.

---

However, writing styled-components CSS like this is unwieldy. Something simple as this:


~~~jsx
styled.div`
  background-color: ${props => props.theme.background};
  font-family: ${props => props.theme.fonts.default};
`
~~~

now looks ridiculous (removed new lines from the strings for legibility):

~~~jsx
styled.div(["background-color: ", "; font-family: ", ";"], [
  props => props.theme.background,
  props => props.theme.fonts.default
])
~~~

egeriis:
@iDuuck There are no way for Flow to know what `props` is in this case, since you're creating a new React component. 

> I don't want type each props in the styled-components files.

Well, how should Flow know what type the function arguments have? 🙂 It can only defer to what `styled.span` call returns, which is `React.ComponentType<{}>`.

It doesn't look like there's type export from this libdef to do something similar to `ComponentType<Props>`, but in any case you need to specify what your props object looks like in all your React components to have Flow coverage on those.

egeriis:
@omninonsense Wouldn't it be possible to export something like `StyledComponentType`, which is a curried version of `ComponentType`? Then we could do something a la:

```js
type Props = {};
const Component: StyledComponentType<Props> = styled.div``;
```

omninonsense:
Also what @egeriis said. In your example, there is no way to know. In most cases you can get away with doing it once:

~~~jsx
styled.div`
  padding: 3px;
  /* more static css */
  ${(props: PropsWithTheme<Props>) => css`
    background-color: ${props.theme.backgroundColor}
    /* more css relying on props */
  `}
`
~~~

omninonsense:
@egeriis Yes, but see https://github.com/flow-typed/flow-typed/pull/2933#issuecomment-456471203 about flow not fully supporting template literals.

egeriis:
@omninonsense Yeah, I read it. Mind slip that I didn't understand it would affect being able to do this.

omninonsense:
I think TypeScript supports something along these lines this:

~~~jsx
const MyDiv: StyledComponentType<'div', Props> = styled.div``;
~~~

It would be nice if we could support this too someday.

egeriis:
@omninonsense Idea of effort required to fix the CI errors?

omninonsense:
I haven't actually looked at fixing it.

I rebased and pushed. _Hopefully_ that fixes it. :man_shrugging:

egeriis:
@omninonsense It didn't work ;)

pascalduez:
Tests are failing for older versions of the typdef, mostly due to recent Flow changes.  
Given it's been tested by some users here and all the work already been done, I would merge this. Opening the path for a follow-up on Flow > 0.75, okay for you @AndrewSouthpaw?

egeriis:
Which Flow versions is this libdef compatible with?

goodmind:
https://github.com/facebook/flow/issues/3936

AndrewSouthpaw:
Seems reasonable to me!

AndrewSouthpaw:
Also, would just like to say, way to go @omninonsense on your commitment on this PR.

PR #2931, opened 2018-11-01
Add libdef for query-string v3.0.x

drwlrsn:

PR #2934, opened 2018-11-02
[react-loadable] Make type application of ReactLoadable.Map more precisely

kogai:
## Why

When a programmer uses [Loadable.Map](https://github.com/jamiebuilds/react-loadable#loading-multiple-resources), he faces a situation where a loaded module hasn't any type-restriction.
(Actually, it may evaluate as `empty`)

Then, the programmer applies its own type arguments into `Loadable.Map` like below.

```js
Loadable.Map<Props, { a: () => Promise<string>, b: () => Promise<number> }>({...})
```

But this application causes type error with a message like `Cannot instantiate `MapOptions` because number [1] is incompatible with string [2] in property `b` of type argument `Loadable.Map`.`.

This causes from a restriction of type argument named `TModules`.
It defined as `MapOptions<TProps, TModules: { [key: string]: * }>`,
When type arguments applied to TModules, It seems Flow may infer resctriction of TModules like `{[string]: () => Promise<string> & () => Promise<number>}`, of course actual value like this can't exist anyay.

Therefore, it would be better to relax restriction of TModules to a type of properties as covariant.
It makes Flow can infer TModules more pragmatically.

FYI: A situation described above can reproduce more simply as below.

```js
type T<S: {[string]: * }> = S;
declare var t: T<{a: string, b: number}>;

(t.a); // Cause error
```

[Try code](https://flow.org/try/#0C4TwDgpgBAKgPAZQFxQN4G0DOwBOBLAOwHMBdFAKigF8A+KAXigQG4AoAEwgGMAbAQxzQAbgKjAU8VHxTZ8xADRQARigIBXALZKIOWm1YAKYADo+ASmZA).

### DIFF
--- a/definitions/npm/react-loadable_v5.x.x/flow_v0.56.0-/react-loadable_v5.x.x.js
+++ b/definitions/npm/react-loadable_v5.x.x/flow_v0.56.0-/react-loadable_v5.x.x.js
@@ -1,3 +1,5 @@
+declare type $Await<T, P: Promise<T>> = T;
+
 declare module 'react-loadable' {
   declare type LoadingProps = {
     isLoading: boolean,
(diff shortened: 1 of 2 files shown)


kogai:
Whoops, something occurred at Travis 🤔 
Several PRs seems to face this problem, too 👀 
(In my local environment, all specs are green)

https://github.com/flow-typed/flow-typed/pull/2933#issuecomment-435117650
https://github.com/flow-typed/flow-typed/pull/2931#issuecomment-435048407

PR #2937, opened 2018-11-02
[formik] Fixed typo in `withFormik` HOC definition

dsizomin:
This pull request fixes a typo in `withFormik` HOC definition.

### Expected
`WithFormikConfig` type and `withFormik` function are polymorphic and expect `Props` and `Values` as type parameters. 
It is expected that `withFormik<Props, Values>` receives argument of type `WithFormikConfig<Props, Values>`.

### Problem
`withFormik<Props, Values>` expects argument of type `WithFormikConfig<`**`React$ComponentType<Props>`**`, Values>`.

### Suggested fix
First type parameter of `withFormik` HOC `React$ComponentType<Props>` needs to be replaced with `Props`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ cli/package-lock.json
 # Editors
 .vscode
 .idea
+*.iml
--- a/definitions/npm/formik_v0.11.x/flow_v0.59.x-/formik_v0.11.x.js
+++ b/definitions/npm/formik_v0.11.x/flow_v0.59.x-/formik_v0.11.x.js
@@ -277,7 +277,7 @@ declare module "formik" {
   declare export function withFormik<
     Props,
     Values,
-  >(WithFormikConfig<React$ComponentType<Props>, Values>): (
+  >(WithFormikConfig<Props, Values>): (
     Component: React$ComponentType<Props>
   ) => React$ComponentType<$Diff<Props, $ObjMap<FormikProps<Values>, TypeOrVoid>>>;
 }


AndrewSouthpaw:
Thanks for your first PR! Can you talk about why it's better to remove `React$ComponentType` in `withFormikConfig` rather than add it in the others? I don't use formik so I'm missing context. 😄

dsizomin:
Hey @AndrewSouthpaw ,

Type `WithFormikConfig` doesn't have any members that rely on `React$ComponentType` type.

In fact, its generic members ([`handleSubmit`](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/formik_v0.11.x/flow_v0.59.x-/formik_v0.11.x.js#L256), [`mapPropsToValues`](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/formik_v0.11.x/flow_v0.59.x-/formik_v0.11.x.js#L261), [`validationSchema`](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/formik_v0.11.x/flow_v0.59.x-/formik_v0.11.x.js#L266) and [`validate`](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/formik_v0.11.x/flow_v0.59.x-/formik_v0.11.x.js#L272)) rely on `Props` (enhanced component props) and/or `Values` (form values).

I think using `Props` as type parameter for `WithFormikConfig` would be proper fix.

@bdrobinson @JCMais What do you guys think?

AndrewSouthpaw:
Sounds good to me, thanks for the response.

PR #2939, opened 2018-11-03
Added types and tests for react-native-fingerprint-scanner

retyui:


AndrewSouthpaw:
Tested locally, all green.

PR #2940, opened 2018-11-06
[react-with-separator] Add definitions

SpainTrain:
Based on https://github.com/LinusU/react-with-separator/blob/v1.1.0/index.d.ts

### DIFF
--- a/definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
+++ b/definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
@@ -1,0 +1,8 @@
+declare module "react-with-separator" {
+  declare module.exports: React$StatelessFunctionalComponent<{
+    children?: React$Node,
+    leading?: boolean,
+    separator: string | React$Element<*>,
+    trailing?: boolean
+  }>;
+}
--- a/definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/test_react-with-separator_v1.x.x.js
+++ b/definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/test_react-with-separator_v1.x.x.js
@@ -1,0 +1,59 @@
+import React from "react";
+import { describe, it } from "flow-typed-test";
+import WithSeparator from "react-with-separator";
+
+describe("react-with-separator", () => {
+  describe("default export WithSeparator", () => {
+    describe("Valid usage", () => {
+      it("Accepts a string as a separator", () => {
+        <WithSeparator separator="|" />;
+      });
+
+      it("Accepts a React Element as a `separator`", () => {
+        <WithSeparator separator={<div />} />;
+      });
+
+      it("Accepts all defined props", () => {
+        <WithSeparator separator="|" leading trailing>
+          {"Hello"}
+        </WithSeparator>;
+      });
+    });
+
+    describe("Invalid usage", () => {
+      it("Rejects usage without `separator` prop", () => {
+        // $ExpectError
+        <WithSeparator />;
+      });
+
+      it("Rejects usage with invalid `Object` `separator` prop", () => {
+        // $ExpectError
+        <WithSeparator separator={{}} />;
+      });
+      it("Rejects usage with invalid `null` `separator` prop", () => {
+        // $ExpectError
+        <WithSeparator separator={null} />;
+      });
+      it("Rejects usage with invalid `void` `separator` prop", () => {
+        // $ExpectError
+        <WithSeparator separator={void 0} />;
+      });
+      it(" Rejects usage with invalid `boolean` `separator` prop", () => {
+        // $ExpectError
+        <WithSeparator separator={true} />;
+      });
+      it("Rejects usage with invalid `children` prop", () => {
+        // $ExpectError
+        <WithSeparator separator="|" children={{}} />;
+      });
+      it("Rejects usage with invalid `leading` prop", () => {
+        // $ExpectError
+        <WithSeparator separator="|" leading={1} />;
+      });
+      it("Rejects usage with invalid `trailing` prop", () => {
+        // $ExpectError
+        <WithSeparator separator="|" trailing={1} />;
+      });
+    });
+  });
+});


definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
```diff
@@ -0,0 +1,26 @@
+type WithSeparatorProps = {
```
AndrewSouthpaw: Please wrap these inside the module definition.

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
```diff
@@ -0,0 +1,26 @@
...
+  children?: React$Node,
+  leading?: boolean,
+  separator: string | React$Element<any>,
```
AndrewSouthpaw: Can you explain why you went with `React$Element<any>` instead of `React$Node`?

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
```diff
@@ -0,0 +1,26 @@
...
+ * needed.
+ */
+
+// Filename aliases
+declare module "react-with-separator/index" {
+  declare module.exports: $Exports<"react-with-separator">;
+}
+declare module "react-with-separator/index.js" {
+  declare module.exports: $Exports<"react-with-separator">;
+}
```
AndrewSouthpaw: I don't think we do this normally, can you talk about a use case for defining it for `index` and `index.js`? @villesau @GAntoine have you seen this sort of thing before?

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/test_react-with-separator_v1.x.x.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+import WithSeparator from "react-with-separator";
```
AndrewSouthpaw: Please use `it/describe` blocks for better organization.

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
SpainTrain: This is auto-generated by flow-typed when it creates the stub.  I can remove it.

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
```diff
@@ -0,0 +1,26 @@
...
+};
+
+type WithSeparator = React$StatelessFunctionalComponent<WithSeparatorProps>;
+
+declare module "react-with-separator" {
+  declare module.exports: WithSeparator;
+}
+
+/**
+ * We include stubs for each file inside this npm package in case you need to
```
SpainTrain: :point_right:

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
AndrewSouthpaw: Ah, cool, do that then. We don't normally include those stubs.

GAntoine:
@AndrewSouthpaw Once your comments are addressed, I'm happy with this. 👍

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
```diff
@@ -0,0 +1,10 @@
+declare module "react-with-separator" {
+  import type { ChildrenArray } from "react";
+
+  declare module.exports: React$StatelessFunctionalComponent<{
+    children?: React$Node,
+    leading?: boolean,
+    separator: ChildrenArray<string | number | React$Element<any>>,
```
pascalduez: I think `separator: React$ElementType` is enough here, given the lib source code.

SpainTrain:
CI seems to be flaky, related to https://github.com/flow-typed/flow-typed/pull/2942

SpainTrain:
nm, retried CI enough to get it to pass :-)

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
```diff
@@ -0,0 +1,10 @@
+declare module "react-with-separator" {
+  import type { ChildrenArray } from "react";
```
AndrewSouthpaw: I'm pretty sure this will be typed as `any`, can you use the internal version?

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/test_react-with-separator_v1.x.x.js
AndrewSouthpaw: Ah. You're correct in that the test runner doesn't actually use them. We use `describe/it` from the test helper file because it just makes tests a bit more expressive. See [here](https://github.com/flow-typed/flow-typed/pull/2945/files#diff-f9434d4cd5132206735a09150d0876f8) for an example usage.

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
SpainTrain: Great point!

It looks like it is typed to not accept `string` either - https://github.com/facebook/flow/blob/41b0eab99cdc5199421f7cccad9e0c4950f8b2f9/tests/react_custom_funs/clone_element.js#L23

SpainTrain: But this case is handled manually - https://github.com/LinusU/react-with-separator/blob/0027cc509653a825cfc77e8963e17fef05acb50f/index.js#L6

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
SpainTrain: AFAIK there is no internal version for `ChildrenArray`; I did try to do that first.  There is an issue to fix this - https://github.com/facebook/flow/issues/4787

But given that `cloneElement` doesn't accept this type, I think this is now moot.

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/test_react-with-separator_v1.x.x.js
SpainTrain: :white_check_mark:

definitions/npm/react-with-separator_v1.x.x/flow_v0.72.x-/react-with-separator_v1.x.x.js
SpainTrain: `cloneElement` only accepts `React$Element` type - https://github.com/facebook/flow/blob/41b0eab99cdc5199421f7cccad9e0c4950f8b2f9/tests/react_custom_funs/clone_element.js

The library explicitly handles the string case - https://github.com/LinusU/react-with-separator/blob/0027cc509653a825cfc77e8963e17fef05acb50f/index.js#L6

SpainTrain:
:white_check_mark: 	All feedback from latest review round has been attended and responded to.

SpainTrain:
:heart:

PR #2941, opened 2018-11-06
Accepting a type T as options to support custom rendered options

dgeyer-mdsol:
Accept any type as options in react-select.
The types added in 
https://github.com/flow-typed/flow-typed/pull/2817
break the implementations where a custom optionRenderer and filterOptions are passed

### DIFF
--- a/definitions/npm/react-select_v1.x.x/flow_v0.53.x-/react-select_v1.x.x.js
+++ b/definitions/npm/react-select_v1.x.x/flow_v0.53.x-/react-select_v1.x.x.js
@@ -1,13 +1,6 @@
 declare module 'react-select' {
-  declare type SelectOption = {
-    value: string | number,
-    label: string,
-    clearableValue?: boolean,
-  };
-
-  declare type Option = {
-    [key: string]: any,
-  };
+  declare type OptionType = Object
+  declare type OptionsType = OptionType[]
 
   declare type Props = {
     // html id(s) of element(s) that should be used to describe this input (for assistive tech)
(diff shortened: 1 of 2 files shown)


pascalduez:
Hi @dgeyer-mdsol,

could you please provide samples reproducing the error, or unit tests, or links to the documentation?   
I don't think using `any` (no typechecking) could be an improvement.   

Actually in #2817 I only unclocked the definitions for Flow >= 0.53 
Granted I had the feeling those definitions are not in great shape.

dgeyer-mdsol:
@pascalduez Thanks for reviewing, I have added some tests and changed the OtionType to be Object, to support custom rendering and filtering.
This is a link to the docs https://react-select.com/props#prop-types

AndrewSouthpaw:
LGTM, @pascalduez it sounds like you have more context, once you give a thumbs up I'll merge it in.

definitions/npm/react-select_v1.x.x/flow_v0.53.x-/react-select_v1.x.x.js
```diff
@@ -1,13 +1,6 @@
...
-  declare type SelectOption = {
-    value: string | number,
-    label: string,
-    clearableValue?: boolean,
-  };
-
-  declare type Option = {
-    [key: string]: any,
-  };
+  declare type OptionType = Object
```
pascalduez: But between `Object` and `{ [key: string]: any }` there's not much differences? 
Except if one need fancy keys.

pascalduez:
Please note that it's the v1 definitions, so the doc is at https://github.com/JedWatson/react-select/tree/v1.x Although I couldn't find more precise refinements.

definitions/npm/react-select_v1.x.x/flow_v0.53.x-/react-select_v1.x.x.js
dgeyer-mdsol: Good observation, actually, I have a case when { [key: string]: any } fails, when options are a Union type. I'll modify the test

dgeyer-mdsol:
@pascalduez @AndrewSouthpaw I have modified the test so it fails when OptionType is  `{[key: string]: any}` Please have a look and let me know any feedback

pascalduez:
It still looks a bit weird to me, especially the union error.   
But should be fine since `Object` is really permissive.   
Maybe @villesau could have a look to get a third eye :)

AndrewSouthpaw:
Sounds good, thanks for the in depth review @pascalduez.

PR #2942, opened 2018-11-06
Try to fix CI issues by reducing worker count

villesau:
This should increase the hash table size from 8 megabytes to 32 megabytes and dir table size from 2 megabytes to 8 megabytes for libdef tests. Let's see if this fixes our CI issues.

See
https://flow.org/en/docs/config/options/#toc-sharedmemory-dep-table-pow-unsigned-integer
https://flow.org/en/docs/config/options/#toc-sharedmemory-hash-table-pow-unsigned-integer

E: Ended up reducing worker count so that there is only one flow instance per version. This not only fixes the CI issues but should also speed up tests in most of the cases.

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -127,13 +127,7 @@ async function getOrderedFlowBinVersions(
 
     const flowBins = apiPayload.data
       .filter(rel => {
-        if (rel.tag_name === 'v0.84.0') {
-          printSkipMessage(
-            rel.tag_name,
-            'https://github.com/facebook/flow/issues/7108',
-          );
-          return false;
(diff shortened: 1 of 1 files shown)


AndrewSouthpaw:
Nice work. @villesau can you make a change to one of the libdefs to fake trigger a libdef test run in CI and see if it passes here? If so, we can know the CI is probably working and revert the change to just be the settings file changes.

villesau:
Looks like 0.85.0 is still stuck. I have few ideas still: Reducing max worker count to 1 should reduce needed memory if that's the problem.

villesau:
@AndrewSouthpaw @GAntoine setting the worker count to 1 worked in my experiment. Removed the hash table memory increases as they didn't seem to work.

PR #2951, opened 2018-11-08
add Checkbox.Group in antd types

ivan-ha:
`Checkbox.Group` is missed in current definition
https://ant.design/components/checkbox/#Checkbox-Group


pascalduez:
A bit sad that those components don't have their props typed though.

ivanhahk01:
Ya, it will take more time.

And I saw it actually missed many of them, planning to add them during weekend when I got time, maybe in another PR?

PR #2953, opened 2018-11-08
Add zipkin defs

briancavalier:
Add defs for [zipkin](https://www.npmjs.com/package/zipkin), largely ported from that package's [existing Typescript defs](https://github.com/openzipkin/zipkin-js/blob/master/packages/zipkin/index.d.ts).

Most of the tests so far focus on places that diverged from the TS defs either out of necessity (e.g. Option type) or as a result of fixing a type that was clearly incorrect (e.g. Option flatMap, Endpoint isEmpty).  There are higher level tests in test_tracer.js derived from examples in the zipkin package's README.

Things that may be worth a closer look:

* Looking at other tests for inspiration, it wasn't entirely clear whether I should use the TDD `describe`/`it` style or not.  Happy to update these tests if `describe`/`it` is preferred.
* The nested parametric Option/Some/None export seemed tricky.  If anyone has suggestions on ways to improve this, I'd be grateful 😄.

### DIFF
--- a/definitions/npm/zipkin_v0.14.x/flow_v0.72.x-/test_model.js
+++ b/definitions/npm/zipkin_v0.14.x/flow_v0.72.x-/test_model.js
@@ -1,0 +1,6 @@
+// @flow
+
+import { model } from 'zipkin'
+
+const e1: model.Endpoint = new model.Endpoint({ serviceName: 'test1' })
+const isEmpty: boolean = e1.isEmpty()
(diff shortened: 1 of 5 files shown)


GAntoine:
Great work @briancavalier! `describe/it` are indeed the preferred styled. Let me know if you're up for making the change, and I'll wait on merging this in till it's done.

briancavalier:
Thanks for confirming, @GAntoine.  Sure, I'll update it to use `describe`/`it`.

PR #2981, opened 2018-11-24
fix(react-apollo): add onSubscriptionData prop to Subscription (test failures are unrelated)

jedwards1211:
and make children optional
https://github.com/apollographql/react-apollo/blob/master/src/Subscriptions.tsx#L31


jedwards1211:
```
 FAIL  src/lib/__tests__/libDefs-test.js
  ● Test suite failed to run
    TypeError: Cannot set property 'gracefulify' of undefined
      1 | // @flow
      2 | 
    > 3 | import fsExtra from 'fs-extra';
        | ^
      4 | 
      5 | import mkdirpCb from 'mkdirp';
      6 | 
```

AndrewSouthpaw:
Looks like you checked in a yarn.lock, maybe that's part of the issue?

jedwards1211:
Maybe, in any case that was a mistake :)

jedwards1211:
Okay, I removed the changes to `yarn.lock`

PR #2986, opened 2018-11-29
Update firebase@5.X.X libdef

necinc:
Regarding the actual firebase implementation, the `getToken` method can return a string
or a null if the `Notification.permission === 'default'`

Link to FCM source: https://github.com/firebase/firebase-js-sdk/blob/master/packages/messaging/src/controllers/base-controller.ts#L77


AndrewSouthpaw:
Fantastic, thanks for your contribution!

Issue #2983, opened 2018-11-25
[ramda_0.x.x] Change assoc type


retyui:
:heavy_check_mark: [__Try current and proposal ->__ ](https://flow.org/try/#0CYUwxgNghgTiAEA3W8C2BPAggZ2wezAC54BvAKHkvgB4AVAGngGViB5AIwCtwAXRgaWIASJgFd2PdAAcQ1IfxDps1JgD51ACgDWi4v3rIIxBthhFmASmIl4AbS3FsPGAEsAdgHMAusfgBfeAAyZnoKKjpGFngObjAeVW1deCdXTwMoI3gTM2ImK1I7B2Tnd29jRgA6KqEAUQAPKDiVVT8yVrIOsDw3J3goeABeNCxcAg0AcjBxxnGARgAmAGZp0ihiWcZ2Ynm-CwBueAB6Q4HVVfXN7cZzOaXx9o0oW3GocZ94N1FUdhAYfbJHs92G9iJ9vr9-oDJiDiqkPP8yF0ejw+vNBsMcPgwBNXjMFstGCQ1vANvAtvAdvsjiczkTiLcCWTtvcOo95s9Xu8wT8YPAAD6w0qQqDs8bArlfHn8wWeBFkY4KxVK5Uq1Vq9UazWqjqgSCwBAAM1EbjiLm6fVGYAizASOnQjhKaUM5VM5jy1kKDrh71o-iCzD2nW6vXY6KglomUzxd0JxNJ5MpB2Op3OJMuFOu9PxLIB7GeU3eKSFgY0eZeMO5EJLZfFoMlVaAA)

At the moment, this method return not what I expect reality:

### Current type:
https://github.com/flow-typed/flow-typed/blob/9902b627c46d7bce635187fe25371a43abd605d6/definitions/npm/ramda_v0.x.x/flow_v0.82.x-/ramda_v0.x.x.js#L1320-L1324

[...]

PR #2989, opened 2018-12-02
[ramda_0.x.x] Update assoc type

retyui:


retyui:
https://github.com/flow-typed/flow-typed/issues/2983

definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_object.js
```diff
@@ -19,13 +19,29 @@ const objMixed: { [k: string]: mixed } = { a: 1, c: "d" };
...
+        c: string
+      });
+
+      (_.assoc("x", { x1: 11 }, { y: ["y1"] }): {
+        x: { [string]: number },
+        y: Array<string>
+      });
+    });
+
+    it("should return sum types when passed key already has in passed object", () => {
```
AndrewSouthpaw: Can you add a comment here about how this is mostly due to limitations of flow and what we can accomplish with the implementation, rather than that it's really the *correct* type?

retyui:
I'm on vacation now and only at the end of the month can I add comments

AndrewSouthpaw:
Alright I won't worry about it for now. If you remember to do it later, great.

PR #2993, opened 2018-12-05
Update type react-onclickoutside

nifgraup:
* based on https://flow.org/en/docs/react/hoc/#toc-injecting-props-with-a-higher-order-component
* add missing props https://github.com/Pomax/react-onclickoutside/blob/master/src/index.js
* fix tests and add new


nifgraup:
the old file doesn't pass lint

flow-typed/definitions/npm/react-onclickoutside_v6.x.x/flow_v0.54.1-/react-onclickoutside_v6.x.x.js
  11:26  error  Parsing error: Unexpected token

   9 |   declare export var IGNORE_CLASS_NAME: string;
  10 | 
> 11 |   declare export default <P, S>(
     |                          ^
  12 |     BaseComponent: Class<React$Component<P, S>>,
  13 |     config?: { excludeScrollbar?: boolean }
  14 |   ) => React$ComponentType<P & OnClickOutsideProps & {

✖ 1 problem (1 error, 0 warnings)

pascalduez:
Just realised that it's a bump from `0.54.1` to `0.57.1` not that impactful I guess.

nifgraup:
@pascalduez I added the old definition again (had to disable pre-commit hook as lint failed)

AndrewSouthpaw:
cc @GAntoine

I'm not sure it'll behave correctly if we have two definitions with unbounded flow limits. Considering 0.54 is so far back, I'd be fine with dropping 0.54 - 0.56 support, so we keep things simple. Thoughts?

GAntoine:
@AndrewSouthpaw Should be fine, the original definition is bound to 0.56.

AndrewSouthpaw:
Cool let's go with that then. @nifgraup can you go ahead and update the PR?

nifgraup:
@AndrewSouthpaw what needs update? I'm not sure if you noticed my fixup ae8c979.

GAntoine:
I think this is good to go as is.

PR #2994, opened 2018-12-05
react-router-dom: fix React-related types

omninonsense:
Fix a few types related to React components

I am not sure what `$Supertype` achieves (see second commit which adds it back, made it a separate commit for the sake of clarity).

When I remove it, flow complains about a few unused suppression comments. :confounded:

### DIFF
--- a/definitions/npm/react-router-dom_v4.x.x/flow_v0.63.x-/react-router-dom_v4.x.x.js
+++ b/definitions/npm/react-router-dom_v4.x.x/flow_v0.63.x-/react-router-dom_v4.x.x.js
@@ -152,11 +152,7 @@ declare module "react-router-dom" {
     location?: Location
   |}>
 
-  declare export function withRouter<WrappedComponent: React$ComponentType<*>>(
-    Component: WrappedComponent
-  ): React$ComponentType<
-    $Diff<React$ElementConfig<WrappedComponent>, ContextRouterVoid>
-    >;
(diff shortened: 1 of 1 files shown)


omninonsense:
Hi @eddedd88!

Could you explain how `$Supertype` works in Flow and what it achieves here? As far as I can tell from a quick Google search, most things point out that it shouldn't be used. Also, it doesn't seem to be documented on the Flow page.

eddedd88:
Hi @omninonsense, to be honest I am not entirely sure what it does, but this is needed to maintain types when chaining multiple higher order components. Is common to use `withStyles` in material-ui so using it along with `withRouter` would make the component lose its types.

I got this logic from material-ui (see below), where it addresses the same thing. I didn't come up with the original logic.

https://github.com/flow-typed/flow-typed/blob/14b502d647d093dcf484b081c81174c1396e4ffc/definitions/npm/%40material-ui/core_v1.x.x/flow_v0.58.x-/core_v1.x.x.js#L2346

omninonsense:
> to be honest I am not entirely sure what it does, but this is needed to maintain types when chaining multiple higher order components.

Ah, okay. Thanks for clearing it up! :)

FezVrasta:
Note that $SuperType has now been deprecated

omninonsense:
> Note that $SuperType has now been deprecated

@FezVrasta You're right! Those should probably be part of a separate PR, since they target a different version of Flow!

PR #2997, opened 2018-12-08
Fix nullability of iterable find utilities

matthewwithanm:
All of these can return `undefined` if the input iterable is empty.


AndrewSouthpaw:
Thanks for the PR! The build is currently failing, you'll need to get the tests to pass first.

AndrewSouthpaw:
Closing due to inactive. Feel free to respond to reopen.

PR #2999, opened 2018-12-10
[luxon_v1.x.x] Add DateTime.fromSeconds

conradreuter:
https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#static-method-fromSeconds


mars-lan:
@conradreuter could you also add  https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-toSeconds?

PR #3000, opened 2018-12-11
fix(react-apollo): ensure subscriptionData types data the same way as query

jedwards1211:
@GAntoine there will be a CI build failure due to a bug in Flow 0.88 not producing an expected error in an unrelated test case.


jedwards1211:
@GAntoine thanks for the quick merge!

PR #3001, opened 2018-12-12
[jquery_v3.x.x] Changed Object to {} where needed

doberkofler:
Changed type Object to {} where needed to support the new behavior of flow 0.88 that now handles Object as an alias for any


AndrewSouthpaw:
Nice work, thanks!

PR #3005, opened 2018-12-13
[lodash_v4.x.x] lodash.values generic parameters

beaucollins:
The existing `values` definition is:

```js
    values(object?: ?Object): Array<any>;
```

So if you use an Object with an indexer definition, the type gets erased when using `lodash.values`

```js
type Ship = {| name: string, type: string |};
type ShipIndex = { [key: string]: Ship };

const ships = { 'x-wing': { name: 'X Wing' , type: 'Fighter' }};

values(ships); // Array<any>; 😢
```

By adding an overload on `values` that accepts indexed types we get to keep the correct type:

```js
values(ships); // Array<Ship>; 🎉
```


beaucollins:
Locally I'm seeing this:

```
handlesflow-typed (master) $ ./build_and_test_cli.sh 
// noise noise noise
All libdefs are named and structured correctly. (Found 755)
```

beaucollins:
@AndrewSouthpaw 

Just to confirm, these failures _already_ exist in f0bbb7e53e7b25d83fc9c53269a99f67bd5e49bd

```
$> cd cli
$> git checkout origin master
HEAD is now at f0bbb7e5 Added type definitions for 'tape-cup' package (v4.7.x) (#3015)
$> echo "// lol" >>  ../definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js 
$> node dist/cli.js run-tests --onlyChanged
```

And voila:

```
Fetching all Flow binaries...

NOISE NOISE NOISE NOISE NOISE    
    
    Found 10 errors and 14 warnings
    
```

So I _at least_ didn't increase any errors or warnings: https://travis-ci.org/flow-typed/flow-typed/jobs/470139319#L1237

beaucollins:
Tests will pass once #3020 lands.

beaucollins:
@AndrewSouthpaw test failures were fixed in #3020 so this is passing now.

I also added a test for the new type definition in 81d6c2d.

hewsut:
@villesau is this ok to go?

villesau:
@beaucollins I think this is a bit problematic, for the same reasons why native `.values` returns mixed. See e.g https://github.com/facebook/flow/issues/2221#issuecomment-323861218

For exact objects it's a different story though, there all possible values are known beforehand.

PR #3006, opened 2018-12-14
Update react-select libdef

kangax:
- Adds `Select.Async`
- Makes `Props` exact so that they're actually type-safe
- Fixes faulty autofocus test (as found by exact `Props`)
- Defines `OptionType` in tests [the same way as it is defined in react-select](https://github.com/JedWatson/react-select/blob/c22d296d50917e210836fb011ae3e565895e6440/src/types.js#L4)
- Adds a test for an invalid prop


GAntoine:
As long as it's running this project's config for prettier they're ok to keep.

Issue #2946, opened 2018-11-07
[react-redux] libdef incompatible with flow v0.85


elvorfirilmathredia:
The [todos-flow example](https://github.com/reduxjs/redux/tree/master/examples/todos-flow) just works with flow v0.78 and the current flow-typed react-redux types, but after upgrading to flow v0.85 there are a lot of flow errors I am not able to fix.

[Full repo here](https://github.com/elvorfirilmathredia/todos-flow) and [full error log here](https://github.com/elvorfirilmathredia/todos-flow/blob/master/flow.log).

```
> todos-flow@0.0.1 flow 
> flow

Error ------------------------------------------------------------------------------- src/containers/FilterLink.js:29:19

Missing type annotation for `CP`. `CP` is a type parameter declared in function type [1] and was implicitly instantiated
at call of `connect` [2].

   src/containers/FilterLink.js:29:19
                          v-------
    29| const connector = connect(
    30|   mapStateToProps,
    31|   mapDispatchToProps
    32| );
        ^ [2]

References:
   flow-typed/npm/react-redux_v5.x.x.js:133:34
                                         v
   133|   declare export function connect<
   134|     Com: ComponentType<*>,
   135|     A,
   136|     S: Object,
   137|     DP: Object,
   138|     SP: Object,
   139|     RSP: Object,
   140|     RDP: Object,
   141|     CP: $Diff<$Diff<ElementConfig<Com>, RSP>, RDP>,
   142|     ST: $Subtype<{[_: $Keys<Com>]: any}>
   143|     >(
   144|     mapStateToProps: MapStateToProps<S, SP, RSP>,
   145|     mapDispatchToProps: MapDispatchToProps<A, DP, RDP>
   146|   ): (component: Com) => ComponentType<CP & SP & DP> & $Shape<ST>;
          --------------------------------------------------------------^ [1]


Error ------------------------------------------------------------------------------- src/containers/FilterLink.js:29:19

Missing type annotation for `DP`. `DP` is a type parameter declared in function type [1] and was implicitly instantiated
at call of `connect` [2].

   src/containers/FilterLink.js:29:19
                          v-------
    29| const connector = connect(
    30|   mapStateToProps,
    31|   mapDispatchToProps
    32| );
        ^ [2]

References:
   flow-typed/npm/react-redux_v5.x.x.js:133:34
                                         v
   133|   declare export function connect<
   134|     Com: ComponentType<*>,
   135|     A,
   136|     S: Object,
[...]

Issue #2752, opened 2018-09-23
--onlyChanged matches too many libdefs


villesau:
--onlyChanged flag matches all the libdefs with same name regardless specified version. This causes unnecessary test failures if previously bad libdefs has been pushed to master: https://github.com/flow-typed/flow-typed/pull/2751

This is because of how parseRepoDirItem function works. By changing how parseRepoDirItem function works, this could be fixed.

PR #3012, opened 2018-12-15
[react-redux_v5.x.x] Add support for Flow 0.89+

peter-leonov:
Hi guys,

Thanks for making Flow better every day!

This PR addresses the issue (#2946) with the very delicious bug fix in Flow 0.85 which required us to explicitly pass types **into** the exported generic types.

These typings allow passing the new placeholder type `_` everywhere but in the first argument `Props`, which is the all props of the component being wrapped. Other types are good to specify explicitly too with a concrete type to get better error messages and avoid "inhabitant islands" of the type graph connected to nowhere.

Here is a simple example:

```js
export Component = connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(WrappedComponent);
```

**One important RFC**: do you think it's better to make type arguments to all forms of `connect()` follow the same pattern or what I've made with requiring only the needed ones is better. I made myself few typos because of sometimes it's `connect<Props, State, …>()` and sometimes it's `connect<Props, Action, …>()`.

Still a bit TODO:

- [x] decide on type arguments order
- [x] update tests
- [x] support `mapStateToProps` and `mapDispatchToProps` [factory functions](https://github.com/reduxjs/react-redux/blob/master/docs/api.md#user-content-arguments)
- [x] add comments in the typings code explaining what to do if Flow errors are misleading (good candidates are all the `{| ...OP, ...DP |}`)
- [x] add a short readme in the typings themselves on how to use `connect()`
- [x] clean up unused types and misleading comment leftovers
- [x] add example on how to type a factory state/dispatch function


peter-leonov:
@villesau @hwdtech @maximeg Dear hardcore `react-redux` typers, could you please take a look at this PR which changes typings for our beloved `react-redux` module. I found you in git blame 🤓

villesau:
Hi, nice work!

Could you provide an example e.g on redux flow-example about the usage: https://github.com/reduxjs/redux/tree/master/examples/todos-flow ? I tried to change it to use these libdefs but couldn't get them working.

villesau:
When you start to work on fixing the tests, I'd suggest to go them trough one test function at a time. Remove the rest of the tests and focus on one. Add type parameters and see how it goes, then move to next one. I guess some `$ExpectError` locations might change, but otherwise I'd expect the specs to cover most of the use cases pretty well.

peter-leonov:
@villesau thanks! Currently I'm fixing the last 🤞 bug dew to a misunderstanding of `mergeProps` nature (never used them) and then I'm all up to have a good example for todos-flow.

BTW, here is the most advanced example with both [state and dispatch](https://github.com/peter-leonov/react-redux-flow-typed/blob/master/tests/dispatch.js#L76) arguments. The rest live in the [tests](https://github.com/peter-leonov/react-redux-flow-typed/tree/master/tests) folder. Yes, a bit simplified 😅 but `export`ed to make Flow ask for explicit typings.

rkurbatov:
@peter-leonov is it possible to add support of the mapStateToProps factory in your rewritten type-definition?

hewsut:
> do you think it's better to make type arguments to all forms of connect() follow the same pattern or what I've made with requiring only the needed ones is better. 

Same pattern. Explicit over magic, please!

peter-leonov:
@rkurbatov sure, everything is possible 😅, adding to the TODO list in the PR.

peter-leonov:
Can anybody help me to get why travis reports errors in `react-redux_v4.x.x/flow_v0.53.x-`?

peter-leonov:
Regarding the order of type arguments. What do you think of this order:
```js
connect<Props, OwnProps, StateProps, DispatchProps, State, Action>(…)
```

Also, as far as I can tell for flow v0.89 only the first two are mandatory to specify, other 4 can be repaced with the new awesome type placeholder:
```js
connect<Props, OwnProps, _, _, _, _>(…)
```

fabiomcosta:
One option is to follow the same order TS uses:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/v5/index.d.ts#L77-L145

DragorWW:
@peter-leonov i check your typing, and find problem with this case:
```
type OwnProps = {
  text: string,
  ...React.ElementProps<'div'>,
};
type Props = {
  ...OwnProps,
  onClick: () => void,
}
connect<Props, OwnProps,  _, _, _, _>(null, {onClick: openModal})(Comp)
```
get error `Flow: inexact `OwnProps` [1] is incompatible with exact object type [2].`

```
  declare export function connect<-P, -OP, -SP, -DP, S, A>(
    mapStateToProps: null | void,
    mapDispatchToProps: MapDispatchToPropsFn<A, OP, DP> | DP,
    mergeProps?: null | void,
    options?: ?Options<S, OP, {||}, {| ...OP, ...DP |}>,
  ): Connector2<OP, ExtendProps<
  P, 
  {| ...OP, ...DP |} // <--- problem in this
>>;
```

peter-leonov:
@DragorWW nice catch! The reason is that there is something I really do not understand in how Flow mixes exact and inexact objects merging. In my typings and also in the tests the `OwnProps` object should be exact (tell me please if this works for you):

```js
type OwnProps = {|
  text: string,
  ...React.ElementProps<'div'>,
|};
type Props = {
  ...OwnProps,
  onClick: () => void,
}
```

This trick also fixes an annoying error like "property x is incompatible with undefined`. I guess, this `undefined` can come from an inexact object. To make typing more stable (and also support the trend on making all the object types strict by default) I just decided to make an assumption that it's ok to have `OwnProps` always strict. Adding this to the PR message.

peter-leonov:
@fabiomcosta Mimicking TS typings is a nice idea!

One issue is that they use different number of arguments for many of the forms of the function calls. I agree with @hewsut in that having implicitness here might harm readability. Also, when I was typing examples myself I found it cumbersome to always look up the right order for each arguments combination.

Another issue is that TS typings are less strict with the dispatch part. They allow using `any` for the action type (like this: `DispatchProp<any>`) while these typings require (an inferable) `Action` type argument making it impossible to dispatch garbage (this revealed some bugs in our project at work).

rkurbatov:
Hmmm... `Connector` type is only declared and not used. `Connector2` is used everywhere instead.

peter-leonov:
@rkurbatov correct, thanks for spotting this out! There is a bit to clean up in there. My goal was to make typing work for our in-company cases and also pass the very tricky tests from `flow-typed` while keeping the type API of `connect()` clean as now we have to specify the types.

Anyway, adding the clean up to the TODO section.

villesau:
> Can anybody help me to get why travis reports errors in react-redux_v4.x.x/flow_v0.53.x-?

@peter-leonov don't worry about that. It's a known issue: https://github.com/flow-typed/flow-typed/issues/2752

villesau:
@peter-leonov aren't I think `mapStateToProps` & `mapDispatchToProps` gets the same props. Should there be only one type parameter for covering both?

lukeapage:
Hi,

I'm trying this on our test project and it seems like there is a problem with default props:

```
type Props = {|
    stringProp: string,
    defaultBooleanProp: boolean,
|};

class ExampleComponent extends React.PureComponent<Props> {
    static defaultProps = {
        defaultBooleanProp: false,
    };

    render() {
        const { defaultBooleanProp, stringProp } = this.props;
        return defaultBooleanProp && stringProp;
    }
}
    type OwnProps = {||};

    const mapStateToProps = (state, props) => (
        {
            stringProp: 'a',
        }
    );

    const Connected = connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(ExampleComponent);
    const c1 = <Connected/>;
```

I think we might need to use abstract component to get the default props out

lukeapage:
Factories finally work! Amazing. I'm not sure where the documentation is but it took me a while till I found the comment about how to make them work...

```
    type OwnProps = {||};
    type StateProps = {|
        stringProp: string,
    |};

    const mapStateToPropsFactory: ({}, OwnProps) => ({}, OwnProps) => StateProps = () => (state, props) => (
        { stringProp: 'a' }
    );

    const ConnectedFn = connect<Props, OwnProps, StateProps, _, _, _>(mapStateToPropsFactory)(ExampleComponent);
```

Its not very obvious.. I guess this should live int he recat-redux repo?

peter-leonov:
Thanks for your input @lukeapage!

> …it seems like there is a problem with default props

Yes, there is one if we want to make Flow infer `OwnProps` for us. I've spent 10+ hours trying to make Flow do it, but it was always something missing (I guess a type loop somewhere). So, to make it happen still in 2018, I've tried to ask a user to provide `OwnProps` and them to be exact.

In the tests and also in our app we use it (will even more dump typings) like this:

```js
type OwnProps = {|
    defaultBooleanProp?: boolean,
|};
type Props = {|
    ...OwnProps,
    stringProp: string,
|};

class ExampleComponent extends React.PureComponent<Props> {};

connect(Props, OwnProps, _, _, _, _)(ExampleComponent);
```

If this does not work, it's a bug then :)

peter-leonov:
> Its not very obvious.. I guess this should live in the recat-redux repo?

I guess it will not hurt to add this example to the typings comment block. Thanks for pointing this out!

lukeapage:
> it seems like there is a problem with default props

I think that the problem is that I was using the components props passed also into the react component - but the Props passed into the connector are the derived props of the component, that is, with Props that have defaults are made optional?

but its more than that - if I change props to make the default prop optional I get

```
null or undefined [1] is incompatible with boolean [2] in property `defaultBooleanProp`.

   flow-typed/definitions/react-redux_v5.x.x.js:99:22
   99|     ): Connector<OP, ExtendProps<P, {| ...OP, ...SP |}>>;
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

References:
   flowexample.js:26:25
   26|     defaultBooleanProp: ?boolean,
                               ^^^^^^^^ [1]
   flowexample.js:21:25
   21|     defaultBooleanProp: boolean,
                               ^^^^^^^ [2]
```

and for reference the original problem is this:

```
Property `defaultBooleanProp` is missing in object type [1] but exists in `Props` [2] in type argument `MP`.

   flow-typed/definitions/react-redux_v5.x.x.js:99:37
   99|     ): Connector<OP, ExtendProps<P, {| ...OP, ...SP |}>>;
                                           ^^^^^^^^^^^^^^^^^^ [1]

References:
   flowexample.js:74:32
   74|     const Connected1 = connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(ExampleComponent);
                                      ^^^^^ [2]

```

lukeapage:
@peter-leonov thanks for your reply - the trouble with your example there is that defaultBooleanProp will be optional when passed into the component, which means that its type on props will be `?boolean` so it can't be used as `boolean` even though we know it will always be boolean.
For boolean this doesn't matter but if its an object, its annoying as it means we have to guard the prop even though it will never be null.

I'm more than happy to provide the typings rather than infer as a first step, I just can't get it to work with inferring them properly.

lukeapage:
The example above does actually show this.

All my problems go away if I type the default prop as optional, but then a new error appears:

```
Cannot extend property `PureComponent` [1] with `ExampleComponent` because in the return value of property `render`:
 - Either null or undefined [2] is incompatible with null [3].
 - Or property `@@iterator` is missing in null or undefined [2] but exists in `$Iterable` [4].

   flowexample.js:29:1
       v-----------------------------------------------------------
   29| class ExampleComponent extends React.PureComponent<Props> {
   30|     // for class components we have to declare defaultProps as a static member
   31|     // the corresponding members in Props will automatically be set as optional
   32|     static defaultProps = {
   33|         defaultBooleanProp: false,
   34|     };
   35|
   36|     render() {
   37|         const { defaultBooleanProp, stringProp } = this.props;
   38|         return defaultBooleanProp && stringProp;
   39|     }
   40| }
       ^

References:
   flowexample.js:29:32
   29| class ExampleComponent extends React.PureComponent<Props> {
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^ [1]
   flowexample.js:26:25
   26|     defaultBooleanProp: ?boolean,
                               ^^^^^^^^ [2]
   react.js:14:5
   14|   | null
           ^^^^ [3]
   react.js:20:5
   20|   | Iterable<?React$Node>;
           ^^^^^^^^^^^^^^^^^^^^^ [4]

```
So its asking me to cast my prop to a boolean even though I know it will never be one.

I'm wondering if we need to grab the wrapped component props through `React.AbstractComponent` ?

Sorry for being verbose, I'm no expert on this!

lukeapage:
Sorry, I give up! I've been trying to make my example from https://github.com/flow-typed/flow-typed/pull/3012#issuecomment-449338436 work in any way, even if I have to make defaultBooleanProp `?boolean` and I cannot get it to work. I've tried lots of different changes, specifying all the params and I've tried making changed to your lib def - I'm completely stuck

lukeapage:
Sorry for all the spam. Okay, so for default props, the Wrapped component props !== the redux props. It works if you define them seperately:

```
// @flow
import { connect } from 'react-redux';

type ComponentProps = {|
    stringProp: string,
    defaultBooleanProp: boolean,
|};

class ExampleComponent extends React.PureComponent<ComponentProps> {
    static defaultProps = {
        defaultBooleanProp: false,
    };

    render() {
        const { defaultBooleanProp, stringProp } = this.props;
        return defaultBooleanProp && stringProp;
    }
}

type OwnProps = {|
    defaultBooleanProp?: boolean,
|};

type Props = {|
    ...OwnProps,
    stringProp: string,
|};

const mapStateToProps = (state, props) => (
    {
        stringProp: 'a',
    }
);

const Connected = connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(ExampleComponent);
const c1 = <Connected/>;
```

I'm happy with the typings, I hope my examples help someone.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
```diff
@@ -0,0 +1,226 @@
...
+    static +WrappedComponent: WC;
+    getWrappedInstance(): React$ElementRef<WC>;
+  }
+
+  declare type Dispatch<A> = (action: A) => A;
+
+  declare type MapStateToProps<-S, -OP, +SP> =
+    | ((state: S, ownProps: OP) => SP)
+    // If you want to use the factory function but get a strange error
+    // like "function is not an object" then just type the factiry function
```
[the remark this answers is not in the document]
peter-leonov: Fixed, thanks!

peter-leonov:
@villesau

> StateProps === DispatchProps?

Oh, if it was so, we would have had the typings in no time 😕 I'll just [appeal to authorities](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/v5/index.d.ts#L77-L145) to prove that `mapStateToProps` != `mapDispatchToProps`. In short one maps state to properties while another maps dispatch to functions.

Regarding the Flow version I can say that it took me two full evenings to try making older flow versions work, but as I heard they all have some limitations in handling exact vs. inexact objects. I guess, the improvement in 0.89 is motivated by that the upcoming versions of Flow will be much more oriented to having everything exact by default (as in TypeScript).

villesau:
@peter-leonov How else it could be? :) You pass props to connected component, and those props will be passed to: mapStateToProps, mapDispatchToProps and to the connected component. If there are some advanced use cases where the  mapStateToProps props and  mapDispatchToProps props would receive different props, please enlighten me!

You might want to _expect_ different props for each: `mapStateToProps` needs prop `a` when `mapDispatchToProps` needs props `b`. But that does not mean that they receive different props: both receives `a` and `b`. I don't see a need to specify different type slice of props for each.

villesau:
@peter-leonov to my understanding, this should be valid (concentrate on different props types):

```

type CommonProps = {|a: number, b: string|};
type MapStateToPropsReturn = {|stateA: number|};
type MapDispatchToPropsReturn = {|dispatchB: string|}

const mapStateToProps = (state: State, ownProps: CommonProps) => {
  return {
    stateA: ownProps.a
  };
};

const mapDispatchToProps = (dispatch: Dispatch, ownProps: CommonProps) => {
  return {
    dispatchB: ownProps.b
  };
};

type ComponentProps = {|...CommonProps, ...MapStateToPropsReturn, ...MapDispatchToPropsReturn|};

const Component = (props: ComponentProps) => <div/>

const connector = connect(
  mapStateToProps,
  mapDispatchToProps
)(Component);
```

peter-leonov:
@villesau yes, if we rename `CommonProps` to `OwnProps` then it is exactly what these typings expect.

peter-leonov:
@villesau 

>  I don't see a need to specify different type slice of props for each.

What do you mean by this? Could you please point to the code? I guess we understand different "props" as props are everywhere here :)

villesau:
@peter-leonov ah, ok! I got mixed up wit all the different props here then. by slices I meant pattern where you would only define type `{a: number}` form `mapStateToProps` and type `{b: string}` to mapDispatchToProps. Never mind that :D

villesau:
Looks pretty good from my side :+1: I'd like to keep this still open for a while if some comments and use cases appears. People tend to use flow & redux in many different ways.

peter-leonov:
@lukeapage good news! Flow is able to understand that the connected component has a static property `defaultProps` and stop requiring those defined there, but the inference works correct only for a simple case without `mapStateToProps` argument (`connect()(WC)`). I'm unable to debug why Flow makes this decision differently in otherwise identical cases. Maybe we just need to wait for v0.90 😅

peter-leonov:
@villesau 

> Looks pretty good from my side 👍

Thanks, it just took 10 full nights to type 😂

> I'd like to keep this still open for a while

Do you think we can add these typings as they have no conflicts with the older ones, which in turn do not work with the new Flow at all? This might help those who are blocked by the lack of, imho, most important typings for a React developer, to upgrade to the latest Flow and provide a valuable feedback. Otherwise it feels like a deadlock 🤓 I'll be able to work on a next PR full time as I'v just got a 2 weeks of vacation 🎉

villesau:
One interesting, though super minor, problem I noticed when trying out with reac-tredux todos-flow example was that these does not accept `store` type. Normally the connect takes store from context, but in tests you might want to pass the store to the connected component.

peter-leonov:
Yeah, having monads would solve this problem. I dream of the day where this would be possible to catch at compile time:
```js
type StateA = {…};
const storeA = createStore<StateA>;

type StateB = {…};
const ConnectedToStateB = connect(…, StateB);

<Provider store={storeA}>
  <ConnectedToStateB />
    ^^^^^^^^^^^^^^^^^^^^ Cannot blah-blah because StateA is incompatible with StateB
</Provider>
```

villesau:
> Do you think we can add these typings as they have no conflicts with the older ones, which in turn do not work with the new Flow at all? This might help those who are blocked by the lack of, imho, most important typings for a React developer, to upgrade to the latest Flow and provide a valuable feedback. Otherwise it feels like a deadlock 🤓 I'll be able to work on a next PR full time as I'v just got a 2 weeks of vacation 🎉

Yeah I guess it makes sense. Last time when fixing react-redux, there came plenty of comments for different use cases. But the existing libdefs were just broken, and test suite was not anywhere near this comprehensive. I'll give a final look today and merge if I don't find anything weird.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+  const mapStateToProps = (state: State, props: MapStateToPropsProps) => {
+    return {
+      fromMapStateToProps: 'str' + state.a
+    }
+  }
+  type MapDispatchToPropsProps = {forMapDispatchToProps: string}
+  const mapDispatchToProps = (dispatch: *, ownProps: MapDispatchToPropsProps) => {
+    return {fromMapDispatchToProps: ownProps.forMapDispatchToProps}
+  }
+  const Connected = connect(mapStateToProps, mapDispatchToProps)(Com);
```
villesau: Shouldn't the new libdefs always require type parameters? Do you know why that's not the case here and in some other places?

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+  //$ExpectError forMergeProps is missing
+  <Connected forMapStateToProps={'data'} />;
+  //$ExpectError forMergeProps is wrong type
+  <Connected forMapStateToProps={'data'} forMapDispatchToProps={'more data'} forMergeProps={'not number'} />;
+
+  const mapDispatchToProps2 = {
+    dispatch1: () => {}
+  };
+  const Connected2 = connect(mapStateToProps, mapDispatchToProps2)(Com);
+  //$ExpectError no dispatch2
```
villesau: when taking this assertion away, the error actually expects `dispatch1`, is passed with `mapDispatchToProps2 `.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+  const mapDispatchToProps = {
+    dispatch1: () => {},
+    dispatch2: () => {}
+  };
+  const mergeProps = (stateProps, dispatchProps, ownProps: {forMergeProps: number}) => {
+    return Object.assign({}, stateProps, dispatchProps, { fromMergeProps: 123 });
+  }
+  const Connected = connect(mapStateToProps, mapDispatchToProps, mergeProps)(Com);
+  <Connected passthrough={123} forMapStateToProps="str" forMergeProps={1234}/>;
+  //$ExpectError no passthrough
```
villesau: here we don't get error from `passtrough`, but

```
Cannot create `Connected` element because property `forMapStateToProps` is missing in props [1] but exists in
    `MapStateToPropsProps` [2] in the second argument.
```

instead.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+  const Connected = connect(mapStateToProps, mapDispatchToProps)(Com);
+  <Connected passthrough={123} forMapStateToProps="str"/>;
+  //$ExpectError no passthrough
+  <Connected/>;
+
+  const mapDispatchToProps2 = {
+    dispatch1: () => {}
+  };
+  const Connected2 = connect(mapStateToProps, mapDispatchToProps2)(Com);
+  //$ExpectError no dispatch2
```
villesau: the error is about missing `dispatch1` instead of `dispatch2`, although `dispatch1` is passed. Something wrong with object types as `mapDispatchToProps` ?

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+      fromMapStateToProps: state.a
+    }
+  }
+  const mapDispatchToProps = {
+    dispatch1: () => {},
+    dispatch2: () => {}
+  };
+  const Connected = connect(mapStateToProps, mapDispatchToProps)(Com);
+  <Connected passthrough={123} forMapStateToProps="str"/>;
+  //$ExpectError no passthrough
```
villesau: getting

```
    
    Cannot create `Connected` element because property `forMapStateToProps` is missing in props [1] but exists in
    `MapStateToPropsProps` [2] in the second argument.
```

instead of error about passtrough. Might be that `<Connected/>;` should have everything except `passtrough` defined here instead.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+
+  const mapDispatchToPropsWithoutDispatch2 = {
+    dispatch1: (num: number) => {}
+  };
+  const Connected2 = connect(null, mapDispatchToPropsWithoutDispatch2)(Com);
+  //$ExpectError no dispatch2
+  <Connected2 passthrough={123}/>;
+
+  const mapDispatchToPropsWithWrongDispatch1 = {
+    //$ExpectError dispatch1 should be number
```
villesau: when removing this expectError, we get

```
annot create `Connected3` element because property `dispatch1` is missing in props [1] but exists in `Props` [2].
```

instead of wrong type in dispatch1

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+  const Connected = connect(null, mapDispatchToProps)(Com);
+  <Connected passthrough={123}/>;
+  //$ExpectError no passthrough
+  <Connected/>;
+
+  const mapDispatchToPropsWithoutDispatch2 = {
+    dispatch1: (num: number) => {}
+  };
+  const Connected2 = connect(null, mapDispatchToPropsWithoutDispatch2)(Com);
+  //$ExpectError no dispatch2
```
villesau: complaining about dispatch1 instead of dispatch2, although dispatch1 is defined

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+    }
+  }
+
+  const mapDispatchToProps = {
+    dispatch1: (num: number) => {},
+    dispatch2: () => {}
+  };
+  const Connected = connect(null, mapDispatchToProps)(Com);
+  <Connected passthrough={123}/>;
+  //$ExpectError no passthrough
```
villesau: this complains about `dispatch1` instead of `passthrough`

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+    }
+  }
+
+  type MapDispatchToPropsProps = {forMapDispatchToProps: string};
+  const mapDispatchToProps = (dispatch: *, ownProps: MapDispatchToPropsProps) => {
+    return {fromMapDispatchToProps: ownProps.forMapDispatchToProps}
+  }
+  const Connected = connect(null, mapDispatchToProps)(Com);
+  <Connected passthrough={123} forMapStateToProps={'data'} forMapDispatchToProps={'more data'} />;
+  //$ExpectError passthrough missing
```
villesau: removing this ExpectError returns 

```
    Error -------------------------------------------------------------------------------------------- test_connect.js:288:3
    
    Cannot create `Connected` element because property `fromMapDispatchToProps` is missing in props [1] but exists in
    `Props` [2].
    
       test_connect.js:288:3
       288|   <Connected forMapStateToProps={'data'} forMapDispatchToProps={'more data'} />;
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ [1]
    
    References:
       test_connect.js:272:37
       272|   class Com extends React.Component<Props> {
                                                ^^^^^ [2]
    
    
```

instead of error about passtrough

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+      fromMapStateToProps: 'str' + state.a
+    }
+  }
+  type MapDispatchToPropsProps = {forMapDispatchToProps: string}
+  const mapDispatchToProps = (dispatch: *, ownProps: MapDispatchToPropsProps) => {
+    return {fromMapDispatchToProps: ownProps.forMapDispatchToProps}
+  }
+  const Connected = connect(mapStateToProps, mapDispatchToProps)(Com);
+  <Connected passthrough={123} forMapStateToProps={'data'} forMapDispatchToProps={'more data'} />;
+  //$ExpectError passthrough missing
```
villesau: removing this causes

```
    Error -------------------------------------------------------------------------------------------- test_connect.js:260:3
    
    Cannot create `Connected` element because property `fromMapDispatchToProps` is missing in props [1] but exists in
    `Props` [2].
    
       test_connect.js:260:3
       260|   <Connected forMapStateToProps={'data'} forMapDispatchToProps={'more data'} />;
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ [1]
    
    References:
       test_connect.js:236:37
       236|   class Com extends React.Component<Props> {
                                                ^^^^^ [2]
```

instead of `passthrough` missing.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+  type State = {a: string}
+  const mapStateToProps = (state: State) => {
+    return {
+      fromStateToProps: state.a
+    }
+  }
+
+  const Connected = connect(mapStateToProps)(Com);
+  <Connected passthrough={123}/>;
+  //$ExpectError component property passthrough not found
```
villesau: getting 

```
    
    Error -------------------------------------------------------------------------------------------- test_connect.js:227:3
    
    Cannot create `Connected` element because property `fromStateToProps` is missing in props [1] but exists in `Props` [2].
    
       test_connect.js:227:3
       227|   <Connected />;
              ^^^^^^^^^^^^^ [1]
    
    References:
       test_connect.js:211:37
       211|   class Com extends React.Component<Props> {
                                                ^^^^^ [2]
    
    
    
    Found 1 error
    
```

instead of missing `passthrough`.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+    }
+  };
+
+  const Connected = connect(mapStateToProps)(Com);
+  <Connected passthrough={123} forMapStateToProps={'data'}/>;
+  //$ExpectError wrong type for passthrough
+  <Connected passthrough={''} forMapStateToProps={'data'}/>;
+  //$ExpectError wrong type for forMapStateToProps
+  <Connected passthrough={123} forMapStateToProps={321} />;
+  //$ExpectError passthrough missing
```
villesau: complains about missing `fromStateToProps` instead of passthrough.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -0,0 +1,495 @@
...
+  };
+  const mapStateToProps = (state: State, props: InputProps) => {
+    return {
+      fromStateToProps: 'str' + state.a
+    }
+  };
+
+  const Connected = connect(mapStateToProps)(Com);
+  <Connected passthrough={123} forMapStateToProps={'data'}/>;
+  //$ExpectError wrong type for passthrough
```
villesau: complains about missing `fromStateToProps` instead of wrong type for `passthrough`

villesau:
@peter-leonov I now went trough the tests. Looks like there are some discrepancies with expectations and real errors. Unfortunately the expectations are pretty binary, so it's not possible to define type of error expected so this kind of stuff gets easily missed. Are the discrepancies are because of lack of type parameters?

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
PinkaminaDianePie: Flow 0.85+ require explicit type parameter while exporting from the module. Nothing is exported here, so Flow doesn't ask about parameters. Anyway, I think it's better to export functions in tests - Flow infers much less amount of info for exported functions, so without export, we can miss something.

peter-leonov:
@villesau very nice manual checking done! I've also done it like this once and found out that the new typings fit the suppression lines in 95% of cases but not the error messages.

> Are the discrepancies are because of lack of type parameters?

Yes, in my experience this is the case. At work I'm using the strictly typed version of this PR's typings and the error messages are amazingly correct. I guess, TypeScript's approach in being explicit might not be so very wrong :)

peter-leonov:
@PinkaminaDianePie @villesau regarding the non required type parameters in the tests. Initially I started playing with these typings with everything exported in my own mini test suite. But then I've seen the tests in `flow-typed` and decided to not break the flow-way and first cover what was there already: a very flexible `connect()` typings which infer almost everything. For this I had to sacrifice the error messages to make it kinda serving both worlds.

I guess it would be cool to still support all the use cases in the exported way as in the non exported. What do you think if I just copy the `test_connect()` and export each and every connected component to prove it's working? I guess, this should have been the first thing to ask :)

villesau:
@peter-leonov what about adding the type parameters to all the connects? Having discrepancies in `// $ExpectError` kinda fight against their purpose.

peter-leonov:
I've pushed the fully exported test suit with few minor bug fixes to the original one and it shows exactly the correct error messages everywhere. But, as usual with Flow, the three mentioned missing `dispatch1` or `dispatch2` show error in the typings file, not in the test source. Flow is Flow :)

peter-leonov:
@villesau 

> what about adding the type parameters to all the connects

Now we've got both variants covered. I strongly believe the smart inference is the main selling point of Flow (at least this is the only last reason I have in front of TS evangelists) and we should test against it. Don't you agree?

villesau:
@peter-leonov then perhaps the error explanations in `// $ExpectError`s should be updated? Now they are at least a bit misleading.

peter-leonov:
I'm not sure how to change them. Current ones at leas point in the right direction. If we just change them what the current Flow version gives us in this situation we would be documenting the bugs in error messages. I expect Flow to simply error out in such type looplike  cases instead of wild guessing.

Anyway, I'm not sure about that. If you know what exactly should be done to get this PR ready to be merged, please, give precise instructions.

peter-leonov:
@villesau

> One interesting, though super minor, problem I noticed when trying out with reac-tredux todos-flow example was that these does not accept `store` type.

I'm not sure if this is what users of `flow-typed` typings for `react-redux` expect, but, it would be really nice to somehow express the dependency of `despatch()` type and action creators in `Props`. I'm currently using similar trick (giving `Dispatchable` instead of pure `Action`) in our in-house version of these typings for `redux-thunk`, but I was not sure if it's too much of a shift for a single release. I guess, too many surprises for Flow users these months 😅

What do you think?

PinkaminaDianePie:
Not sure about the contribution process to `flow-typed`, but can it be done via multiple PRs? Just merge this one with main stuff, and do other minor tweaks in separate PRs?

I checked definitions from this PR on my pet project (only for some components) and everything worked. I'd like to see this merged, so I will be able to use it everywhere :) And no matter how many unfinished things left - it's still much better than previous definitions, just because they are broken with v0.85x+. And lots of other people can't update flow mainly because of `react-redux`, so merging this PR can help them. If there are any inconsistencies, they could be fixed later.

villesau:
@peter-leonov 

> I'm not sure how to change them. Current ones, at least, point in the right direction. If we just change them to what the current Flow version gives us, in this situation, we would be just documenting the bugs in the error messages. I expect Flow one day to simply error out in such type loop looking cases instead of the wild guessing it does now.

Two approaches: A) remove the misleading tests, or B) link the `// $ExpectError`s to corresponding flow bug reports so that future contributors knows what to expect or C) change the message of `// $ExpectError` to correspond current behaviour. IMO B would be the best option here. We would have expectations and explanation on what the expect _should_ be expecting.

`react-redux` libdefs are probably the most important libdefs besides `lodash` and builting react defs for React devs. That's why it is especially important that the libdefs are as maintainable and explanatory as possible. before last rewrite about a year ago, the libdefs were in such bad shape that both libdefs _and_ tests had to be rewritten; tests no longer documented the behaviour in sane way nor were very understandable. The test suite is exceptionally large for react-redux, so keeping them maintainable is crucial for providing high quality libdefs.

villesau:
> I'm not sure if this is what users of flow-typed typings for react-redux expect, but, it would be really nice to somehow express the dependency of despatch() type and action creators in Props. I'm currently using similar trick (giving Dispatchable instead of pure Action) in our in-house version of these typings for redux-thunk, but I was not sure if it's too much of a shift for a single release. I guess, too many surprises for Flow users these months 😅

If you don't have clear way to fix this smoothly, this is by no means any kind of blocker. It can be sorted out later.

But I would imagine that having some sort of `{...Props, {store?: ?StoreType}}` could help here. The store type is anyways one of the type parameters so no need to add new type parameters at least.

peter-leonov:
> C) change the message of `// $ExpectError` to correspond current behaviour.

Done.

> IMO B would be the best option here.

Completely correct. IMHO we can do this in another PR with the distilled bug examples reported and linked.

> `react-redux` libdefs are probably the most important libdefs besides `lodash` and builtin react defs for React devs. That's why it is especially important that the libdefs are as maintainable and explanatory as possible.

I undersign every word! And I tried my best to make everything documented in this PR and in the libdefs sources.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -308,22 +342,33 @@ function testMapDispatchToPropsPassesActionCreators() {
   };
   const Connected = connect(null, mapDispatchToProps)(Com);
   <Connected passthrough={123}/>;
+  // Here Flow reports that `dispatch1` is missing from props,
+  // while the real error is still the missing passthroug property,
+  // giving the passthrough property fixes the wrongly titled error.
+  //$ExpectError no passthrough
```
villesau: duplicate `//$ExpectError no passthrough` causes unnecessary errors. Once this is fixed, the PR is good to go!

peter-leonov: Sorry, did not get, what needs to be changed?

villesau: @peter-leonov one line below there is second `//$ExpectError`: This one does not actually match anything.

peter-leonov: Now I see 🤦‍♀️ Fixed!

peter-leonov:
@villesau

> One interesting, though super minor, problem I noticed when trying out with reac-tredux todos-flow example was that these does not accept `store` type.

I've replaced `Action` type with `Dispatch` type as this changes nothing for this stage, but gives the opportunity to fix that "super minor problem" you've mentioned above and also makes these libdefs more flexible for `redux-thunk` users.

I hope, we are totally fine by now 😎

PinkaminaDianePie:
Thank you so much, guys! I was checking this thread much more often than my social networks :D

rkurbatov:
Awesome work! We have a huge project that uses flow so I have something to do during upcoming holidays :D

peter-leonov:
Thank a lot for all of you guys for your help and support! Bug reports are very welcome 😅

P.S. Updated the PR description with two examples.

peter-leonov:
> @villesau
> 
> > One interesting, though super minor, problem I noticed when trying out with reac-tredux todos-flow example was that these does not accept `store` type.
> 
> I've replaced `Action` type with `Dispatch` type as this changes nothing for this stage, but gives the opportunity to fix that "super minor problem" you've mentioned above and also makes these libdefs more flexible for `redux-thunk` users.

For those who are interested in stricter action type checking or use `redux-thunk` a lot with the new libdefs, please, also follow the https://github.com/flow-typed/flow-typed/pull/3035.

Ashoat:
Thanks for the new and improved libdef! Unfortunately you stripped away the default exports, which breaks compatibility with the CommonJS distro of `react-redux`. Since Node still can't handle the ES6 distro of `react-redux` (since it attempts a named import of `react`, which is CommonJS-only, and Node can't handle named imports of CommonJS modules), this breaks compatibility of this libdef with Node's experimental ES6 support. More context is available [here](https://github.com/flow-typed/flow-typed/pull/2018).

I'll put up a PR momentarily to reintroduce the default export.

kangax:
@peter-leonov I'm getting "Missing type annotation for OP. OP is a type parameter declared in function type [1] and was implicitly instantiated at call of connect [2]." with quite a lot of `connect()`'s in our app. Adding types to MSP (and/or MDP) doesn't change anything. Anything else I can do to fix it?

This is on Flow 0.89 + the redux typedefs from this PR.

AlimovSV:
@peter-leonov Will you have the chance to add support of react-redux@6?

peter-leonov:
@all: I'm having an exam later this month so will not be of much help, sorry.

@kangax could you please show some portion of the code producing this error?

@AlimovSV I have not planed it, but will be happy to take a look at your PR ;)

fabiomcosta:
@kangax I think this is happening because the new types require using Flow types when calling `connect`.
See https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js#L6-L8

`connect<Props, OwnProps, _, _, _, _>(…)`

Where you can define `Props` and `OwnProps` as:
```
type OwnProps = {|
  // define props that don't come from your redux store
|};
type Props = {|
  ...OwnProps,
  // define props that don't come from your redux store
|};
```

On the link you'll also see that OP -> OwnProps.
I honestly think we should just use the full names on the type definitions, it would make the errors messages a bit more understandable.

PR #3017, opened 2018-12-19
Lodash tests failing with flow v0.89.0

beaucollins:
I tried adding an overload of `lodash.values` (#3005 ) however the [CI failed](https://travis-ci.org/flow-typed/flow-typed/jobs/470129954#L938).

I found it highly suspicious that a single addition to `lodash.values` would cause this so I tested a branch that introduces no changes and the same errors exist.

This PR is to show the failures exist in `flow-typed/flow-typed master`.


beaucollins:
Will be fixed with #3019

PR #3020, opened 2018-12-19
Fixing lodash defs for flow 0.88.0+

beaucollins:
Now that `Function` and `Object` are equivalent to `any` type many Lodash typedefs no longer work as expected.

For example this is no longer valid:

```
isObject(value: Object): true;
isObject(value: any): boolean;
```

Also, now that `Function` is the same as `any`, many of the `null` tests failed because with v0.88.0 and onward this:

```js
unary(Function): Function;
```

becomes the same as:

```js
unary(any): any;
```

We can fix some of these by defining the types a bit better. For instance this better describes what `.unary` actually does.

```js
unary<T, R>((T, ...any[]) => R): (T) => R;
```


definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -158,7 +158,7 @@ declare module "lodash" {
 
   declare type NestedArray<T> = Array<Array<T>>;
 
-  declare type matchesIterateeShorthand = Object;
+  declare type matchesIterateeShorthand = {[id: any]: any};
```
beaucollins: Since `Object` now means `any` we can require a very liberal indexed object definition instead.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -746,8 +746,8 @@ declare module "lodash" {
     shuffle<T>(array: ?Array<T>): Array<T>;
     shuffle<V, T: Object>(object: T): Array<V>;
     size(collection: $ReadOnlyArray<any> | Object | string): number;
-    some<T>(array: ?$ReadOnlyArray<T>, predicate?: Predicate<T>): boolean;
```
beaucollins: By making the `null` case overload first, the null case test passes successfully.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -776,22 +776,22 @@ declare module "lodash" {
...
+    flip<R>(func: (...any[]) => R): (...any[]) => R;
+    memoize<A, R>(func: (...A) => R, resolver?: (...A) => mixed): (...A) => R;
+    negate<A, R>(predicate: (...A) => R): (...A) => boolean;
+    once<A, R>(func: (...A) => R): (...A) => R;
     overArgs(func?: ?Function, ...transforms?: Array<Function>): Function;
     overArgs(func?: ?Function, transforms?: ?Array<Function>): Function;
-    partial(func: Function, ...partials: any[]): Function;
-    partialRight(func: Function, ...partials: Array<any>): Function;
+    partial<R>(func: (...any[]) => R, ...partials: any[]): (...any[]) => R;
+    partialRight(func: (...any[]) => any, ...partials: Array<any>): Function;
```
beaucollins: `Function` is now `any`, this at least requires a valid function.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -891,12 +891,10 @@ declare module "lodash" {
     isNull(value: any): false;
     isNumber(value: number): true;
     isNumber(value: any): false;
-    isObject(value: Object): true;
-    isObject(value: any): false;
+    isObject(value: any): boolean;
```
beaucollins: `Object` is now `any` so `isObject(value: Object): true` no longer works.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -1066,6 +1064,7 @@ declare module "lodash" {
     ): Object;
     at(object?: ?Object, ...paths: Array<string>): Array<any>;
     at(object?: ?Object, paths: Array<string>): Array<any>;
+    create(prototype: void | null, properties: void | null): {};
```
beaucollins: Explicitly adding the `null` case for `create` which allows the test to pass.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -1283,8 +1282,8 @@ declare module "lodash" {
       iteratee?: ?OIteratee<*>,
       accumulator?: ?any
     ): {};
-    unset(object: Object, path?: ?Array<string> | ?string): boolean;
     unset(object: void | null, path?: ?Array<string> | ?string): true;
+    unset(object: Object, path?: ?Array<string> | ?string): boolean;
```
beaucollins: Moving the `null` case first allows the null test to pass.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -1393,13 +1392,13 @@ declare module "lodash" {
     cond(pairs?: ?NestedArray<Function>): Function;
     conforms(source?: ?Object): Function;
     constant<T>(value: T): () => T;
-    defaultTo<T1: string | boolean | Object, T2>(
+    defaultTo<T1: void | null, T2>(value: T1, defaultValue: T2): T2;
```
beaucollins: Move the `null` case to appear first and remove `Object` from the non-null case since it means `any`.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -1608,7 +1607,7 @@ declare module "lodash/fp" {
 
   declare type NestedArray<T> = Array<Array<T>>;
 
-  declare type matchesIterateeShorthand = Object;
+  declare type matchesIterateeShorthand = {[string | number]: any};
```
beaucollins: Now that `Object` means `any` an indexed object type more accurately describes that the `matchesIterateeShorthand` type can be.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -2358,7 +2357,7 @@ declare module "lodash/fp" {
     delay(wait: number, func: Function): TimeoutID;
     flip(func: Function): Function;
     memoize<F: Function>(func: F): F;
-    negate(predicate: Function): Function;
+    negate<A, R>(predicate: (...A) => R): (...A) => R;
```
beaucollins: Should be:

```suggestion
    negate<A, R>(predicate: (...A) => R): (...A) => boolean;
```

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -3154,18 +3152,18 @@ declare module "lodash/fp" {
     cond(pairs: NestedArray<Function>): Function;
     constant<T>(value: T): () => T;
     always<T>(value: T): () => T;
-    defaultTo<T1: string | boolean | Object, T2>(
+    defaultTo<T1: void | null, T2>(defaultValue: T2): (value: T1) => T2;
```
beaucollins: Place the `null` cases first and remove `Object` since it means `any`.

definitions/npm/lodash_v4.x.x/flow_v0.63.x/lodash_v4.x.x.js
```diff
@@ -6081,3 +6079,4 @@ declare module "lodash/fp/toPath" {
 declare module "lodash/fp/uniqueId" {
   declare module.exports: $PropertyType<$Exports<"lodash/fp">, "uniqueId">;
 }
+// lol
```
beaucollins:
```suggestion
```

definitions/npm/lodash_v4.x.x/flow_v0.63.x/test_lodash-v4.x.x.js
```diff
@@ -106,9 +106,6 @@ opaque type O = string;
 const v: { [O]: number } = { x: 1, y: 2 };
 find(v, { x: 3 });
 
-// $ExpectError undefined. This type is incompatible with object type.
-var result: Object = find(users, "active");
```
beaucollins: Since `Object` now means `any` this will never cause an error.

villesau:
@beaucollins what's the status of this? Good to go from your side?

villesau:
@beaucollins some feedback from quick test:
rather than `debounce<A, R>(func: (...A) => R, wait?: number, options?: DebounceOptions): (...A) => R;`

this seems to work better:

`debounce<F: (...any[]) => any>(func: F, wait?: number, options?: DebounceOptions): F`

Same apply for others where function is restructured. Rather have it as a single type. That way Flow does not require type parameters.

beaucollins:
@villesau perhaps do both. I believe the typed parameters and return values should carry through and then it can fall back to the `F: (...any[]) => any)` one.

villesau:
@beaucollins `F: (...any[]) => any)` carries trough. `F` is type parameter that carries the type, and `: (...any[]) => any)` restricts it. It carries the function parameters and return value, so it does everything you would expect `(...A) => R` to do. Is there some reason why you would want to use `(...A) => R` rather than single type parameter for the function?

beaucollins:
@villesau you're suggestion is implemented. These changes should be "good enough" to keep us going.

There are a lot of definitions in this library that still use `Function` and `Object` so there's more work to do but I'll leave them for future PR's.

villesau:
Nice, thanks for the contribution @beaucollins !

PR #3025, opened 2018-12-20
Add webpack devServer configuration options

langri-sha:
Add options for the `{ devServer }` configuration. See [documentation](https://webpack.js.org/configuration/dev-server/#devserver).

Extended the simple shape test to make sure all the `{ devServer }` options are optional.


langri-sha:
I found the Prettier rules for ESLint here and will try to match up https://github.com/flow-typed/flow-typed/blob/f0bbb7e53e7b25d83fc9c53269a99f67bd5e49bd/cli/.eslintrc#L16-L22 :+1:.

langri-sha:
Thanks @pascalduez, I've amended the changes and have a cleaner diff now.

PR #3026, opened 2018-12-20
fix(sequelize): add a Transaction.afterCommit typedef

jedwards1211:
It has been ported from v5 back to v4.42.0: https://github.com/sequelize/sequelize/commit/d2428dd580c1adb0a4c763a30499da2d87b19b3d
Note that because of the drawbacks of how flow-typed typedef versioning is designed, it will fail to give users of < 4.42.0 an error if they use an `afterCommit` hook, even though it should because it isn't present in earlier versions. ¯\_(ツ)_/¯
@GAntoine once again, there are unrelated buggy flow errors in 0.88 and 0.89 😕

### DIFF
--- a/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/sequelize_v4.x.x.js
+++ b/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/sequelize_v4.x.x.js
@@ -7208,6 +7208,11 @@ declare module "sequelize" {
     rollback(): Promise<void>,
 
     /**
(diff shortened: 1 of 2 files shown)


jedwards1211:
@AndrewSouthpaw as I said:
> once again, there are unrelated buggy flow errors in 0.88 and 0.89 😕
There is nothing I can do here about these bugs, unfortunately.

AndrewSouthpaw:
Derp, sorry, missed that.

PR #3031, opened 2018-12-24
[http-errors] Allow message in constructor

102:
Looks like constructor actually allows custom error message https://github.com/jshttp/http-errors/blob/master/index.js#L135 & https://github.com/jshttp/http-errors/blob/master/index.js#L183


villesau:
Nice, thanks!

PR #3034, opened 2018-12-26
[sequelize] Add findByPk function to Model (#3033)

### LINKED ISSUE #3033: [sequelize] Add findByPk to Model
In the current version of Sequelize, Model.findById is deprecated and replaced with Model.findByPk. But when i use findByPk in my code, Flow warns that it isn't defined on Model.

Scimonster:
Fix #3033


AndrewSouthpaw:
Build is broken but seems unrelated to the PR, merging...

PR #3035, opened 2018-12-30
[react-redux_v5.x.x] Connect the `Dispatch` type with actions in dispatch params

peter-leonov:
Long time no see :)

This PR solves the problem with actions in dispatch props have currently no connection to the `Dispatch` type argument of the `connect()` function. In practice this means that Flow will not stop us from returning any random type from an action creator given in `mapDispatchToProps`. Here is an example:

```js
// first valid action with action creator of our app
type ValidAction1 = {|
  +type: 'VALID_ACTION_1'
|}
const validAction1 = () => ({type 'VALID_ACTION_1'})
// second valid action with action creator of our app
type ValidAction2 = {|
  +type: 'VALID_ACTION_2'
|}
const validAction2 = () => ({type 'VALID_ACTION_1'})

// combined type of all valid actions of our app
type Action = ValidAction1 | ValidAction2

// the dispatch type of our application, which should accept only valid actions of our app
// (vanilla dispatch() returns action as is)
type Dispatch = Action => Action

// here is an example of an invalid action creator (possibly a typo or copy-paste problem)
const invalidAction = () => { foo: 'bar' }

// no own props
type OwnProps = {||}
// this component is going to only dispatch things
type Props = {
  action1: typeof action1,
  action2: typeof action2,
  invalidAction: typeof invalidAction
}

// ... other boilerplate code here ...

// ERROR: currently at this point Flow does not warn us about the `invalidAction`, but is should warn
connect(Props, OwnProps, _, _, _, Dispatch)
```

We do check the `mapStateToProps` to get the real state type (if it's present in the code and given or inferable from the call) but fo dispatch the `Dispatch` type is just bogus and was added to be typed in the bright future. Hopefully, it is now :)

Also, I've fixed some minor things:
* not passing dispatch to `config` and `mergeProps` arguments if no `mapStateToProps` parameter passed;
* fixed type arguments order for the `mergeProps` part of typings (was a typo);
* pass the empty object type `{||}` as `stateProps` to `mergeProps` and `config` instead of wrongly inferred `empty` type for the non present `StateProps` and `DispatchProps`.

As this PR makes the typings significantly stricter (and safer) but a bit harder to drop into a random project (especially, after the habits we've developed with the very forgiving mode of Flow <v0.85) we should discuss if and how to make this change smooth 😄


definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
```diff
@@ -52,6 +52,8 @@ declare module "react-redux" {
     // and provide the StateProps type to the SP type parameter.
     | ((state: S, ownProps: OP) => (state: S, ownProps: OP) => SP);
 
+  declare type Bind<D> = <A, R>((...A) => R) => (...A) => $Call<D, R>;
```
peter-leonov: This is what [`bindActionCreators()`](https://redux.js.org/api/bindactioncreators) actually does.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connectExported.js
```diff
@@ -601,3 +601,770 @@ function testHoistConnectedComponent() {
...
+  }
+
+  const Connected = connect<Props, OwnProps, _,_,_,Dispatch>(
+    mapStateToProps,
+    mapDispatchToPropsFn,
+    mergeProps
+  )(Com);
+  e.push(Connected);
+  <Connected passthrough="foo" />;
+}
```
peter-leonov: What do you think, is it time to split the tests for `connect` in multiple files by topic? Like `test_connectMergeProps.js`, `test_connectStateOnly.js`, `test_connectConfig.js`.

EugeneZ:
This looks great. Massive thanks for all the work you've been doing on these typings! I think there is one overload of `connect` missing, the one where `mapStateToProps` is provided, *and* the mapDispatchToProps is a map. (And I guess by that logic a corresponding overload where the third and fourth arguments are provided.)

In our code, at least, this seems to be the most commonly used overload.

peter-leonov:
@EugeneZ 🤦‍♀️ yes, you're right. It was not exactly forgotten, but embedded into the function case and thus typed incorrectly in case of thunks. Thanks a lot for trying out the yet very raw material! And happy new year 🎄🎅🍋!

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -548,4 +548,140 @@ function testHoistConnectedComponent() {
   <Connected passthrough={123} passthroughWithDefaultProp={456} forMapStateToProps={'data'}/>;
   // OK with declared static property
   Connected.myStatic;
+  //$ExpectError property `myStatic1` is missing in statics
```
villesau: you mean `notStatic`? :)

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connectExported.js
villesau: I think that could make sense! Good idea :+1:

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
peter-leonov: Eagle eye achievement unlocked!

peter-leonov:
Rebased to hopefully get green tests. Will split into files asap.

villesau:
Nice, GJ!

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
Hypnosphi: Should the corresponding change be made to [redux libdef](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/redux_v3.x.x/flow_v0.55.x-/redux_v3.x.x.js#L50-L51) as well?

PR #3037, opened 2019-01-02
[bcrypt_v{2,3}.x.x] Make bcrypt work for 2.x and 3.x versions

nvie:


villesau:
@nvie flow-typed expects major version to be present, so the proper way in this case is to copy the libdefs over with new name. It's reasonable since normally major versions has breaking api changes.

nvie:
Thanks for the suggestion, @villesau — I'm still learning the rules 👍  This is now addressed in 4fa5db21e26c4f929e94df121562c775fcd7099d.

nvie:
🙌

PR #3038, opened 2019-01-02
Copied Jasmine@2.4.x to Jasmine@2.9.x

Jason94:
- I did not change the definitions at all because 2.9 should be
  backwards compatible with 2.4.
- Verified that the bindings work for simple tests with the default
  expectations.

### DIFF
--- a/definitions/npm/jasmine_v2.9.x/flow_v0.25.x-/jasmine_v2.9.x.js
+++ b/definitions/npm/jasmine_v2.9.x/flow_v0.25.x-/jasmine_v2.9.x.js
@@ -1,0 +1,107 @@
+/* eslint-disable flowtype/no-weak-types */
+
+type JasmineExpectType = {
+  not: JasmineExpectType,
+  toBe(value: mixed): void,
+  toBeCloseTo(num: number, delta: mixed): void,
+  toBeDefined(): void,
+  toBeFalsy(): void,
+  toBeGreaterThan(number: number): void,
+  toBeLessThan(number: number): void,
+  toBeNull(): void,
+  toBeTruthy(): void,
+  toBeUndefined(): void,
+  toContain(str: string): void,
+  toEqual(value: mixed): void,
+  toHaveBeenCalled(): void,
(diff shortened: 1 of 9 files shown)


Jason94:
I'm not sure exactly why this is failing. The warnings are occurring in the 2.4 folder as well, which I did not change, so I suspect that the issue was existing. 

There are 10 warnings that each of the `$ExpectError` tags are not doing anything. For example:
```
Warning --------------------------------------------------------------------------------- test_jasmine_v2.4.x_it.js:33:1
   
   Unused suppression comment.
   
      33| // $ExpectError number. This type is incompatible with function type.
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

However, if I remove the suppression then I get an actual error:
```
 * jasmine_v2.4.x/flow_v0.25.x- (flow-v0.87.0): Unexpected Flow errors(2):
   Error ---------------------------------------------------------------------------------- test_jasmine_v2.4.x_it.js:34:13
   
   Cannot call `xit` with `12` bound to `fn` because number [1] is incompatible with function type [2].
   
      test_jasmine_v2.4.x_it.js:34:13
      34| xit('desc', 12);
                      ^^ [1]
   
   References:
      jasmine_v2.4.x.js:35:40
      35| declare function xit(name: string, fn: Function): void;
                                                 ^^^^^^^^ [2]
   
   
   
   Found 1 error
```
Is it possible that the functions from jasmine are colliding with the syntatic sugar functions provided by flow-typed (https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#use-describe-and-it-blocks-to-limit-scope)?

AndrewSouthpaw:
Nope, the flow-typed sugar functions must be imported so they don't collide with jasmine/jest/etc.

My guess is that the behavior of `Function` changed in `0.89`, since I'm noticing the errors only crop up in the latest flow verison. I know they've been making a bunch of changes lately so I'd start digging in there.

AndrewSouthpaw:

PR #3041, opened 2019-01-04
[lodash_v4.x.x] Add potentially-void return types for lodash functions

nvie:
Functions like `maxBy()`, `first()`, `last()`, etc. may return `undefined` potentially, typically when called on the empty array.  This is a common case that requires explicit handling at call sites when using these APIs.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
@@ -260,7 +260,7 @@ declare module "lodash" {
       fromIndex?: ?number
     ): -1;
     // alias of _.head
-    first<T>(array: ?$ReadOnlyArray<T>): T;
+    first<T>(array: ?$ReadOnlyArray<T>): T | void;
     flatten<T, X>(array?: ?Array<Array<T> | X>): Array<T | X>;
     flattenDeep<T>(array?: ?(any[])): Array<T>;
     flattenDepth(array?: ?(any[]), depth?: ?number): any[];
@@ -318,7 +318,7 @@ declare module "lodash" {
     ): Array<T>;
     join<T>(array: Array<T>, separator?: ?string): string;
     join<T>(array: void | null, separator?: ?string): "";
-    last<T>(array: ?$ReadOnlyArray<T>): T;
+    last<T>(array: ?$ReadOnlyArray<T>): T | void;
     lastIndexOf<T>(array: Array<T>, value?: ?T, fromIndex?: ?number): number;
     lastIndexOf<T>(array: void | null, value?: ?T, fromIndex?: ?number): -1;
     nth<T>(array: T[], n?: ?number): T;
@@ -935,12 +935,12 @@ declare module "lodash" {
     ceil(number: number, precision?: number): number;
     divide(dividend: number, divisor: number): number;
     floor(number: number, precision?: number): number;
-    max<T>(array: ?Array<T>): T;
-    maxBy<T>(array: ?$ReadOnlyArray<T>, iteratee?: Iteratee<T>): T;
+    max<T>(array: ?Array<T>): T | void;
+    maxBy<T>(array: ?$ReadOnlyArray<T>, iteratee?: Iteratee<T>): T | void;
     mean(array: Array<*>): number;
     meanBy<T>(array: Array<T>, iteratee?: Iteratee<T>): number;
-    min<T>(array: ?Array<T>): T;
-    minBy<T>(array: ?$ReadOnlyArray<T>, iteratee?: Iteratee<T>): T;
+    min<T>(array: ?Array<T>): T | void;
+    minBy<T>(array: ?$ReadOnlyArray<T>, iteratee?: Iteratee<T>): T | void;
     multiply(multiplier: number, multiplicand: number): number;
     round(number: number, precision?: number): number;
     subtract(minuend: number, subtrahend: number): number;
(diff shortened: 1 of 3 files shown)


villesau:
@nvie there has been similar pull requests also earlier which we didn't end up merging. See the reasoning from previous one: https://github.com/flow-typed/flow-typed/pull/2720

nvie:
Hmm. That's a big bummer 😞  We've had several `undefined` errors in production because of this bug, and we finally decided to sit down and fix this.  I had no idea that this was deliberately returning `T`, not `void` (perhaps an idea to document that in a comment?).

In my personal opinion, checking for existence isn't too big of a deal in practice.  Often, you can simply do `|| 0` or something at the end, but we definitely want to handle these cases explicitly to avoid the runtime errors that we use a type system for in the first place.  (We know the indexer property is a weird one for the same reason, and we avoid it as much as possible because of that.)

villesau:
> Hmm. That's a big bummer 😞 We've had several undefined errors in production because of this bug, and we finally decided to sit down and fix this.

It's design decision straight from Flow, not a bug. Although I'm not too big of a fan of it either, for sake of consistency it makes sense.

>  I had no idea that this was deliberately returning T, not void (perhaps an idea to document that in a comment?).

Good idea!

> In my personal opinion, checking for existence isn't too big of a deal in practice.

Agreed. I'm more and more leaning towards opinion that _any_ array index access should return maybe type and be explicitly checked. I don't have experience of the effect in practice though, so it's just an opinion not based on any practice or facts.

> We know the indexer property is a weird one for the same reason, and we avoid it as much as possible because of that.

I've made an issue to flow about indexer object indexer properties btw: https://github.com/facebook/flow/issues/6980

nvie:
Yeah, I can see the point in both arguments.  Thanks for making that PR against Flow 🙏 — I really hope that the indexer property access logic can and will be changed eventually 🤞 (perhaps when you opt-in to flow-strict mode or something?)

I'm also definitely in the correctness camp.  (When people say ease of use, I only hear "less typing", but also realize it's going to be "more time spent hunting bugs in production".)

The problem with the consistency argument is that it quickly becomes philosophical, and you get to pick the thing you would like it to be consistent with in your argument.  For example—yes, it's consistent with the indexer property, but not with methods like `find(): T | void` anymore.  It all boils down to the assumptions about what each such method _should_ do, and what's important to the argument maker.

Lastly, in the case of `_.first()`, the `undefined` behaviour is even in the Lodash documentation:

```javascript
_.head([1, 2, 3]);
// => 1

_.head([]);
// => undefined
```

Some people (including us) picked this API for this behaviour specifically.

Anyway — I just wanted to add these considerations to this discussion.  I definitely see both sides of the argument here.  Internally, we can still use a light "correctness" wrapper around `lodash` for the time being.

villesau:
`.find` is very good point, haven't thought about that and it's not been raised up before. That's clearly inconsistent behaviour compared to indexer access.

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
```diff
@@ -260,10 +260,10 @@ declare module "lodash" {
       fromIndex?: ?number
     ): -1;
     // alias of _.head
-    first<T>(array: ?$ReadOnlyArray<T>): T;
+    first<T>(array: ?$ReadOnlyArray<T>): T | void;
     flatten<T, X>(array?: ?Array<Array<T> | X>): Array<T | X>;
     flattenDeep<T>(array?: ?(any[])): Array<T>;
-    flattenDepth(array?: ?(any[]), depth?: ?number): any[];
```
jedwards1211: I wonder if this would work?
```js
flattenDepth(array: (any[]), depth?: ?number): any[];
flattenDepth(array?: ?(any[]), depth?: ?number): ?any[];
```
This would be ideal, because it wouldn't force null checks in cases where the `array` passed is guaranteed to be defined, but would require null checks otherwise.

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
```diff
@@ -318,7 +318,7 @@ declare module "lodash" {
     ): Array<T>;
     join<T>(array: Array<T>, separator?: ?string): string;
     join<T>(array: void | null, separator?: ?string): "";
```
jedwards1211: notice this overloading here -- maybe we can apply similar overloading for other functions that tolerate nullish arguments.

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
```diff
@@ -806,7 +806,7 @@ declare module "lodash" {
     wrap(value?: any, wrapper?: ?Function): Function;
 
     // Lang
-    castArray(value: *): any[];
+    castArray(value: *): Array<mixed>;
```
jedwards1211: FWIW changing this to `mixed` is out-of-scope for the title of this PR

definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js
jedwards1211: nevermind, I'm forgetting that lodash always returns an array for cases like this

AndrewSouthpaw:
Thanks for seeing both sides of the argument. I hear you on the arguments in favor of correctness.

I just want to add here that "ease of use" is also about "ease of adoption." If a person is already using lodash throughout their codebase and not null checking (because who does that 100%?), then if we forced null checks they would get a deluge of errors that would need fixing, even though their code isn't broken.

I hear you on the pain of hunting for production bugs. But I think that's better than not getting flow adopted in a codebase. I know for at my work if we had to do null checks everywhere, then flow likely would not have been adopted. It's not just about "more typing," it's about remembering to do something each time and add more noise to the code.

I also recognize I seem to be the minority opinion a lot of the time (except for folks at Flow and their stance on array index access), and can be overruled.

AndrewSouthpaw:

PR #3043, opened 2019-01-05
Add clearStore method to ApolloClient in react-apollo.

lasersox:
The clearStore method is also available on the ApolloClient. See: https://github.com/apollographql/apollo-client/blob/183c81dbf0da2d8bc905cd2881dcaea9213f18f6/packages/apollo-client/src/ApolloClient.ts#L488


AndrewSouthpaw:
Hmm, build is broken, but appears to be from before. Merging anyway.

If someone is willing to fix the broken tests that would be *amazing*.

PR #3046, opened 2019-01-06
[node-cache_v4.x.x] Add libdefs for node-cache

nvie:


definitions/npm/node-cache_v4.x.x/flow_v0.25.x-/node-cache_v4.x.x.js
```diff
@@ -0,0 +1,24 @@
...
+|};
+
+declare module "node-cache" {
+  declare class NodeCache {
+    constructor(options?: nodeCache$Options): NodeCache;
+    get(key: string): mixed | void;
+    set(key: string, val: mixed, ttl?: number): boolean;
+  }
+
+  declare module.exports: Class<NodeCache>;
```
villesau: same as in https://github.com/flow-typed/flow-typed/pull/3047#discussion_r245523167, you can make the types better scoped using `declare export`

nvie: Thanks for the feedback – addressed in 9cb06b1d680f617de27d739842c229b147627259! 🙏

nvie: Thanks – this is fantastic feedback.  It's been really confusing to understand the implications of these statements, but this really helps.  I've addressed this in bee7bb2377bff62172b1a240335915c58c74cb1b now – does this look better now?

definitions/npm/node-cache_v4.x.x/flow_v0.25.x-/node-cache_v4.x.x.js
```diff
@@ -0,0 +1,24 @@
+// @flow strict
```
villesau: Although I don't think it makes any harm, `// @flow strict` is not actually needed in libdefs.

nvie: Cheers, good to know 👍 — fixed in 572f99a

PR #3047, opened 2019-01-06
[randomatic_v3.x.x] Add libdefs for randomatic

nvie:


definitions/npm/randomatic_v3.x.x/flow_v0.25.x-/randomatic_v3.x.x.js
```diff
@@ -0,0 +1,20 @@
...
+ * by Vincent Driessen, 2018-12-21
+ */
+
+type $npm$randomatic$Options = {|
+  chars?: string,
+  exclude?: string,
+|};
+
+declare module 'randomatic' {
+  declare module.exports: (
```
villesau: Prefer `declare export default` (search examples from this repo if needed), that way you can define `$npm$randomatic$Options` as private type `Options`. Now `$npm$randomatic$Options` is actually globally defined.

If you define any `declare export ...` in the module declaration, that makes all the other `declare`s private.

nvie: Thanks – this is fantastic feedback.  It's been really confusing to understand the implications of these statements, but this really helps.  I've addressed this in bee7bb2377bff62172b1a240335915c58c74cb1b now – does this look better now?

nvie: Addressed in 8f80ce67b00937e117ac4630397b1e2106ee35d6 – thanks!

PR #3056, opened 2019-01-08
[cookie-parser_v1.x.x] Add libdefs for cookie-parser

nvie:


definitions/npm/cookie-parser_v1.x.x/flow_v0.25.x-/cookie-parser_v1.x.x.js
```diff
@@ -0,0 +1,14 @@
...
+ * Flow libdef for 'cookie-parser'
+ * See https://www.npmjs.com/package/cookie-parser
+ * by Vincent Driessen, 2018-12-21
+ */
+
+declare module 'cookie-parser' {
+  declare function cookieParser(
+    secret?: string | Array<string>,
+    options?: mixed,
+  ): express$Middleware;
```
nvie: Mostly just to learn about best practices here… how would one go about typing this return type?  It should be equal to whatever this is:
https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/express_v4.x.x/flow_v0.32.x-/express_v4.x.x.js#L103-L105

How can one refer to a type defined elsewhere? What's the community best practice here?

definitions/npm/cookie-parser_v1.x.x/test_cookie-parser.js
```diff
@@ -0,0 +1,16 @@
...
+import type { Middleware } from "express";
+
+describe("cookie-parser", () => {
+  it("errors", () => {
+    // $ExpectError
+    cookieParser(42);
+  });
+
+  it("works", () => {
+    (cookieParser(): Middleware);
```
nvie: Same here… how would I be able to import this in the test case?

definitions/npm/cookie-parser_v1.x.x/flow_v0.25.x-/cookie-parser_v1.x.x.js
villesau: Right now the best practice is unfortunately to copy the relevant types over. The CLI cannot resolve type dependencies. See e.g this issue for further details: https://github.com/flow-typed/flow-typed/issues/1857

nvie: OK, thanks for providing some background material here.  For now, I've copied over the necessary type defs from `express` manually just to try how that would work (see ba756cc1975bcf39fd38a07c5145ae65b8e86cff), but it's indeed hard to see how this will scale.  It seems very laborious to keep updating all libdefs that depend on express when the express annotations get updated…

definitions/npm/cookie-parser_v1.x.x/flow_v0.25.x-/cookie-parser_v1.x.x.js
```diff
@@ -0,0 +1,278 @@
+/**
+ * Flow libdef for 'cookie-parser'
+ * See https://www.npmjs.com/package/cookie-parser
+ * by Vincent Driessen, 2018-12-21
+ */
+
+import * as http from 'http';
+import type { Socket } from 'net';
```
villesau: When defining imports outside `declare module` in libdefs, they are defined as `any`. you need to move these inside `declare module`

nvie: Ohhhh, this explains a lot! I was indeed seeing some stuff from express as `any`.  I just copied that from express itself, following that pattern: https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/express_v4.x.x/flow_v0.32.x-/express_v4.x.x.js#L1-L2 — probably a good idea to fix that in there then too?

nvie: `express` is using `default module.exports` instead of `declare export default ...` – is that at all significant to the behaviour of `import` outside of `declare` blocks?

nvie: How do you declare imports inside modules?

Both

```
declare module 'express' {
  declare import * as http from 'http';
}
```

and

```
declare module 'express' {
  import * as http from 'http';
}
```

Seem to be invalid syntax…?

villesau: You can only import types there, so something like `import type {...} from 'http'`

nvie: Oh wow. That is so much nicer — thanks!

In general, while ramping up at learning how to make contributions here, I have come across documentation that suggests what's possible, but not what are best practices.  For example, the docs cover quite a bit on describing global types, or module.exports. Also, the empty stubs generated by the tool use `module.exports` so at first I thought this was the _only_ way they could be used.  Also, a lot of existing definitions use these patterns, so copying those bad practices over also doesn't help (like in this `'express'` case).  Learning more about all this, I wish I could have started with the best practices you're teaching me about here, because they're way nicer :)

villesau: @nvie made a PR to remove the outdated instructions about globabl types: https://github.com/flow-typed/flow-typed/pull/3068

Feel free to improve CONTRIBUTING.md with your learnings!

PR #3059, opened 2019-01-08
[enzyme] Add new definition - Wrapper.renderProp

adamesque:
Added def for new `renderProp` method in enzyme 3.8:
- https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/renderProp.md
- https://github.com/airbnb/enzyme/blob/master/docs/api/ReactWrapper/renderProp.md

I tried to add a test or two but wasn't able to run them locally? On a fresh clone, `./build_and_test_cli.sh` failed during runTests-test & install-test 🤷‍♂️.


pascalduez:
The tests are failing, but they already are on master... Most like because React is imported outside the module definition, hence resolving to `any`.

PR #3060, opened 2019-01-08
[cloudinary_v1.x.x] Add libdefs for cloudinary

nvie:
Note that this library has only been partially typed.  I did include all the objects that are exported from the top level of this module, but marked them `__TODO__` (alias for `any`).  I'm not interested in typing those APIs out completely, as I don't personally use/need those APIs.  However, someone who's interested in them can easily update this definition later on.

I figured having a partially typed API is better than having no typing at all?  This way, coverage can be incrementally improved upon over time.  Is this considered a good or bad practice?

### DIFF
--- a/definitions/npm/cloudinary_v1.x.x/flow_v0.25.x-/cloudinary_v1.x.x.js
+++ b/definitions/npm/cloudinary_v1.x.x/flow_v0.25.x-/cloudinary_v1.x.x.js
@@ -1,0 +1,111 @@
+/**
+ * Flow libdef for 'cloudinary'
+ * by Vincent Driessen, 2019-01-08
+ *
+ * See https://www.npmjs.com/package/cloudinary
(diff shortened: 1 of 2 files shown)


definitions/npm/cloudinary_v1.x.x/flow_v0.25.x-/cloudinary_v1.x.x.js
```diff
@@ -0,0 +1,111 @@
...
+    | 'pad'
+    | 'lpad'
+    | 'mpad'
+    | 'fill_pad'
+    | 'crop'
+    | 'thumb'
+    | 'imagga_crop'
+    | 'imagga_scale';
+
+  // Source: https://cloudinary.com/documentation/image_transformation_reference
```
villesau: Good work with references!

PR #3068, opened 2019-01-11
update CONTRIBUTING.md

villesau:
Global variables should be avoided since nowadays flow support also private types. @AndrewSouthpaw could you check the grammar? :)

This came up in this PR: https://github.com/flow-typed/flow-typed/pull/3056#discussion_r246948475

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,7 +132,6 @@ declare module 'other-module' {
 
 ...but you would be wrong. Flow silently converts `MyType` to be typed `any`, and then sadness ensues.
 
-
 **But wait, I want my React types!**
 
 Good news! You can use the raw, private React types (e.g. `React$Node`, `React$ComponentType`) directly without importing them. You can also import types built into Flow *inside* the module declaration:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -229,22 +229,9 @@ Using `mixed` in place of `any` for the return type of a function or the type of
...
-type $npm$MyModule$Options = {
-  option1: number,
-  option2: string,
-};
-
-declare module "MyModule" {
-  declare function doStuff(options: $npm$MyModule$Options): void;
-}
-```
+Sometimes you see global definitions like `$npm$ModuleName$`. This is due to fact that in the past Flow didn't support private types. **Global types should not be used anymore**. Since then Flow has added support for `declare export` which defines that every type which doesn't have it are defined as private and can't be imported, see https://flow.org/en/docs/libdefs/creation/#toc-declaring-an-es-module for details.
```
pascalduez: "due to **the** fact"?

pascalduez: "which **means** that every type..."?

Issue #3072, opened 2019-01-13
[react-redux] injecting props HOC and forced exact type of OwnProps


jacekkolasa:
Currently OwnProps have to be Exact Type. Is it possible to use current connect<....> typings to create injecting-props HOC, like `injectProp` example in HOC docs?

https://flow.org/en/docs/react/hoc/
```
function injectProp<Config: {}>(
  Component: React.AbstractComponent<Config>
): React.AbstractComponent<$Diff<Config, {foo: number | void}>> {
  return function WrapperComponent(
    props: $Diff<Config, {foo: number | void}>,
  ) {
    return <Component {...props} foo={42} />;
  };
}
```
this doesn't work and complains about `OwnProps is incompatible with exact type`
```
type Props = { ...OwnProps, foo: number };
type OwnProps = {};
const mapStateToProps = () => ({ foo: 5 });
export default connect<Props, OwnProps, _, _, _, _>(mapStateToProps);
```
(cc @peter-leonov)

peter-leonov:
Quick guess:

```js
-type OwnProps = {};
+type OwnProps = {||};
```

jacekkolasa:
@peter-leonov yes, exactly. I cannot do `type OwnProps = {||};` because then i cannot pass any other props into the decorated component, like:
`let usage = <DecoratedConnectedComponent propToBePassedThru="bar" />;` - here i want `ConnectedComponent` to use `propToBePassedThru` but `injectProp` should not know about it

peter-leonov:
This little test works fine:

```js
// @flow
import * as React from "react";
import { connect } from "react-redux";

export let e = [];

function checkSimplePropertyInjection() {
  type OwnProps = {|
    foo: number,
    bar: string,
  |};
  type Props = { ...OwnProps, foo: number };
  const mapStateToProps = () => ({ foo: 5 });

  class Com extends React.Component<Props> {}

  const Connected = connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(Com);

  <Connected foo={42} bar="str" />;
  //$ExpectError property `foo` is missing in props [1] but exists in `OwnProps`
  <Connected bar="str" />;
  e.push(Connected);

  function injectProp<Config: {}>(
    Component: React.AbstractComponent<Config>,
  ): React.AbstractComponent<$Diff<Config, { foo: number | void }>> {
    return function WrapperComponent(
      props: $Diff<Config, { foo: number | void }>,
    ) {
[...]

PR #3076, opened 2019-01-14
[react-redux_v5.x.x] Better support for HOC

peter-leonov:
Addressing an issue with HOC described in #3072 by making it possible to provide inexact `OwnProps` type parameter and also by adding tests for the basic HOC use cases.

More example are welcome!

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
@@ -67,30 +67,36 @@ declare module "react-redux" {
     static +WrappedComponent: WC;
     getWrappedInstance(): React$ElementRef<WC>;
   }
+  // The connection of the Wrapped Component and the Connected Component
+  // happens here in `MP: P`. It means that type wise MP belongs to P,
+  // so to say MP >= P.
   declare type Connector<P, OP, MP: P> = <WC: React$ComponentType<P>>(
     WC,
   ) => Class<ConnectedComponent<OP, WC>> & WC;
 
   // No `mergeProps` argument
 
+  // Got error like inexact OwnProps is incompatible with exact object type?
+  // Just make the OP parameter for `connect()` an exact object.
+  declare type MergeOP<OP, D> = {| ...$Exact<OP>, dispatch: D |};
+  declare type MergeOPSP<OP, SP> = {| ...$Exact<OP>, ...SP |};
+  declare type MergeOPDP<OP, DP> = {| ...$Exact<OP>, ...DP |};
+  declare type MergeOPSPDP<OP, SP, DP> = {| ...$Exact<OP>, ...SP, ...DP |};
+
   declare export function connect<-P, -OP, -SP, -DP, -S, -D>(
     mapStateToProps?: null | void,
     mapDispatchToProps?: null | void,
     mergeProps?: null | void,
-    options?: ?Options<S, OP, {||}, {| ...OP, dispatch: D |}>,
-    // Got error like inexact OwnProps is incompatible with exact object type?
-    // Just make your OP parameter an exact object.
-  ): Connector<P, OP, {| ...OP, dispatch: D |}>;
+    options?: ?Options<S, OP, {||}, MergeOP<OP, D>>,
+  ): Connector<P, OP, MergeOP<OP, D>>;
 
   declare export function connect<-P, -OP, -SP, -DP, -S, -D>(
     // If you get error here try adding return type to your mapStateToProps function
     mapStateToProps: MapStateToProps<S, OP, SP>,
     mapDispatchToProps?: null | void,
     mergeProps?: null | void,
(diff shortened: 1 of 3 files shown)


definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
```diff
@@ -67,41 +67,45 @@ declare module "react-redux" {
...
   ) => Class<ConnectedComponent<OP, WC>> & WC;
 
   // No `mergeProps` argument
 
+  // Got error like inexact OwnProps is incompatible with exact object type?
+  // Just make the OP parameter for `connect()` an exact object.
+  declare type MergeOP<OP, D> = {| ...$Exact<OP>, dispatch: D |};
+  declare type MergeOPSP<OP, SP> = {| ...$Exact<OP>, ...SP |};
+  declare type MergeOPDP<OP, DP> = {| ...$Exact<OP>, ...DP |};
+  declare type MergeOPSPDP<OP, SP, DP> = {| ...$Exact<OP>, ...SP, ...DP |};
```
villesau: What a name 😂

villesau: should SP and DP be wrapped to `$Exact` too? Or is it enough that OP is wrapped?

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -594,7 +638,7 @@ function testHoistConnectedComponent() {
 
   const Connected = connect<Props, OwnProps, _,_,_,_>(mapStateToProps)(Com);
   e.push(Connected);
-  // $ExpectError should be OK without passthroughWithDefaultProp
+  // OK without passthroughWithDefaultProp
```
villesau: Nice!

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
peter-leonov: To be honest, I really do not know. For the known use cases it's enough. Also, the `OP` is an explicit type parameter while all the others get inferred, and we all know how bad Flow might be in reporting errors if there are multiple layers of type transformations on top. This is why I try to keep it as simple as possible.

peter-leonov:
@jacekkolasa what do you think of all this? Does this PR fix anything? 😅

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
villesau: Ok, makes totally sense! IT can be improved later on if necessary.

jbrown215: FWIW, I'm currently working on spread semantics and you won't have to do guesswork about how flow infers the types for spreads anymore.

peter-leonov: Thanks a lot for making Flow better! Keep up the great job! Can I somehow help you (say, with providing some use cases) or could it be possible to track the progress somewhere to get the typings prepared a bit better this time 😅?

jbrown215: Your types as-is should work. I'll write up something more formally when I'm further along.

Basically, spreads for exact objects will stay the same (modulo objects with indexers, that remains to be seen) and spreads with inexact objects will actually work without inferring mixed everywhere. But, there will be more errors when spreading inexact objects.

jacekkolasa:
@peter-leonov yes, it fixes the problem where `connect` just decorates any component. That's the first issue from https://github.com/flow-typed/flow-typed/issues/3072. Big thanks!

dannyharding10:
@peter-leonov I've been looking around for best practices when it comes to typing connect since flow v0.85.0. Would you be able to provide a simple example so I can go about fixing the hundreds of flow errors I have now? I would prefer not to do `connect<*, *, *, *, *, *>(mapStateToProps)(MyComponent)`

benfjohnson:
Came here looking for an answer to the same question as @dannyharding10

villesau:
@dannyharding10 @benfjohnson does this blog post help you to do the update: https://medium.com/flow-type/upgrading-flow-codebases-40ef8dd3ccd8 ?

benfjohnson:
That is incredibly helpful!

unknownbreaker:
The medium article advice does not eliminate the Flow `connect()` error for me. In fact, Flow's error says `Cannot resolve name "Props"` and `Cannot resolve name "OwnProps"`.

Also, `connect<*, *, *, *, *, *>(mapStateToProps)(MyComponent)` does not work either.

Currently using `react-redux_v7.x.x`
My version of flow is 0.134.0

PR #3093, opened 2019-01-22
Update types to latest release redux-saga and lock Flow.js version < 0.87

retyui:
- Links to documentation: https://github.com/redux-saga/redux-saga/tree/master/docs/api
- Link to GitHub or NPM: https://github.com/redux-saga/redux-saga
- Type of contribution: fix and refactor


retyui:
### flow.js 0.88 chnagelog:
> Made Function and Object types be aliases for any. They were always unsafe types, just like any, but they had peculiar behavior. This change revealed places where they were handled improperly within Flow, and ended up surfacing type errors that were previously missed.

When you unlock the Flow.js version types 0.88+  fro redux-saga_v1 and run test
```sh
node cli/dist/cli.js run-tests redux-saga_v1
```
You will see more erros associated with the changes described above.

Perhaps there are some recommendations to fix them?

Issue #2876, opened 2018-10-23
Flow v.0.84.0 explicit inexact objects


jbrown215:
We will be releasing flow 0.84.0 this week, which includes the `...` syntax for inexact objects (for more info, see the [blog post](https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf)).  AFAIK, flow-typed doesn't require [babel support](https://github.com/babel/babel/pull/8884) to land before we can start upgrading the syntax here. I'm aiming to have the `flow-upgrade` command ready end of week to help you guys move through this as quickly as possible.

I'm opening this issue as a request for help, as I am not able to see this through myself. I imagine a script can be written with the following logic:
```
for each flow-typed library definition:
  copy the folder to one for flow v0.84.0
  run `flow-upgrade` in that new folder
```

jbrown215:
`flow-upgrade` is blocked on the new flow parser being deployed, which should happen today or tomorrow. As soon as that happens, I'll make sure everything is ok with flow-upgrade and release.

I've noticed a few bugs in the tools `flow-upgrade` relies on, so I'm fixing those as I find them. A notable such example is that `recast` doesn't traverse opaque types.

The good news is that this change is not breaking-- we are still parsing anything with just curly braces `{}` as inexact objects, so you have at the very least 2 weeks from when the new version of flow is released to make sure this gets done.

villesau:
Ok, that's good to know! 2 weeks should be enough for us to migrate the libdefs to new format. Eaxct by default change is great BTW! :)

jbrown215:
Thanks for the help here! it would be useless if it broke all of the types our community has made until now

villesau:
@jbrown215 Turns out latest release is not stable in Travis: https://github.com/flow-typed/flow-typed/issues/2916

We might need to disable checks for 0.84.0, which means that we can't migrate libdefs to 0.84 either: We can't test them.

jbrown215:
No worries. A lot of things have to go right on our side before we're ready to parse objects as exact by default. You'll likely have a few releases to get this done. As soon as the change lands in master I'll update you here if this issue is still open.

villesau:
Ok, good! And thanks for keeping us updated! We started to skip 0.84.0 to get tests succeed again. Looks like it worked.

villesau:
[...]

PR #3098, opened 2019-01-26
[@zeit/title_v3.4.x] Add definitions

pascalduez:
- Links to documentation: https://github.com/zeit/title
- Link to GitHub or NPM: https://github.com/zeit/title

### DIFF
--- a/definitions/npm/@zeit/title_v3.4.x/flow_v0.30.x-/title_v3.4.x.js
+++ b/definitions/npm/@zeit/title_v3.4.x/flow_v0.30.x-/title_v3.4.x.js
@@ -1,0 +1,7 @@
+declare module '@zeit/title' {
+  declare type Options = {|
+    +specials: $ReadOnlyArray<string>,
+  |};
+
+  declare module.exports: (str: string, options?: Options) => string;
+}
--- a/definitions/npm/@zeit/title_v3.4.x/test_title_v3.4.x.js
+++ b/definitions/npm/@zeit/title_v3.4.x/test_title_v3.4.x.js
@@ -1,0 +1,33 @@
+// @flow
+
+import { describe, it } from 'flow-typed-test';
+import title, { type Options } from '@zeit/title';
+
+describe('The `title` function', () => {
+  it('should accept strings as first parameter', () => {
+    title('');
+    // $ExpectError
+    title(123);
+  });
+
+  it('should accept an `Options` object as second parameter', () => {
+    title('', { specials: [''] });
+    // $ExpectError
+    title('', {});
+    // $ExpectError
+    title('', []);
+    // $ExpectError
+    title('', { specials: '' });
+  });
+
+  it('should enforce immutability for its `Options` parameter', () => {
+    let options: Options = { specials: [''] };
+    title('', options);
+    // $ExpectError
+    options.specials.push('');
+    title('', options);
+
+    let newOptions = { specials: [...options.specials, ''] };
+    title('', newOptions);
+  });
+});


definitions/npm/@zeit/title_v3.4.x/flow_v0.30.x-/title_v3.4.x.js
```diff
@@ -0,0 +1,7 @@
+declare module '@zeit/title' {
+  declare type Options = $Exact<{
+    specials: Array<string>,
```
villesau: Why $Exact and not `{||}`? Specials could perhaps be immutable `+specials`, as well as the type could be `$ReadOnlyArray`?

pascalduez: > Why $Exact and not `{||}`
Not real reason, but I was thinking, isn't it a bit more "future proof" when they will remove the `{||}` notation?

Updating the rest, nice suggestions!

pascalduez:
> Would perhaps be good to encourage some "best practices" regarding immutability, exactness etc.

I'm all for this.

Just wondering in this particular case whether it's that useful. Let's say one wants to reuse and mutate on options object. 

```js
let options = { specials: [''] };

title('', options);

options.specials.push(''); // Error

title('', options);
````

So we enforce immutability here, which is a good thing I guess.

```js
let options = { specials: [''] };

title('', options);

let newOptions = { specials: [...options.specials, ''] };

title('', newOptions);
```

definitions/npm/@zeit/title_v3.4.x/flow_v0.30.x-/title_v3.4.x.js
villesau: There will be codemod to get rid of it. It needs to be run against the whole repo neverthless. IT at least removes the pipes, not sure about $Exact, see https://github.com/flow-typed/flow-typed/issues/2876

villesau:
It actually will not error, see: https://flow.org/try/#0PTAEAEDMBsHsHcBQiAmBTAxtAhgJzaJAK4B2GALgJawmjloDO5AFJSQA5HkBcoAJACU02FAHkS0AJ4BBXLmySAPCSIBbAEZpcAPgCUvAG6xKKANzIMNJqDy5QAXlABtALrnbAOk4MAFswCMAEwAzLrm9EzMtmFAA

Taking in immutable types ensures that the value is not mutated inside the function. If you take in mutable types, you can't pass immutable values to it: https://flow.org/try/#0PTAEAEDMBsHsHcBQiAmBTAxtAhgJzaJAK4B2GALgJawmjloDO5AFJSQA5HkBcoAgrlzYAngB4SRALYAjNLgB8ASl4A3WJRQBuZBhpNCtALyhmeXLwAkAJTTYUAeRLRhAoWIky5S0IfmgA3ogAkPRMpoKK2gC+iEA

pascalduez:
My test where like: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDyOGAlnAHYDOYAvGACQCiAHgIYDGGAPAN6pgwAair4OpNjCoAuRgCU8bACaEKMLAEEATlrZZuVDFtIUA5gD4ANKgC+5gNzoleDjDZaCUAK4Uu5CmBkGDB4ABSGWrIRJqaWYHAk-lQA-LLEZJRUAJRRRjGOqCEY8YmZdGC8YKIuElKyANoA5I0AumA2BUEhoc1xCRnUWQX9SQB01eKSVKM4XlQAFj2NQ+hAA

villesau:
I guess that's because Flow refines the type if no explicit type defined. If the type is explicitly defined, it does not error: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDyOGAlnAHYDOYAvGACQCiAHgIYDGGAPAN6pgwAair4OpNjCoAuRgCU8bACaEKMLAEEATlrZZuVDFtIUA5gD4ANKgC+5gNzoleDjDZaCUAK4Uu5CmBkGDB4ABSGWrIRJqaWYHAk-lQA-LLEZJRUAJRRRjGO2PhEiZkATHSMrJw8-IKiLhJSstq6+tFmVrYO6ByZGGBQAfShCRnUaSXUpVl05mC1gaTBYQDkK3GjSVmOgpuZAHT14pJU+zheVAAWoWvbtugPQA

PR #3101, opened 2019-01-28
Redux reducers: Make actions contravariant

craigfowler:
Redux reducers can legitimately be contravariant
with regard to the actions they accept.

The action is used only in a parameter
position and constructs like this would be
useful, particularly when a base action
typedef (specifying a string 'type' property) is
used.

```js
const reducers : Array<Reducer<TState,BaseAction>> = [];
const specificReducer : Reducer<TState,TDerivedAction> = getSpecificReducer();
reducers.push(specificReducer);
```

- Type of contribution: Small feature enhancement to existing definition.

**Other notes**
FWIW I noticed that the PR changed the line break at the end of the file. That wasn't intended and was added by the editor I used. Feel free to revert that bit out if you want (I created/submitted this PR from a mobile device & editor).

### DIFF
--- a/definitions/npm/redux_v4.x.x/flow_v0.55.x-/redux_v4.x.x.js
+++ b/definitions/npm/redux_v4.x.x/flow_v0.55.x-/redux_v4.x.x.js
@@ -24,7 +24,7 @@ declare module 'redux' {
     replaceReducer(nextReducer: Reducer<S, A>): void
   };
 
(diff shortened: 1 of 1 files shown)


AndrewSouthpaw:
Also seems like build is failing, so you'll want to get those tests to pass.

AndrewSouthpaw:

PR #3107, opened 2019-01-30
[enzyme_v3.x.x] Fix instance error

### LINKED ISSUE #925: Enzyme: Wrapper#instance() returns React$Component instead of the original component class

HsuTing:
- Links to documentation: https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/instance.md
- Link to GitHub or NPM: https://github.com/airbnb/enzyme
- Type of contribution: fix

Other notes:
Fix #925

### DIFF
--- a/definitions/npm/enzyme_v2.3.x/flow_v0.53.x-/enzyme_v2.3.x.js
+++ b/definitions/npm/enzyme_v2.3.x/flow_v0.53.x-/enzyme_v2.3.x.js
@@ -6,7 +6,7 @@ declare module "enzyme" {
     index: number
   ) => boolean;
   declare type NodeOrNodes = React.Node | Array<React.Node>;
-  declare type EnzymeSelector = string | Class<React.Component<*, *>> | Object;
+  declare type EnzymeSelector = string | Class<React.Component<*, *>> | {};
 
   // CheerioWrapper is a type alias for an actual cheerio instance
   // TODO: Reference correct type from cheerio's type declarations
--- a/definitions/npm/enzyme_v3.x.x/flow_v0.53.x-/enzyme_v3.x.x.js
+++ b/definitions/npm/enzyme_v3.x.x/flow_v0.53.x-/enzyme_v3.x.x.js
@@ -1,5 +1,5 @@
 declare module "enzyme" {
-  declare type PredicateFunction<T: Wrapper> = (
+  declare type PredicateFunction<T: Wrapper<*>> = (
     wrapper: T,
     index: number
   ) => boolean;
@@ -10,7 +10,7 @@ declare module "enzyme" {
   // TODO: Reference correct type from cheerio's type declarations
   declare type CheerioWrapper = any;
 
-  declare class Wrapper {
+  declare class Wrapper<RootComponent> {
     find(selector: EnzymeSelector): this,
     findWhere(predicate: PredicateFunction<this>): this,
     filter(selector: EnzymeSelector): this,
@@ -53,7 +53,7 @@ declare module "enzyme" {
     setState(state: {}, callback?: () => void): this,
     setProps(props: {}, callback?: () => void): this,
     setContext(context: Object): this,
-    instance(): React$Component<*, *>,
+    instance(): React$ElementRef<RootComponent>,
     update(): this,
     debug(options?: Object): string,
     type(): string | Function | null,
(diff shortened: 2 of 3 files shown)


wchargin:
This PR makes it illegal to pass a `React$Node` to `render()`, yet we’ve
been doing so until now and haven’t seen any runtime type errors. Is
this change intentional, or a side-effect of the larger change?

HsuTing:
@wangbinyq 
I think it's a side-effect of my pr. I change  `React$Element` from `React$Node` in `render()` to make wrapper being able to get the type and assign the type to `instance`. Is it possible for you to provide an example? Maybe I can clarify the problem with the example.

wchargin:
Sure—consider the following module:

```javascript
// @flow

/*:: import type {Node as ReactNode} from "react"; */
const React = require("react");
const Enzyme = require("enzyme");
const Adapter = require("enzyme-adapter-react-16");

class MyComponent extends React.Component /*:: <{}> */ {
  render() {
    return React.createElement("div", {}, "Hello from MyComponent!");
  }
}

function makeNode1() {
  return React.createElement(MyComponent);
}

function makeNode2() {
  return React.createElement("div", {}, "Hello from a simple <span>!");
}

function makeNode(which /*: boolean */) /*: ReactNode */ {
  return which ? makeNode1() : makeNode2();
}

function test() {
  Enzyme.configure({adapter: new Adapter()});
  for (const which of [false, true]) {
    console.log(Enzyme.render(makeNode(which)).html());
  }
  console.log("OK");
}

test();
```

Running `node test.js` completes successfully:

```
$ node test.js 
Hello from a simple &lt;span&gt;!
Hello from MyComponent!
OK
```

But running `yarn flow check` gives the following errors:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ test.js:29:31

Cannot call Enzyme.render with makeNode(...) bound to node because:
 • inexact $Iterable [1] is incompatible with exact React.Element [2].
 • inexact React.Portal [1] is incompatible with exact React.Element [2].
 • inexact boolean [1] is incompatible with exact React.Element [2].
 • inexact null [1] is incompatible with exact React.Element [2].
 • inexact number [1] is incompatible with exact React.Element [2].
 • inexact string [1] is incompatible with exact React.Element [2].

     test.js
 [1]  22│ function makeNode(which /*: boolean */) /*: ReactNode */ {
      23│   return which ? makeNode1() : makeNode2();
      24│ }
      25│
      26│ function test() {
      27│   Enzyme.configure({adapter: new Adapter()});
      28│   for (const which of [false, true]) {
      29│     console.log(Enzyme.render(makeNode(which)).html());
      30│   }
      31│   console.log("OK");
      32│ }

     flow-typed/npm/enzyme_v3.x.x.js
 [2] 125│     node: React$Element<T>,

Found 6 errors
```

<details>
<summary>package.json contents</summary>

```json
{
  "name": "tmp.KjpKoG7s4R",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "enzyme": "^3.10.0",
    "enzyme-adapter-react-16": "^1.14.0",
    "flow-bin": "^0.105.2",
    "react": "^16.9.0",
    "react-dom": "^16.9.0"
  }
}
```

</details>

Using `React.Element<any>` instead of `ReactNode` fixes the error, but
using `any` generally loses type information, so I’m a bit nervous about
making that change. Is that the right thing to do? Is it safe?

Our [real-world example][1] is pretty similar to this.

[1]: https://github.com/sourcecred/sourcecred/blob/acdcbc29ede149116cf61a04c3ecb2a7f4196ab1/src/plugins/github/render.js#L82-L93

wchargin:
As just a partial fix for our use case (selfishly), it looks like the
change to `render` specifically could be averted without inconveniencing
others, because `render` returns a `CheerioWrapper`, not a `Foo<T>` for
any `Foo`, so the generics aren’t actually buying you anything here.

Opened #3522.

PR #3138, opened 2019-02-06
[typescript] Add definitions

goodmind:
- Links to documentation: https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API
- Link to GitHub or NPM: https://github.com/Microsoft/TypeScript

Other notes: I don't expect this to get merged lol, generated enums in this works bad and no one would review 10000 lines, also it uses spread but I don't remember  in what version spread was added, so I started with default `flow_v0.25.x`


villesau:
Should we merge this? 🤔

AndrewSouthpaw:
What's the use case?

goodmind:
@villesau No idea, did you review 10k lines?
@AndrewSouthpaw  I use this in `flowgen`, but I'm ok with this being internal types

AndrewSouthpaw:
So this is a libdef generated by a tool that will be using the libdef to provide type definitions to what it's translating.

There's so many layers. 😆 

I'd keep it as internal types for now unless you think other people working on `flowgen` need this?

goodmind:
@AndrewSouthpaw or if anyone else wants to use typescript compiler

AndrewSouthpaw:
I'm not compelled to merge it, but if someone else feels strongly I can be on board. Others?

villesau:
Apparently there is already an use case for it: https://github.com/joarwilk/flowgen/tree/master/flow-typed

So why not! Though, would be good to have few smoke tests at least.

Mouvedia:
> I don't remember in what version spread was added

I do: 0.42.

goodmind:
I'll change minimum version then

AndrewSouthpaw:
Works for me. Go ahead and add the minimum version, a couple smoke tests, and we can call it good.

AndrewSouthpaw:
The minimum version would be `[version]-` with a hanging dash. Check other libdefs for examples. :-)

Issue #2842, opened 2018-10-15
Question: shared types between packages


Ashoat:
I help maintain the Flow libdefs for `react-navigation`. For our upcoming 3.0 release, we're planning on splitting the project into multiple npm packages. We'd ideally be able to define types that are shared across packages only once, but my understanding is that it's not yet possible to import types between libdefs.

My question is: what is the best practice for this today? Is copy-pasting types between libdefs the only way to do it? If so, does anybody know of a build system that automates/simplifies this process?

Ashoat:
Prior discussions:

- https://github.com/flow-typed/flow-typed/issues/16
- https://github.com/flow-typed/flow-typed/issues/1795
- https://github.com/flow-typed/flow-typed/issues/1857

Flow issue:

- https://github.com/facebook/flow/issues/3883

villesau:
Hi @Ashoat, thanks for asking!

TL;DR: Currently copy-pasting is indeed the best practice.

We have experimental branch that would support lerna monorepo structure for libdefs. The benefit is that it would allow to push all the libdefs to npm, and define cross dependecies between libdefs: https://github.com/flow-typed/flow-typed/tree/v3-experimental In addition to that CLI will need to be changed to support resolving libdef dependencies. cc @adityavohra7

Ashoat:
So let's say we have libdefs for `react-navigation`, `react-navigation-stack`, and `react-navigation-tabs`. The idea is that we (React Navigation team) would create a lerna monorepo holding all three libdefs, and they would be able to cross-reference each other? And libdefs will no longer be hosted on this `flow-typed` repo, but instead published to npm?

LoganBarnett:
@Ashoat are these projects for `react-navigation` going to use Flow directly? If so, we have some mechanisms to import the types defined in the source directly. It gets around the libdef problem. Generally we want to use  libdefs for libraries that don't have their own types defined.

Ashoat:
No, they don’t use Flow directly.

AndrewSouthpaw:
Also worth noting the experimental branch is far from finished, so I wouldn't bank on it any time soon. :-/

Ashoat:
Wondering if there has been any update on this work. It's by far my #1 complaint about working with Flow...

AndrewSouthpaw:
No progress has been made AFAIK, @Ashoat. 🙁

Brianzchen:
[...]

PR #3139, opened 2019-02-07
[react-navigation] Introduce libdef for @react-navigation/core

Ashoat:
This is unfortunately copy-pasted from the `react-navigation` libdef due to #2842.


definitions/npm/@react-navigation/core_v3.x.x/flow_v0.60.x-/core_v3.x.x.js
```diff
@@ -0,0 +1,745 @@
...
+    | ''
+    | $ReadOnlyArray<StyleObj>
+    | { [name: string]: any };
+  declare type ViewStyleProp = StyleObj;
+  declare type TextStyleProp = StyleObj;
+  declare type AnimatedTextStyleProp = StyleObj;
+
+  // This one is too large to copy. Actual definition is in
+  // react-native/Libraries/Animated/src/nodes/AnimatedValue.js
+  declare type AnimatedValue = Object;
```
pascalduez: You know that from Flow 0.75 this is an alias for `any`, right?

definitions/npm/@react-navigation/core_v3.x.x/flow_v0.60.x-/core_v3.x.x.js
```diff
@@ -0,0 +1,745 @@
...
+
+  declare type HeaderForceInset = {
+    horizontal?: string,
+    vertical?: string,
+    left?: string,
+    right?: string,
+    top?: string,
+    bottom?: string,
+  };
```
pascalduez: We ask to use exact objects now on, if possible.

definitions/npm/@react-navigation/core_v3.x.x/flow_v0.60.x-/core_v3.x.x.js
Ashoat: Didn't know that! Thanks for the pointer.

What would you suggest doing in this case? The source type is [here](https://github.com/facebook/react-native/blob/master/Libraries/Animated/src/nodes/AnimatedValue.js#L74). My instinct is to type just the methods [listed in the documentation](http://facebook.github.io/react-native/docs/animatedvalue.html), though this will be a fairly brittle solution, and may require updating this libdef whenever React Native's Animated API changes.

definitions/npm/@react-navigation/core_v3.x.x/flow_v0.60.x-/core_v3.x.x.js
Ashoat: Good call - this can definitely be refined a bit.

Ashoat:
I just put up three PRs for the requested follow-ups. Please merge this one first - afterwards, I will update the three PRs so that they touch both `react-navigation` and `@react-navigation/core` libdefs, and then we can merge the three new PRs.

definitions/npm/@react-navigation/core_v3.x.x/flow_v0.60.x-/core_v3.x.x.js
pascalduez: I'm afraid I don't have the right answer here, and that might even change when exact objects will become the default. I tend to use `{}` instead, but seems fragile as well.  
Maybe @jbrown215 could gives use some hints ;)

PR #3143, opened 2019-02-10
[react-redux] Pass dispatch prop when no dispatch function provided.

AndrewSouthpaw:
Provides a `dispatch` property when no dispatch function is provided.

Downside is that it requires people to specify a `dispatch` property when they're using exact props on `Props`, but I think that's worth the cost of being able to correctly type the component. We use this pattern throughout our codebase.

Also apparently it makes it no longer possible to infer types in this use case. Didn't seem like a terrible cost.

@peter-leonov @villesau

### DIFF
--- a/definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
+++ b/definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
@@ -79,7 +79,7 @@ declare module "react-redux" {
   // Got error like inexact OwnProps is incompatible with exact object type?
   // Just make the OP parameter for `connect()` an exact object.
   declare type MergeOP<OP, D> = {| ...$Exact<OP>, dispatch: D |};
-  declare type MergeOPSP<OP, SP> = {| ...$Exact<OP>, ...SP |};
+  declare type MergeOPSP<OP, SP, D> = {| ...$Exact<OP>, ...SP, dispatch: D |};
   declare type MergeOPDP<OP, DP> = {| ...$Exact<OP>, ...DP |};
   declare type MergeOPSPDP<OP, SP, DP> = {| ...$Exact<OP>, ...SP, ...DP |};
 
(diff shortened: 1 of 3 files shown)


definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connectInfer.js
```diff
@@ -75,45 +75,6 @@ function doesNotRequireDefinedComponentToTypeCheck5case() {
   connect(mapStateToProps, mapDispatchToProps, mergeProps)(Component);
 }
 
-function testExactProps() {
```
AndrewSouthpaw: This test now fails with the new change; types can no longer be inferred for exact props.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
```diff
@@ -79,7 +79,7 @@ declare module "react-redux" {
   // Got error like inexact OwnProps is incompatible with exact object type?
   // Just make the OP parameter for `connect()` an exact object.
   declare type MergeOP<OP, D> = {| ...$Exact<OP>, dispatch: D |};
-  declare type MergeOPSP<OP, SP> = {| ...$Exact<OP>, ...SP |};
+  declare type MergeOPSP<OP, SP, D> = {| ...$Exact<OP>, ...SP, dispatch: D |};
```
peter-leonov: 👍

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
```diff
@@ -722,3 +724,23 @@ function checkIfStateTypeIsRespectedAgain() {
   <Connected />;
   e.push(Connected);
 }
+
+function testPassingDispatchPropWithoutDispatchFunction() {
+  type OwnProps = {||}
+  type Props = {| ...OwnProps, dispatch: any |};
```
peter-leonov: I'd also add a test with an exact type for `dispatch` to test that the typings pass the correct type explicitly. I haven't seen any tests for this case in the pre-0.89 version and neither added them myself.

peter-leonov:
Nice catch! LGTM 👍

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/test_connect.js
AndrewSouthpaw: √  Also tested that `dispatch` is not passed when `mapDispatchToProps` function is provided.

definitions/npm/react-redux_v5.x.x/flow_v0.89.x-/react-redux_v5.x.x.js
hally9k: @AndrewSouthpaw I just spent a while tracking this down as it was tripping me up for a while and I couldn't understand why I should be expected to add a dispatch property to my exact Props type, a helpful comment above this line wouldn't go amiss. Thanks for your work by the way, highly appreciated! 👍

PR #3147, opened 2019-02-11
[react-navigation] Refine AnimatedValue type

Ashoat:
Lots of copy-pasting going on here, unfortunately. Can't wait for #2842 to be resolved.


Ashoat:

PR #3148, opened 2019-02-11
[react-navigation] Introduce SafeAreaViewInsets type

Ashoat:
This type is used in two separate places in the libdef. I've updated it to use `$Shape` and an exact type.

### DIFF
--- a/definitions/npm/@react-navigation/core_v3.x.x/flow_v0.60.x-/core_v3.x.x.js
+++ b/definitions/npm/@react-navigation/core_v3.x.x/flow_v0.60.x-/core_v3.x.x.js
@@ -630,8 +630,8 @@ declare module '@react-navigation/core' {
     // The value that represents the progress of the transition when navigation
     // state changes from one to another. Its numeric value will range from 0
     // to 1.
-    //  progress.__getAnimatedValue() < 1 : transtion is happening.
-    //  progress.__getAnimatedValue() == 1 : transtion completes.
+    //  progress.__getAnimatedValue() < 1 : transition is happening.
+    //  progress.__getAnimatedValue() == 1 : transition completes.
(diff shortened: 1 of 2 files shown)


Ashoat:

definitions/npm/@react-navigation/core_v3.x.x/flow_v0.60.x-/core_v3.x.x.js
```diff
@@ -720,6 +711,16 @@ declare module '@react-navigation/core' {
     config?: NavigationTabRouterConfig
   ): NavigationRouter<*, *>;
 
+  declare type _SafeAreaViewForceInsetValue = 'always' | 'never' | number;
+  declare type _SafeAreaViewInsets = $Shape<{|
```
pascalduez: Not sure I get it, or maybe I wasn't clear enough in the previous review.   `$Shape<{|`   
Exact objects should be preferred as default, but if some cases really needs to be more relaxed, then a normal object will be fine.

Ashoat:
I’m confused... are there any changes you’d like to see? Do you want me to explain what I did? There were two types representing the same thing, I got rid of one of them and refined the other one. It is correct to use `$Shape` here because all of the properties are optional, and it’s correct to use an exact type because no other properties should be possible.

pascalduez:
@Ashoat Thanks for the reasoning behind this.
Fact is this is an unusual usage and it picked my attention. 

`$Shape` is already exact by default, so `$Shape<{ }>` should be enough. 

Usually I would have used optional properties instead, but granted that if all properties in a  type are optional, `$Shape` probably makes more sense.

Ashoat:
Didn't realize that `$Shape` is exact by default! Will get rid of the exact part.

PR #3149, opened 2019-02-11
[react-navigation] Add tests for routers

Ashoat:
These tests cover the "core" router functionality.


Ashoat:

PR #3155, opened 2019-02-14
Add static now() to moment libdef

LulzAugusto:
- Links to documentation: https://momentjs.com/docs/#/customization/now/
- Link to GitHub or NPM: https://github.com/moment/moment/

Other notes:
Noticed the definition for `moment.now()` was missing while working on a project.


LulzAugusto:
Should I fix the error causing the build to fail? ☝️ (It wasn't introduced in my PR)

PR #3159, opened 2019-02-16
[lodash_v4.x.x] unify path argument to $ReadOnlyArray<string | number…

kubijo:
- Links to documentation: https://lodash.com/docs/4.17.11
- Link to GitHub: https://github.com/lodash/lodash
- Type of contribution: fix & refactor

Other notes:
I've took the liberty to unify (hopefully) all path arguments to one re-used type.

Previously there were several inline syntactical variants of the following:

```javascript
string[] | string
```

That is not true since array paths with mixed-in numbers wouldn't work, yet they do.
The new type after my change is following:

```javascript
$ReadOnlyArray<string | number> | string | number;
```

- I've found out that this was already used in some places (get/set IIRC).
- Here's a [flow.try](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgAKAhhgBZgC8YAJAEp7EAmA8gHYxYCCATj8VgA8AZww8AlmwDmYAD5g2AVwC2AIzw8AfHLCiJ0nUrUaA3OgDGcNqLDEAXEVIVqAbQDkxNwBowARh9uAMxuALpmltYYYDhUYO6ePv5gQaFmqFCKbOYY4lZgUngYABRwDsRsWD44DiTkAJRgAN4AvugFxY22Di5JAEw+gSFgzT7x3n4hdWbtRT0+-WCDAb5uU6gzc2ALS35TQA)
- You can also try that this is true in the web console at: https://lodash.com/docs/4.17.11
    ```javascript
    _.get({ a: [1, 2, 3] }, ['a', 1]);
    _.get([1, 2, 3], '1');
    _.get([1, 2, 3], 1);
    ```
- I've only updated definitions for flow versions >= 0.38 as that is the first mention of $ReadOnlyArray [in flow changelog](
https://github.com/facebook/flow/blob/acd54fda1b335082fea29ab23ac6b163c8f4948f/Changelog.md#v0380)

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.38.x-v0.46.x/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.38.x-v0.46.x/lodash_v4.x.x.js
@@ -1,4 +1,6 @@
+
 declare module "lodash" {
+  declare type Path = $ReadOnlyArray<string | number> | string | number;
   declare type TemplateSettings = {
     escape?: RegExp,
     evaluate?: RegExp,
@@ -381,12 +383,12 @@ declare module "lodash" {
     includes(str: string, value: string, fromIndex?: number): boolean;
     invokeMap<T>(
       array: ?Array<T>,
-      path: ((value: T) => Array<string> | string) | Array<string> | string,
+      path: ((value: T) => Path) | Path,
       ...args?: Array<any>
     ): Array<any>;
     invokeMap<T: Object>(
       object: T,
-      path: ((value: any) => Array<string> | string) | Array<string> | string,
+      path: ((value: any) => Path) | Path,
       ...args?: Array<any>
     ): Array<any>;
     keyBy<T, V>(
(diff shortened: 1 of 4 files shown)


kubijo:
As a sidenote… I've also tried to do this with lodash-es, but that has **a lot** of errors with functions that I didn't touch and I couldn't be bothered to the the big investigation on whats wrong there.

Whats the idea on reporting this type of issue?
Do I try to fish-out original contributor from `git blame` and go to him, or do I just report it here?

definitions/npm/lodash_v4.x.x/flow_v0.38.x-v0.46.x/lodash_v4.x.x.js
```diff
@@ -1,3 +1,5 @@
+type Path = $ReadOnlyArray<string | number> | string | number;
```
villesau: This defines _global_ type named `Path` that would be available in all the files without explicitly importing it. In order to make private type, you need to move this inside the module declaration.

kubijo: I see ... I thought that you have to `declare` the type for it to be global and that the bare `type` is scoped to file (given it's not exported).

The reason for which I've moved it outside of the module declaration is because it's used in `lodash/fp` as well, so this seemed cleaner this way as opposed to copying.

Are you sure about the global issues?
If so, I'll move it gladly…

kubijo: So I've changed it anyway to move along … however I'm still curious about that scoping issue

villesau: Flow has a bit of special handling for library definitions. I went and tried out how similar definition to Path behves, and I didn't need to import it in any files. No `declaration` was needed.

kubijo: I see. Thanks for clarification!

PR #3157, opened 2019-02-15
Updated Express definition for Flow 0.93

daviduzumeri:
- Links to documentation:
- Link to GitHub or NPM: 
- Type of contribution: fix

Other notes: Updated reference to `http$IncomingMessage` to match Flow 0.93 requiring 0-1 type arguments

PR #3164, opened 2019-02-19
Add v0.93.x for cookie-parser

edahlseng:
- Links to documentation:
- Link to GitHub or NPM: 
- Type of contribution: fix

Other notes:
It looks like in flow v0.93.x, `http$IncomingMessage` is now a parameterized generic type. This pull request copies the existing `cookie-parser` definitions, and updates every instance of `http$IncomingMessage` with `http$IncomingMessage<>` (which is also what `express` does, since the types are copied from `express`).

(Side note: I notice that there's a large number of types that have been copied from `express`. Is there any way to reference those types without copying them? That may have helped to prevent this issue!)


pascalduez:
See 
https://github.com/facebook/flow/pull/7215   
https://github.com/flow-typed/flow-typed/pull/3157    

> (Side note: I notice that there's a large number of types that have been copied from `express`. Is there any way to reference those types without copying them? That may have helped to prevent this issue!)

Not at the moment, see #2842

PR #3165, opened 2019-02-19
Update styled-components for flow v0.92.x- for tagged template literals

edahlseng:
- Links to documentation:
- Link to GitHub or NPM: 
- Type of contribution: fix

Other notes:
Flow v0.92.x included some fixes to tagged template literals that made the existing `styled-components` definitions unusable when using `styled-components` interpolations. This pull request fixes that, and also adds the ability for a `theme` and `props` type to be specified when using `styled`. Example:

```js
const ComponentStyled = styled<ThemeType, PropsType, _>(Component)`
    color: ${props => props.theme.red};
`
```


definitions/npm/styled-components_v4.x.x/flow_v0.92.x-/styled-components_v4.x.x.js
```diff
@@ -0,0 +1,410 @@
...
+    rect:                     StyledComponentType<StyledElementType<'rect'>, Theme>,
+    stop:                     StyledComponentType<StyledElementType<'stop'>, Theme>,
+    svg:                      StyledComponentType<StyledElementType<'svg'>, Theme>,
+    text:                     StyledComponentType<StyledElementType<'text'>, Theme>,
+    tspan:                    StyledComponentType<StyledElementType<'tspan'>, Theme>
+  }
+
+  declare export default StyledComponentList & {
+    [[call]]: <S : string>(S) => $ElementType<StyledComponentList, S>,
+    [[call]]: <T: Theme, Props: {}, C: React$ComponentType<Props>>(C) => StyledComponentType<React$ComponentType<{...$Exact<Props>, className?: string}>, T>
```
edahlseng: Is there a better way to write this type? The `C` generic in this line was needed in order for Flow to properly infer the type of the component input to this function. Unfortunately, this means that when there's a desire to explicitly pass in `T` and `Props`, one has to write the following:

```js
const ComponentStyled = styled<ThemeType, PropsType, _>(Component)``;
```

The above example contains an unnecessary `_` for the third generic parameter. It appears that Flow doesn't allow generic defaults to be added in a libdef?? I'm not exactly sure what's going on/how to make this better. Any pointers would be greatly appreciated!

edahlseng:
Just fixed some of the failing tests, and added some additional tests to cover the interpolations work added in this PR. There are a number of expected errors that aren't actually error that I'm currently having trouble with. Will call out a question inline.

definitions/npm/styled-components_v4.x.x/flow_v0.92.x-/styled-components_v4.x.x.js
```diff
@@ -0,0 +1,410 @@
...
```
edahlseng: There are a few warnings that remain in the tests, that are due to this line, as best as I can tell. Editing this line as follows removes the warnings:
```suggestion
    [[call]]: <T: Theme, Props: {}, C: React$ComponentType<Props>>(C) => StyledComponentType<React$ComponentType<Props>, T>
```

But even editing this line to below still causes the warnings:
```suggestion
    [[call]]: <T: Theme, Props: {}, C: React$ComponentType<Props>>(C) => StyledComponentType<React$ComponentType<$Exact<Props>>, T>
```

I don't understand why this last example still causes errors. I don't see how `$Exact<Props>` should behave any differently than `Props` for the component type (or at least, I would expect it to cause _additional_ errors to arise, not fewer). Can anyone shed some light into why this isn't working as expected?

omninonsense:
Hi! I think is based off the the original typedef for v4 (but for an older version of flow) that I wrote. This seems to build on top of that. These look OK overall, but there's quite a few "ugly" hacks in what I wrote to work around some flow limitations (and it uses some undocumented/deprecated utility types on top of that) this that can go away, and also the whole thing can be simplified with a bit of refactoring.

I think a lot of the issues you've raised in the PR are related to these problems with the old typedefs. :point_up:

edahlseng:
@omninonsense thanks for the insight! I'll take a look at cleaning up the styles across the board, and see if that helps. Do you have any pointers on where to give attention first (other than following the comments)?

omninonsense:
Hey @edahlseng 

I originally started to write down a response to:

> Do you have any pointers on where to give attention first (other than following the comments)?

but it turned out to be too long and incoherent to actually be useful to you. So, I started a separate PR https://github.com/flow-typed/flow-typed/pull/3228. Can you please add your changes there since new libdef might be a bit easier to understand and work with than the old ones (the ones you're currently working with), and then we can merge it in all together? Just bare in mind, there's a few bugs in the current defs that I haven't gotten around to fixing yet.

Thanks!

edahlseng:
@omninonsense, that sounds great! Will add to that PR!

edahlseng:
Closing in favor of #3228.

PR #3168, opened 2019-02-19
Fix koa flow definition

franciscomorais:
- Links to documentation:
- Link to GitHub or NPM: 
- Type of contribution: fix

Other notes:
This PR fixes the `req` type in the `Request` definition.


pascalduez:
Tests are failing, looks like you missed several other `http$IncomingMessage` to update.

FireyFly:
Looks like the remaining `http$IncomingMessage` to update are in the test itself, rather than the libdef. Presuambly those errors would be resolved with the splitting into files for different Flow version ranges.

PR #3173, opened 2019-02-22
Add $Log and $Interval to angular 1.5.x type; fix test warnings

rgeraldporter:
- Links to documentation: https://docs.angularjs.org/api/ng/service/$interval and https://docs.angularjs.org/api/ng/service/$log
- Link to GitHub or NPM: https://www.npmjs.com/package/angular
- Type of contribution: addition (and fix for warnings)

Other notes:

Added types for `$log` and `$interval` (`$Log` and `$Interval` respectively).

Also there were many test warnings due to frequent use of `Object` and `Function` causing `$ExpectedError` not to throw where it should.

My solution was to replace all `Object` with `{}`, and any `Function` that were causing warnings were changed to `(...args: Array<*>) => *`.

All `Object` cases were replaced, not all `Function` were though.

Feedback very welcomed.

### DIFF
--- a/definitions/npm/angular_v1.5.x/flow_v0.47.x-/test_angular_interval_log_v1.5.x.js
+++ b/definitions/npm/angular_v1.5.x/flow_v0.47.x-/test_angular_interval_log_v1.5.x.js
@@ -1,0 +1,111 @@
+// @flow
+import type { $Interval, $Log, AngularPromise } from 'angular';
+
+function testInterval($interval: $Interval) {
+  // can be called with two args: fn, delay
+  $interval(() => {}, 0);
+  // can be called with two args
+  $interval(() => {}, 0);
+  // third argument is a number
+  $interval(() => {}, 0, 1);
+  // fourth argument is a boolean
+  $interval(() => {}, 0, 1, true);
+  // remaining arguments are passed to the function
+  $interval((a, b, c) => {}, 0, 1, false, 'a', 'b', 'c');
+  //$ExpectError third argument must be a number
+  $interval(() => {}, 0, true);
+  //$ExpectError fourth argument must be a boolean
+  $interval(() => {}, 0, 1, 2);
+  //$ExpectError must have two arguments
+  $interval();
+  //$ExpectError must have two arguments
+  $interval(() => {});
+  //$ExpectError must have two arguments minimum where fn is Function and delay is number
+  $interval(123, 1);
+  //$ExpectError must have two arguments minimum where fn is Function and delay is number
+  $interval(() => {}, 'string');
+  // returns back angular promise
+  ($interval(() => {}, 0): AngularPromise<*>);
+}
+
+function testLog($log: $Log) {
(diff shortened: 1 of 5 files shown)


definitions/npm/angular_v1.5.x/flow_v0.47.x-/test_angular_v1.5.x.js
```diff
@@ -108,7 +108,7 @@ describe("service", () => {
       "bar",
       "bazz",
       (bar, bazz) => {
-        return a => a + 1;
+        return { foo: "bar" };
```
rgeraldporter: One side note: this was a test defect: `Object` was allowing a number to be returned so this never threw an error until I changed `Object` to `{}`.

definitions/npm/angular_v1.5.x/flow_v0.47.x-/test_angular_interval_log_v1.5.x.js
```diff
@@ -0,0 +1,111 @@
+// @flow
+import type { $Interval, $Log, AngularPromise } from 'angular';
+
+function testInterval($interval: $Interval) {
```
AndrewSouthpaw: Since you're creating a new test file, can you use `describe/it` for better test organization?

rgeraldporter: Yes, good point.

rgeraldporter:
@AndrewSouthpaw Giving `(*) => *` a try again, I rediscovered what might have put me off that path:

```
    Unexpected token >
    
       189|     module: (...a: Array<string | ((*) => * | {}>) => () => void
                                                           ^
```

Where as 

```
    module: (...a: Array<string | ((...args: Array<*>) => *) | {}>) => () => void
```

is fine.

So far the other places it seems to take though. Thanks.

rgeraldporter:
@AndrewSouthpaw thanks for that, making those changes now.

definitions/npm/angular_v1.5.x/flow_v0.47.x-/test_angular_interval_log_v1.5.x.js
rgeraldporter: Actually, do you have any ideas how to get that to work with angular's dependency injection? I suspect that's why most the other files don't have describe/it...

rgeraldporter:
Made those changes, except one spot where the tests would fail with that change.

definitions/npm/angular_v1.5.x/flow_v0.47.x-/test_angular_interval_log_v1.5.x.js
rgeraldporter: Never mind, figured it out. No need for injection, just need to name the args right.

AndrewSouthpaw:
Thanks for making the changes!

PR #3175, opened 2019-02-23
[lodash-es_v4.x.x] add defs for standalone lodash-es modules

dszakallas:
Sadly some module bundlers still can't manage to shrink size to a minimum when using named imports from `lodash-es`, see e.g. https://github.com/rollup/rollup/issues/1403 and https://github.com/rollup/rollup/issues/691. Because of this, it makes sense to include standalone module definitions as well.

- [x] Add all standalone lodash modules
- [x] Test


pascalduez:
@dszakallas Will you add some tests (a couple would be enough)? It's listed in your description. Otherwise looks good.

dszakallas:
Some of the original tests are failing. I didn't add new failures.

pascalduez:
Yeah, the tests are in a pretty bad shape... :\

PR #3176, opened 2019-02-23
[firebase_v5.x.x] Add missing firestore definitions and fix incorrect definitions

SamChou19815:
Changelog:
- Added `enableNetwork`, `disableNetwork` and `enablePersistence` methods for `firebase.firestore` that currently exist in TS but not in flow-typed.
  - https://github.com/firebase/firebase-js-sdk/blob/4d82d1e6e697d2a3fef09757a7cac3700116c052/packages/firestore-types/index.d.ts#L119-L139
  - https://github.com/firebase/firebase-js-sdk/blob/4d82d1e6e697d2a3fef09757a7cac3700116c052/packages/firestore-types/index.d.ts#L188-L204
- Added missing type `firebase.firestore.Settings` and `firebase.firestore.PersistanceSettings`.
  - https://github.com/firebase/firebase-js-sdk/blob/4d82d1e6e697d2a3fef09757a7cac3700116c052/packages/firestore-types/index.d.ts#L40-L99
- Added many missing `isEqual` methods that exist in TS but not in flow-typed.
- Added the `path` field to `firebase.firestore.DocumentReference`.
  - https://github.com/firebase/firebase-js-sdk/blob/4d82d1e6e697d2a3fef09757a7cac3700116c052/packages/firestore-types/index.d.ts#L584-L605
- Changed `documentId` method in `FieldPath` from instance method to static method to match the definition in TS.
  - https://github.com/firebase/firebase-js-sdk/blob/4d82d1e6e697d2a3fef09757a7cac3700116c052/packages/firestore-types/index.d.ts#L1293-L1312
- Changed many unclear type `Function` to `() => void` to match the definition in TS.

Links:
- Links to documentation: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts
- Link to GitHub or NPM: https://github.com/firebase/firebase-js-sdk/blob/master/packages/firestore-types/index.d.ts
- Type of contribution: addition & fix

### DIFF
--- a/definitions/npm/firebase_v5.x.x/flow_v0.34.x-/firebase_v5.x.x.js
+++ b/definitions/npm/firebase_v5.x.x/flow_v0.34.x-/firebase_v5.x.x.js
@@ -567,9 +567,12 @@ declare class $npm$firebase$firestore$FieldPath {
   documentId(): typeof $npm$firebase$firestore$FieldPath;
 }
 
-declare interface $npm$firebase$firestore$FieldValue {
-  delete(): $npm$firebase$firestore$FieldValue;
-  serverTimestamp(): $npm$firebase$firestore$FieldValue;
+declare class $npm$firebase$firestore$FieldValue {
+  static delete(): $npm$firebase$firestore$FieldValue;
+  static serverTimestamp(): $npm$firebase$firestore$FieldValue;
+  static arrayUnion(...elements: any[]): $npm$firebase$firestore$FieldValue;
+  static arrayRemove(...elements: any[]): $npm$firebase$firestore$FieldValue;
+  isEqual(other: $npm$firebase$firestore$FieldPath): boolean;
 }
 
 declare type $npm$firebase$firestore$FirestoreError =
--- a/definitions/npm/firebase_v5.x.x/flow_v0.34.x-/firebase_v5.x.x.js
+++ b/definitions/npm/firebase_v5.x.x/flow_v0.34.x-/firebase_v5.x.x.js
@@ -871,7 +871,7 @@ declare module '@firebase/firestore' {
     DocumentReference: typeof $npm$firebase$firestore$DocumentReference,
     DocumentSnapshot: typeof $npm$firebase$firestore$DocumentSnapshot,
     FieldPath: typeof $npm$firebase$firestore$FieldPath,
-    FieldValue: $npm$firebase$firestore$FieldValue,
+    FieldValue: typeof $npm$firebase$firestore$FieldValue,
     Firestore: typeof $npm$firebase$firestore$Firestore,
     FirestoreError: $npm$firebase$firestore$FirestoreError,
     GeoPoint: typeof $npm$firebase$firestore$GeoPoint,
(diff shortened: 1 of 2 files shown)


definitions/npm/firebase_v5.x.x/flow_v0.34.x-/firebase_v5.x.x.js
```diff
@@ -541,24 +544,28 @@ declare interface $npm$firebase$firestore$DocumentChange {
...
     onError?: $npm$firebase$firestore$observerError
-  ): Function;
+  ): () => void;
   set(data: Object, options?: {| merge?: boolean, mergeFields?: string[] |} | null): Promise<void>;
   update(...args: Array<any>): Promise<void>;
 }
 
 declare class $npm$firebase$firestore$DocumentSnapshot {
-  data(): Object;
+  data(): Object | void;
```
pascalduez: Would be nice to replace `Object` while at it, cause it's an alias of `any` from Flow >= 0.75.  
`{ +[string]: mixed }` ?

SamChou19815: Although it is more sound to declare this as `{ +[string]: mixed }`, it might be too noisy for existing codebases.
If you see the _Listen to multiple documents in a collection_ in [official docs](https://firebase.google.com/docs/firestore/query-data/listen), the usage of the `data()` method is similar to that of `JSON.parse`.

Here is a example to show that it might be too noisy for existing codebases:
https://flow.org/try/#0PQKgBAAgZgNg9gdzAZwC4CcCWBjVYTABQhqAngA4CmYAIgIap1gC8YA3mANQDaaWAdgHMAugC4wAW0wAPSgBMwAXwDcxKAFd+uTHH5hBlVPUYAKAJTjjTNoTBhUAC3SIw-SkgCi6Z+nPKwhIrEZFRgAAqU6Mi6LOwAPrZc-HQSlOJ8mEKqdpx0BuL86hIARpGqcSrE2Lpo7K4plAA0YHnUimDiEVExrAZGDHR+hEA

I'm also thinking of remove the newly added `void` case because there are many places where you definitely know that the data exist. The example shown in the official docs is one such instance.

My proposed solution is `{| +[string]: any |}`. I use exact object notation so that it can be compatible with exact type declaration. Here is the demo:
https://flow.org/try/#0PTAEAEDMBsHsHdQGcAuAnAlgYxQKFygJ4AOApqACICGKVoAvKAN4A+oA1ANqqYB2A5gF0AXKCq9CoFgF8A3PkgBXXjgyxeofqRTVaACgCUo3XSa5QoFAAs0CUL1KIAomltpDs0Lmn4iZUAAKpGhI6gzMLOYcvFQAtqSiPBgC8hbsVFqivIqxAEbB8jLyuFjqqMz2caQANGJaoNKiQSFhjFo6NFQeuEA

If you like it, I can quickly make the change.

pascalduez: Your right `mixed` is more suited for inputs than return types. 
`{| +[string]: any |}` seems reasonable in such case.  

Seems to fit their TS declarations also:
https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts#L1556
https://github.com/firebase/firebase-js-sdk/blob/master/packages/firebase/index.d.ts#L754

SamChou19815: I also converted those unclear object types in input positions to `{ +[string]: mixed }`. I think it's good for merging now.

PR #3180, opened 2019-02-27
Update react-navigation_v3.x.x.js(missing props)

Fausto95:
- Links to documentation:
- Link to GitHub or NPM: https://github.com/react-navigation/react-navigation-tabs/blob/542a5cc35e19edbc272e4946d0754ec8c6d1941e/src/navigators/createBottomTabNavigator.js#L54

Other notes:

### DIFF
--- a/definitions/npm/react-navigation_v3.x.x/flow_v0.60.x-/react-navigation_v3.x.x.js
+++ b/definitions/npm/react-navigation_v3.x.x/flow_v0.60.x-/react-navigation_v3.x.x.js
@@ -1203,6 +1203,7 @@ declare module 'react-navigation' {
       jumpToIndex: (index: number) => void,
     }) => void,
     renderIcon: (scene: TabScene) => React$Element<*>,
+    getButtonComponent: (scene: TabScene) => React$Element<*>,
     labelStyle?: TextStyleProp,
     iconStyle?: ViewStyleProp,
   };
@@ -1231,6 +1232,7 @@ declare module 'react-navigation' {
     }) => void,
     getTestIDProps: (scene: TabScene) => (scene: TabScene) => any,
     renderIcon: (scene: TabScene) => React$Node,
+    getButtonComponent: (scene: TabScene) => React$Node,
     style?: ViewStyleProp,
     animateStyle?: ViewStyleProp,
     labelStyle?: TextStyleProp,


Ashoat:
Awesome, thanks so much for adding this missing prop!

Can you give the [`CONTRIBUTING`](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-navigation_v3.x.x/CONTRIBUTING.md) guidelines a quick read? The first step to making this change is actually to submit a PR on the mainline `react-navigation` repo.

Fausto95:
https://github.com/react-navigation/react-navigation/pull/5648
Merge!

definitions/npm/react-navigation_v3.x.x/flow_v0.60.x-/react-navigation_v3.x.x.js
```diff
@@ -1231,6 +1232,7 @@ declare module 'react-navigation' {
     }) => void,
     getTestIDProps: (scene: TabScene) => (scene: TabScene) => any,
     renderIcon: (scene: TabScene) => React$Node,
+    getButtonComponent: (scene: TabScene) => React$Node,
```
pascalduez: Why one is returning a React.Element and the other a React.Node, can you point us to some code/docs?

Ashoat:
Hey @Fausto95, sorry the review in React Navigation took so long! Next time try to tag me in the PR and I can try to get it reviewed and merged more quickly.

Curious about the question @pascalduez asked as well.

definitions/npm/react-navigation_v3.x.x/flow_v0.60.x-/react-navigation_v3.x.x.js
pascalduez: A quick search through the react-navigation github org shows it should be more `React.ComponentType<*>`

Fausto95: Both `React$Node` & `React$Element` are acceptable as React Component return type
This `React.ComponentType` might be the newer type declaration

pascalduez: They are three distinct types serving different purposes.  
From what I understand `getButtonComponent` is not rendering anything, it returns a component. 

Again, please point us to some docs or material asserting those changes.

Ashoat: I looked into this (sorry for the delay). `_TabBarTopProps` does not actually even have a `getButtonComponent` prop ([source](https://github.com/react-navigation/react-navigation-tabs/blob/master/src/views/MaterialTopTabBar.js)), whereas `_TabBarBottomProps` does and it is a `React.ComponentType` as you said ([source](https://github.com/react-navigation/react-navigation-tabs/blob/master/src/views/BottomTabBar.js)).

I think we can probably just close this PR - I'll put up a new one in a sec.

Ashoat:
https://github.com/flow-typed/flow-typed/pull/3259

PR #3183, opened 2019-03-01
Fix typing on enzyme's Wrapper.find(Component)

jdelStrother:
- Links to documentation: https://airbnb.io/enzyme/docs/api/selector.html
- Link to GitHub or NPM: https://github.com/airbnb/enzyme/blob/master/packages/enzyme/src/ReactWrapper.js

Other notes:

When using Enzyme's `find` methods:
```js
const wrapper = mount(<RootComponent />)
const child = wrapper.find(SomeDeeperComponent)
```

\- prior to this PR, `child` has a type of `Wrapper<RootComponent>`, but it ought to have a type of `Wrapper<SomeDeeperComponent>`.

This PR special-cases the find/parent/closest/children methods so that if you pass a component type to them, the result will be of that component type.

---

Enzyme also supports passing strings or object literals to `find` - eg `wrapper.find(".foo")` - in which case we don't have enough type information to know what the result will be.  The type declaration currently returns `this` - ie, `Wrapper<RootComponent>`:
```js
  declare class Wrapper<RootComponent> {
    find(selector: UntypedSelector): this,
    find<T: React$ElementType>(selector: T): ReactWrapper<T>
```

\- but perhaps we ought to change it to something like:

```js
  declare class Wrapper<RootComponent> {
    find(selector: UntypedSelector): ReactWrapper<React$ElementType>,
    find<T: React$ElementType>(selector: T): ReactWrapper<T>
```

@marudor it looks like you created that typing originally - any opinions?

### DIFF
--- a/definitions/npm/enzyme_v3.x.x/flow_v0.53.x-/enzyme_v3.x.x.js
+++ b/definitions/npm/enzyme_v3.x.x/flow_v0.53.x-/enzyme_v3.x.x.js
@@ -4,16 +4,19 @@ declare module "enzyme" {
     index: number
   ) => boolean;
   declare type NodeOrNodes = React$Node | Array<React$Node>;
-  declare type EnzymeSelector = string | {} | React$ElementType;
+  declare type UntypedSelector = string | {[key: string]: number|string|boolean};
+  declare type EnzymeSelector = UntypedSelector | React$ElementType;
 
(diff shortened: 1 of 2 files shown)


marudor:
I'm open for both. But this is already a great addition. Thanks.

PR #3186, opened 2019-03-02
[verror_v1.10.x] Add definitions

pascalduez:
- Links to documentation: https://github.com/joyent/node-verror
- Link to GitHub or NPM: https://github.com/joyent/node-verror

Other notes:

I struggled a bit making it accept commonjs default and named exports as well as constructor and callable signature for classes...

It currently works for commonjs, but named exports for ES modules are not recognized.
`import VError, { WError } from 'verror'; // Error` 

One would need to:
```
import VError from 'verror';
const { WError } = VError;
````
If someone see a better solution.

See https://github.com/joyent/node-verror/blob/master/lib/verror.js#L16-L23


pascalduez:
Note added.

jedwards1211:
How common is it for people to use `require` in flow-checked code?

PR #3189, opened 2019-03-03
[mocha_v6.x] Add definitions for mocha v6 (taken from mocha_v5.x)

AVykhrystyuk:
- Links to documentation: https://github.com/mochajs/mocha
- Link to GitHub or NPM: https://github.com/mochajs/mocha

### DIFF
--- a/definitions/npm/mocha_v6.x.x/flow_v0.28.x-/mocha_v6.x.x.js
+++ b/definitions/npm/mocha_v6.x.x/flow_v0.28.x-/mocha_v6.x.x.js
@@ -1,0 +1,219 @@
+declare interface $npm$mocha$SetupOptions {
+  slow?: number;
+  timeout?: number;
+  ui?: string;
+  globals?: Array<any>;
+  reporter?: any;
+  bail?: boolean;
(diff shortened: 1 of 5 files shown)


pascalduez:
Thanks for hte PR!  
Could you detail a bit any additions/changes if any? 
Actually, the best when copying an exiting definitions, is to split the actions into commits. 
One commit for copying/bumping, other ones for updates.

AlexeyVykhrystyuk-GB:
> Actually, the best when copying an exiting definitions, is to split the actions into commits.

Ahh sorry for that, for sure will do it next time the way you described

There was not changes at all, this is just a copy of mocha_v5.x with bumping

AVykhrystyuk:
> Actually, the best when copying an exiting definitions, is to split the actions into commits.

Ahh sorry for that, for sure will do it next time the way you described

There were no changes at all, this is just exact copy of mocha_v5.x with the version bumping

pascalduez:
> There were no changes at all, this is just exact copy of mocha_v5.x with the version bumping

All right, thanks!

PR #3114, opened 2019-01-31
[query-string_v6.x.x] Allow number, boolean, null, and void for stringify values

amccloud:
- Links to documentation: https://github.com/sindresorhus/query-string#falsy-values
- Link to GitHub or NPM: https://github.com/sindresorhus/query-string
- Type of contribution: fix

Other notes:

This is a followup to https://github.com/flow-typed/flow-typed/pull/2657

According to the query-string docs boolean, number, null, and void are also [valid values into `stringify`](https://github.com/sindresorhus/query-string#falsy-values)

```js

//=> 'foo=false'

queryString.stringify({foo: null});
//=> 'foo'

queryString.stringify({foo: undefined});
//=> ''
```

cc: @callumlocke @villesau

PR #3193, opened 2019-03-07
[query-string] Fix a problem with typed query objects

julienw:
This fixes a problem with #3114. I think the added test clearly shows the problem, but please ask if you need more information.


julienw:
Only tests for query-string_v5 (that I didn't touch) are failing.

PR #3194, opened 2019-03-07
Further improved knex support for Flow

daviduzumeri:
- Links to documentation: https://knexjs.org/
- Link to GitHub or NPM: https://www.npmjs.com/package/knex
- Type of contribution: fix

Other notes:

Sorry for the second PR on this so soon -- I started applying the new definition to some more microservices I was refactoring to be fully flow-typed and found a few more inconsistencies between the docs and the defs.

(Out of curiosity -- I'd be lying out of my ass if I didn't say I'm using the TypeScript definitions as a guide -- what's the general thoughts about breaking changes by migrating/transforming the TS defs to Flow format, since those are maintained by the actual project?)


pascalduez:
>  I started applying the new definition to some more microservices I was refactoring to be fully flow-typed and found a few more inconsistencies between the docs and the defs.

That's a pretty good sign I would say, there's no better tests than dog fooding :)

> what's the general thoughts about breaking changes by migrating/transforming the TS defs to Flow format, since those are maintained by the actual project?

Whatever makes the current def more accurate and up to date would be welcome I would say, especially when they are maintained by the project?
There are some tools for the conversion also.

PR #3196, opened 2019-03-07
Antd updates

ChristopherChudzicki:
- Links to documentation: https://ant.design/docs/react/introduce
- Link to GitHub or NPM: https://github.com/ant-design/ant-design
- Type of contribution: addition

Other notes: This PR adds a few missing `antd` components, as well as props for some of them.

PR #3197, opened 2019-03-11
antd handler fixes

ChristopherChudzicki:
- Links to documentation: https://ant.design/docs/react/introduce
- Link to GitHub or NPM: https://github.com/ant-design/ant-design
- Type of contribution: fix

In #3196, I erroneously made event arguments in a few event handlers optional. They are always provided and so should be required. 

Sorry for the error, I had a misunderstanding about matching function types.

### DIFF
--- a/definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
+++ b/definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
@@ -398,8 +398,8 @@ declare module "antd" {
     okText?: string,
     okType?: 'primary' | 'dashed' | 'ghost' | 'danger' | 'default',
     title?: 'string' | React$Node,
-    onCancel?: (event: SyntheticEvent<HTMLElement>) => void,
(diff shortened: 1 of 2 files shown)


definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
```diff
@@ -398,8 +398,8 @@ declare module "antd" {
     okText?: string,
     okType?: 'primary' | 'dashed' | 'ghost' | 'danger' | 'default',
     title?: 'string' | React$Node,
-    onCancel?: (event?: SyntheticEvent<HTMLElement>) => void,
-    onConfirm?: (event?: SyntheticEvent<HTMLElement>) => void,
+    onCancel?: (event: SyntheticEvent<HTMLElement>) => void,
```
pascalduez: If we can't know the element type in advance, I would rather use the `SyntheticEvent<>` form.

ChristopherChudzicki: @pascalduez Thanks, changed.

PR #3201, opened 2019-03-12
Add defs to Chai and LRUCache

rifflock:
- Links to documentation:
https://www.chaijs.com/api/bdd/#method_match
- Link to GitHub or NPM: 
https://www.npmjs.com/package/lru-cache/v/4.1.5
https://www.npmjs.com/package/chai

addition | fix

Other notes:
Added `prune` to LRU_Cache
Fixed chai to support `matches` alias


AndrewSouthpaw:
Thanks for the PR! Currently your build is failing, and we'll need that to pass to merge it in. You can check out the build in Travis, or read about running libdef tests locally in CONTRIBUTING.md.

Sometimes the build was already failing for that libdef; if you can at least establish that your changes didn't introduce any new errors that would be fine too. Thanks!

rifflock:
Looks like a bunch of warnings around unnecessary ExpectError comments that were there before. I'll remove them.

pascalduez:
@rifflock It's most likely due to Flow > 0.75 aliasing `Object`, `Function` to `any`. So those comments are not to remove! Instead we should replace those types with stronger ones.

rifflock:
Oh yeah, removing those comments just made it worse.

I just had to make that change in one of our internal packages when jumping from 0.66 to 0.94, let me see if I can do the same with the Chai package.

rifflock:
Okay!
I squashed my merges because there were some useless changes in there that I reverted.
It was indeed the change in 0.88 that caused the failure. Unrelated to my changes, but better to fix it now.

AndrewSouthpaw:
Brilliant, thanks for your work!

PR #3214, opened 2019-03-15
Update bluebird def to include `onCancel`

EvHaus:
- Links to documentation: See http://bluebirdjs.com/docs/api/cancellation.html for documentation on his feature.
- Link to GitHub or NPM: https://github.com/petkaantonov/bluebird/blob/master/docs/docs/api/cancellation.md


definitions/npm/bluebird_v3.x.x/flow_v0.70.x-/bluebird_v3.x.x.js
```diff
@@ -189,7 +189,8 @@ declare class Bluebird$Promise<+R> extends Promise<R> {
   constructor(
     callback: (
       resolve: (result?: $Promisable<R>) => void,
-      reject: (error?: any) => void
+      reject: (error?: any) => void,
+      onCancel?: (fn?: Function) => void,
```
pascalduez: `Function` is aliased to `any` from Flow 0.75, so here `() => void` should fo it I guess.

EvHaus: Fixed.

PR #3220, opened 2019-03-19
[koa_v2] support custom middleware context

### LINKED ISSUE #3212: Koa: how to extend context?
Koa's middlewares add new fields to context sometimes. Is there canonical way of representing this in a type-safe manner?

I can do this dynamically via refinements (checking that property is set and that it is an `instanceof` proper class), but it would be much nicer if I could specify that `app.use(someMiddleware)` changes app's context and further middlewares get this new context.

Another option would be to just declare presence of additional fields on the level of middleware, but I'm not sure if that is possible. My naive attempts at this failed with loud errors

Here's what I tried:

```
/* @flow */
import type { Context } from "koa";
import type Redis from "ioredis";


export type AppContext = {
  ...Context,
  dbs: {
    redis: Redis
  }
}
```

```
async function health(ctx: AppContext, next: () => Promise<void>) {…}

app.use(addRedis);
app.use(health);  // uses `ctx.dbs.redis` defined by `addRedis` above
```

and here are the errors:

```
Cannot call app.use with health bound to fn because:
 • property inspect is read-only in Context [1] but writable in AppContext [2] in the first argument.
 • property toJSON is read-only in Context [1] but writable in AppContext [2] in the first argument.
 • undefined [3] is incompatible with function type [4] in property accepts of the first argument.
 • undefined [3] is incompatible with function type [5] in property accepts of the first argument.
 • undefined [3] is incompatible with function type [6] in property accepts of the first argument.
 • undefined [7] is incompatible with function type [8] in property acceptsCharsets of the first argument.
 • undefined [7] is incompatible with function type [9] in property acceptsCharsets of the first argument.
 • undefined [7] is incompatible with function type [10] in property acceptsCharsets of the first argument.
 • undefined [11] is incompatible with function type [12] in property acceptsEncodings of the first argument.
 • undefined [11] is incompatible with function type [13] in property acceptsEncodings of the first argument.
 • undefined [11] is incompatible with function type [14] in property acceptsEncodings of the first argument.
… (more of these)
```

HsuTing:
- Links to documentation: https://github.com/koajs/koa/tree/master/docs/api
- Link to GitHub or NPM: https://github.com/koajs/koa

Other notes:
I try to fix #3212 , but this pr have some problem.
1. flow not support [generic default parameters](https://github.com/facebook/flow/issues/6875) in class. So, we can only use `app.use<{}>(...`, not `app.use<>(...`.

2. If user do not use generic in function, they will not get any error, for example:

```js
app.use((ctx, next) => {
  (ctx.body: number); // this should throw error, but it do not throw.
});

app.use<{}>((ctx, next) => {
  (ctx.body: number); // this will throw error.
});
```
I can not find the more better way to modify this type.


pascalduez:
@HsuTing I think your overriding the class generic with the method one.  
Would that work?  
https://flow.org/try/#0CYUwxgNghgTiAEkoGdnxANxAOwC7IBIBRLPIgWwEtdcQZ4BvAXwCgXQk5FpV4BBAA4CIlMFFyUA9tgA8AYXgBeRkwB86AB61swNJhz5ipXBWq16DFvHgBXZCAAUtZLgBc8OQEp3uABaVkABoWVmsWCBBceCghJXhsEAB3fiERMQlpGVUHTwBuK3gWGIEAOjtHZjyWAHpq+GINAXATGBhJGCKhMvsHAEYAJgBmPKA

pascalduez:
Maybe `use<C: {}>(fn: Middleware<C>): this`, at least it will report errors if omitted. 

https://flow.org/try/#0CYUwxgNghgTiAEkoGdnxANxAOwC7IBIBRLPIgWwEtdcQZ4BvAXwCgXQk5FpV4BBAA4CIlMFFyUA9tnQAPWtmBpMOfMVK4K1WvQYt48AK7IQAHgDCALkZMAfAApayXNfMBKa7gAWlZABoWVgMWCBBceCgheABeeGwQAHd+IRExCWl7NwBufXgWSIEAOmMQe2ZslgB6SvhiWQFwTRgYSRh8oWKTewBGACYAZmygA

PR #3226, opened 2019-03-22
Add sentry-expo package.

AndrewSouthpaw:
- Links to documentation: https://github.com/getsentry/react-native-sentry/blob/master/lib/Sentry.d.ts
- Link to GitHub or NPM: https://www.npmjs.com/package/sentry-expo

As mentioned in the libdef comments, `sentry-expo` mostly delegates to the `react-native-sentry` package. Adding the definitions from there into this libdef.


definitions/npm/sentry-expo_v1.11.x/flow_v0.83.x-/sentry-expo_v1.11.x.js
```diff
@@ -0,0 +1,87 @@
...
+  declare export type SentrySeverity = "fatal" | "error" | "warning" | "info" | "debug" | "critical"
+
+  declare type SentryLog = {
+    None: 0,
+    Error: 1,
+    Debug: 2,
+    Verbose: 3
+  }
+
+  declare type SentryOptions = {
```
pascalduez: Exact + covariant keys? ;)

definitions/npm/sentry-expo_v1.11.x/flow_v0.83.x-/sentry-expo_v1.11.x.js
```diff
@@ -0,0 +1,87 @@
...
+    logLevel?: SentryLog;
+    instrument?: boolean;
+    disableNativeIntegration?: boolean;
+    ignoreModulesExclude?: [string];
```
pascalduez: The TS declarations are `string[]` which is the same as `Array<string>`. 
`[string]` will only allows one item per array, is that intended?  

https://flow.org/try/#0DYUwLgBGIM5gXBA2nATgSwHYHMC6EBeZAcmIBoJTcBuIA

definitions/npm/sentry-expo_v1.11.x/flow_v0.83.x-/sentry-expo_v1.11.x.js
```diff
@@ -0,0 +1,87 @@
...
+
+    static config(dsn: string, options?: SentryOptions): Sentry;
+
+    static isNativeClientAvailable(): boolean;
+
+    static crash(): void;
+
+    static nativeCrash(): void;
+
+    static setEventSentSuccessfully(callback: Function): void;
```
pascalduez: Maybe `() => void` instead of `Function`?

definitions/npm/sentry-expo_v1.11.x/flow_v0.83.x-/sentry-expo_v1.11.x.js
AndrewSouthpaw: I happen to despise covariant keys, we did that for a while in a project of mine and then it got us into snarls of problems because Flow is subtle and strange. I'd be happy to make them exact but I'd only make them covariant if there's a compelling reason to do so.

definitions/npm/sentry-expo_v1.11.x/flow_v0.83.x-/sentry-expo_v1.11.x.js
AndrewSouthpaw: Woops, seems like a copy error. Nice catch.

definitions/npm/sentry-expo_v1.11.x/flow_v0.83.x-/sentry-expo_v1.11.x.js
AndrewSouthpaw: 👍

definitions/npm/sentry-expo_v1.11.x/flow_v0.83.x-/sentry-expo_v1.11.x.js
pascalduez: I understand.  
Just [we try to enforce it](https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#prefer-immutability) for options like objects now.

AndrewSouthpaw: Ah I didn't see that one. Good to know for next time.

PR #3228, opened 2019-03-23
styled-components: keeping up with flow

omninonsense:
Modernise the typedefs for styled-components to support new flow which has more type level soundness around some React abstractions.

Only starting this, but below is a (non-finalised) list of ideas/todos I'd like to add.

Todo:

 - [x] Pin flow versions
 - [x] Remove type-level lies
 - [x] Add support for forward refs
 - [x] Fix `withTheme` HOC
 - [x] Fix template string literals
 - [x] Add support for styled-components inside interpolations
 - [x] Add back react native stuff
 - [x] Decrease version requirements to Flow v0.92.0 before merging
 - [x] Fix `attrs` API (partially)

---

If anyone else wants to contribute, please do!


omninonsense:
Right… Unless I don't know something, or am having a brain fart, I think that currently, it's not possible to fully support the `styled-components` API. The best I can do it support it without `attrs`. The problem is that there's no way to describe a function which also has a parameter on it (`attrs`) without using something like:

~~~js
  declare export type StyledComponent<StyleProps: {}, OwnProps: {}, Instance> = React$AbstractComponent<StyleProps & OwnProps, Instance>
  declare interface StyledFactory<StyleProps, OwnProps, Theme, Instance> {
    [[call]]: TaggedTemplateLiteral<StyleProps & OwnProps & ThemeProps<Theme>, StyledComponent<StyleProps, OwnProps, Instance>>;
    +attrs: <A: {}>(A | (OwnProps & StyleProps) => A) => React$AbstractComponent<$Diff<OwnProps & StyleProps, A>, Instance>;
  }
~~~

However, this is not useful, since there's no way to delay specialising this generic.

The alternative approach I found, which is kinda verbose, but works well with type-checking the template, is this:

~~~js
  declare interface StyledShorthands {
    div: <S, O, T, I>(strings: string[], ...interpolations: Interpolation<S & O & ThemeProps<T>>[]) => StyledComponent<S, O, BuiltinElementType<'div'>>,
    span: <S, O, T, I>(strings: string[], ...interpolations: Interpolation<S & O & ThemeProps<T>>[]) => StyledComponent<S, O, BuiltinElementType<'span'>>
  }
~~~

(`BuiltinElementType<>` is just a helper that maps `div` → `HTMLDivElement`, etc)

I think we probably need a way to create an anonymous interface maybe? So we can do something like

~~~js
div: interface <S, O, T, I> {
  [[call]]: <S, O, T, I>(strings: string[], ...interpolations: Interpolation<S & O & ThemeProps<T>>[]) => StyledComponent<S, O, BuiltinElementType<'div'>>,
  +attrs: <A: {}>(A | (S & O) => A) => React$AbstractComponent<$Diff<S & O, A>, I>;
}
~~~

Alternatively, we need some syntax that allows us to something like

~~~js
  declare interface StyledShorthands {
    div<S, T>: StyledFactory<S, {}, T, BuiltinElementType<'div'>>
   // ...
  }
~~~

---

Does someone know of syntax/ways of achieving this?

I'll tag the maintainers in case they might have an idea: @AndrewSouthpaw, @GAntoine, @villesau

omninonsense:
Might've found a way; I'll play with it a bit. There's a lot of duplication and it's a bit ugly, but we can address that later maybe.

omninonsense:
Found a way to get rid of the duplication and some of the uglyness as well. I _think_… No `:tada:` emoji yet.

omninonsense:
Got some stuff working on https://github.com/omninonsense/flow-typed/tree/nino--flow92-sc4-round2, but there's still stuff missing. I'll merge it into this branch once I do some cleanup there.

I probably won't have time to look at this until next weekend, so if someone wants to help, please do :)

edahlseng:
Hi @omninonsense, just coming up to speed with this PR. It looks like all of the changes from nino--flow92-sc4-round2 have already been added to this branch, and all of the tests are passing. Unless I'm missing something? I'd love to help out with this PR wherever needed, let me know what's still pending!

Side note: looks like the tests are also passing on flow v0.92.0, so the lower bound for these changes could be dropped to that!

definitions/npm/styled-components_v4.x.x/flow_v0.95.2-/styled-components_v4.x.x.js
```diff
@@ -0,0 +1,444 @@
+// @flow
+
+declare module 'styled-components' {
+  // Could this use `React$ElementRef<>` instead?
+  // It might be more correct and future-proof
+  // Since it won't rely on us keeping this up to date
+  // TODO: Experiment/Look into this
```
edahlseng: In most cases, I don't believe using `React$ElementRef<>` will make a practical difference, as (for example) `React$ElementRef<'div'>` will return `HTMLDivElement`, which is what is already listed here. To your point about future-proofing these types, though, I could see a benefit for using `React$ElementRef`, in case some of the elements that don't have specialized element types gain those in the future (I imagine the chances of that are rare, but possible?). Given that, I'd say +1 for using `React$ElementRef`!

villesau:
Ok, looks like @edahlseng has already noticed this :) I was about to ping you since you have already discussed about similar stuff in here: https://github.com/facebook/flow/issues/7306 I believe you have better understanding on tagged template literals & styled-components with Flow than I do.

omninonsense:
Hi!

> Hi @omninonsense, just coming up to speed with this PR. It looks like all of the changes from nino--flow92-sc4-round2 have already been added to this branch, and all of the tests are passing. Unless I'm missing something? I'd love to help out with this PR wherever needed, let me know what's still pending!

You're right; I did merge `nino--flow92-sc4-round2` into `nino--flow92-sc4` this morning. It seemed to be OK, so I added it back in. There's a few tests/things I want to add (they're on a post-it note) and gotta fix/add back in the React native stuff.

> Side note: looks like the tests are also passing on flow v0.92.0, so the lower bound for these changes could be dropped to that!

I'm aware of that, the reason I pinned it to v0.95.2 was to reduce the console output when running `node cli/dist/cli.js run-tests styled-components_v4`, since it displays errors for all the versions. I'll add a to-do to remember to move down the v0.92 before merging.

definitions/npm/styled-components_v4.x.x/flow_v0.95.2-/styled-components_v4.x.x.js
omninonsense: I am not sure which one of these two we should do then, though:

 1.  `div: 'div'` and then use the `V` inside `React$ElementRef` when building `ConvenientShorthands`. This is slightly more "convenient" in case Flow devs decide to change their mind (wouldn't be the first time), but it's slightly less obvious WTF is going on.

 2.  `div: React$ElementRef<'div'>`, and just use that as `V` when building `ConvenientShorthands`. This has the benefit of being a little bit more explicit, but it might require more "work"<sup>[1](#workbitch)</sup> in the future to keep up

----

<a name="workbitch">1:</a> "Work": Well, it will probably just be running `sed` to search & replace stuff

edahlseng: Given that these defs will be community maintained by several different people, the clarity offered by option 2 may be worth the extra work. Especially considering that extra work could be automated, as you mentioned.

omninonsense: > Given that these defs will be community maintained by several different people, the clarity offered by option 2 may be worth the extra work. Especially considering that extra work could be automated, as you mentioned.

Yes, that's a very good point.

omninonsense: I'll mark this as resolved, as we have https://github.com/omninonsense/flow-typed/pull/1

omninonsense:
Just an apology, this week[end] was a bit of a mess, so I had almost paid no attention to this. I'll do better, I promise :sweat_smile:

definitions/npm/styled-components_v4.x.x/flow_v0.95.2-/styled-components_v4.x.x.js
```diff
@@ -0,0 +1,418 @@
...
+    tspan: React$ElementRef<'tspan'>,
+
+    // Deprecated, should be HTMLUnknownElement, but Flow doesn't support it
+    keygen: React$ElementRef<'keygen'>,
+    menuitem: React$ElementRef<'menuitem'>,
+  }
+
+  declare type BuiltinElementType<ElementName: string> = $ElementType<BuiltinElementInstances, ElementName>
+
+  declare export type Interpolation<P> =
```
w01fgang: It looks like there is missing `React$ComponentType<P>` https://github.com/styled-components/styled-components/blob/master/packages/styled-components/src/types.js#L9

omninonsense:
> Just an apology, this week[end] was a bit of a mess (in a _good_ way), so I had almost paid no attention to this. I'll do better, I promise

Completely forgot I won;t

definitions/npm/styled-components_v4.x.x/flow_v0.95.2-/styled-components_v4.x.x.js
omninonsense: @w01fgang This is a good point! Although it's a "yes-but-no" type of situation. So, we can use other _styled components_ as selectors (which will expand to the class name). However, the problem with what you linked (and what styled-components did) is that it's not true: it doesn't accept just any component, a subset of components (if even that is true).

Basically we need to accept something like this:

~~~
interface {
  +styledComponentId: string;
}
~~~

I am not sure if this is _correct_, though; styled-components could just use `.toString()` blindy (that gives us the the selector which is just `".${this.styledComponentId}"`). It could also do something else entirely, I haven't checked the code recently.

But, I will add it for now. And also make sure the component factory has it described at the type level.

benfjohnson:
Hi @omninonsense, any chance the fixes in place here might also be applicable to the 3.x.x branch of styled-components? We're seeing the same issue with this version. Cheers!

edahlseng:
@omninonsense let's get this PR through this week! Is the todo list in the PR description up to date? I can jump on a few things to help get this wrapped up!

omninonsense:
> @omninonsense let's get this PR through this week! Is the todo list in the PR description up to date? I can jump on a few things to help get this wrapped up!

Hi @edahlseng! I've got a lot goin on at the moment. I'm really sorry for not giving this PR a bit more love. I'll double-check the to-do and push what I have done tomorrow evening. We can then coordinate and split up tasks. Does that sound ok? :slightly_smiling_face:

edahlseng:
@omninonsense, all good, I understand the demands on time! :) That sounds like a good plan!

omninonsense:
So, I removed the "Functional API" from the list, because I forgot what it is… Maybe it's I imagined exists?

I have no idea how to add the `as` prop. The way they do it in TypeScript looks a bit insane, and I'm not sure we can even port that at all, or if that works.

I have never used React native, so if someone can please help with that.

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
```diff
@@ -0,0 +1,423 @@
...
+  // This is a bit hard to read. Not sure how to make it more readable. I think adding line-breaks makes it worse.
+  declare type ThemeProps<T> = {
+    theme: T
+  }
+
+  declare export function withTheme<Theme, Config: {}, Instance>(Component: React$AbstractComponent<Config, Instance>): React$AbstractComponent<$Diff<Config, ThemeProps<Theme | void>>, Instance>
+
+  declare export type StyledComponent<Props, Theme, Instance> = React$AbstractComponent<Props, Instance> & InterpolatableComponent
+  declare interface StyledFactory<StyleProps, OwnProps, Theme, Instance> {
+    [[call]]: TaggedTemplateLiteral<StyleProps & OwnProps & ThemeProps<Theme>, StyledComponent<StyleProps & OwnProps, Theme, Instance>>;
```
omninonsense: Maybe we should use `{...T1, ...T2}` instead of `T1 & T2` because of [intersection behaviour on exact object types](https://flow.org/en/docs/types/objects/#toc-exact-object-types). I tried a quick search and replace locally (maybe I effed up), but got errors; I'll double-check tomorrow morning.

omninonsense:
> Hi @omninonsense, any chance the fixes in place here might also be applicable to the 3.x.x branch of styled-components? We're seeing the same issue with this version. Cheers!

Hi @benfjohnson! Could you open a new issue for this, please? That way we can track/discuss things there?

---

I'm not too familiar with API chances in styled-components between the versions, but I from what I remember, I think it should be safe to backport these libdefs with some minor changes.

edahlseng:
@omninosense, regarding the remaining two items:
* React Native: I don’t use React Native myself, so the best I could do here is copy and paste and make sure the tests pass. Let me know if you’d like me to jump in with that or if you’re happy to!
* `as` prop: Did we support this prop previously in these flow-typed defs? I don’t think so, but please correct me if I’m wrong! If that’s the case then I’d recommend we open a new issue for that. There’s a lot of good changes here already that would be a huge step forward even without that prop!

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
edahlseng: I also saw some weirdness with trying to use spread with these defs at one point. I’ll double check when I’m at my computer next to see if that’s still the case.

edahlseng:
@omninonsense, I've submitted a PR against this branch for the React Native types. (I focused on just achieving parity with what previously existed, rather than full feature set). From the checklist on this PR, the only other outstanding item is the `as` prop, but I agree with your assessment about it not being worth the effort to support at the moment.

Given the above, once the React Native PR is merged, this PR should be good to go. I'm going to take these types and apply them to an existing codebase to see if there are any other conditions that we've neglected to cover, but I think we are getting close!

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
```diff
@@ -0,0 +1,423 @@
...
+  declare export var ThemeProvider: React$ComponentType<{children?: ?React$Node, theme: mixed | (mixed) => mixed}>
+
+  // This is a bit hard to read. Not sure how to make it more readable. I think adding line-breaks makes it worse.
+  declare type ThemeProps<T> = {
+    theme: T
+  }
+
+  declare export function withTheme<Theme, Config: {}, Instance>(Component: React$AbstractComponent<Config, Instance>): React$AbstractComponent<$Diff<Config, ThemeProps<Theme | void>>, Instance>
+
+  declare export type StyledComponent<Props, Theme, Instance> = React$AbstractComponent<Props, Instance> & InterpolatableComponent
```
edahlseng: When testing this out in a real codebase, I'm seeing several instances of the following:

```
Cannot create CustomComponent element because InterpolatableComponent [1] is not a React component.
```

Is there a better way that we can handle this?

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
edahlseng: I am also getting errors when attempting to use spread. I think you're right that we should be using the spread syntax, but we'll have to see what it will take to make that work...

benfjohnson:
> Hi @benfjohnson! Could you open a new issue for this, please? That way we can track/discuss things there?
> 
> I'm not too familiar with API chances in styled-components between the versions, but I from what I remember, I think it should be safe to backport these libdefs with some minor changes.

I've created the issue https://github.com/flow-typed/flow-typed/issues/3339 to backport the fixes here to 3.x.x. Thanks again for all the hard work @omninonsense, @edalseng!

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
omninonsense: @edahlseng Hmm. Can you provide the code that causes this error, if it's not too much code? Maybe we could make a test for that.

omninonsense: I think a fix might be to do:

~~~js
declare class InterpolatableComponent extends React$Component<Props, any> {
  +styledComponentId: string;
}
~~~

I think using the interface might have been a _bigger_ hack than this anyway. I'll try this approach and see if the current tests pass, but it would be nice to add your case with the failing tests.

edahlseng: Sounds good, I'll work on pulling out a minimal test case!

edahlseng:
When testing this out on an existing codebase I'm getting a number of the following errors:

```
Missing type annotation for SP. SP is a type parameter declared in function type [1] and was implicitly instantiated
at call of styled [2].

Missing type annotation for Th. Th is a type parameter declared in function type [1] and was implicitly instantiated
at call of styled [2].
```

These errors make sense with the new changes, but I wonder if there are opportunities for us to improve the developer experience for people who are suddenly going to see a ton of these errors. A first order change may be to update the names (e.g. `SP` to `StyledProps`, and `Th` to `Theme`), so that people have a better understanding of what the error is complaining about. Ideally, though, we would also be able to set defaults for these arguments, so that Flow complains about the errors at the site of the interpolation instead. As far as I'm aware, though, there's no way to add default arguments in lib defs. @omninonsense, do you know of any workarounds?

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
```diff
@@ -0,0 +1,431 @@
...
+  declare interface StyledFactory<StyleProps, OwnProps, Theme, Instance> {
+    [[call]]: TaggedTemplateLiteral<StyleProps & OwnProps & ThemeProps<Theme>, StyledComponent<StyleProps & OwnProps, Theme, Instance>>;
```
edahlseng: @omninonsense, I'm looking into options for replacing the intersections here with object spread instead. As I do that, I have a question about the design intent here:

It looks like `StyleProps` here is intended to only be the props used within the styled-component interpolations, is that correct? Would it make sense for us to pass in `StyleProps` as _all_ of the props that the styled component exports? As an example of how this could be used:

```js
type Props = {name : string}
  type StyleProps = { ...Props, color?: string}
  type Theme = {
    accent: string
  }

  const Hello = (p: Props) =>
    <div>Hello {p.name}</div>

  const StyledHello = styled<StyleProps, Theme, Props, *>(Hello)`
    color: ${props => props.color || props.theme.accent};
  `
```

At first glance it looks like this might simplify implementation a bit, as well as be simpler for consumers to maintain (they wouldn't have to update the `StyleProps` if they wanted to use an existing `Prop` in an interpolation). Thoughts?

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
omninonsense: So, this should probably be fixed by 121ec90. Would still be nice to have that test in there, so we know if it breaks in the future.

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
omninonsense: I like your idea better. I can't remember where I was going with the original idea!

omninonsense: > When testing this out on an existing codebase I'm getting a number of the following errors:
> 
> ```
> Missing type annotation for SP. SP is a type parameter declared in function type [1] and was implicitly instantiated
> at call of styled [2].
> 
> Missing type annotation for Th. Th is a type parameter declared in function type [1] and was implicitly instantiated
> at call of styled [2].
> ```
> 
> These errors make sense with the new changes, but I wonder if there are opportunities for us to improve the developer experience for people who are suddenly going to see a ton of these errors. A first order change may be to update the names (e.g. `SP` to `StyledProps`, and `Th` to `Theme`), so that people have a better understanding of what the error is complaining about. Ideally, though, we would also be able to set defaults for these arguments, so that Flow complains about the errors at the site of the interpolation instead. As far as I'm aware, though, there's no way to add default arguments in lib defs. @omninonsense, do you know of any workarounds?

I don't think there are any workarounds. I guess passing in an empty object might work (haven't tried)… Although, I am a bit reluctant on adding that in, since I am not sure if it will contribute much. IMO, people can always make their own utility types to help with these. Something along these lines:

~~~
type Theme = {| … |}
type MySC<Props> = StyledComponent<Props, Theme, *>
~~~

and they should be able to use it as `const msc: MySC<SomeProps> = styled(…)`

omninonsense:
> When testing this out on an existing codebase I'm getting a number of the following errors:
> 
> ```
> Missing type annotation for SP. SP is a type parameter declared in function type [1] and was implicitly instantiated
> at call of styled [2].
> 
> Missing type annotation for Th. Th is a type parameter declared in function type [1] and was implicitly instantiated
> at call of styled [2].
> ```
> 
> These errors make sense with the new changes, but I wonder if there are opportunities for us to improve the developer experience for people who are suddenly going to see a ton of these errors. A first order change may be to update the names (e.g. `SP` to `StyledProps`, and `Th` to `Theme`), so that people have a better understanding of what the error is complaining about. Ideally, though, we would also be able to set defaults for these arguments, so that Flow complains about the errors at the site of the interpolation instead. As far as I'm aware, though, there's no way to add default arguments in lib defs. @omninonsense, do you know of any workarounds?

@edahlseng I don't know of any workarounds. I guess passing in an empty object might work (haven't tried)… Although, I am a bit reluctant on adding that in, since I am not sure if it will contribute much. IMO, people can always make their own utility types to help with these. Something along these lines:

~~~
type Theme = {| … |}
type MySC<Props> = StyledComponent<Props, Theme, *>
~~~

and they should be able to use it as `const msc: MySC<SomeProps> = styled(…)`

omninonsense:
There seems to be some issues with flow 0.99 - 0.100, the `withTheme` doesn't work properly with functional components. I think this might be flow's fault. Opened a bug so we can track it here:

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
omninonsense: @edahlseng Implemented this in 45f8151. Thoughts?

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
omninonsense: Should be fixed in 
45f8151

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
omninonsense: **Note**: This also switched to using spread from using intersections, so we should have less "surprises"

omninonsense:
Is there a way to remove the duplication in `styled-components/native`? Could we just do something like `import type … from 'styled-components'` since we already declared that module in the file above, or will flow not like this?

pascalduez:
> Is there a way to remove the duplication in styled-components/native? Could we just do something like import type … from 'styled-components' since we already declared that module in the file above, or will flow not like this?

@omninonsense If both declarations are in the same file it will work, otherwise not.

omninonsense:
> > Is there a way to remove the duplication in styled-components/native? Could we just do something like import type … from 'styled-components' since we already declared that module in the file above, or will flow not like this?
> 
> @omninonsense If both declarations are in the same file it will work, otherwise not.

This is kind of disappointing, even though in this case it works in our favour! Thanks for the tip! I might "migrate" to this approach tomorrow (depending on how hungover I am) or sometime next week then

omninonsense:
@edahlseng Hi! Did you have time to review the latest changes by any chance? :grinning:

edahlseng:
Hi @omninonsense, thanks for being patient with my super delayed reply 😬.

I applied these changes to our component library codebase at work today and made all of the necessary updates to match the updated libdef. Using Flow v0.101.1, everything worked except for default props. I haven't looked into why that's the case yet, but I need to run for today. I've put up a [small PR](https://github.com/omninonsense/flow-typed/pull/3) to add a test case for that and will take another look this weekend if possible. Once default props is fixed up I think this is good to merge!

PierreCapo:
@omninonsense @edahlseng Thanks a lot for your awesome work, that is really needed ❤. Is this PR is still going to fix some issues on react-native stuffs ?

vasilich6107:
@omninonsense thanks for your work!
Do you have an example of using `withTheme`?

gwing33:
I was testing out this code on `v0.102.0` and was running into this issue. I poked around trying to figure out, but wasn't able to figure out what the solution should be.

```
Missing type annotation for V. V is a type parameter declared in function type [1] and was implicitly instantiated at
$ObjMap [2].

 [2] 269│     declare type ConvenientShorthands = $ObjMap<
     270│         BuiltinElementInstances,
 [1] 271│         <V>(
     272│             V
     273│             // TODO: Figure out how to add support for `attrs` API here
     274│             // TODO: Would be nice if we could find a way to use `StyledFactory` here, but I am too dumb to figure it out.
     275│         ) => <StyleProps, Theme>(
     276│             string[],
     277│             ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]
     278│         ) => StyledComponent<StyleProps, Theme, V>
     279│     >;
```

PierreCapo:
I have had the same issue too. I believe that pr should be merged to provide until v101 support which is super great. And then someone can open a new PR with change for v102 and upcoming flow versions ?

gwing33:
I'm not versed well enough with Flow generics/util functions, so please enlighten me if I'm wrong. But it seems like the `V` lost scope and isn't getting passed to the next function. After lots of tinkering around this works, however, I'm not sure if this is what we want/is correct:

```
declare type ConvenientShorthands = $ObjMap<
    BuiltinElementInstances,
    <V>(
        V
    ) => <StyleProps, Theme, V>(
        string[],
        ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]
    ) => StyledComponent<StyleProps, Theme, V>
>;
```

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
```diff
@@ -0,0 +1,443 @@
...
+    static +styledComponentId: string;
+  }
+
+  declare export type Interpolation<P> =
+                                       | ((executionContext: P) => Interpolation<any>) // eslint-disable-line flowtype/no-weak-types
+                                       | InterpolatableComponent<any> // eslint-disable-line flowtype/no-weak-types
+                                       | CSSRules
+                                       | KeyFrames
+                                       | string
+                                       | number
```
maxammann: I think types like `boolean` is missing here. Consider the following styled component:

```
type EnabledElementPropsType = {|selected: boolean, children: React.Node, theme: ThemeType|}

export const EnabledElement: StyledComponent<EnabledElementPropsType, ThemeType, *> = styled(Element)`
  color: ${props => props.theme.textColor};
  ${props => props.selected && `background-color: ${props.theme.colors.backgroundAccentColor}`};
`
```

I think it should be allowed to add a prop with the type "boolean".

omninonsense: These are not _props_ interpolations that styled components understands (when using the tagged template literal syntax). So, this doesn't make sense to styled components (AFAIK):
~~~js
const Something = styled.div`
  ${true}
`
~~~

omninonsense:
> I was testing out this code on `v0.102.0` and was running into this issue. I poked around trying to figure out, but wasn't able to figure out what the solution should be.
> 
> ```
> Missing type annotation for V. V is a type parameter declared in function type [1] and was implicitly instantiated at
> $ObjMap [2].
> 
>  [2] 269│     declare type ConvenientShorthands = $ObjMap<
>      270│         BuiltinElementInstances,
>  [1] 271│         <V>(
>      272│             V
>      273│             // TODO: Figure out how to add support for `attrs` API here
>      274│             // TODO: Would be nice if we could find a way to use `StyledFactory` here, but I am too dumb to figure it out.
>      275│         ) => <StyleProps, Theme>(
>      276│             string[],
>      277│             ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]
>      278│         ) => StyledComponent<StyleProps, Theme, V>
>      279│     >;
> ```

Hi! Thanks for catching this. Do you still have the code that produced this? Asking so we can add it to the tests, too? The tests seem to pass fine for 0.102

omninonsense:
> Is this PR is still going to fix some issues on react-native stuffs ?

Hi @PierreCapo!

I am not too familiar with how react native works. I'd prefer to keep the react-native typing improvements for another PR, since I am not really qualified to write (or review) them.

omninonsense:
> Hi @omninonsense, thanks for being patient with my super delayed reply .
> 
> I applied these changes to our component library codebase at work today and made all of the necessary updates to match the updated libdef. Using Flow v0.101.1, everything worked except for default props. I haven't looked into why that's the case yet, but I need to run for today. I've put up a [small PR](https://github.com/omninonsense/flow-typed/pull/3) to add a test case for that and will take another look this weekend if possible. Once default props is fixed up I think this is good to merge!

It looks like Flow doesn't support `defaultProps` on `React$AbstracComponent` intentionally… This appears to be a problem when using `React.memo`  and `React.forwardRef`, too. See https://github.com/facebook/flow/issues/7467#issuecomment-463049743

Since we _have_ to use an abstract component because styled components uses `React.forwardRef`. Uh. Not sure what to do about this. Anyone got any ideas?

definitions/npm/styled-components_v4.x.x/flow_v0.92.0-/styled-components_v4.x.x.js
maxammann: You are right, I had to change my code to `  ${props => props.selected ? `background-color: ${props.theme.colors.backgroundAccentColor}` : ''};` which makes a lot more sense.Thank you!

edahlseng:
Hi @omninonsense, thanks for that. Of course it doesn't support default props... :/

Looking through that PR that you linked to, I actually think that we could get away by migrating all of our default props over to default function arguments, since we've migrated most of our components to functional components anyways. I'll double check tomorrow and will reply back here. If that works, then at least we would have a way to support default props, which is encouraging for this PR. That still makes it a pain for users to migrate to these updated defs, though, although I don't currently see a good way around that.

gwing33:
@omninonsense I was able to make some time to dive back into this. I created a test project and at first look everything was functioning as expected, but I was able to reproduce the error when I add the `suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe` then using it on an untyped styled definition:

```
// @flow
import styled, { type StyledComponent } from 'styled-components';

export const Div: StyledComponent<{}, {}, HTMLDivElement> = styled.div``;

// $FlowFixMe
export const IgnoreDiv = styled.div``;
```

Now, if you keep the `$FlowFixMe` but add a type to the `IgnoreDiv`, everything works:
```
// @flow
import styled, { type StyledComponent } from 'styled-components';

export const Div: StyledComponent<{}, {}, HTMLDivElement> = styled.div``;

// $FlowFixMe
export const IgnoreDiv: StyledComponent<{}, {}, HTMLDivElement> = styled.div``;
```

I'm not sure this is an issue with your code, this might need to be reported as a bug to the Flow project, but someone more involved probably knows better than I.

lyleunderwood:
@gwing33 whenever some JS code fails to be typed due to a type error, flow does essentially (simplified for our purposes here) two things:

1. report the error to the user,
2. for anything that could not be correctly type due to the error, type each of those things as `any`,

then it carries on as if nothing has happened. this means that when you use `$FlowFixMe` you are (generally speaking) introducing one or more `any`s somewhere into the type flow.

- are you receiving an error if you remove the `$FlowFixMe`?
- if so, what error?
- can you reproduce the issue you're concerned with without using `$FlowFixMe`?

omninonsense:
> @omninonsense I was able to make some time to dive back into this. I created a test project and at first look everything was functioning as expected, but I was able to reproduce the error when I add the `suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe` then using it on an untyped styled definition:
> 
> ```
> // @flow
> import styled, { type StyledComponent } from 'styled-components';
> 
> export const Div: StyledComponent<{}, {}, HTMLDivElement> = styled.div``;
> 
> // $FlowFixMe
> export const IgnoreDiv = styled.div``;
> ```
> 
> Now, if you keep the `$FlowFixMe` but add a type to the `IgnoreDiv`, everything works:
> 
> ```
> // @flow
> import styled, { type StyledComponent } from 'styled-components';
> 
> export const Div: StyledComponent<{}, {}, HTMLDivElement> = styled.div``;
> 
> // $FlowFixMe
> export const IgnoreDiv: StyledComponent<{}, {}, HTMLDivElement> = styled.div``;
> ```
> 
> I'm not sure this is an issue with your code, this might need to be reported as a bug to the Flow project, but someone more involved probably knows better than I.
> 
> Edit: this is an issue in `v0.101` also.

Hmm. I don't this this has anything to do with these typedefs. What error do you get from Flow without the annotation _and_ the `$FlowFixMe` comment? It might just be that you're getting a "unused suppression comment" error, because Flow can infer the type just fine (I haven't tested this yet; it's just a hunch/theory)?

gwing33:
@lyleunderwood I appreciate your feedback/insight here. To answer and give more context:

> are you receiving an error if you remove the $FlowFixMe?

If I don't define anything, then I receive 3 errors for every one:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ flow-typed/npm/styled-components_v4.x.x.js:269:41

Missing type annotation for V. V is a type parameter declared in function type [1]
and was implicitly instantiated at $ObjMap [2].

     266│         ) => React$AbstractComponent<$Diff<StyleProps, A>, Instance>;
     267│     }
     268│
 [2] 269│     declare type ConvenientShorthands = $ObjMap<
     270│         BuiltinElementInstances,
 [1] 271│         <V>(
     272│             V
     273│             // TODO: Figure out how to add support for `attrs` API here
     274│             // TODO: Would be nice if we could find a way to use `StyledFactory` here, but I am too dumb to figure it out.
     275│         ) => <StyleProps, Theme>(
     276│             string[],
     277│             ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]
     278│         ) => StyledComponent<StyleProps, Theme, V>
     279│     >;
     280│
     281│     declare interface Styled {
     282│         <StyleProps, Theme, ElementName: $Keys<BuiltinElementInstances>>(


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/index.js:6:26

Missing type annotation for StyleProps. StyleProps is a type parameter declared in
function type [1] and was implicitly instantiated at encaps tag [2].

     src/index.js
       3│
       4│ export const Div: StyledComponent<{}, {}, HTMLDivElement> = styled.div``;
       5│
 [2]   6│ export const IgnoreDiv = styled.div``;

 [1] 275│         ) => <StyleProps, Theme>(
     276│             string[],
     277│             ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]
     278│         ) => StyledComponent<StyleProps, Theme, V>


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/index.js:6:26

Missing type annotation for Theme. Theme is a type parameter declared in function
type [1] and was implicitly instantiated at encaps tag [2].

     src/index.js
       3│
       4│ export const Div: StyledComponent<{}, {}, HTMLDivElement> = styled.div``;
       5│
 [2]   6│ export const IgnoreDiv = styled.div``;

 [1] 275│         ) => <StyleProps, Theme>(
     276│             string[],
     277│             ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]
     278│         ) => StyledComponent<StyleProps, Theme, V>
```

> can you reproduce the issue you're concerned with without using $FlowFixMe?

As listed in the error above, `$FlowFixMe` fixes all but 2 of these errors. Now if i were to `export const IgnoreDiv: any = styled.div``;` that fixes the problem all together.

But there is a problem here, in that my I have thousands of styled components which are currently untyped in our repo because I'm on an old Flow version. In order to upgrade effectively as it stands, that means I need to go through and add `any` or correctly define all of the component. I think this is a poor way of upgrading because it forces a lot of manual change. Flow has a [tool to easily add comments](https://github.com/facebook/flow/blob/master/packages/flow-dev-tools/bin/tool) to make upgrading easier, and they do recommend using that to more easily upgrade.

By adding `$FlowFixeMe`, it is much more descriptive (than `any`) that a change needs to be made, and we can do that when we make an actual meaningful change to the file rather than doing this all at once.

`any` is a perfectly reasonable result if I put `$FlowFixMe` on something, but for some reason the `V` generic seems to be getting confused in this translation.

gwing33:
Yeah, my reported bug shouldn't prevent merging. I just wanted to report what I was running into and I appreciate anyone who gave feedback.

omninonsense:
I'll give everything another glance this evening and remove the WIP tag if everything is alright. I'll also comment the failing test for `defaultProps` instead of delete it, just so people are aware of this issue.

omninonsense:
Also, I think this `defaultProps` thing might slightly break integration with [styleguidist](https://react-styleguidist.js.org/) since that relies on `defaultProps` to generate some documentation. But I'll try and sort that out with them on their side, if possible.

omninonsense:
Whew. I'm removing the draft/wip status on this. Hopefully it can get merged. The failing CI pipelines are unrelated to this PR as far as I can tell!

omninonsense:
Hi @vasilich6107!

Sorry for only replying just now. I hope you figured it out in the meantime, but just in case:

~~~ts
// define Theme shape somewhere
type Props = {|
  theme: Theme,
  otherProp: any
|}

const ThemedComp = withTheme<Theme>((props: Props) => {
  const { theme, otherProp } = props
  // ...
})
~~~

goodmind:
This `defaultProps` issue looks off, why AbstractComponent can't support it? Isn't default argument isn't the same thing, it would be recreated with each call of component

omninonsense:
Hi goodmind!

> This `defaultProps` issue looks off, why AbstractComponent can't support it?

Thanks for chipping in. I am not too happy about this `defaultProps` either to be honest, but I don't think there's a straightforward way around it. I linked an issue in the comment above (here's the link again: https://github.com/facebook/flow/issues/7467#issuecomment-463049743).

I don't think Flow team is too keen on fixing this at the moment since they don't even seem to consider it a bug (see https://github.com/facebook/flow/issues/7467#issuecomment-511863362). :(

> Default argument in destructuring isn't the same thing, it would be recreated with each call of component.

This depends on what it is and how it's passed, I guess. If it's expensive to construct, it could be done like this I guess:

~~~tsx
type Props = {person: Person}

const kPerson: Person = { /* … */ }
const Comp = ({person = kPerson}: Props) => <div>{/* … */}</div>
~~~

For primitives it doesn't really matter.

Another thing that could be done, is something something similar to what react does with default props:

~~~tsx
const defaultProps = {
  person: { /* … */ }
}

const Comp = (props: Props) => {
  const { person } = {...defaultProps, ...props}
  // …
}
~~~
---

> Can you post more examples with styled-components

Not sure how to post an example with styled-components, _but_ here is a simplified example on flow.org/try that illustrates the problem:

~~~tsx
/* @flow */

import * as React from 'react'

type Props = {
  a?: number
}

const C: React.AbstractComponent<Props, void> = (c) => <div />

C.defaultProps = {
  a: 10
}

const c = <C />
~~~

[link to the example](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBdkwBDAZzACUBTUgY2KnzizAHJ87H31CBPHDTAAFFjkoBeMAG9UYMgH4AXGAB2AVywAjGvlQBfdPThryxAMKraDQgDoAgtvP5bF1njU01hADxi4CQAaMAA3OAwAEwA+MGkACnoASjjY30iMULBgaPQLO0iaKFINGEIAiTjZeTJVAEYABkNjU3MweirfC2zooA)

---

> I think `attrs` can serve the same purpose too

Yes I think you're right.

goodmind:
@omninonsense would it make sense if `defaultProps` type was `Partial<Props>` in `AbstractComponent`?

omninonsense:
Since this has been "hovering" here for quite a while, and it might've slipped through the cracks, I'm pinging you guys @AndrewSouthpaw, @GAntoine, @villesau! 😁 

Also, I'd like to thank everyone for being so patient and helping with this, especially @edahlseng! ❤️

villesau:
Wow, great job everyone! :)

AndrewSouthpaw:
Amazing work y'all.

PierreCapo:
Hello all!
Everything is working alright and thanks a lot for the PR!
@omninonsense I have this issue when using the new flow definition for styled-components:

```
Cannot call styled(...).attrs(...) because AbstractComponent [1] is not a function.

      20│ const HeaderCell: React$ComponentType<Object> = styled(Text).attrs(() => ({
      21│   numberOfLines: 1
      22│ }))`
      23│   text-align: center;

     react-native-core-library/flow-typed/npm/styled-components_v4.x.x.js
 [1] 382│     +attrs: <A: {}>(A | (StyleProps => A)) => React$AbstractComponent<$Diff<StyleProps, A>, Instance>;
```
Have you faced this problem and do you know how to fix it?

melissafzhang:
Hi! I'm trying to install the new libdef using `flow-typed install styled-components@4.x.x` but getting the error `!! No flow@v0.92.0-compatible libdefs found in flow-typed for the explicitly requested libdefs. !!`. Has this been released yet?

melissafzhang:
Hi! Thanks for working on this! I'm getting some flow errors and would love your help @omninonsense 

```
Cannot call styled(...) because a callable signature is missing in StyledFactory [1].

```

Here's a minimal repro
```
// @flow
import styled from 'styled-components';

type Props = {|
  isWidget?: boolean,
|};

class MyComponent extends React.Component<Props> {
  render() {
    return <StyledNewComponent />;
  }
}

const NewComponent: React.ComponentType<*> = () => {
  return null;
};

type StyleProps = {| ...Props, noMarginBottom?: boolean |};

type Theme = {
  size: { huge: string },
};

const StyledNewComponent = styled<StyleProps, Theme, Props, *>(NewComponent)`
  flex-direction: column;
  align-items: center;
  margin: ${p => (p.isWidget ? `auto` : 0)} 0
    ${p => (p.isWidget && p.noMarginBottom ? `auto` : p.theme.size.huge)} 0;
`;

export default MyComponent;
```

melissafzhang:
@PierreCapo looks like the type for attrs is wrong since it doesn't return a function. I used this type and it seems to fix the error
```
  declare interface StyledFactory<StyleProps, Props, Theme, Instance> {
    [[call]]: TaggedTemplateLiteral<PropsWithTheme<StyleProps, Theme>, StyledComponent<StyleProps, Theme, Instance>>;
    +attrs: <A: {}>(A | (StyleProps) => A) =>StyledFactory<$Diff<StyleProps, A>, Props, Theme, Instance>;
  }
```

melissafzhang:
@omninonsense I'm seeing an issue with nested interpolations in a tagged template literal. 
Flow version: 0.92.0
Styled components version: 4.3.2
###Error:
```Cannot create Content element because InterpolatableComponent [1] is not a React component.

     src/components/BlockingMessage/BlockingMessage.js
      25│ class MyComponent extends React.Component<Props> {
      26│   render() {
      27│     return (
      28│       <Content isWidget={this.props.isWidget}>
      29│         hello
      30│       </Content>
      31│     );
      32│   }
      33│ }

     ../types/flow-typed/npm/styled-components_v4.x.x.js
 [1] 245│   declare export type StyledComponent<Props, Theme, Instance> = React$AbstractComponent<Props, Instance> & InterpolatableComponent<Props>
```

###Code
```
type Props = {
  isWidget?: boolean,
};


class MyComponent extends React.Component<Props> {
  render() {
    return (
      <Content isWidget={this.props.isWidget}>
        hello
      </Content>
    );
  }
}

const Content = styled(Flex).attrs(() => ({
  auto: true,
  column: true,
  align: 'center',
  justify: 'center',
}))`
  max-width: 420px;
  margin: 0 auto;
  padding: ${p => p.theme.size.huge};
  text-align: center;
  ${p =>
    p.isWidget && `* { width: 100%;} & >:last-child { margin-top: auto; }`};
`;
```

omninonsense:
> @PierreCapo looks like the type for attrs is wrong since it doesn't return a function. I used this type and it seems to fix the error
> 
> ```
>   declare interface StyledFactory<StyleProps, Props, Theme, Instance> {
>     [[call]]: TaggedTemplateLiteral<PropsWithTheme<StyleProps, Theme>, StyledComponent<StyleProps, Theme, Instance>>;
>     +attrs: <A: {}>(A | (StyleProps) => A) =>StyledFactory<$Diff<StyleProps, A>, Props, Theme, Instance>;
>   }
> ```

Ah I wasn't aware `attrs` accepted a function like that. Have you made a PR to fix this, or do you want me to make another PR?

omninonsense:
Hi @melissafzhang!

> @omninonsense I'm seeing an issue with nested interpolations in a tagged template literal.
> Flow version: 0.92.0
> Styled components version: 4.3.2
> 
> ### Error:
> ```
> 
>      src/components/BlockingMessage/BlockingMessage.js
>       25│ class MyComponent extends React.Component<Props> {
>       26│   render() {
>       27│     return (
>       28│       <Content isWidget={this.props.isWidget}>
>       29│         hello
>       30│       </Content>
>       31│     );
>       32│   }
>       33│ }
> 
>      ../types/flow-typed/npm/styled-components_v4.x.x.js
>  [1] 245│   declare export type StyledComponent<Props, Theme, Instance> = React$AbstractComponent<Props, Instance> & InterpolatableComponent<Props>
> ```
> 
> ### Code
> ```
> type Props = {
>   isWidget?: boolean,
> };
> 
> 
> class MyComponent extends React.Component<Props> {
>   render() {
>     return (
>       <Content isWidget={this.props.isWidget}>
>         hello
>       </Content>
>     );
>   }
> }
> 
> const Content = styled(Flex).attrs(() => ({
>   auto: true,
>   column: true,
>   align: 'center',
>   justify: 'center',
> }))`
>   max-width: 420px;
>   margin: 0 auto;
>   padding: ${p => p.theme.size.huge};
>   text-align: center;
>   ${p =>
>     p.isWidget && `* { width: 100%;} & >:last-child { margin-top: auto; }`};
> `;
> ```

Not sure what the error is, since it seems like the formatting got a bit broken, but, the problem is likely caused by the wrong return type of this function:

~~~
p => p.isWidget && `* { width: 100%;} & >:last-child { margin-top: auto; }`
~~~

The return type is `boolean | string` (`boolean` when `p.isWidget` is falsy, and `string` when it's truthy). Booleans are not valid interpolation values (AFAIK).

melissafzhang:
@omninonsense that fixed my error. Thanks!

gajus:
```
declare export type Interpolation<P> =
                                      | ((executionContext: P) => Interpolation<any>) // eslint-disable-line flowtype/no-weak-types
                                      | InterpolatableComponent<any> // eslint-disable-line flowtype/no-weak-types
                                      | CSSRules
                                      | KeyFrames
                                      | string
                                      | number
```

Forgot to add `boolean` type.

omninonsense:
> ```
> declare export type Interpolation<P> =
>                                       | ((executionContext: P) => Interpolation<any>) // eslint-disable-line flowtype/no-weak-types
>                                       | InterpolatableComponent<any> // eslint-disable-line flowtype/no-weak-types
>                                       | CSSRules
>                                       | KeyFrames
>                                       | string
>                                       | number
> ```
> 
> Forgot to add `boolean` type.

Hi @gajus! See some of the previous comments in the PR about this (like https://github.com/flow-typed/flow-typed/pull/3228#issuecomment-517666007), but I'll reiterate:

`boolean` is left out intentionally because it's not a valid interpolation. `true` results in `"true;"` (which is invalid CSS), and `false` results in nothing/empty. So, adding `boolean` to the interpolations would result in styled-components emitting erroneous CSS.

_Maybe_ we could only add `false` (and even `null` and `void` since they act the same), but then we're relying on a quirk that's not documented (at least I can't find it being documented). _But_ if we do this, then we deviate from the internal styled-components types implementation ([see here](https://github.com/styled-components/styled-components/blob/560213155847ca3a546a1fbe1151b6aae15ef6a8/packages/styled-components/src/types.js#L6-L10))

lyleunderwood:
Can somebody spell out for me exactly how the `Instance` type parameter of `StyledComponent` should be used (possibly in relation to the `Instance` type parameter on `withTheme`)? I understand that they're basically just being passed along to `React.AbstractComponent`, but I'm too dumb to fully understand what `Instance` is being used for there, either. Take pity on me.

omninonsense:
Hi @lyleunderwood!

> Can somebody spell out for me exactly how the `Instance` type parameter of `StyledComponent` should be used (possibly in relation to the `Instance` type parameter on `withTheme`)? I understand that they're basically just being passed along to `React.AbstractComponent`, but I'm too dumb to fully understand what `Instance` is being used for there, either. Take pity on me.

You can think of "instance" as what the `ref` will be of the component.

~~~jsx
const Input: StyledComponent<Props, Theme, HTMLInputElement> = styled.input``
const ref = React.useRef<?HTMLInputElement>(null)
// ...
return <Input ref={ref} />
~~~

Although, you can do something like this, and flow will infer it automatically, since the libdefs should take care of this:

~~~jsx
const C1: StyledComponent<Props, Theme, _> = styled.input`` // ref is HTMLInputElement
const C2: StyledComponent<Props, Theme, _> = styled(ClassComponent)`` // ref is instance of the class
const C3: StyledComponent<Props, Theme, _> = styled(FunctionalComponent)`` // ref is null
~~~

lyleunderwood:
Thanks.

lyleunderwood:
Actually, @omninonsense, whenever I use `_` in this type parameter position I get `Unexpected reserved type`. :stuck_out_tongue:

lyleunderwood:
Hm, `_` seems to work fine for the `Instance` type parameter for `withTheme`.

Anyway, it seems like the key is to get all my `withTheme`s sorted out first, then I no longer need to stipulate the `StyledComponent`s explicitly.

omninonsense:
> Hm, `_` seems to work fine for the `Instance` type parameter for `withTheme`.
> 
> Anyway, it seems like the key is to get all my `withTheme`s sorted out first, then I no longer need to stipulate the `StyledComponent`s explicitly.

Hmm. You could try using `*` instead of `_` for now. You can also do something like this:

~~~jsx
export type MyStyledComponet<Props, Instance> = StyledComponent<Props, YourThemeShapeHere, Instance>
~~~

 and then use that type instead, if it makes things simpler.

lyleunderwood:
Here are some examples of what I'm doing to get everything working with this libdef. Let me know if I could be doing any of this stuff better. It seems like so far so good.

```javascript
// where `Icon = React$Component<IconProps>` (from a libdef)
export const ButtonIcon =
  styled<IconProps, Theme, IconProps, React.ComponentType<IconProps>>(Icon)`
`;
```

and

```javascript
// where Button is an imported functional component
const BadgedButton: StyledComponent<BadgedButtonProps, Theme, React.ComponentType<typeof Button>> = 
  styled(Button)`
```

I'm not entirely sure if the `typeof` above is correct, because `Button` is already a `React.AbstractComponent`, but flow was complaining about it being a value.

Then my `withTheme` calls all look like this:

```javascript
const Tab = withTheme<Theme, TabProps, _>(
```

goodmind:
@omninonsense looks like this type 
```js
declare export type StyledComponent<Props, Theme, Instance> = React$AbstractComponent<Props, Instance> & InterpolatableComponent<Props>
```

breaks jsx instantiation

goodmind:
@omninonsense 
```js
declare interface StyledFactory<StyleProps, Props, Theme, Instance> {
    [[call]]: TaggedTemplateLiteral<PropsWithTheme<StyleProps, Theme>, StyledComponent<StyleProps, Theme, Instance>>;
    +attrs: <A: {...}>(A | (StyleProps) => A) => React$AbstractComponent<$Diff<StyleProps, A>, Instance>;
  }
```

What `Props` generic does? It isn't used

omninonsense:
> @omninonsense looks like this type
> 
> ```js
> declare export type StyledComponent<Props, Theme, Instance> = React$AbstractComponent<Props, Instance> & InterpolatableComponent<Props>
> ```
> 
> breaks jsx instantiation, because of intersection, if I have complex spreaded type as props

Do you have an example that reproduces this? I wanna find a way of making it work, and adding it to the tests. The reason this exists is because we want to use `InterpolatableComponent` as part of the `Interpolation` union. I guess we can simplify this type by removing `InterpolatableComponent`, and using `React$ComponentType<any>` in the `Interpolation` union, but that's kind of a type-level lie and it means it won't catch some errors.

omninonsense:
> @omninonsense
> 
> ```js
> declare interface StyledFactory<StyleProps, Props, Theme, Instance> {
>     [[call]]: TaggedTemplateLiteral<PropsWithTheme<StyleProps, Theme>, StyledComponent<StyleProps, Theme, Instance>>;
>     +attrs: <A: {...}>(A | (StyleProps) => A) => React$AbstractComponent<$Diff<StyleProps, A>, Instance>;
>   }
> ```
> 
> What `Props` generic does? It isn't used

Ah really good catch! It does nothing, Lol 🤣 It's just left-overs from some simplification/refactoring.

I'll make a PR that removes this and fixes the `attrs` thing to accept a function argument, too.

goodmind:
@omninonsense check my `.attrs` PR

PR #3229, opened 2019-03-23
[koa_v2]: Add custom context

HsuTing:
- Links to documentation:  https://github.com/koajs/koa/tree/master/docs/api
- Link to GitHub or NPM: https://github.com/koajs/koa

Other notes:
Like #3220


HsuTing:
@pascalduez Owing to #3221 , koa_v2 has the new version for flow >= 94.
I made the same modification, but this do not work.
I have no idea about this error. In #3221 , this just modify the server type.

pascalduez:
@HsuTing Because `{}` is too loose. So `ctx.body` is not typed and accept anything. 
So the fail assertion in the tests are no longer true.

HsuTing:
Yes, I found that, and I found the [testing](https://travis-ci.org/flow-typed/flow-typed/jobs/510329025#L639-L666) in #3220  did not run `koa_v2.0.x/flow_v0.93.x-flow_v0.93.x` and `koa_v2.x.x/flow_v0.93.x-flow_v0.93.x`. So I did not notice this did not work.

HsuTing:
After I tried, I think if the generic parameter is not passed, it will give `any` type.
https://flow.org/try/#0CYUwxgNghgTiAEkoGdnxANxAOwC7IBIBRLPIgWwEtdcQZ4BvAXwCgXcBPABwQGEB7PCAAeueAF5GLePABG-YBwBcM5LhiVsAc3gAfeACkAygHkAcidkArcGP3HzAQRgwoHPfGwBXCBAA0LEwA3GycPIamZhIeahraHt7ksnQe8vwQIFDYCT4QHg4W1rb5kc6uHCFhCAWWNmBikgzwANqxmloAuioF8MHs3NWlLm7RZW4APAUAfCH94QCylMDAGQDusCDjvNHMUxLSMvAAFPXCKgJCovAAZPC8fp4iuCpHAJQSewAKMPxUyJsYfhLKbvcRfH5-AFA4BTXSApazUBIOCIaCoeCOLhcCCUMBQXCUQToUQ4YBoTA4fDEUi4CjUWj0BgHLz-LYqXZHABm2BUi2Waw2WxBKlwAAtKMgAqwZCwMmIoFjotgQKsMVicXiCYI3iEZQquAA6FkgI4nXDCUF7JkyU4G+SKaIARl1iHNdoU7kkAHJaGovSEmK8gkA

pascalduez:
Yeah all those defs are confusing, what's the difference between `2.0.x` and `2.x.x`? 
Also first time I see something like `flow_v0.93.x-flow_v0.93.x`, should probably be only `flow_v0.93.x` but well.

HsuTing:
I do not know why here has `2.0.x` and `2.x.x`. Before I modified, those folders had existed here.

HsuTing:
[Here](https://github.com/flow-typed/flow-typed/pull/874) is the reason why we have `koa_v2.x.x` and `koa_v2.0.x`.

pascalduez:
Thinking this is not the right solution, and a breaking change for current users. If they don't pass the parameter, they will loose the current `Context` typing.

pascalduez:

PR #3230, opened 2019-03-24
Changed `and` and `or` to reflect actual behaviour

lensbart:
`and` and `or` don’t necessarily return `true` or `false`, instead they return their first or second argument. (The docs are a bit unclear here as well, but the correct behaviour is explained when you click “expand parameters”.)

- Links to documentation: [and](https://ramdajs.com/docs/#and), [or](https://ramdajs.com/docs/#or)
- Link to GitHub or NPM: [and](https://github.com/ramda/ramda/blob/v0.26.1/source/and.js), [or](https://github.com/ramda/ramda/blob/v0.26.1/source/or.js)
- Type of contribution: fix

Other notes: thanks for maintaining this repository

### DIFF
--- a/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/ramda_v0.26.x.js
+++ b/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/ramda_v0.26.x.js
@@ -1881,7 +1881,7 @@ declare module ramda {
     fns: Array<(...args: Array<T>) => boolean>
   ): (...args: Array<T>) => boolean;
 
-  declare function and<X>(x: X): <Y>(y: Y) => X | Y;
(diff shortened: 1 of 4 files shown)


pascalduez:
Hi, thanks for your contribution! 
Would be cool to straighten a bit the tests fo those methods then, not sure they reflect this changes properly.

lensbart:
Consider it done, monsieur Duez!

pascalduez:
I'll let @AndrewSouthpaw have the final look, I'm not familiar with Ramda.   
And the Prettier changes are making the review a bit though.

lensbart:
Yes, my apologies for that — but figured since the project has a `.prettierrc` file I could just as well apply those rules? (Happened by default on save)

PR #3236, opened 2019-03-26
[@storybook/react v5.x.x] Use React$Node instead of a bespoke type

kejistan:
Cleaning this up since the bespoke `Renderable` type isn't accurate (it can be any react element, e.g. a string, or a number would be valid). Also gets rid of the existential type, so thats nice.

- Links to documentation: https://storybook.js.org/docs/basics/introduction/
- Link to GitHub or NPM: https://www.npmjs.com/package/@storybook/react
- Type of contribution: fix

Other notes:

### DIFF
--- a/definitions/npm/@storybook/react_v5.x.x/flow_v0.72.x-/react_v5.x.x.js
+++ b/definitions/npm/@storybook/react_v5.x.x/flow_v0.72.x-/react_v5.x.x.js
@@ -9,7 +9,7 @@ declare module '@storybook/react' {
     | Iterable<?Renderable>;
   declare type RenderCallback = (
     context: Context
-  ) => React$Node;
+  ) => Renderable;
   declare type RenderFunction = () => Renderable;
 
   declare type StoryDecorator = (
--- a/definitions/npm/@storybook/react_v5.x.x/flow_v0.72.x-/react_v5.x.x.js
+++ b/definitions/npm/@storybook/react_v5.x.x/flow_v0.72.x-/react_v5.x.x.js
@@ -2,16 +2,15 @@ type NodeModule = typeof module;
 
 declare module '@storybook/react' {
   declare type Context = { kind: string, story: string };
-  declare type Renderable = React$Element<*>;
   declare type RenderCallback = (
     context: Context
-  ) => Renderable | Array<Renderable>;
-  declare type RenderFunction = () => Renderable | Array<Renderable>;
+  ) => React$Node;
+  declare type RenderFunction = () => React$Node;
 
   declare type StoryDecorator = (
     story: RenderFunction,
     context: Context
-  ) => Renderable | null;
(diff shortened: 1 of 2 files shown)


pascalduez:
Hi, 
I wrote those types, and I spent quite some time testing with real code and reading the storybook sources. So we will need strong proofs (working code) to consider this change. 
In my testing it only properly accept elements.  
Also the [TS definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/storybook__react/index.d.ts) are not using `React$Node`.

kejistan:
I have some stories that are returning strings, which work, but aren't `React.Element<*>`. Why would storybook depend on these being react elements rather than nodes?

kejistan:
> Also the [TS definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/storybook__react/index.d.ts) are not using `React$Node`.

The TS defintions are using `React.ComponentType | JSX.Element`, which is not the same as `React.Element<*>`. `JSX.Element` is `React.Node`, `React.ComponentType` is the same in TS and Flow. Translated to Flow this would literally be: `React$ComponentType<{}> | React$Node`

pascalduez:
Let's forget the TS thing.   
In my testing (real code), even a string wasn't working properly, it was logging warnings or errors in the console or something.

kejistan:
I'd prefer not to put this on hold over this. In my own testing none of the `React.Node` types cause errors or warnings:

```js
import {storiesOf} from "@storybook/react";

const stories = storiesOf("Test", module);

stories.add("String", () => "test string");
stories.add("Number", () => 99);
stories.add("Boolean", () => true);
stories.add("Null", () => null);
stories.add("Array", () => ["string1", "string2"]);
```

pascalduez:
https://github.com/pascalduez/storybook-flow-test

`null` crashes  
`boolean` crashes   
`Portal` crashes

kejistan:
Dug into this more. Looks like storybook is [doing some sanity checking](https://github.com/storybooks/storybook/blob/a27a523c7a977039f01d39cdededa1662bdf1ab8/app/react/src/client/preview/element_check.js#L15) on the final result of rendering the story (which also includes any decorators on the story, which was my problem). In spirit it seems pretty clear that they're trying to enforce that stories return a `React.Node`, but in practice the checks don't match the types exactly. How does converting this to `React$Element<any> | string | number | $ReadOnlyArray<React$Element<any> | string | number>` sound? It'd match the checks that storybook is doing, and it'd let people include `string` or `number` elements as well.

pascalduez:
> and it'd let people include string or number elements as well.

Yes, apparently we need to allow those. 

Maybe we can do something similar to the `React$Node` type:

```
  declare type Renderable =
    | string
    | number
    | React$Element<any>
    | Iterable<?Renderable>;
```

PR #3251, opened 2019-04-03
[pell_v1.x.x] Add definitions

ChillyBwoy:
See https://github.com/jaredreich/pell


definitions/npm/pell_v1.x.x/flow_v0.69.x-/pell_v1.x.x.js
```diff
@@ -0,0 +1,103 @@
+declare module 'pell' {
+  declare type pell$defaultActions =
```
pascalduez: You are scoped inside the module declaration, no need to prefix with `peel$`, `DefaultActions` will be fine.

PR #3256, opened 2019-04-07
[ramda_0.26.x] Refactor pipe()

retyui:

Other notes:
Now fist function can get unlimited arguments

[FLOW.JS TRY >](https://flow.org/try/#0FDAmFMGMBsEMCdwAIAuBPADsgqgOwWgGK4A8AggDRIBKAfEgLxIAUsAXEmQJSP3UDcIdFiQBZWAGtwABQCWWYoyTl4AcwDOFauBQBXeLlrNgSUywB0lhBo5k16ng3qxcaKibN4CxEi7c0dfUNgR3pmZktza3Vbe1CAvQMuQRAIGARkADcEJGkASWkAUQ4Abw9TEgAhKgBhKgARKkKqQhaAOQBGDgirDgAqeOqkQjaAVg4veCJSVqQ+2iNys1NYACMOEY73ZeXVyAn8KZ8hmtptndNIUAPvUjqkerOkJeWIG6PSRqRCp5ezcAAZu9piRmsNfhdTADVBsxhQllwOAASGqwaDQEjiKRyBS4dpbYZjM7AAC+IDAUDgiCQ2XgSABuC6LHYSFwugAtqtwPAKOskOoUPBZLhVBR9khVgB7SXQcAuCjXJAAbRKAB9UOABRwpTK5bgkKqSQBdeJszncwRpKlZHIMgBM3VgTLNXPg8QFQpFlspGRpttwAGZHUyPcLVPEdbKXN70tTafTcAAWYPa6VR3DxJWhkVGmPWv10hnjZlMrOCsMm3jKl3c3MU2M2wu4ABsKerHNdlacysjerrwEgktwAo1ApxyCY+SKxmWDK2S3t8NngaXZgZidXUNwo03Ceb8OSIGYKE1KHHzC2AHI7ZeKILdOAqCqT1r6Wj1OBjYjTErey4TSkx6nueV43ne8APk+JQvigHAAu+n4mhwEppnqh4wSBFDXre96Psq0GnnBCFUEgkoYOoMRIB0X78EAA)

Problems:
- When announcing an increasing number of arguments, the amount of code increases exponentially as it requires a single generic type per argument to the function

- Sometimes you need to specify the type for the function since the flow cannot decide which type to use when intersecting the number of types
```js
<A, B, C, D, E> ...
<A0, A1, A2, B> ...
```

image

### DIFF
--- a/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/ramda_v0.26.x.js
+++ b/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/ramda_v0.26.x.js
@@ -311,49 +311,50 @@ declare module ramda {
       args: [A, B, C, D, E, F, G, H, I]
     ) => () => R);
 
+  declare type $MakePipeFn = <Args,Return>(
+    (...args: Args) => any,
+    UnaryFn<any, Return>
+  ) => ((...args: Args) => Return);
+
+  declare type $PipeFn<FirstFn, LastFn> = $Call<$MakePipeFn, FirstFn, LastFn>;
+
   declare var pipe: {
-    <A, B, C, D, E, F, G>(ab: UnaryFn<A, B>, bc: UnaryFn<B, C>, cd: UnaryFn<C, D>, de: UnaryFn<D, E>, ef: UnaryFn<E, F>, fg: UnaryFn<F, G>): UnaryFn<A, G>,
-    <A0, A1, B, C, D, E, F, G>(ab: (A0, A1) => B, bc: UnaryFn<B, C>, cd: UnaryFn<C, D>, de: UnaryFn<D, E>, ef: UnaryFn<E, F>, fg: UnaryFn<F, G>): (A0, A1) => G,
-    <A0, A1, A2, B, C, D, E, F, G>(ab: (A0, A1, A2) => B, bc: UnaryFn<B, C>, cd: UnaryFn<C, D>, de: UnaryFn<D, E>, ef: UnaryFn<E, F>, fg: UnaryFn<F, G>): (A0, A1, A2) => G,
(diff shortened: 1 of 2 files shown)


retyui:
@LoganBarnett Could you have a look?

LoganBarnett:
@retyui I recall running into similar issues with `partial` and so I put together a [code generator](https://github.com/LoganBarnett/typedef-gen/blob/master/index.org) that relieves me of the nasty business of maintaining such error-prone and repetitive type declarations.

Without some way of saying "give me the arguments that this function takes", you either have to spell out every argument type in order to preserve them, or resort to `any` like you do here.

Using your example:
``` javascript
(testPipe(1,'2',2, [{test: false}]):  [boolean]); // passes, but should fail on the 3rd arg
```

I suspect that's why your exact-type 4th arg example passes the type check, even though the 4th arg doesn't match the type.

I could look at making the generator include `pipe` and submit a PR for what it emits. Would that work for you? I imagine it would take me no more than a day or two (I'm in the middle of a move, so there's some padding there).

I like the recursive type approach with `$Call` - I don't think I've seen that before!

retyui:
@LoganBarnett I recheck your example

```js
(testPipe(1,'2',2, [{test: false}]):  [boolean]); 
```
>  passes, but should fail on the 3rd arg

And I see error:

image

LoganBarnett:
Ah my mistake. I tried it in a local test bed using an older Flow - I do this to catch coverage loss that the Try page doesn't do very well.

Everything works as expected, and it even catches the inexact argument too!

I noticed you removed the 1-function `pipe` calls in the tests. Just thinking out loud here: Those are nonsensical uses of `pipe` since such an operation really should return the original function (what else would you be piping to?). Since that's the case I think it makes sense to make it into a type error.

This PR looks good to me!

PR #3262, opened 2019-04-10
[apollo-client] Fix issue with query/mutate functions and passing variables

### LINKED ISSUE #2888: [react-apollo] false negative when using client with variables
When working with `ApolloClient` programmatically to query with variables, flow throws an error despite being the documented data structure.

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/components/XXXXXXX.js:66:18

Cannot call client.query with object literal bound to options because an indexer property is missing in object
literal [1] but exists in object type [2] in property variables.

     src/components/XXXXXXX.js
      63│     const response = await client.query({
      64│       query: FETCH_USER,
      65│       fetchPolicy: 'no-cache',
 [1]  66│       variables: {
      67│         userId: user.id,
      68│         type: USER_TYPE.ADMIN,
      69│       },
      70│     });
      71│   };

     flow-typed/npm/react-apollo_v2.x.x.js
 [2] 262│     variables?: { [key: string]: any };
```

ganemone:
- Links to documentation: https://www.apollographql.com/docs/react/api/apollo-client#ApolloClient.query
- Link to GitHub or NPM: https://github.com/apollographql/apollo-client/
- Type of contribution: fix

This fixes an issue where variables could not be passed to query/mutate functions and adds a regression test.

Fixes #2888

### DIFF
--- a/definitions/npm/apollo-client_v2.x.x/flow_v0.57.x-/apollo-client_v2.x.x.js
+++ b/definitions/npm/apollo-client_v2.x.x/flow_v0.57.x-/apollo-client_v2.x.x.js
@@ -270,7 +270,7 @@ declare module "apollo-client" {
   };
 
   declare interface ModifiableWatchQueryOptions {
-    variables?: { [key: string]: any };
+    variables?: any;
     pollInterval?: number;
     fetchPolicy?: FetchPolicy;
     errorPolicy?: ErrorPolicy;
@@ -310,7 +310,7 @@ declare module "apollo-client" {
 
   declare export interface SubscriptionOptions {
     query: DocumentNode;
-    variables?: { [key: string]: any };
+    variables?: any;
   }
 
   declare export type FetchPolicy =
@@ -368,7 +368,7 @@ declare module "apollo-client" {
 
   declare export type PureQueryOptions = {
     query: DocumentNode,
-    variables?: { [key: string]: any }
+    variables?: any,
   };
 
   declare export type ApolloQueryResult<T> = {
@@ -495,7 +495,7 @@ declare module "apollo-client" {
 
   declare interface GraphQLRequest {
     query: DocumentNode;
-    variables?: { [key: string]: any };
+    variables?: any;
     operationName?: string;
     context?: { [key: string]: any };
     extensions?: { [key: string]: any };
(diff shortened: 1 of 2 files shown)


indeyets:
@ganemone add `[apollo-client]` to the title please

ganemone:
> We generally avoid using `any` where possible, and this fix loses type information. E.g. we do know the `variables` option would be an object.
> 
> Can you work toward a solution that addresses the problem without dropping knowledge of it being an object?

That is fair. There is a solution which would let us not use `any` however it would require significant changes that would be breaking for many users as it would moving many of the exports to be type aliases rather than interfaces. I'm happy to do that but I thought I would introduce the less intrusive option first.

AndrewSouthpaw:
Sure, that's understandable. Generally we've suggested moving away from type interfaces when you can get away with it, because they've often triggered more slowness in flow.

Do you suspect that many people have actually been using that exported type as an interface? How bad of a breaking change would it be?

ganemone:
Now that I think about it, it might not be that big of a breaking change since it was broken when passing variables in the first place. And it should be non breaking unless they are using features like `extends`.

I updated the diff to move the affected exports from interfaces to types and removed usages of `any`.

definitions/npm/apollo-client_v2.x.x/flow_v0.57.x-/apollo-client_v2.x.x.js
```diff
@@ -495,7 +494,7 @@ declare module "apollo-client" {
 
   declare interface GraphQLRequest {
     query: DocumentNode;
-    variables?: { [key: string]: any };
+    variables?: any;
```
AndrewSouthpaw: Change this line back?

definitions/npm/apollo-client_v2.x.x/flow_v0.57.x-/apollo-client_v2.x.x.js
```diff
@@ -278,8 +278,7 @@ declare module "apollo-client" {
     notifyOnNetworkStatusChange?: boolean;
   }
 
-  declare export interface WatchQueryOptions
-    extends ModifiableWatchQueryOptions {
+  declare export type WatchQueryOptions = ModifiableWatchQueryOptions & {
```
AndrewSouthpaw: Generally I subtype by doing:

```js
type Subtype = {
  ...$Exact<Supertype>,
  foo: string
}
```

There are times when doing an intersection creates weird outcomes.

ganemone:
Any further comments?

AndrewSouthpaw:
Thanks!

PR #3264, opened 2019-04-17
[ramda] improve _.pick definition

moroine:
Links to documentation: https://ramdajs.com/docs/
Link to GitHub or NPM: https://github.com/ramda/ramda
Type of contribution: fix

This is not perfect yet as following code won't throw error:

```javascript
const pck = _.pick(["a", "d"], { a: 1, b: 2, c: 3, d: '4' });

// should not work
const b = pck.b;
```

But at least it will throw an error when the key didn't exist in original object


moroine:
@AndrewSouthpaw could you have a look at it?

PR #3267, opened 2019-04-19
Export types from winston libdef

dmitry-zakablukov:
- Links:
https://github.com/winstonjs/logform
https://github.com/winstonjs/winston
- Type of contribution: addition and refactor

Changes:
- All winston types are exported from libdef to make them available in external code
- Callable signature added for winston.format()
- has() method added for winston.Container

### DIFF
--- a/definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
+++ b/definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
@@ -1,93 +1,93 @@
-declare type $winstonLevels = {
-  [string]: number
-};
+declare module 'winston' {
+  declare export type $winstonLevels = {
+    [string]: number
+  };
 
-declare type $winstonNpmLogLevels = {
-  error: number,
-  warn: number,
-  info: number,
-  verbose: number,
-  debug: number,
-  silly: number
-};
+  declare export type $winstonNpmLogLevels = {
+    error: number,
+    warn: number,
+    info: number,
+    verbose: number,
+    debug: number,
+    silly: number
+  };
 
-declare type $winstonInfo<T: $winstonLevels> = {
-  [optionName: string]: any,
-  level: $Keys<T>,
-  message: string
-};
+  declare export type $winstonInfo<T: $winstonLevels> = {
+    [optionName: string]: any,
+    level: $Keys<T>,
+    message: string
+  };
 
-declare type $winstonFormat = Object;
+  declare export type $winstonFormat = Object;
 
-declare type $winstonFileTransportConfig<T: $winstonLevels> = {
-  filename: string,
-  level?: $Keys<T>
-};
+  declare export type $winstonFileTransportConfig<T: $winstonLevels> = {
+    filename: string,
+    level?: $Keys<T>
+  };
 
-declare class $winstonTransport {
-  level?: string;
-  silent?: boolean;
-}
+  declare export class $winstonTransport {
+    level?: string;
+    silent?: boolean;
+  }
 
-declare class $winstonFileTransport<T> extends $winstonTransport {
-  constructor($winstonFileTransportConfig<T>): $winstonFileTransport<T>;
-}
+  declare export class $winstonFileTransport<T> extends $winstonTransport {
+    constructor($winstonFileTransportConfig<T>): $winstonFileTransport<T>;
+  }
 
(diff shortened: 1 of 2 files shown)


definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
```diff
@@ -1,100 +1,100 @@
...
+  declare export class $winstonContainer<T> {
+    constructor(config?: $winstonLoggerConfig<T>): $winstonContainer<T>;
+    add(loggerId: string, config?: $winstonLoggerConfig<T>): $winstonLogger<T>;
+    get(loggerId: string): $winstonLogger<T>;
+    has(loggerId: string): boolean;
+  }
 
-declare module "winston" {
-  declare module.exports: $winstonDefaultLogger & {
+  declare export default $winstonDefaultLogger & {
```
pascalduez: The problem is that this will break badly for any commonjs user. 
Also the lib is written using commonjs.

dmitry-zakablukov: Is there any way to export types in Common JS module?

pascalduez: Did some quick testing and it's fine to use either `declare type` or `declare export type` in a commonjs module declaration, but that works only with a `type`.

So for classes one could use aliases:
```
declare class $WinstonTransport {}
declare export type WinstonTransport = typeof $WinstonTransport;
```
Unless there's a better solution.

dmitry-zakablukov: I made Common JS module with type exports as you suggested.

definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
```diff
@@ -65,6 +65,7 @@ declare type $winstonConfigSubModule = {
 };
 
 declare type $winstonFormatSubModule = {
+  ((info: Object) => Object): () => $winstonFormat,
```
pascalduez: `Object` is `any` from Flow 0.75 on. No way to type it more precisely? We often use `{}` or `{ +[string]: mixed }` depending on the context.

definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
```diff
@@ -84,9 +85,26 @@ declare class $winstonContainer<T> {
...
+  declare export type $WinstonTransport = typeof $winstonTransport;
+  declare export type $WinstonFileTransport<T: $WinstonLevels> = $winstonFileTransport<T>;
+  declare export type $WinstonConsoleTransportConfig<T: $WinstonLevels> = $winstonConsoleTransportConfig<T>;
+  declare export type $WinstonConsoleTransport<T: $WinstonLevels> = $winstonConsoleTransport<T>;
+  declare export type $WinstonLoggerConfig<T: $WinstonLevels> = $winstonLoggerConfig<T>;
+  declare export type $WinstonLogger<T: $WinstonLevels> = $winstonLogger<T>;
+  declare export type $WinstonConfigSubModule = $winstonConfigSubModule;
+  declare export type $WinstonFormatSubModule = $winstonFormatSubModule;
+  declare export type $WinstonDefaultLogger = $winstonDefaultLogger;
+  declare export type $WinstonContainer<T: $WinstonLevels> = $winstonContainer<T>;
```
pascalduez: The `$winston` prefix was used because those types are global. 
In a module scope this is not necessary.  
I would rather remove them personally.  `Levels`, `NpmLogLevel`, `Format`, etc.

definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
dmitry-zakablukov: In other places of this file an `Object` type is used, so I used it too to keep the module style.

definitions/npm/winston_v3.x.x/flow_v0.34.x-/winston_v3.x.x.js
dmitry-zakablukov: Prefix removed.

Issue #3237, opened 2019-03-28
[sequelize] should use commonjs export format, not ES Module


zeorin:
Sequelize uses commonjs export format.
The libdef here uses the ES Module format.

Importing as follows:
```
const Sequelize = require('sequelize');
```
Doesn't work in Flow (though the code runs just fine), you have to import as:
```
const Sequelize = require('sequelize').default;
```
Luckily, Sequelize manually added a `default` prop to their `module.exports`, so the 2nd version works.

However, it'd be better if the types were congruent with the actual export format.

retyui:
@zeorin 

We can't mix two syntaxes `ES`  and `commonjs` modules
Flow.js imposes [restrictions](https://flow.org/en/docs/libdefs/creation/#toc-declaring-a-commonjs-module)

> CommonJS modules have a **single value** that is exported (the `module.exports` value).

zeorin:
I'm not suggesting that we mix the export formats. Just that we pick the one that the actual package uses, so that there isn't a mismatch, that still allows one to export everything the actual JS package exports, the same way.

Based on my reading of other libdefs, choosing to use the CommonJS export format does not preclude using `declare export type`.

Clearly, trying to use Sequelize as per their docs:
```
const Sequelize = require('sequelize');
```
and not having this work for Flow despite having the libdef installed, is not what we want.

> CommonJS modules have a single value that is exported (the module.exports value).

Sequelize _uses a CommonJS_ export, therefore it exports a single value. Surely we can do the same? I don't see a problem.

retyui:
> Sequelize uses a CommonJS export, therefore it exports a single value. Surely we can do the same? I don't see a problem.

Now we export a lot of types:
https://github.com/flow-typed/flow-typed/blob/890709941d934f31381dd646747fd1f6d2200fa0/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/test_sequelize.js#L4-L35

With commonjs modules it is impossible to do:


![image](https://user-images.githubusercontent.com/4661784/56644661-cb0f3b00-6684-11e9-899b-68d27dcb70f4.png)

retyui:
Fixed in: https://github.com/flow-typed/flow-typed/pull/3276

retyui:
I was mistaken and eventually found a solution

jedwards1211:
sorry, this was not a solution, it broke working code that imports via ES6

jedwards1211:
[...]

PR #3276, opened 2019-04-26
[sequelize_v4] fix: Convert module from ESM to CommonJS

retyui:
- Links to documentation: https://github.com/sequelize/sequelize/tree/v4/docs
- Link to GitHub or NPM: https://github.com/sequelize/sequelize
- Type of contribution: fix | refactor

Other notes:
via: https://github.com/flow-typed/flow-typed/issues/3237

### DIFF
--- a/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/test_sequelize.js
+++ b/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/test_sequelize.js
@@ -1,9 +1,9 @@
 // @flow
+const Sequelize = require("sequelize");
+const { Model, Transaction } = Sequelize;
 
-import Sequelize, {Model, Transaction} from "sequelize";
-import type {fn} from 'sequelize'
-import type { DefineAttributes } from "sequelize";
 import type {
+  DefineAttributes,
   HasOneGetOne,
   HasOneSetOne,
   HasOneCreateOne,
--- a/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/sequelize_v4.x.x.js
+++ b/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/sequelize_v4.x.x.js
@@ -1379,7 +1379,7 @@ declare module "sequelize" {
     field: string,
   }
 
-  declare export class Association<Source: Model<any>, Target: Model<any>> {
+  declare class $Association<Source: Model<any>, Target: Model<any>> {
     constructor(source: Class<Source>, target: Class<Target>, options?: AssociationOptions): this;
     static BelongsTo: typeof BelongsTo;
     static HasOne: typeof HasOne;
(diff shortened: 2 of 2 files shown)


jedwards1211:
@pascalduez @GAntoine hey guys, maybe these defs work better for people using CommonJS, but they cause a bunch of new errors in working code for people using ES6 imports.  Really aggravating.  Can we revert this and try again to support CommonJS without breaking ES6 imports?

definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/test_sequelize.js
```diff
@@ -1,9 +1,9 @@
 // @flow
+const Sequelize = require("sequelize");
+const { Model, Transaction } = Sequelize;
 
-import Sequelize, {Model, Transaction} from "sequelize";
```
jedwards1211: @pascalduez deleting this line is unacceptable and it should have never passed review.

jedwards1211:
@pascalduez I'd like to take on the review responsibilities for the sequelize type defs.  I'm the one who put in tons of hours to create them in the first place and keep them updated.

retyui:
Now I managed to find out that if I use the declaration object

We can work with CommonJS and ES modules `sequelizeObject`

But if we use the class then the behavior changes 

```js
// tmpSeq_vx.x.x.js
declare module 'sequelizeObject' {
  declare module.exports: {
    Model1: number,
  };
}

declare module 'sequelizeClass' {
  declare class S {
    Model2: number;
  }

  declare module.exports: typeof S;
}

declare module 'sequelizeClass3' {
  declare class S {
    Model2: number;
  }

  declare module.exports: S;
}
//----------------------------------------------------------------------------
// test_tmpSeq.js 
import { Model1 } from 'sequelizeObject'; // Work
import S2, { 
Model2  // error "Cannot import `Model2` because there is no `Model2` export in `sequelizeClass`"
} from 'sequelizeClass'; 
import S3, { 
Model3  // work
} from 'sequelizeClass3'; 
new S2();
new S3(); // error "Cannot call `S` because `S` [1] is not a function."
```

jedwards1211:
I'm not 100% sure what you're trying to say but doesn't that example show that none of the three ways can fully support both CommonJS and ES modules?

definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/test_sequelize.js
pascalduez: Agree, we should test both commonjs and ES modules when doing those kind of conversions.

pascalduez:
> @pascalduez I'd like to take on the review responsibilities for the sequelize type defs. I'm the one who put in tons of hours to create them in the first place and keep them updated.

I'm all for it, a bit of ownership will be beneficial.

pascalduez:
> I'm assuming there isn't, which is why I don't think it's possible to support both CommonJS and ES modules

Looks like the same issue I had wit the `verror` libdef, Flow can't properly model all those cases.

jedwards1211:
I'll watch this repo and set an email rule to filter out most other packages (there are a few others I should watch)

jedwards1211:
or do you know any kind of bot I can set up to only watch changes to certain type defs?  Unfortunately there's no way to make a 100% perfect email filter rule...

pascalduez:
If I'm not mistaken, TS (definitely typed) has a bot for assigning reviewers based on contributions. 
Also we should maybe require at least two approvals per PR... Might considerably bump merging time though.

GAntoine:
@jedwards1211 You can add a CONTRIBUTING.md file to the `sequelize` folder, and mark yourself as the person to ping when there are changes. It's also a good place to detail best practices and pitfalls for that specific PR.

@pascalduez I'll add a [CODEOWNERS](https://help.github.com/en/articles/about-code-owners) file sometime this week, which will autorequest a review from specific people in certain cases.

PR #3223, opened 2019-03-22
[redux_v3-4] Fix errors

retyui:
- Links to documentation: https://redux.js.org/
- Link to GitHub or NPM: https://github.com/reduxjs/redux
- Type of contribution: fix

---

### Other notes:
When we run `node cli/dist/cli.js run-tests redux_v4` a command
You will see more error about:
- deprecated `$Subtype` 
- Object is any

PR #3288, opened 2019-04-30
Annotate redux v3 for flow past 0.89

wgao19:
- Links to documentation: https://redux.js.org/
- Link to GitHub or NPM: https://github.com/reduxjs/redux

---
Other notes:

- Adding since `$Subtype` is deprecated after 0.89
- Rest annotations follow https://github.com/flow-typed/flow-typed/pull/3223
- This is my first contribution to this repo. I'm not sure if I've done everything correctly: 
  My `$ ./build_and_test_cli.sh` failed on a few libraries that are not related. But I expect things should work since most of my definitions follow existing ones 😅

### DIFF
--- a/definitions/npm/redux_v3.x.x/flow_v0.89.x-/redux_v3.x.x.js
+++ b/definitions/npm/redux_v3.x.x/flow_v0.89.x-/redux_v3.x.x.js
@@ -1,0 +1,93 @@
+declare module 'redux' {
+  /*
+
+    S = State
+    A = Action
+    D = Dispatch
+
+  */
+
+  declare export type DispatchAPI<A> = (action: A) => A;
+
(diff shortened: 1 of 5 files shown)


drwlrsn:
Thanks @wgao19 this is really appreciated.

pascalduez:
Thanks @wgao19 !

Some notes for later hopefully ;)

- It's better to split into commit when upgrading a Flow version like this, one for the copying, one for the changes, it makes reviewing the changes easier and safer.

- Tests should be `wrapped in `describe, it` blocks](https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#use-describe-and-it-blocks-to-limit-scope).

wgao19:
Thank you @pascalduez! Will take note next time :]

PR #3290, opened 2019-04-30
[jest_v24.x.x] Update JestMatcher definition

mdesrosiers:
### Description
Update the Flow type for Jest's Matcher object. The current definition is a function accepting two argument but it should take an arbitrary number of parameters which are passed to the matcher.

- Links to documentation: https://jestjs.io/docs/en/expect#custom-matchers-api
- Link to GitHub or NPM: https://github.com/facebook/jest/blob/master/packages/expect/src/jestMatchersObject.ts#L84
- Type of contribution: fix

### Other notes
- Renamed parameters to received and actual
- Changed actual to a rest parameter

### DIFF
--- a/definitions/npm/jest_v24.x.x/flow_v0.39.x-/jest_v24.x.x.js
+++ b/definitions/npm/jest_v24.x.x/flow_v0.39.x-/jest_v24.x.x.js
@@ -124,8 +124,8 @@ type JestMatcherResult = {
 };
 
 type JestMatcher = (
-  actual: any,
-  expected: any
+  received: any,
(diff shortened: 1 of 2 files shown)


pascalduez:
Hi @mdesrosiers,

could you please point us to documentation, code, any material asserting those changes?
Also a brief explanation might help as well.

Note: Tests already failing on master for lower than 24.x versions.

mdesrosiers:
Thanks @pascalduez , I updated the description. For the tests, I'm not seeing any failures in the CI build, can you point me to where it fails?

PR #3300, opened 2019-05-03
[ramda_v0.26.x] Refactor pipe() function

retyui:

Motivation:
There is no possibility to specify the type for incoming arguments and return value:

This RP add this feature:
```js
type State = {
  navigation: {
    componentId: string,
  },
};
// typeof getComponentId === `State => string`
export const getComponentId = pipe<[State], string, _, _, _>(
  prop('navigation'),
  prop('componentId'),
);

```


retyui:
Blocked by https://github.com/flow-typed/flow-typed/pull/3286

retyui:
@LoganBarnett ^

LoganBarnett:
I'm not sure if Github's interface is being strange or I'm doing something wrong. I see in one case you have a test, but if I ask it to show me the aggregate of all the changes then I don't see the changes to the test suite anymore.

If the test changes were removed, could you comment as to why that was? If possible it would be great to capture something to help enforce these changes. Other than that, this looks good!

retyui:
There is no test changes

PR #3305, opened 2019-05-06
useCacheUntil option in install command 

sambathl:
useCacheUntil option to install command, since it is 1 minute by hard coded which isn't good

### DIFF
--- a/cli/src/commands/install.js
+++ b/cli/src/commands/install.js
@@ -31,6 +31,7 @@ import {
 import {
   getCacheRepoDir,
   _setCustomCacheDir as setCustomCacheDir,
+  CACHE_REPO_EXPIRY,
 } from '../lib/cacheRepoUtils';
 
 import {getRangeLowerBound} from '../lib/semver';
@@ -55,7 +56,8 @@ export type Args = {
   cacheDir?: mixed, // string
   packageDir?: mixed, // string
   ignoreDeps?: mixed, // Array<string>
-  rootDir?: mixed, // string
+  rootDir?: mixed, // string,
+  useCacheUntil?: mixed, // seconds
 };
 export function setup(yargs: Yargs) {
   return yargs.usage(`$0 ${name} - ${description}`).options({
@@ -104,6 +106,11 @@ export function setup(yargs: Yargs) {
       describe: 'Directory of .flowconfig relative to node_modules',
       type: 'string',
     },
+    useCacheUntil: {
+      alias: 'u',
+      describe: 'Use cache until specified time in milliseconds',
+      type: 'number',
+    },
   });
 }
 export async function run(args: Args) {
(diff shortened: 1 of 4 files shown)


pascalduez:
I'm not familiar with the CLI codebase, but this seems like a good addition.

pascalduez:
I'm not familiar with the CLI codebase, but this seems like a good addition.

cli/src/lib/npm/npmLibDefs.js
```diff
@@ -329,12 +329,15 @@ async function _npmExists(pkgName: string): Promise<Function> {
   return got(pkgUrl, {method: 'HEAD'});
 }
 
+export const CACHE_REPO_EXPIRY = 1000 * 60;
```
villesau: this is already defined in 
https://github.com/flow-typed/flow-typed/blob/master/cli/src/lib/cacheRepoUtils.js#L11

Maybe export that and use it instead?

cli/src/commands/install.js
```diff
@@ -147,6 +154,7 @@ export async function run(args: Args) {
     overwrite: Boolean(args.overwrite),
     skip: Boolean(args.skip),
     ignoreDeps: ignoreDeps,
+    useCacheUntil: Number(args.useCacheUntil) | CACHE_REPO_EXPIRY,
```
villesau: Did you forgot one pipe `||`? `|` is bitwise OR.

cli/src/lib/npm/npmLibDefs.js
```diff
@@ -329,12 +329,15 @@ async function _npmExists(pkgName: string): Promise<Function> {
...
+
 export async function findNpmLibDef(
   pkgName: string,
   pkgVersion: string,
   flowVersion: FlowVersion,
+  useCacheUtil: number = CACHE_REPO_EXPIRY,
```
villesau: I think this is now third place where default `CACHE_REPO_EXPIRY` is set.

Other places being:
- https://github.com/flow-typed/flow-typed/pull/3305/files#diff-8fda159841136e265107671260f39d80R157
- https://github.com/flow-typed/flow-typed/blob/master/cli/src/lib/cacheRepoUtils.js#L79

would https://github.com/flow-typed/flow-typed/blob/master/cli/src/lib/cacheRepoUtils.js#L79 be enough?

sambathl:
@villesau okay to merge ? :smile_cat:

cli/src/lib/npm/npmLibDefs.js
```diff
@@ -333,8 +334,9 @@ export async function findNpmLibDef(
...
 ): Promise<null | NpmLibDef> {
-  const libDefs = await getCacheNpmLibDefs();
+  const libDefs = await getCacheNpmLibDefs(useCacheUtil);
```
villesau: should this be `useCacheUntil` instead of `useCacheUtil`? :)

sambathl:
@villesau completed review remarks :smile:

villesau:
Thanks @GAntoine!

PR #3307, opened 2019-05-07
Remove 'unmounted' from react-transition-group

jdelStrother:
react-transition-group only calls its child-function if the state is one of entering/entered/exiting/exited - not 'unmounted'.  (Transition's render() function bails early if the state is 'unmounted' - https://github.com/reactjs/react-transition-group/blob/65cd9f83309ade93aa9c21d6f6b3c18cbf9e5f27/src/Transition.js#L330-L334)

The documentation example here - http://reactcommunity.org/react-transition-group/transition - would complain with `Cannot get transitionStyles[state] because property unmounted is missing in object literal`.


jdelStrother:
(I've also submitted a documentation fix for react-transition-group here - https://github.com/reactjs/react-transition-group/pull/498)

jdelStrother:
That’s now merged

PR #3308, opened 2019-05-07
Revert "[sequelize_v4] fix: Convert module from ESM to CommonJS"

pascalduez:
Reverts flow-typed/flow-typed#3276

### DIFF
--- a/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/sequelize_v4.x.x.js
+++ b/definitions/npm/sequelize_v4.x.x/flow_v0.42.x-/sequelize_v4.x.x.js
@@ -1379,7 +1379,7 @@ declare module "sequelize" {
     field: string,
   }
 
-  declare class $Association<Source: Model<any>, Target: Model<any>> {
+  declare export class Association<Source: Model<any>, Target: Model<any>> {
     constructor(source: Class<Source>, target: Class<Target>, options?: AssociationOptions): this;
     static BelongsTo: typeof BelongsTo;
     static HasOne: typeof HasOne;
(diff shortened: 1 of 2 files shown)


jedwards1211:
thanks, i was intending to make myself earlier but somehow failed to fetch the commit in question so that I could revert it

pascalduez:
@jedwards1211 Tests are failing (recent Flow versions), if I'm merging we should have a follow-up PR to update it. Or at least scope the def to lower Flow versions, WDYT?

jedwards1211:
let me take a look

pascalduez:
There's `$Subtype` and probably `Function`.

jedwards1211:
Some of the errors are due to flow's ever-increasing bugs.  For example, it's not our fault that Flow is failing to decide which case to select below (there's no way the object literal could be a function)

```
    Error ---------------------------------------------------------------------------------------- test_sequelize.js:1574:16
    
    Could not decide which case to select. Since case 1 [1] may work but if it doesn't case 2 [2] looks promising too. To
    fix add a type annotation to array literal [3].
    
       test_sequelize.js:1574:16
                            v
       1574|     reversed : {
       1575|       order : [['id', 'DESC']]
       1576|     }
                 ^
    
    References:
       sequelize_v4.x.x.js:5278:26
       5278|     [scopeName: string]: AnyFindOptions | Function
                                      ^^^^^^^^^^^^^^ [1]
       sequelize_v4.x.x.js:5278:43
       5278|     [scopeName: string]: AnyFindOptions | Function
                                                       ^^^^^^^^ [2]
       test_sequelize.js:1575:15
       1575|       order : [['id', 'DESC']]
                           ^^^^^^^^^^^^^^^^ [3]
```

jedwards1211:
I can make a follow-up PR to replace `$Subtype` and `Function` though

jedwards1211:
Do you have any idea how to refine member types in derived classes?  When I get rid of `$Subtype` it complains about `BelongsTo` declaring `associationType: 'BelongsTo'` whereas the base class (`Association`) declares `associationType: 'BelongsTo' | 'HasOne' | ...`

jedwards1211:
Nevermind, changing to covariant `+associationType` in the base and derived classes fixes it

PR #3314, opened 2019-05-08
[antd_v3.x.x] Add definition for Input.Password and Fix definition error for Form.create

c10h22:
1. Add definition for Input.Password added in v3.12.0 

- [Links to documentation](https://ant.design/components/input/#Input.Password-(Added-in-3.12.0)): 
- [Link to GitHub or NPM](https://github.com/ant-design/ant-design/blob/master/components/input/Password.tsx): 

2. Fix flow error when using Form.create - mixed is not a function
- [Links to documentation](https://ant.design/components/form/#Form.create(options)): 
- [Link to GitHub or NPM](https://github.com/ant-design/ant-design/blob/2d8396bad79a086ae57b3ef462e681e81786f02d/components/form/Form.tsx#L137): 
- Type of contribution: fix

### DIFF
--- a/definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
+++ b/definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
@@ -298,6 +298,7 @@ declare module "antd" {
   declare export class Input extends React$Component<InputProps> {
     static Search: typeof InputSearch;
     static TextArea: typeof InputTextArea;
+    static Password: typeof InputPassword;
   }
 
   declare class InputSearch extends React$Component<{}> {
@@ -305,6 +306,13 @@ declare module "antd" {
   }
 
   declare class InputTextArea extends React$Component<{}> {}
+
+  declare type InputPasswordProps = {
+    visibilityToggle?: boolean
+  };
+  
+  // Added in 3.12.0
+  declare class InputPassword extends React$Component<InputPasswordProps> {}
 
   declare export class Layout extends React$Component<{}> {
     static Content: typeof LayoutContent;
--- a/definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
+++ b/definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
@@ -208,7 +208,7 @@ declare module "antd" {
     valuePropName?: string
   };
 
-  declare type WrappedFormUtils = {
+  declare export type WrappedFormUtils = {
     getFieldDecorator(
       id: string,
       options?: GetFieldDecoratorOptions
@@ -247,6 +247,17 @@ declare module "antd" {
     validateFieldsAndScroll(options?: Object, callback?: ValidateCallback): void
   };
 
+  declare interface RcBaseFormProps {
+    wrappedComponentRef?: any;
+  }
+
+  declare interface FormComponentProps extends RcBaseFormProps {
+    form: WrappedFormUtils;
+  }
+
+  declare type FormWrappedProps<T> = <C: React$ComponentType<T>>
+  (component: C) => C;
+
   declare export type FormProps = {
     className?: string,
     form?: WrappedFormUtils,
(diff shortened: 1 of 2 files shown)


pascalduez:
Ping @reyronald

definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
```diff
@@ -247,6 +247,17 @@ declare module "antd" {
     validateFieldsAndScroll(options?: Object, callback?: ValidateCallback): void
   };
 
+  declare interface RcBaseFormProps {
+    wrappedComponentRef?: any;
```
pascalduez: `React.Ref`?

c10h22: Followed the way it was defined in antd Form here: https://github.com/ant-design/ant-design/blob/2d8396bad79a086ae57b3ef462e681e81786f02d/components/form/Form.tsx#L193

c10h22: `wrappedComponentRef` is used instead of `withRef` following [this note](https://github.com/react-component/form#note-use-wrappedcomponentref-instead-of-withref-after-rc-form140).

c10h22: Mentionned in antd [Form documentation](https://ant.design/components/form/#Form.create(options)) as well:
``
class CustomizedForm extends React.Component { ... }

// use wrappedComponentRef
const EnhancedForm =  Form.create()(CustomizedForm);
<EnhancedForm wrappedComponentRef={(form) => this.form = form} />
this.form // => The instance of CustomizedForm
`

pascalduez: Yes, but `wrappedComponentRef` is forwarded to the inner component `ref`, so it accepts a `React.Ref`. Anyway, let's stick to the lib definitions.

c10h22: We can replace `any` -> `React.Ref` it's a good idea I think even though in antd is using more generic type.

definitions/npm/antd_v3.x.x/flow_v0.25.x-/antd_v3.x.x.js
```diff
@@ -247,6 +247,17 @@ declare module "antd" {
...
+  }
+
+  declare interface FormComponentProps extends RcBaseFormProps {
+    form: WrappedFormUtils;
+  }
+
+  declare type FormWrappedProps<T> = <C: React$ComponentType<T>>
+  (component: C) => C;
```
c10h22: @pascalduez I think we have an issue here, the returned type is not `C` but it's more complex than that as [defined here](https://github.com/ant-design/ant-design/blob/2d8396bad79a086ae57b3ef462e681e81786f02d/components/form/interface.ts)

```javascript
// Copy from @types/react-redux https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-redux/index.d.ts
export type Matching<InjectedProps, DecorationTargetProps> = {
  [P in keyof DecorationTargetProps]: P extends keyof InjectedProps
    ? InjectedProps[P] extends DecorationTargetProps[P]
      ? DecorationTargetProps[P]
      : InjectedProps[P]
    : DecorationTargetProps[P]
};

export type GetProps<C> = C extends React.ComponentType<infer P> ? P : never;

export type ConnectedComponentClass<C extends React.ComponentType<any>, P> = React.ComponentClass<
  JSX.LibraryManagedAttributes<C, P>
> &
  hoistNonReactStatics.NonReactStatics<C> & {
    WrappedComponent: C;
  };

export type FormWrappedProps<TOwnProps extends WrappedFormInternalProps> = <
  C extends React.ComponentType<Matching<TOwnProps, GetProps<C>>>
>(
  component: C,
) => ConnectedComponentClass<C, Omit<TOwnProps, keyof WrappedFormInternalProps>>;
```

I don't know how I can fix that, any help is welcome

Issue #2991, opened 2018-12-03
Flow is removing the $Supertype and $Subtype utility types


dsainati1:
Heads up: in an upcoming release we are going to start linting against the $Supertype and $Subtype utility types, and eventually remove them entirely.

jbrown215:
Note that this affects the types for ReactRedux.

In a recent change, Flow changed the type parameter variance on ComponentType to be contravariant (releasing in 0.89), so getting rid of the Supertype there should be trivial.

wchargin:
Thanks for the heads up!

As far as I understand it, there are safe and sound ways to use
`$Supertype` and `$Subtype` for which I do not know any alternatives.
For instance, quoting from:
<https://github.com/facebook/flow/pull/6887#issuecomment-422225975>

> ```js
> /**
>  * Merge maps without mutating the arguments.
>  *
>  * Merges multiple maps, returning a new map which has every key from
>  * the source maps, with their corresponding values. None of the inputs
>  * are mutated. In the event that multiple maps have the same key, an
>  * error will be thrown.
>  */
> export function merge<K, V>(
>   maps: $ReadOnlyArray<Map<$Subtype<K>, $Subtype<V>>>
> ): Map<K, V> {
>   // Implementation elided (available upon request). It has no casts.
> }
> ```
> 
> The use of `$Subtype`, instead of fixed type parameters `InK` and `InV`,
> is useful because it enables heterogenously typed maps with a common
> upper bound:
> 
> ```js
>     it("allows upcasting the type parameters", () => {
>       const numberMap: Map<number, number> = new Map().set(1, 2);
>       const stringMap: Map<string, string> = new Map().set("one", "two");
>       type NS = number | string;
>       const _unused_polyMap: Map<NS, NS> = MapUtil.merge([
>         numberMap,
>         stringMap,
>       ]);
>     });
> ```

What is the migration plan for such usages?

jbrown215:
@wchargin: This is actually related to the conversation we had about the variance on ReadOnlyMap.

For some background, the reason `$Subtype` and `$Supertype` exist was to have something like specifying variance on type parameters before we had the `+` and `-` support.

Note that `K` and `V` are in invariant positions in your `maps` parameter. So when we check `Map<X, Y> <: Map<$Subtype<K>, $Subtype<V>>`, we check

[...]

PR #3321, opened 2019-05-10
Remove Supertype from styled-components v4+

tobilen:
- Links to documentation: https://flow.org/en/docs/types/utilities/#toc-supertype
- Type of contribution: fix

Other notes:
Subtype and supertype are deprecated. In this case, `$Shape` is an easy drop-in replacement.


retyui:
it's a bad idea to replace to `$Shape`

tobilen:
well, it is my understanding that `supertype` never really worked as intended and was - in practice - equivalent to an `any` type. so i'd disagree with replacing it with `shape` being a bad idea. could you elaborate why you think that?

what i do agree with is that the typing here is too loosecould be stricter. The component being returned should be one that expects the hocs props minus the props only relating to the hoc itself. I'll try to see if i can fit that into  a flowtype within the next few days

psychobolt:
Related to #2991

tobilen:
im still planning on doing this, sorry for not having gotten around to it yet. hopefully i'll get some time for it this week

tobilen:
once again i am confused by the fact that tests for definitions i did not touch are suddenly failing. the v4 stuff looks good to me though

AndrewSouthpaw:
> this will be irrelevant if #3321 gets merged.

If which one gets merged? You were referring to this PR. :-)

tobilen:
well, technically i am still correct then! 😆 
i meant this one though: https://github.com/flow-typed/flow-typed/pull/3228

AndrewSouthpaw:
Considering that'll create a big merge conflict with the other one, and the other one does more than this PR, I'd favor driving effort toward completing that WIP. 

Any objections?

tobilen:
nope, let's do it 👍

Issue #3008, opened 2018-12-15
[express_v4.16.x] Deprecated utility. Using `$Subtype` types is not recommended


doberkofler:
When using flow 0.89 with the `deprecated-utility` lint rule enabled (related to #2991) the following warnings are shown:
```
Warning ---------------------------------------------------------------------- interfaces.flow/express_v4.16.x.js:123:12

Deprecated utility. Using `$Subtype` types is not recommended! (`deprecated-utility`)

   123|       req: $Subtype<express$Request>,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^


Warning ---------------------------------------------------------------------- interfaces.flow/express_v4.16.x.js:129:12

Deprecated utility. Using `$Subtype` types is not recommended! (`deprecated-utility`)

   129|       req: $Subtype<express$Request>,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^


Warning ---------------------------------------------------------------------- interfaces.flow/express_v4.16.x.js:192:12

Deprecated utility. Using `$Subtype` types is not recommended! (`deprecated-utility`)

   192|       req: $Subtype<express$Request>,
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
```

daviduzumeri:
So now that Flow 0.98 is producing errors for `deprecated-utility` errors, is there any plan on how to deal with this? Adding properties to the request object is a fairly common practice, so simply reverting it to allowing `express$Request` on its own would probably break a huge majority of applications using this type definition. Should we just give up on flow-typing Express?

doberkofler:
@daviduzumeri Sorry I cannot help on this as I have transitioned to TypeScript

pbondoer:
Same happening with `express-v4.x.x`

borlaym:
So now it is an error by default. Anyone have an idea how to handle it?

emmenko:
I turned off the deprecated utility lint rule

```
[lints]
deprecated-utility=off
```

Someone would need to update the express types I guess, then we can turn the rule back on.

daviduzumeri:
The question is _how_ to update the express type, as simply making it `$Request` would probably break a very high number of users of this type definition. (Personally, I'm taking this as the over-the-ledge kick to get everyone onboard with moving to TypeScript, which seems to be run by people who are actually interested in maintaining a type definition community).

AndrewSouthpaw:
[...]

PR #3337, opened 2019-05-18
[express] Remove Subtype for recent versions of flow

pbondoer:
Following discussion in #3008 here's a PR to remove `$Subtype` from the recent flow defs for `express`.

Test pass locally, let me know if there's anything else I'm missing.

### DIFF
--- a/definitions/npm/express_v4.16.x/flow_v0.93.x-/express_v4.16.x.js
+++ b/definitions/npm/express_v4.16.x/flow_v0.93.x-/express_v4.16.x.js
@@ -117,13 +117,13 @@ declare class express$Response extends http$ServerResponse mixins express$Reques
 declare type express$NextFunction = (err?: ?Error | "route") => mixed;
 declare type express$Middleware =
   | ((
-      req: $Subtype<express$Request>,
+      req: express$Request,
       res: express$Response,
       next: express$NextFunction
     ) => mixed)
   | ((
       error: Error,
-      req: $Subtype<express$Request>,
+      req: express$Request,
       res: express$Response,
       next: express$NextFunction
     ) => mixed);
@@ -186,7 +186,7 @@ declare class express$Router extends express$Route {
   param(
     param: string,
     callback: (
-      req: $Subtype<express$Request>,
+      req: express$Request,
       res: express$Response,
       next: express$NextFunction,
       id: string
--- a/definitions/npm/express_v4.x.x/flow_v0.93.x-/express_v4.x.x.js
+++ b/definitions/npm/express_v4.x.x/flow_v0.93.x-/express_v4.x.x.js
@@ -101,8 +101,8 @@ declare class express$Response extends http$ServerResponse mixins express$Reques
 
 declare type express$NextFunction = (err?: ?Error | 'route') => mixed;
 declare type express$Middleware =
-  ((req: $Subtype<express$Request>, res: express$Response, next: express$NextFunction) => mixed) |
-  ((error: Error, req: $Subtype<express$Request>, res: express$Response, next: express$NextFunction) => mixed);
+  ((req: express$Request, res: express$Response, next: express$NextFunction) => mixed) |
+  ((error: Error, req: express$Request, res: express$Response, next: express$NextFunction) => mixed);
 declare interface express$RouteMethodType<T> {
   (middleware: express$Middleware): T;
   (...middleware: Array<express$Middleware>): T;
@@ -152,7 +152,7 @@ declare class express$Router extends express$Route {
   param(
     param: string,
     callback: (
-      req: $Subtype<express$Request>,
+      req: express$Request,
       res: express$Response,
       next: express$NextFunction,
       id: string


mgreenw:
I'm worried that just removing `$Subtype` won't quite do it because a common usage is to add properties to the `req` parameter as it is passed along, in which case a subclass of `Request` is commonly used. I think `Class<express$Request>` would be the most appropriate thing here as that allows subclasses of the request class.

Please let me know if it's more appropriate to post in #3337!

AndrewSouthpaw:
@mgreenw Could you write out a test or something that @pbondoer could use to fit the new typedef? I think we're both operating on poor understanding of how express is used these days, and a real-world typing example would be helpful, if the current tests do not cover your concern.

mgreenw:
@AndrewSouthpaw Absolutely -- here is an example of how the former `$Subtype` feature could be used in an express application.

### Example

First, we _declare_ (not define) a subclass of `express$Request`.

```declare export class AuthorizedRequest extends express$Request {
  user: {
    id: number,
    ipAddress: string,
  };
}```

Any middleware can modify the `req` object directly, so when we know that we are using a middleware _after_ one that has been modified, we can declare the type of the `req` parameter to be `AuthorizedRequest`:

```app.use((req: AuthorizedRequest, res: $Response, next: NextFunction) => {
  console.log(user.id, user.ipAddress);
});```

### API Design

Unfortunately, the `express` API was not designed with Flow or static analysis in mind, and this API does not lend itself to type inference because of the cascading nature of the middleware. If it were to be redone, the `req` object should not be modified directly but should be immutably copied when necessary. Then, downstream middleware functions would be able to infer the type of the `req` object from the upstream modifications.

Therefore, the above example is not ideal in any way. However, it is the only way at this point to set request context (like a user id) and pass it to downstream using normal express middleware. Flow complains, however, if we type the `req` object in the example above as a normal `$Request` object because it does not contain the `user` property.

### Tests

I tried testing my above `Class<T>` idea and unfortunately it is not compatible with the example I have. The current state of this PR (removing $Subtype<>) also breaks the example. I will keep investigating alternatives...

**Comment by AndrewSouthpaw** [comment]
Sweet. Thanks for the breakdown!

In your exploration, make sure you try out type spreads:

```js
type MySubtypeT = {
  ...$Exact<SomeTypeT>,
  foo: string,
}
```

pbondoer:
Hey there, sorry I couldn't get to this earlier. Life kind of took over.

Thank you very much for the example! This should be more than enough for me to test it out.

MatrixFrog:
Anything we can do to help push this over the finish line?

mgreenw:
@MatrixFrog @pbondoer I've tried a bunch of combinations of `Class<T>` and generic bounds and I cannot figure out how to remove `$Subtype<express$Request>` without causing the case I described to fail as well. @jbrown15, do you have any suggestions on how to handle this case?

LoganBarnett:
We have a repo that solved this problem a while back (probably before `$SubType` was introduced to the libdef, or I just didn't know I could utilize it).

We have a custom libdef:

``` javascript
// "-ext" added because it will squash the existing express. This includes the
// json body that comes from adding the built in json body parser at runtime.
declare module 'express-ext' {
  declare type JsonRequest = { body: mixed }
}
```

Then in our implementation:

``` javascript
import type { JsonRequest } from 'express-ext'
import type { $Application, $Request, $Response } from 'express'

...

app.post('/foos', (req: $Request & JsonRequest, res: $Response) => {
    const jsonReq: JsonRequest = req
    // Now you can use jsonReq.body to get a mixed-ish thing.
})
```

Could we do something like this? I realize it isn't the prettiest, but like what was mentioned earlier: Express wasn't really designed with static typing in mind. The middleware, request chains you can modify, etc don't really lend themselves well to saying "this is the stuff I know will always be there" and instead becomes highly dependent on the sequence in which things are applied.

LoganBarnett:
Something I forgot to add: This would mean we'd just remove `$Subtype` and present an alternative using existing Flow mechanisms that the consumers could provide - they are already providing their own types and assuming risks of that being wrong anyways.

It's worth noting that my example is fairly old at this point, and you might be able to get away with fewer hijinks to make it work.

mgreenw:
@LoganBarnett That's a nifty solution. I think that could totally work.

AndrewSouthpaw:
I wonder if a helper type could be provided, something like

```js
type ExtendedRequest<T> = { ...$Exact<$Request>, ...$Exact<T> }
```

mgreenw:
@AndrewSouthpaw That looks like it'd work, but do you think that helper belongs in this express `flow-typed` repo or should it be defined manually per use?

AndrewSouthpaw:
I dunno, it wouldn't hurt to be in the repo to help nudge people into that direction. None of this is obvious. Invariably people will open up the libdef for express in anger as they try to extend the Request. If there was a `<marquee>THIS IS HOW YOU EXTEND THE REQUEST</marquee>` that might help. Sure people could just write it out manually but it's one more way to make it a little more plug'n'play.

LoganBarnett:
Re: @AndrewSouthpaw's "nudge" - Maybe we can provide some examples via the built in middlewares, such as `body-parser`? Maybe it's not built-in anymore and I'm dating myself :P Either way I think that's good to have. I wish we had a `USAGE.md` or similar that would blast you on any given update or install of a libdef. Think Homebrew's "caveats" section. It comes in handy a lot!

Supporting middelware definitions in `express` proper is likely a rabbit hole. We can let the community contribute libdefs against the middleware modules directly (in flow-typed). `express` shouldn't be less for it and nobody has to get entangled with a middleware unrelated to their own.

Ashoat:
I tried working on this today, after making an ill-informed attempt to just strip `$Subtype` in #3532 last week.

1. I found that I needed to parameterize `express$Middleware` to allow for custom properties on the response and request objects.
2. That ended up complicating the typings for the functions that take variable middlewares, as each of their parameters needed a different parameterization.
3. I tried to type each variant individually (à la reselect). That seemed to mostly work, but I couldn't get `test_overrideUseMethodClassExtension.js` to work. Updating the `declare class test_express$CustomApplication` to match my new uber-complex each-variant-individually-typed `express$Application`, it ended up triggering a bunch of errors due to incorrect variant pairings.
4. I then tried to just get rid of the `declare class test_express$CustomApplication` statement, relying on the type parameters properly identifying the custom response and request shapes. This mostly worked, but I couldn't get the final `$ExpectError` in `test_overrideUseMethodClassExtension.js` to trip up. I suspect this may be because of a Flow bug, and I opened https://github.com/facebook/flow/issues/8110 to track

Can share my current state if anybody wants to provide feedback, or pick up where left off. Also curious on any feedback to the Flow issue above, and whether or not I'm misusing type parameter refinement in some way.

pascalduez:
Closing in  favor of #3578

Issue #2411, opened 2018-06-22
Ramda's ifElse incompatible with test


vorobeez:
Flow throws me the folowing error:
```
Cannot call R.ifElse because rest array [1] has an unknown number of
elements, so is incompatible with tuple type [2].

     src/validator.js
       66│
       67│ const checkMaxPrecision = (maxPrecision) => R.ifElse(
       68│   R.test(new RegExp(`^\\d+\\.?\\d{0,${maxPrecision}}$`)),
       69│   R.always(undefined),
       70│   R.always(VALIDATION_ERROR.MAX_PRECISION(maxPrecision)),
       71│ );
       72│

     ../../flow-typed/npm/ramda_v0.x.x.js
 [2]   31│   declare type __CurriedFunction1<A, R, AA: A> = (...r: [AA]) => R;
         :
 [1] 1829│     cond: (...args: Array<any>) => boolean,
```

I don't understand is it error in my code or in ramda's types? How can i fix this?

Please, help.

AndrewSouthpaw:
@LoganBarnett

LoganBarnett:
Apologies, I must've missed the email notification for this.

@vorobeez are you still seeing this issue with a recent `ramda` libdef? We see a lot of issues relating to using rest args in the libdef and we've been working on cleaning them as we go, but there's a lot left as well.

winghimjns:
I am having this issue as well and I have just solved it. I think there must be better way but at least the flow error is gone.

I think Ramda team didn't expect to use Ramda with Flow because in production this would never cause any typing problems.

at line 68, the type checker expects a function takes unlimited arguments and return a boolean (`(...args: Array<any>) => boolean`). But your function composed with `R.test` would output a function of `(arg: any) => boolean`.

In your case, `R.test(new RegExp(...))` returns a function accepts one arg only whilst flow thinks it should take unlimited args.

I found this is annoying as well as it should never be a problem to run. But anyway I solved it by composing a function to change the type of the original function.

So I built this:

```js
function ary<I, T>(fn: (a: I) => T): (...rest: Array<I>) => T {
    return (...rest) => fn(rest[0]);
}
```

And in your case, line 68 should exactly be changed to:

```js
   ary<boolean>(R.test(new RegExp(`^\\d+\\.?\\d{0,${maxPrecision}}$`))),
```

LoganBarnett:
[...]

PR #3338, opened 2019-05-20
fix Ramda ifElse to work with variadic arguments

LoganBarnett:
- https://ramdajs.com/docs/#ifElse
- #2411
- Type of contribution: fix

Other notes:

This fixes the issue found in #2411 where `test` could not be used with `ifElse` because `test` is automatically curried, which relies upon variadic arguments.

As another note, I noticed as part of working on this that some of our tests have a lot of cross-dependencies. For example, the `ifElse` test also depended on `is`. I worry that if something happens with `is`, the test for `ifElse` can become invalid. Perhaps instead we can write our tests by only using a single Ramda function at a time, so our test doesn't potentially become a lie built atop another lie. For example, if `is` returns or takes `any` it could invalidate the `ifElse` test entirely. It also conflates what's really being tested. In the case of `test` from #2411, if it dropped variadic argument support while the `ifElse` test I used actually used `test`, then the test would silently become invalid yet pass.

TLDR; Going forward, let's use only one Ramda function at a time during an assertion.

Thoughts?

From the commit message:

This is very typical when working with curried functions, particularly from within Ramda itself. The prior definition of ifElse used to work in type refinement scenarios (although really, it appears to be coercing to any or otherwise losing type information).

The original test for ifElse is broken because it relies upon such refinements or loss of type information.

The tests for ifElse are now expressed in a way that is independent of other Ramda functions.

### DIFF
--- a/definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
@@ -1,7 +1,6 @@
 /* @flow */
 /*eslint-disable no-undef, no-unused-vars, no-console*/
 import _, {
-  always,
   compose,
   curry,
   filter,
--- a/definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
@@ -87,15 +87,14 @@ describe('ifElse', () => {
 
   it('works with variadic arg inputs', () => {
     const ifElseFn = ifElse(
-      // The original motivation was to work with "test". Future changes could
-      // invalidate the nature of the test, so we need to mimic the form of test
-      // at time of writing. The error will be along the lines of this:
+      // Without this test, variadic argument input functions will produce an
+      // error like this:
       //
       // Cannot call `ifElse` because rest array [1] has an unknown number of
       // elements, so is incompatible with tuple type [2].
       (...args: [string]): boolean => true,
-      always(undefined),
-      always('some error'),
+      (s: string) => s.toUpperCase(),
+      (s: string) => s.toLowerCase(),
     );
     const result: void | string = ifElseFn('input')
   })
@@ -103,8 +102,8 @@ describe('ifElse', () => {
   it('preserves the types used with variadic argument condition functions', () => {
     const ifElseFn = ifElse(
       (...args: [string]): boolean => true,
-      always(undefined),
-      always('some error'),
+      (s: string) => s.toUpperCase(),
+      (s: string) => s.toLowerCase(),
     );
     // $ExpectError
     const result: void | string = ifElseFn(5)
--- a/definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
+++ b/definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
@@ -1,6 +1,17 @@
 /* @flow */
 /*eslint-disable no-undef, no-unused-vars, no-console*/
-import _, { compose, pipe, curry, filter, find, repeat, zipWith } from 'ramda';
+import _, {
+  always,
+  compose,
+  curry,
+  filter,
+  find,
+  ifElse,
+  pipe,
+  repeat,
+  zipWith,
+} from 'ramda';
+import { describe, it } from 'flow-typed-test';
 
 const ns: Array<number> = [1, 2, 3, 4, 5];
 const ss: Array<string> = ['one', 'two', 'three', 'four'];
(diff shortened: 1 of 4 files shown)


definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
```diff
@@ -61,11 +72,58 @@ const def1: number = defaultTo42(undefined);
...
+    // $ExpectError
+    const result: void | string = ifElseFn(5)
+  })
+
+  // This was the original test for ifElse, which depends on the implementation
+  // of "is". "is" in its present form does not perform type refinement, so the
+  // stricter form of ifElse fails the check with this test. This test below
+  // captures the issue, and is left for reference and/or if some brave soul
+  // wishes to expand the definition of ifElse (and possibly also "is") to
+  // refine the predicate correctly. At time of writing, $Pred and $Refine are
```
AndrewSouthpaw: > and/or if some brave soul wishes to expand the definition

😂

definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
```diff
@@ -61,11 +72,58 @@ const def1: number = defaultTo42(undefined);
...
+    // Because the object literal ({}) isn't a number it
+    // $ExpectError
+    const ie: number | string = incCount({});
+  })
+
+  it('works with variadic arg inputs', () => {
+    const ifElseFn = ifElse(
+      // The original motivation was to work with "test". Future changes could
+      // invalidate the nature of the test, so we need to mimic the form of test
+      // at time of writing. The error will be along the lines of this:
```
AndrewSouthpaw: I'm a little confused because we're writing a test about `test` in the context of `ifElse`. Would you mind expanding a bit here and disambiguating? I'm still having trouble understanding this point.

AndrewSouthpaw:
> TLDR; Going forward, let's use only one Ramda function at a time during an assertion.

Agreed!

AndrewSouthpaw:
Also... considering your mention of [nmn's solution about variadic generics](https://github.com/facebook/flow/issues/1251#issuecomment-289924710) I'm a little surprised you didn't need to provide a `$ReadOnlyArray<mixed>`.

definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
LoganBarnett: Ah - this is perhaps the wrong kind of context I'm providing. The original ticket this is from is essentially "`test` does not work with `ifElse`", when really the problem is `ifElse` doesn't work with varargs. This might have worked as a self-made PR comment instead of a comment in the code, since the ticket itself and a PR comment are ephemeral to each other compared to the code that will outlast the others' relevances.

I'll remove this comment.

LoganBarnett:
I remember taking out `$ReadOnlyArray<mixed>` because of existing tests that broke from it. It might have been a red herring though, because of the `is` call kind of falsely doing a refinement. My memory isn't the most trustworthy thing though. I tried to reproduce the issue with `$ReadOnlyArray<mixed>` in [this try](https://flow.org/try/#0PTAEBUAsEsGdQC4FNYPgQQE4HN4GMB7AO1UwENoikATRA0AEgCUkzqB5IgGwE8tyeAHgC20AB40AfACgQESElAAHTAQBGSLsNBlMsOClAB3BUVA8CAVx0JkwpQjqhLsRQgXLdZYUmR7QlIgKsmAImJaKZES0AGZkXK6geGSu8IHuimrkRHiQoDGWOQjQxPAAtGVBSDw6mIr69rwhoKIStJSorLQEMcp1rpgAbjQAdNLS1Eh4XLqKBUUlZtAxAKIJSIJYuABcjCxsnLz8ZEKtUgA0oABClwDCkgAU0qAvScTUuw8j37o7oFuwACUoAAvJJQGoCAQuKwiM9QIDPk9XvkAIyfb4jX6wXYA4Fg67w-Hgh4xABMGJ+OBx-2pxNAt3pXypfzxoPBV1AAB8GcCAGTwl4PZEotGUrHU3F09nXc6C17k8XYqW4em3eVMzHK2mqmWcnmMgDc8Mm01m+UKeGKxACq3Wm0ley6hz4mAEInEF1lDMe8sI0SVjrZBMh0NhcpRMXRoGZEtZ0oJN3lipjWqDCfB6peiNTLJpwY53IZxuk-tQb2ioNta1cItj2oA2qRKNgALo50MwqIysIRCMxmnNojYemwEYIAgAVSUSiQmFuKSQD0B-Yeg7CLdH44IABkCEY5wvayvpICS2XHP1LFwELtBgRoLQeUPsFX-dQHgByShKSwIT+As0DArGIs5WisboEJgIQXqAV43neD5PqAL5vu8DwAKyAaWpSOJQeC3FYRCOCC1brA8YgypRACkoBkqCpEAAyXJRBKUQA1KAqIsVR24AMobsOy5njhJB4UguxEJYwgaJgRaoaR+GEYUCAPKiIlyMBoFTAgEGqNBciwdAEmgFJMlzvJgmvopOTKcRDwAN4AL5nkAA) but I cannot any longer. Both `Args` and `Args: $ReadOnlyArray<mixed>` work here. Do you have a preference on which you'd like to see?

definitions/npm/ramda_v0.x.x/flow_v0.82.x-/test_ramda_v0.x.x_logic.js
LoganBarnett: Okay this should be addressed now. There is a comment there still, but the `test` context is now gone.

LoganBarnett:
I forgot to apply the changes to the latest Ramda (0.26.x). It's just a straight copy-pasta.

AndrewSouthpaw:
>  It's just a straight copy-pasta.

https://www.youtube.com/watch?v=cE1FrqheQNI

PR #3379, opened 2019-06-08
Compression

edoardo-bluframe:
- Links to documentation: https://github.com/expressjs/compression
- Link to GitHub or NPM: https://github.com/expressjs/compression

Other notes:


edoardo-bluframe:
@AndrewSouthpaw Hey TODOs are `express` `req` and `res`

I imported from `express` and then you guys told me on channel that we cannot import at this stage!

I guess my question is: Do we wanna copy and paste from the `express` libdef?

Does that create a problem for maintaining 2 identical definitions?? :smile:

AndrewSouthpaw:
Oh! I understand. Go ahead and do something like this:

```js
type ExpressReq = any // fake import from express libdef, since we can't have dependencies

const whatever = (req: ExpressReq) => (...)
```

I'd probably drop the TODO as well, but at least we need to know the TODO is because you're faking an import from a library libdef, not anything else (e.g. you just didn't implement the req type).

PR #3383, opened 2019-06-12
make ramda ascend and descend return curried functions

LoganBarnett:
- Links to documentation: https://ramdajs.com/docs/#ascend https://ramdajs.com/docs/#descend
- Type of contribution: fix

Other notes:

Returning a curried function makes these functions work with `sort`, which matches usage as well as the documented examples for these functions.

### DIFF
--- a/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_relation.js
+++ b/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_relation.js
@@ -74,8 +74,8 @@ describe('ascend', () => {
 
 describe('descend', () => {
   /**
-   * In the examples for ascend, its result is plugged into sort. In order
-   * for this to function properly ascend needs to yield a function that can
+   * In the examples for descend, its result is plugged into sort. In order
+   * for this to function properly descend needs to yield a function that can
    * take two arguments.
    */
   it('works as an input to sort', () => {
--- a/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_relation.js
+++ b/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_relation.js
@@ -2,8 +2,10 @@
 /*eslint-disable no-undef, no-unused-vars, no-console*/
 import { it, describe } from "flow-typed-test";
 import _, {
+  ascend,
   compose,
   curry,
+  descend,
   filter,
   find,
   gt,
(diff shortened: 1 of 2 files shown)


definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_relation.js
```diff
@@ -46,6 +50,50 @@ const eqb: boolean = _.eqBy(Math.abs, 5, -5);
 
 const es: boolean = _.equals([1, 2, 3], [1, 2, 3]);
 
+describe('ascend', () => {
+  /**
+   * In the examples for ascend, its result is plugged into sort. In order
+   * for this to function properly ascend needs to yield a function that can
+   * take two arguments.
+   */
```
AndrewSouthpaw: Awesome. I love your comments.

It might make more sense to put those in the implementation though?

AndrewSouthpaw:
Also seems like build failed, can you investigate and see if that's a regression?

definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_relation.js
LoganBarnett: I agree!

The comment has been added to the implementation with some additional context that I think makes more sense there. I left the comments in the tests because it still feels like a surprising case that should be called out, and `it` blocks don't capture the detailed intent as well. Feel free to be very picky on what you'd like to see there though.

LoganBarnett:
The build failures are the same as local. In a recent Flow update, `$Enum` was removed due to deprecation in favor of `$Keys`. If it's all the same to you, I'd prefer to chase this down in a subsequent pull request since this pull request doesn't actually introduce the problem. I worry it may balloon into something that will distract from the intent of this pull request.

Assuming that's fine, I can pursue that as my next pull request, which I think I can have something for this week.

As an aside, are we happy with out-of-band upgrades to `flow-bin` causing breaks that aren't immediately discovered? I don't mean that in snark - I accept the possibility this is just the workflow that's optimal given many disparate modules and contributors in this repository. But if it's not optimal, maybe we could have a process where we introduce new versions of Flow as their own pull requests (and thus have build failures), and we flag down contributors to catch their individual areas of responsibility up.

I'm happy chipping in either way though. Just spitballing :)

LoganBarnett:
It was way easier than I thought. After #3406 is merged (assuming all is well) we should be clear of build errors in this pull request.

LoganBarnett:
@AndrewSouthpaw the Ramda build should be passing again since that other pull request was merged. Can you try to run it again? I don't seem to have the ability to kick it off.

AndrewSouthpaw:
Looks like the build is good to go now.

Issue #3386, opened 2019-06-14
Move FlowDefinitelyTyped to flow-typed organization


goodmind:
Would you accept my repository [FlowDefinitelyTyped](https://github.com/goodmind/FlowDefinitelyTyped) into flow-typed organization?

Opposed to this repository this would be repository of _horrible quality_ definitions (autogenerated with [flowgen](https://github.com/joarwilk/flowgen) from DT), that can be used as blueprints for contributing to this repository

/cc @villesau @GAntoine @AndrewSouthpaw

villesau:
I think that's an interesting idea! Maybe we could build some tooling around it to make picking the libdef easier? Something like "We couldn't find first class libdefs for you, but looks like there is DefinitelyTyped definition. Would you want to import transformed version of it for basis of a new libdef? Yes/No"

goodmind:
@villesau it would be perfect for create-stub

AndrewSouthpaw:
So are you envisioning this repo as containing TS definitions that we then convert on the fly from the CLI to flow when we can't find the libdef in flow-typed?

AndrewSouthpaw:
If so, is there a reason we wouldn't want to pull directly from the DefinitelyTyped repo?

goodmind:
Now I don't think we need repo actually

villesau:
Yeah on-the-fly transformation sounds a bit more lightweight. But the CLI should communicate super clearly that it's machine transformation and TS->Flow might not be accurate and manual reviewing is required.

AndrewSouthpaw:
Cool. :) Closing this issue.

PR #3387, opened 2019-06-14
Add flowgen to create-stub in CLI

goodmind:

Other notes:

Related issue #3386 

`flow-typed create-stub ajv --typescript` would use `flowgen`
`flow-typed install` wouldn't use `flowgen` (or you would need to ask user on each untyped dependency to use typescript or not?)

This supports DT, `typings` and `types` fields in `package.json`, or `index.d.ts` if field not found


goodmind:
@Ville @AndrewSouthpaw

AndrewSouthpaw:
Huh I thought I commented. Can you add a test or some other way of verifying that it works?

goodmind:
@AndrewSouthpaw it depends on what .d.ts file contains. `flowgen` already has ton of tests

AndrewSouthpaw:
Sure, that makes sense. Tests would be great, and also I get that because they're not in place you wouldn't want to make new ones. Would you mind just having a screen gif or at least a demonstrated output so we can know it works?

Also, the build is broken, related to flow typing issues from your code. Those will need to be fixed first.

goodmind:
https://gist.github.com/goodmind/7b6233decb2def0597f1a0852078ab0e

goodmind:
It can of course generate nonsense or even throw errors, this is why it is opt-in

goodmind:
@AndrewSouthpaw build was broken because of old libdefs, well that's no surprise :\

AndrewSouthpaw:
Oh, can you add some notes in the `README` on the new usage? This will be helpful to incoming users.

goodmind:
@AndrewSouthpaw done

README.md
```diff
@@ -109,6 +109,15 @@ a short grace period during a libdef installation, but sometimes it is useful to
 do this update manually. Use this command if you want to download the most
 recent definitions into the cache for yourself.
 
+##### `flow-typed create-stub package-specification [--typescript]`
+
+Creates `any`-typed libdef that you can fill in.
+
+If the `--typescript` flag is specified, the libdef would be converted from `typings` field in package.json, `index.d.ts` file or `@types/package-name` package to flow-typed libdef format.
+
+Note that this option can produce unpredictable output because TypeScript-to-Flow converter is not perfect.
```
villesau: Maybe also mention that manual fixes and careful reviewing of the output is needed, and potentially suggestion about contributing them back to flow-typed?

cli/src/lib/npm/npmLibDefs.js
```diff
@@ -408,8 +408,8 @@ export async function getInstalledNpmLibDefs(
               flowVerMatches == null
                 ? matches[3]
                 : flowVerMatches[3] == null
-                  ? flowVerMatches[2]
-                  : `${flowVerMatches[2]}-${flowVerMatches[4]}`;
+                ? flowVerMatches[2]
+                : `${flowVerMatches[2]}-${flowVerMatches[4]}`;
```
villesau: interesting change from prettier O_o I really like the previous formatting more.

README.md
goodmind: Isn't this already adressed with `can produce unpredictable output`

villesau: I think it can't be emphasised enough.

goodmind: I updated it

cli/src/lib/npm/npmLibDefs.js
goodmind: This is just prettier hating nested ternaries

goodmind:
@AndrewSouthpaw @villesau

PR #3404, opened 2019-06-24
[@testing-library/react_v8.x.x] Make function returns generic

tobilen:
- Links to documentation:
- Link to GitHub or NPM: 

Other notes:
Currently, something like this leads to a flow error:
```jsx
  it('checks whether a button is disabled', async () => {
    const { getByText } = render(
      <button disabled>Button</button>,
    );
    const button = getByText('Button');
    expect(button.disabled).toBe(true);
  });
```

Because according to the types, `getByText` always returns an `HTMLElement` and they do not have the `disabled` attribute.

This PR introduces generics so the type of the element needs to be passed as a flow parameter, meaning you can do this:
```jsx
  it('checks whether a button is disabled', async () => {
    const { getByText } = render(
      <button disabled>Button</button>,
    );
    const button = getByText<HTMLButtonElement>('Button');
    expect(button.disabled).toBe(true);
  });
```

### DIFF
--- a/definitions/npm/@testing-library/react_v8.x.x/flow_v0.67.1-/react_v8.x.x.js
+++ b/definitions/npm/@testing-library/react_v8.x.x/flow_v0.67.1-/react_v8.x.x.js
@@ -12,86 +12,181 @@ declare module '@testing-library/react' {
 
   declare type SelectorMatchOptions = { selector?: string } & TextMatchOptions;
 
-  declare type GetByText = (
-    text: TextMatch,
-    options?: SelectorMatchOptions
-  ) => HTMLElement;
-
-  declare type QueryByText = (
-    text: TextMatch,
-    options?: SelectorMatchOptions
-  ) => ?HTMLElement;
-
-  declare type AllByText = (
-    text: TextMatch,
-    options?: SelectorMatchOptions
-  ) => Array<HTMLElement>;
-
-  declare type GetByBoundAttribute = (
-    text: TextMatch,
-    options?: TextMatchOptions
-  ) => HTMLElement;
-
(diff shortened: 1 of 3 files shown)


pascalduez:
Hi @tobilen,

this looks like a huge step backward DX wise as well as a consequent breaking change.  
Having to pass a generic for every query.  
We should at least set a default, and maybe a constraint.

```
declare type GetByText<T: HTMLElement = HTMLElement> = T;
````

File: definitions/npm/@testing-library/react_v8.x.x/flow_v0.67.1-/react_v8.x.x.js [hunk 1]
```diff
@@ -12,86 +12,181 @@ declare module '@testing-library/react' {
...
-  declare type AllByBoundAttribute = (
-    text: TextMatch,
-    options?: TextMatchOptions
-  ) => Array<HTMLElement>;
+  declare type GetByText<T> = T;
+  declare type QueryByText<T> = ?T;
+  declare type AllByText<T> = Array<T>;
+  declare type GetByBoundAttribute<T> = T;
+  declare type QueryByBoundAttribute<T> = ?T;
+  declare type AllByBoundAttribute<T> = Array<T>;
```
pascalduez: I find those quite confusing (naming), because there are now only return types.

tobilen:
i would absolutely *love* to set a default! in fact this is what i first tried, the default param has to be set within the `GetsAndQueries` object though.

Unfortunately, i kept getting eslint as well as flow errors (unexpected token `=`) when trying to set a default param like you suggest though. Happy for any help here

tobilen:
🤷‍♂

PR #3417, opened 2019-07-02
Drop version command

### LINKED ISSUE #3412: CLI includes both the command as well as an option to show version information
Presently, `flow-typed version` and `flow-typed --version` serves the same purpose.
I would prefer to have only the `--version` option.

jamesgeorge007:
> Fixes #3412

`--version` already serves the purpose as the `version` command.


jamesgeorge007:
cc @AndrewSouthpaw

PR #3418, opened 2019-07-02
feat(cli): flow-typed info to show up local env info

### LINKED ISSUE #3413: [CLI] Having a dedicated command to show up local environment information
## Proposed API

`flow-typed info`

Check out [envinfo](https://www.npmjs.com/package/envinfo)

jamesgeorge007:
FIxes #3413 

addition

## Proposed API

`flow-typed info`


jamesgeorge007:
cc @AndrewSouthpaw

AndrewSouthpaw:
BTW @jamesgeorge007 I get notified on all activity on the repo, no need to specifically tag me for new PRs. :-)

AndrewSouthpaw:
Were the lints from our repo prettier settings or something else?

jamesgeorge007:
@AndrewSouthpaw Hope everything's fine now :thinking:

pascalduez:
I'm personally not so keen on merging this, would like @villesau review first.   
Why would we need such infos for flow-typed, like browsers and such.... :thinking:

jamesgeorge007:
`npx envinfo` logs all those general information regarding the local environment. It only shows up local and global versions of certain utilities by default. IMO we could customize the output of the respective command so that only what required shows up :thinking:

villesau:
I don't understand the need for this either. flow-typed is not a tool for checking your local machine information. Also bloating the dependencies sounds risky.

AndrewSouthpaw:
I don't care either way, I was fine with it being a thing, if y'all are strongly opposed we can close it.

jamesgeorge007:
Presently, the repository lacks an issue template which is certainly good to have. Most repositories out there, especially scaffolding utilities are shipped with a dedicated `info` command/option which proves useful while submitting bug reports. It is a common practice to prompt the user in order to paste the results of `<cli> info` within the environment info section as part of the issue template which helps to better understand the issue.

cli/src/commands/info.js
```diff
@@ -0,0 +1,22 @@
...
+  'Print debugging information concerning the local environment';
+export function setup(yargs: Yargs) {
+  return yargs.strict();
+}
+export async function run(argv: Argv) { // eslint-disable-line
+  console.log('\nEnvironment Info:');
+  const data = await envinfo.run({
+    System: ['OS', 'CPU'],
+    Binaries: ['Node', 'Yarn', 'npm'],
+    npmGlobalPackages: ['flow-typed'],
```
jamesgeorge007: Removed all those browser information showing up

jamesgeorge007:
@AndrewSouthpaw If this still doesn't make sense, you may close this PR :+1:

jamesgeorge007:
@AndrewSouthpaw If this still doesn't make sense, you may close the PR :+1:

AndrewSouthpaw:
> especially scaffolding utilities are shipped with a dedicated info command/option which proves useful while submitting bug reports. It is a common practice to prompt the user in order to paste the results of <cli> info within the environment info section

Yeah that was my experience as well, `flow` does this for example. Telling people to run `npx envinfo` is probably sufficient though if we don't want to complicate the cli.

@jamesgeorge007 any concerns about going with that as a solution? If that's the case you could rewrite to add the issue template and such.

jamesgeorge007:
Closing in favour of #3451

PR #3425, opened 2019-07-03
fix ramda lenses

LoganBarnett:
- Links to documentation:
  - https://ramdajs.com/docs/#lens
  - https://ramdajs.com/docs/#lensIndex
  - https://ramdajs.com/docs/#lensPath (not fixing this one yet)
  - https://ramdajs.com/docs/#lensProp
  - https://ramdajs.com/docs/#over
  - https://ramdajs.com/docs/#set
  - https://ramdajs.com/docs/#view
- Type of contribution: fix

Other notes:

The lens functions (`lens`, `lensProp`, `lensIndex` (excluding `lensPath` for today)) weren't producing a correct type in the output, so some modification was needed. The lens types relate to the lens operation functions: over, set, and view. All of the lens operation functions also needed modification to work with the new lens types.

These types should closely reflect Haskell's lens system, which itself inspires Ramda.

These forms aren't considered to be absolutely correct, but should prove more correct than the prior declarations. I do not claim to be an expert on lenses and additional scrutiny over this pull request would be much appreciated.

This also introduces Functor types, which were helpful in reasoning about lenses. Other places in Ramda respect fantasy-land types as well as a generic concept of a Functor (it has a "map"). These should be useful in other Ramda definitions going forward, as many of them also can operate on Functors.

### DIFF
--- a/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/ramda_v0.26.x.js
+++ b/definitions/npm/ramda_v0.26.x/flow_v0.76.x-/ramda_v0.26.x.js
@@ -11,6 +11,12 @@ declare type $npm$ramda$Placeholder = { "@@functional/placeholder": true };
 declare opaque type $npm$ramda$Reduced<T>;
 
 declare module ramda {
+  declare type FunctorObj<A> = {
+    map: (<B>(A => B) => FunctorObj<B>),
+  }
+  declare type FunctorFantasyLand<A> = {
+    'fantasy-land/map': (<B>(A => B) => FunctorFantasyLand<B>),
+  }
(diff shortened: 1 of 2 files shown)


LoganBarnett:
Also

mind=blown

goodmind:
This https://github.com/facebook/flow/pull/7879 or this https://github.com/facebook/flow/pull/7860 should enable typing `lensPath`

AndrewSouthpaw:
Every time I heard reference to "fantasy-land whatevers" I just laugh to myself: no wonder people think functional programming is ludicrous. 😉

LoganBarnett:
@goodmind Nice! I'll watch those so we might later take a pass at `lensPath` with them :)

PR #3426, opened 2019-07-04
Generate stub declarations for directory modules

knu:
This change makes `flow-type install` create stubs for `core-js/stable` and `core-js/es/symbol` rather than just `core-js/stable/index` and `core-js/es/symbol/index`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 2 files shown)


knu:
I had to manually add a stub declaration for `core-js/stable` because the stub file `flow-typed/npm/core-js_vx.x.x.js` which flow-typed generated did not include it.

I'd appreciate someone could help me add necessary tests if this change looks okay.

AndrewSouthpaw:
cc @villesau or @GAntoine

knu:
Added a CHANGELOG entry for this.

PR #3432, opened 2019-07-09
[jest_v24.x.x][jest-dom_v4.x.x] Update definitions

pascalduez:
- Links to documentation: https://github.com/testing-library/jest-dom/blob/master/extend-expect.d.ts
- Link to GitHub or NPM: https://github.com/testing-library/jest-dom


AndrewSouthpaw:
I'm assuming the build is dead due to already bad tests. If you're willing to take a stab at fixing those that'd be amaaaaaazing.

PR #3434, opened 2019-07-09
[just-safe-get_v1.3.x] Add new type definitions

AlexMSmithCA:
- Links to documentation: [README.md](https://github.com/angus-c/just/blob/master/packages/object-safe-get/README.md)
- Link to GitHub or NPM: [GitHub](https://github.com/angus-c/just) | [NPM](https://www.npmjs.com/package/just-safe-get)
- Type of contribution: **new definition** | addition | fix | refactor

### DIFF
--- a/definitions/npm/just-safe-get_v1.3.x/flow_v0.98.x-/just-safe-get_v1.3.x.js
+++ b/definitions/npm/just-safe-get_v1.3.x/flow_v0.98.x-/just-safe-get_v1.3.x.js
@@ -1,0 +1,11 @@
+declare module 'just-safe-get' {
+  declare export default function get(
+    obj: mixed,
+    key: string | Array<string>
+  ): mixed;
+
+  declare export default function get<TReturn>(
+    obj: mixed,
+    key: string | Array<string>
+  ): TReturn;
+}
--- a/definitions/npm/just-safe-get_v1.3.x/test_just-safe-get_v1.3.x.js
+++ b/definitions/npm/just-safe-get_v1.3.x/test_just-safe-get_v1.3.x.js
@@ -1,0 +1,59 @@
+// @flow
+
+import { describe, it } from 'flow-typed-test';
+import get from 'just-safe-get';
+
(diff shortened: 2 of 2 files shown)


definitions/npm/just-safe-get_v1.3.x/flow_v0.98.x-/just-safe-get_v1.3.x.js
```diff
@@ -0,0 +1,11 @@
...
+  declare export default function get(
+    obj: mixed,
+    key: string | Array<string>
+  ): mixed;
+
+  declare export default function get<TReturn>(
+    obj: mixed,
+    key: string | Array<string>
+  ): TReturn;
+}
```
AndrewSouthpaw: This comes up often in other libraries, whether to offer `mixed` or `any` or your strategy of a parameterized generic.

I fall on the "keep it useful and call it `any`" so you're not doing a bunch of type refinements all over the place. Something as low-level as `get` may get very frustrating to use if you have to provide the type always.

Just something to consider. I'll merge it in as-is because I have no vested interest in the usage of this particular libdef. 😉

AlexMSmithCA: Yeah, that's a reasonable point.  It isn't always clear where to fall on the trade-off between soundness and ease-of-use.

I'm partial to the parameterized generic as it defers the concrete typing to the consumer without necessitating any other type refinements.  But if that isn't a standard way of handling this, it may throw some folks off.

AndrewSouthpaw: Yeah it's a long-standing disagreement within the flow community. :-)

I don't think there's really a standard, though I would argue I see more cases of type looseness than correctness. That's at least the case for libraries like `axios` and `ramda`.

Shrug. There's no right answer. You do what's good for you, and if people dislike they can argue. 😆

PR #3451, opened 2019-07-14
feat(chore): create issue template

jamesgeorge007:
Related #3413
Follow up of #3418

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -10,5 +10,4 @@
 **If this is a feature request, what is motivation or use case for changing the behavior?**
 
 **Local Environment Information**
-Is this issue specific to the CLI you may provide the results of `npx envinfo`
-<!-- GitHub gist is the intended medium  -->
+If this issue is specific to the CLI, provide the results of running `npx envinfo` from the command line in a GitHub gist.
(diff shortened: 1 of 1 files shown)


jamesgeorge007:
cc @AndrewSouthpaw

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+
+**If the current behavior is a bug, please provide the steps to reproduce.**
+<!-- A great way to do this is to provide your configuration via a GitHub gist. -->
+
+**What is the expected behavior?**
+
+**If this is a feature request, what is motivation or use case for changing the behavior?**
+
+**Local Environment Information**
+Paste the results of `npx envinfo` here
```
AndrewSouthpaw: I don't think it's always required... Maybe say something around, "if this issue has to do with the CLI" include the results. And maybe encourage pasting it in a gist rather than polluting the issues?

jamesgeorge007:
@AndrewSouthpaw Just refactored based on my assumptions :thinking:

pascalduez:
I'm discovering this template [in action](https://github.com/flow-typed/flow-typed/issues/3463), nice !

However, looking at the output of `npx envinfo`, as previously said, most of the exposed data is useless. Whereas the infos we would need are `flow-typed -v`, `flow -v`.

jamesgeorge007:
@pascalduez I backed the dedicated `info` command just with this intend so as to show up only the necessary information.

PR #3471, opened 2019-07-24
[numeral_v2.x.x] Describe modules with predefined locales

shau-kote:
- Links to documentation: [numeraljs.com](http://numeraljs.com/) (doesn't describe it)
- Link to GitHub or NPM: [github.com/adamwdraper/Numeral-js](https://github.com/adamwdraper/Numeral-js)

Other notes:

Numeral comes with bunch of "side-effect only" modules which register predefined locales. 
One cannot use `numeral.locale('pt-br');` without `import 'numeral/locales'` before it (or `import 'numeral/locales/pt-br'`).
There is some info about it in [this issue](https://github.com/adamwdraper/Numeral-js/issues/517).

### DIFF
--- a/definitions/npm/numeral_v2.x.x/flow_v0.25.x-/numeral_v2.x.x.js
+++ b/definitions/npm/numeral_v2.x.x/flow_v0.25.x-/numeral_v2.x.x.js
@@ -69,3 +69,144 @@ declare var numeral: Numeral;
 declare module 'numeral' {
     declare module.exports: typeof numeral
 }
+
+
+declare module 'numeral/locales' {
+  declare module.exports: void
+}
+
+declare module 'numeral/locales/bg' {
+  declare module.exports: void
+}
+
+declare module 'numeral/locales/chs' {
+  declare module.exports: void
+}
+
+declare module 'numeral/locales/cs' {
+  declare module.exports: void
+}
+
+declare module 'numeral/locales/da-dk' {
+  declare module.exports: void
+}
+
+declare module 'numeral/locales/de-ch' {
+  declare module.exports: void
+}
+
+declare module 'numeral/locales/de' {
+  declare module.exports: void
+}
+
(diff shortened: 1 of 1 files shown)


definitions/npm/numeral_v2.x.x/flow_v0.25.x-/numeral_v2.x.x.js
```diff
@@ -69,3 +69,144 @@ declare var numeral: Numeral;
 declare module 'numeral' {
     declare module.exports: typeof numeral
 }
+
+
+declare module 'numeral/locales' {
+  declare module.exports: void
```
pascalduez: Instead of `void` shouldn't this be `$PropertyType<$Exports<'numeral'>, 'locales'>`?

definitions/npm/numeral_v2.x.x/flow_v0.25.x-/numeral_v2.x.x.js
```diff
@@ -69,3 +69,144 @@ declare var numeral: Numeral;
 declare module 'numeral' {
     declare module.exports: typeof numeral
```
goodmind: Where is types?

definitions/npm/numeral_v2.x.x/flow_v0.25.x-/numeral_v2.x.x.js
shau-kote: I don't think so. 
Locale modules don't export anything, they just register their locale settings via `numeral.register()`. That's why I call them "side-effect only modules".
You can see it in [one of the locale modules](https://github.com/adamwdraper/Numeral-js/blob/master/src/locales/pt-br.js).

pascalduez: All right, thanks.

PR #3476, opened 2019-07-27
[ramda] improve definitions

moroine:
Links to documentation: https://ramdajs.com/docs/
Link to GitHub or NPM: https://github.com/ramda/ramda
Type of contribution: FIX

**Change:**
- Supports placeholder `__` in curried functions
- Add `then` definition
- Support `$ReadOnlyArray` in `groupBy`, `concat`, `without`,
- Improve `lensPath` & `over` defs
- Improve `replace` def


AndrewSouthpaw:
cc @LoganBarnett

definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_object.js
```diff
@@ -100,9 +114,21 @@ describe("Object", () => {
 
     it('requires the lens works with the functor provided (object)', () => {
       type Data = { a: string }
+      type NestedData = { a: { b: { c: { d: { e: string } } } } };
       const data: Data = {
         a: 'foo',
       }
+      const nestedData: NestedData = {
```
LoganBarnett: Was this building up to a test case? I don't see it used anywhere.

LoganBarnett:
These are some huge improvements. Thanks @moroine!! This looks good to me.

Some thoughts that I don't believe should hold up this PR:
- [Ramda's placeholder](https://ramdajs.com/docs/#__) allows multiple uses of itself in a curried function, per the example on the link. This pull request adds solid support. It's not exhaustive yet and maybe something we can add down the road.
- `lensPath` goes out to 5 args, which should cover a wide spread of cases. The other (merged) lens PR has some proposals to Flow itself, which might make the definitions more scalable. I think it's quite a ways away from being anything we can use. Even if we do, we support older versions of Flow.

definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_object.js
moroine: Oh, yes was to test `lensPath`. I'll do the change

moroine:
@LoganBarnett 

> * [Ramda's placeholder](https://ramdajs.com/docs/#__) allows multiple uses of itself in a curried function, per the example on the link. This pull request adds solid support. It's not exhaustive yet and maybe something we can add down the road.

Yeah, I forgot about it. But that's a lot of cases! Also on my project, I start to have some flowjs recursion limit and I had to give some hint like (with much more complex types):

```javascript
const n: number = reduce<number, number>(max, 0, [1, 2, 3])
```

I guess that's because, for any object, flowJs has to consider eventual placeholder... I would love to have kind of `$Not<>` helper (I'll try to implement it, but I have to learn Ocaml first ^^)

```javascript
  declare type __CurriedFunction2<A: $Not<$npm$ramda$Placeholder>, B: $Not<$npm$ramda$Placeholder>, R, AA: A, BB: B> = (
    ((...r: [AA]) => CurriedFunction1<BB, R>) &
    ((...r: [$npm$ramda$Placeholder, BB]) => CurriedFunction1<AA, R>) &
    ((...r: [AA, BB]) => R)
  );
```

> * `lensPath` goes out to 5 args, which should cover a wide spread of cases. The other (merged) lens PR has some proposals to Flow itself, which might make the definitions more scalable. I think it's quite a ways away from being anything we can use. Even if we do, we support older versions of Flow.

Actually, changes made for `lensPath` are here to "fix" new errors introduces after https://github.com/flow-typed/flow-typed/pull/3425

definitions/npm/ramda_v0.26.x/flow_v0.76.x-/test_ramda_v0.26.x_object.js
moroine: Updated

LoganBarnett:
> Actually, changes made for lensPath are here to "fix" new errors introduces after #3425

My apologies - I think my phrasing was unclear earlier. I feel like your `lensPath` change are a huge win. While it goes out to 5 layers deep I think that's plenty, and we have a pattern we can duplicate if we need to expand it :) My interpretation of the old `lensPath` was that it was incorrect and broken in a different way. I'd also spent a rather embarrassing amount of time wrapping my head around lenses and the Ramda libdefs for them. I needed to wrap up what I had and iterate on `lensPath` later (or in this case you got to it first!).

LoganBarnett:
@AndrewSouthpaw LGTM :)

moroine:
@LoganBarnett fixed merge conflicts and applied changes for flow v104

LoganBarnett:
Looks like there's a build failure. At first blush it appears to be unrelated to these changes - likely something that was brought in from the merge.

So far I think everything looks good. Github's view post-merge is very strange. It's like the diff still targets the old HEAD of master, and shows the changes brought in from the merge as well. Is that normal?

moroine:
@LoganBarnett when I merged master into the branch, master branch was failing https://travis-ci.org/flow-typed/flow-typed/builds/566203127

I just updated the branch

LoganBarnett:
Looks clean! @AndrewSouthpaw when time permits, I think this is good to merge if you have no other concerns.

moroine:
@LoganBarnett @AndrewSouthpaw any update?

LoganBarnett:
@villesau I think Andrew is out on (I assume a much deserved) vacation. I get looped into a lot of the Ramda PRs and I gave it my stamp of approval. Could we merge it please? I'm happy to help address any issues that might arise when Andrew returns :)

Thanks!

PR #3494, opened 2019-08-05
[backbone_v1.x.x] Fix tests errors

retyui:
Because in [flow 0.88](https://github.com/facebook/flow/blob/master/Changelog.md#0880)
 - Object and Function is any


villesau:
Why `backbone_v1.x.x/flow_v0.104.x-/backbone_v1.x.x.js` is deleted?

retyui:
To support flow 0.88+ need a lot of effort

PR #3501, opened 2019-08-08
lodash: More methods that take $ReadOnlyArray instead of Array

cakoose:
These methods do not modify the input array.


villesau:
Nice, thanks for the contribution @cakoose! :)

PR #3506, opened 2019-08-10
Update react-dnd_v2.x.x.js to avoid $Supertype

amcvitty:
While upgrading flow through 0.98, which turns on deprecated-utility by default I came across this use of the Supertype utility. 

Substituting in $Shape seems to work well

- Links to documentation:
- Link to GitHub or NPM: 
- Type of contribution:  fix

### DIFF
--- a/definitions/npm/react-dnd_v2.x.x/flow_v0.104.x-/react-dnd_v2.x.x.js
+++ b/definitions/npm/react-dnd_v2.x.x/flow_v0.104.x-/react-dnd_v2.x.x.js
@@ -116,7 +116,7 @@ declare module "react-dnd" {
     spec: DragSourceSpec<OP>,
     collect: DragSourceCollector<CP>,
     options?: DndOptions<OP>
-  ): Connector<$Supertype<OP & CP>, CP>;
+  ): Connector<$Shape<OP & CP>, CP>;
 
   // Drop Target
   // ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


retyui:
> Substituting in $Shape seems to work well

the tests don't agree with you)

goodmind:
$Shape sucks, I would rather use deprecated features, and deprecate $Shape instead

amcvitty:
So - I did run the tests without any changes and they failed, so while this
doesn't fix the flow-typed tests, it doesn't cause breakage. I appreciate
that might not be the standard required to incorporate this change for
everyone, but I hope the info is helpful to anyone who needs something to
make their flow build work.

I'm afraid I don't really have the knowledge of the react-dnd library or
flow to be able to do better here

amcvitty:
I'll not bother taking this further

AndrewSouthpaw:
Oops, I seem to have missed this. I was on vacation during this month.

The issue appears to be that the libdef was not corrected for 0.53.x-v0.103.x. If that can be fixed, we'd be in better shape.

valscion:
> The issue appears to be that the libdef was not corrected for 0.53.x-v0.103.x. If that can be fixed, we'd be in better shape.

Seems like [the CI run](https://travis-ci.org/flow-typed/flow-typed/jobs/570192141) shows that this change is not enough to make the existing tests pass, so this is not only a problem with libdefs for older versions:

Job__7911_1_-_flow-typed_flow-typed_-_Travis_CI

I'm taking a look at the `react-dnd` types right now

valscion:
Argh I've been struggling for the last hour to get to even run this change locally but I can't seem to get ANY version of flow to be installed and I don't have time to debug this any further what is happening.

PR #3515, opened 2019-08-16
static attrs to html element

Gaushao:
can i do it?


goodmind:
`Object` is just `any` and `$Shape` is too dangerous

PR #3516, opened 2019-08-16
Add missing properties for Subscription

aramk:
Aligned the definition for RxJS v5.

- Links to documentation: https://rxjs-dev.firebaseapp.com/api/index/class/Subscription
- Link to GitHub or NPM: 
- Type of contribution: fix

### DIFF
--- a/definitions/npm/rxjs_v5.0.x/flow_v0.34.x-v0.103.x/rxjs_v5.0.x.js
+++ b/definitions/npm/rxjs_v5.0.x/flow_v0.34.x-v0.103.x/rxjs_v5.0.x.js
@@ -1523,9 +1523,13 @@ declare class rxjs$ReplaySubject<T> extends rxjs$Subject<T> {
   ): void;
 }
 
-declare class rxjs$Subscription {
+declare class rxjs$Subscription {       
+  static EMPTY: rxjs$Subscription;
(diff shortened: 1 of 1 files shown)


definitions/npm/rxjs_v5.0.x/flow_v0.34.x-v0.103.x/rxjs_v5.0.x.js
```diff
@@ -1523,9 +1523,13 @@ declare class rxjs$ReplaySubject<T> extends rxjs$Subject<T> {
   ): void;
 }
 
-declare class rxjs$Subscription {
+declare class rxjs$Subscription {       
+  static EMPTY: rxjs$Subscription;
+  closed: Object;
```
pascalduez: `Object` is now an alias to `any`. 
Also looks like the doc is erroneous https://github.com/ReactiveX/rxjs/blob/ffc4e688fa83cb44bff7e5deacc4063f8ea1e2fa/src/internal/Subscription.ts#L28

aramk: Thanks for pointing that out, committed a change.

PR #3522, opened 2019-08-21
[enzyme_v3.x.x] Enable static rendering of arbitrary `React$Node`s

wchargin:
Summary:
The `render` function currently accepts a `React.Component<T>`, but can
actually accept an arbitrary React node, which is sometimes necessary.
This is true of `shallow` and `mount` as well, but the change is easiest
for `render` because there is no component type information to thread
through to the return type.

See discussion:
<https://github.com/flow-typed/flow-typed/pull/3107#issuecomment-523328443>

Test Plan:
Updated tests; the new test fails before this change and passes after
it.

wchargin-branch: enzyme-render-node


HsuTing:
@pascalduez Yes. I think I made this bug accidentally in #3107. Thank @wchargin

wchargin:
Thank you!

PR #3531, opened 2019-08-23
Adds some minor clarifications to Contribution guide

deiga:
- Links to documentation:
- Link to GitHub or NPM: 

Other notes:


goodmind:
Why change to 0.80?

villesau:
@goodmind we do not actually run tests for older versions so they are not supported, only served as is.

PR #3536, opened 2019-08-29
Lodash: Changes to XOR that take $ReadOnlyArray instead of Array

anu365:


jedwards1211:
Looks like there are other methods we should make the same change to

anu365:
yeah, we can do that.. i would raise a new PR soon.

PR #3537, opened 2019-08-29
fix git package support; fixes #3363

### LINKED ISSUE #3363: Flow-typed install crashes if you have a package installed from a git URL
Here is the crash:
```
UNCAUGHT ERROR: Error: Malformed npm package name! Expected the name to be formatted as <PKGNAME>_v<MAJOR>.<MINOR>.<PATCH> but instead got mtwins
    at parsePkgNameVer (/Users/noah.allen/.config/yarn/global/node_modules/flow-typed/dist/lib/npm/npmLibDefs.js:295:11)
    at _callee6$ (/Users/noah.allen/.config/yarn/global/node_modules/flow-typed/dist/lib/npm/npmLibDefs.js:620:38)
    at tryCatch (/Users/noah.allen/.config/yarn/global/node_modules/regenerator-runtime/runtime.js:45:40)
    at Generator.invoke [as _invoke] (/Users/noah.allen/.config/yarn/global/node_modules/regenerator-runtime/runtime.js:271:22)
    at Generator.prototype.(anonymous function) [as next] (/Users/noah.allen/.config/yarn/global/node_modules/regenerator-runtime/runtime.js:97:21)
    at asyncGeneratorStep (/Users/noah.allen/.config/yarn/global/node_modules/flow-typed/dist/lib/npm/npmLibDefs.js:56:103)
    at _next (/Users/noah.allen/.config/yarn/global/node_modules/flow-typed/dist/lib/npm/npmLibDefs.js:58:194)
    at <anonymous>
```

As you can see, it got "mtwins" instead of what it was expecting. This matches a package I have linked through git: (note, I do have several more dependencies that I did not list here)
```json
  "dependencies": {
    "lottie-react-native": "^2.6.1",
    "rn-tooltip": "git+https://github.com/mtwins/rn-tooltip.git"
  },
```

This happens when running:
-  `flow-typed install lottie-react-native` (a package I recently installed).
- `flow-typed install`

To get around this, I have to:
1. `rm -rf flow-typed`
2. `flow-typed install`. This works successfully, unlike before deleting the folder

kirillDanshin:
Signed-off-by: Kirill Danshin <kirill@danshin.pro>

- Links to documentation: –
- Link to GitHub or NPM: –
- Type of contribution: fix

Other notes:

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -54,7 +54,8 @@
     "eslint-plugin-flowtype": "^2.50.3",
     "eslint-plugin-prettier": "^2.7.0",
     "flow-bin": "^0.81.0",
-    "jest": "^23.6.0"
+    "jest": "^23.6.0",
+    "yarn": "^1.17.3"
   },
   "keywords": [
     "flow",
--- a/cli/src/lib/npm/__tests__/npmLibDefs-test.js
+++ b/cli/src/lib/npm/__tests__/npmLibDefs-test.js
@@ -393,6 +393,7 @@ describe('npmLibDefs', () => {
           major: 0,
           minor: 0,
           patch: 0,
+          prerel: '',
         },
       });
     });
--- a/cli/src/lib/npm/__tests__/npmLibDefs-test.js
+++ b/cli/src/lib/npm/__tests__/npmLibDefs-test.js
@@ -386,6 +386,17 @@ describe('npmLibDefs', () => {
       });
     });
 
+    it('parses libs installed from git', () => {
+      expect(parsePkgNameVer('git@github.com:flow-typed/flow-typed.git')).toEqual({
+        pkgName: 'flow-typed',
+        pkgVersion: {
+          major: 0,
+          minor: 0,
+          patch: 0,
+        },
+      });
+    });
+
     it('parses wildcard minor libs', () => {
       expect(parsePkgNameVer('lib_v1.x.x')).toEqual({
         pkgName: 'lib',
--- a/cli/src/lib/npm/npmLibDefs.js
+++ b/cli/src/lib/npm/npmLibDefs.js
@@ -187,8 +187,18 @@ async function getCacheNpmLibDefs(cacheExpiry) {
 }
 
 const PKG_NAMEVER_RE = /^(.*)_v\^?([0-9]+)\.([0-9]+|x)\.([0-9]+|x)(-.*)?$/;
+const PKG_GIT_RE = /^([\w\-]+)@([\w\.]+):([\w\-]+)\/([\w\-]+)(?:\.git)$/;
 function parsePkgNameVer(pkgNameVer: string) {
   const pkgNameVerMatches = pkgNameVer.match(PKG_NAMEVER_RE);
+  const pkgNameGitMatches = pkgNameVer.match(PKG_GIT_RE);
+
+  if (pkgNameGitMatches != null) {
+    return {
+      pkgName: pkgNameGitMatches[4],
+      pkgVersion: {major: 0, minor: 0, patch: 0, prerel: ''},
+    };
+  }
+
   if (pkgNameVerMatches == null) {
     throw new ValidationError(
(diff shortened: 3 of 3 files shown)


kirillDanshin:
build failed due to `sudo -E apt-add-repository -y "ppa:ubuntu-toolchain-r/test` error which is totally unrelated to this PR.

kirillDanshin:
@mars-lan as you can see, I already did

AndrewSouthpaw:
Sorry for the delay! I've re-run the tests and it looks like it needs to be tweaked before the build can pass. https://travis-ci.org/flow-typed/flow-typed/jobs/578422641

kirillDanshin:
@AndrewSouthpaw looks like I've fixed the tests

cli/package.json
```diff
@@ -54,7 +54,8 @@
     "eslint-plugin-flowtype": "^2.50.3",
     "eslint-plugin-prettier": "^2.7.0",
     "flow-bin": "^0.81.0",
-    "jest": "^23.6.0"
+    "jest": "^23.6.0",
+    "yarn": "^1.17.3"
```
serhalp: Random passerby comment: Is this intentional? Why would you ever need to install yarn as a dependency (installed via itself), unless your module was using it programmatically (which doesn't appear to be the case)? Looks like an accident?

kirillDanshin: it is intentional and allows you to still use npm scripts (see https://github.com/flow-typed/flow-typed/blob/master/cli/package.json#L16-L25), which for unknown reason depend on yarn, when you don't have global yarn installed.

serhalp: > for unknown reason depend on yarn

This repo uses yarn for development. It has `yarn.lock` files. It [uses `yarn` in the script you're supposed to run to contribute](https://github.com/flow-typed/flow-typed/blob/e6c6a9001fe8627c89a7e363f03f5063b7b153d6/build_and_test_cli.sh#L5-L8) ([reference in CONTRIBUTING.md](https://github.com/flow-typed/flow-typed/blob/e6c6a9001fe8627c89a7e363f03f5063b7b153d6/CONTRIBUTING.md#running-libdef-tests)). By using `npm install` you're bypassing the `yarn.lock`.

Admittedly CONTRIBUTING.md should explicitly mention that this repo uses yarn and you should have it globally installed.

Again I'm just a random passerby who was following an issue, but installing yarn as a local dependency unnecessarily is asking for trouble. Most people will end up with two different versions of yarn. It also fundamentally doesn't make sense, because you need yarn to install yarn this way, if you're doing things correctly, at which point a yarn dev dependency clearly isn't necessary.

kirillDanshin: @serhalp actually, there is no reference of word "yarn" in the whole CONTRIBUTING.md you referenced.

image

the only place in docs about yarn is the README.md, where yarn is used in the example to install flow-typed.

There's no reason why yarn should be used to run npm scripts (see the reference in my previous message). there's no docs stating that you must have `yarn` globally installed to run `npm run test-quick`, so it may confuse new contributors and be somewhat unpleasant.  this package.json change will not install yarn globally, but it ensures that if you don't use yarn, any npm script still should work for you.

serhalp: > actually, there is no reference of word "yarn" in the whole CONTRIBUTING.md you referenced.

That's actually exactly what I said:
> Admittedly CONTRIBUTING.md should explicitly mention that this repo uses yarn and that you should have it globally installed.

It's clearly just missing from there. It's indirectly referenced rather than directly.

I don't know how to explain this any better without repeating myself. There are multiple package managers out there. They are not compatible with each other. All developers on a repo must use the same, blessed package manager. This repo uses yarn, so it has a `yarn.lock`, its `package.json` scripts use `yarn run`, etc. Developers should _not_ used npm. The only issue is this is poorly documented. Specifying `yarn` as a dev dependency which gets installed by `npm` for some `yarn` for others is _not a good practice_. This is hardly subjective.

PR #3546, opened 2019-09-02
Fix @reach/router `Link` base type

edoroshenko:
- Type of contribution: fix

Other notes: This change fixes the problem when `flow@1.106.x` expects all props of `HTMLAnchorElement` to be present in a `Link` instance.

### DIFF
--- a/definitions/npm/@reach/router_v1.1.x/flow_v0.104.x-/router_v1.1.x.js
+++ b/definitions/npm/@reach/router_v1.1.x/flow_v0.104.x-/router_v1.1.x.js
@@ -71,7 +71,7 @@ declare module '@reach/router' {
   |}>;
 
   declare export type LinkProps<State> = {
-    ...$Shape<HTMLAnchorElement>,
+    ...React$ElementProps<string>,
     children: React$Node,
     innerRef?: React$Ref<HTMLAnchorElement>,
     getProps?: (props: {


definitions/npm/@reach/router_v1.1.x/flow_v0.104.x-/router_v1.1.x.js
```diff
@@ -71,7 +71,7 @@ declare module '@reach/router' {
   |}>;
 
   declare export type LinkProps<State> = {
-    ...$Shape<HTMLAnchorElement>,
+    ...React$ElementProps<string>,
```
Hypnosphi: Why not `...React$ElementProps<a>`?

Hypnosphi: `@reach/router` spreads `Link` props over `<a/>` element: https://github.com/reach/router/blob/master/src/index.js#L385-L398

edoroshenko: My thought was to make it more generic. I could imagine that `Link` component would render another element like `span`.

Hypnosphi: `@reach/router` is very accessibility-centric, so I don't think they will ever allow using spans as links

Hypnosphi: They actually even advertise using native links as a feature =)
https://reach.tech/router/accessibility

bluie1983:
These people are stealing my personal information

pascalduez:
Doesn't feel like the right fix. 

`React$ElementProps<string>` nor `React$ElementProps<'a'>` does not bring all the same validation.

[Current issue reproduction](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBdkwBDAZzACUBTUgY2KnzizAHJ87H30ATGvRikuYIRUoAJACoBZADIBBAHb0AFgQCiMGlhrLiAb1RgxakeRqEAXGHKF8GZQHMA3CbFwCfcrfuOXd1M+RGV4Uj4-Byc3D3NyNSiA2NMNeySYoLA0wmVSPQzAuK4oQpTskuEXMqy9PgxSGo88grtootMCDGcnJtMcCQRvPrABwjUWmhG8ImmWQjh6OBgRrhW25KyuADcRyxF1PfMcKY3Mj0IRZysRwhoADxszjrBCAE8TkYBXS3xJpoAvuh3icwAAFFg4SgAXjAxlMADokQASADKxxoAB4ZAoVOotDo9AYAHwAGlQQNMqCgX1UhAwcGUYGkNHsAAocJDfOCuQBKOEeLiEL5-MCY0hwpEIzlwKEAsDAYlUoGoYDAMDIzT3E6MTT4Fj4WxqEp2DRfGB8MAAIxoz2cqExLPsFRoUGhhgAjAAmADM8sVQA). 

`  ...$Shape<HTMLAnchorElement>` is quite weird to be honest.

edoroshenko:
@pascalduez what would be the right one?

Hypnosphi:
> `React$ElementProps<string>` nor `React$ElementProps<'a'>` does not bring all the same validation.

Oh I see. Looks like flow doesn't validate DOM props at all:

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBdkwBDAZzACUBTUgY2KnzizAHJ87H31hgwAEgCiADxw1Gw-C3wAuMAAsuUMOUVwArjAAmYAEY01hfBgB2Ac1QAeUkpUBeAN4BGAEwBmAL5hgAPlQgA

pascalduez:
> what would be the right one?

I guess `React$ElementProps<'a'>` is indeed what we want, the issue with merging now is we loose the DOM props validation. But this is broken anyway from Flow 0.106.x so...

pascalduez:
What's missing to be able to move here, I guess adding tests to assert the proper intended behaviour?

PR #3552, opened 2019-09-05
[webpack_v4.x.x] fix webpack type error with WebpackError

HsuTing:
- Links to documentation: https://webpack.js.org/api/node/#error-handling
- Type of contribution: fix

### DIFF
--- a/definitions/npm/webpack_v4.x.x/flow_v0.104.x-/webpack_v4.x.x.js
+++ b/definitions/npm/webpack_v4.x.x/flow_v0.104.x-/webpack_v4.x.x.js
@@ -5,7 +5,10 @@ declare module 'webpack' {
   declare class WebpackError extends Error {
     constructor(message: string): WebpackError;
     inspect(): string;
+    details: string;
   }
+
+  declare type webpackError = WebpackError;
 
   declare interface Stats {
     hasErrors(): boolean;
--- a/definitions/npm/webpack_v4.x.x/flow_v0.104.x-/webpack_v4.x.x.js
+++ b/definitions/npm/webpack_v4.x.x/flow_v0.104.x-/webpack_v4.x.x.js
@@ -2,13 +2,13 @@ import * as http from 'http';
 import fs from 'fs';
 
 declare module 'webpack' {
-  declare class WebpackError extends Error {
+  declare class $WebpackError extends Error {
     constructor(message: string): WebpackError;
     inspect(): string;
     details: string;
   }
 
-  declare type webpackError = WebpackError;
+  declare type WebpackError = $WebpackError;
 
   declare interface Stats {
     hasErrors(): boolean;
--- a/definitions/npm/webpack_v4.x.x/flow_v0.71.x-v0.103.x/webpack_v4.x.x.js
+++ b/definitions/npm/webpack_v4.x.x/flow_v0.71.x-v0.103.x/webpack_v4.x.x.js
@@ -5,7 +5,10 @@ declare module 'webpack' {
   declare class WebpackError extends Error {
     constructor(message: string): WebpackError;
     inspect(): string;
+    details: string;
   }
+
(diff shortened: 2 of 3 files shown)


definitions/npm/webpack_v4.x.x/flow_v0.104.x-/webpack_v4.x.x.js
```diff
@@ -5,8 +5,11 @@ declare module 'webpack' {
   declare class WebpackError extends Error {
     constructor(message: string): WebpackError;
     inspect(): string;
+    details: string;
   }
 
+  declare type webpackError = WebpackError;
```
pascalduez: Why the alias? Also types should really be capitalised.

HsuTing: Like the [testing](https://github.com/flow-typed/flow-typed/pull/3552/files#diff-d079baff9a1264ee7406ddb1ee7f83a7R49-R67), I want to use the `webpack` compiler. If we do not export `webpackError`, we should write the `webpackError type` every time.

pascalduez: Why not importing the `WebpackError` class type and using it? 

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVATApgYxgQwCcsw98BncsAdSwCMAHfHAawFFDC5CwsAPAC5YAdhiocuPAN6owpOMPIDCAVxwDuACgC2WSvgDmWAFxglhAJbCDASlO1Gzdp24BuWWCvkGuAZrtmylYG7nLYAvgWMOSm5sHuAL4emLgExGBQKsLqFgpgQkqafPb0TKwS3AEAbnAWGOg4Ckq8nGAAvGDCWEgOZc6SmgDkgzboBX5YnKNAA

pascalduez: Another solution: 
```
declare class $WebpackError {}
declare type WebpackError = $WebpackError
```

HsuTing: > Why about importing the `WebpackError` class with `typeof` and using it?

I had tried this:

```js
import webpack, {
  typeof WebpackError as webpackErrorType,
  type Stats as statsType,
} from 'webpack';
```

But I got the error:

```js
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ server/koa-react/src/utils/buildJs.js:5:10

Cannot import WebpackError because there is no WebpackError export in webpack.

     2│
     3│ import debug from 'debug';
     4│ import webpack, {
     5│   typeof WebpackError as webpackErrorType,
     6│   type Stats as statsType,
     7│ } from 'webpack';
     8│
```

I am not sure why we can not import `WebpackError`. Maybe `class` can not be exported in the `declare module`? If we can import `WebpackError` directly, that will be the best solution.

HsuTing: > Another solution:
> 
> ```
> declare class $WebpackError {}
> declare type WebpackError = $WebpackError
> ```

Done

pascalduez: `import typeof WebpackError from 'webpack'` worked for me. 
But probably the `$` solution is the easiest.

HsuTing:
Thanks.

PR #3564, opened 2019-09-19
[hoist-non-react-statics v3.x.x] Add defs

d4wei:
- Link to GitHub or NPM: https://github.com/mridgway/hoist-non-react-statics

Other notes:
The `v2.x.x` definitions for this package were very loose and I tried as much as possible to tighten things up in this updated version, taking inspiration from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/hoist-non-react-statics/index.d.ts.

What I was able to do:
- Hoisted statics are no longer just a blanket `$Shape` call over source component statics and are much closer to their actual shape, providing more meaningful typechecks.

What I was not able to do and why:
- React statics that may be present but not always (depending on whether the component has been `memo`'ed or `forwardRef`'ed) are not taken into account with precision like everything else, because we can't actually tell if a component has gone thru `memo` or `forwardRef` calls with our current defs for them. Let me know if you know a way to tell!
- Results from functional components still cannot be typechecked in any meaningful way, just like in `v2.x.x` 😢 I spent some time playing with extracting statics from functions, ran into a lot of quirks. Hopefully I can circle back to this later and find a way (or find things magically fixed), but for now, I don't think this should block everything else. I'm not making anything worse, after all 😆

### DIFF
--- a/definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
+++ b/definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
@@ -1,0 +1,86 @@
+declare module 'hoist-non-react-statics' {
+  /**
+   * Inspired by DefinitelyTyped/types/hoist-non-react-statics/index.d.ts
+   *
+   * Unfortunately, unlike in TypeScript, current flow definitions for React does not allow us to tell whether
+   * a React$ComponentType is “clean”, or had been the result of a call to React.memo or React.forwardRef.
+   * Therefore we’ll only be able to precisely filter out statics that are common to all 3 cases, and will
+   * have to live with maybes for everything else. This is not 100% precise, but is better than a blanket
+   * $Shape call on the source component’s statics like what we were doing before in the v2.x.x defs.
+   */
+
+  // declare type REACT_STATICS = {
+  //   childContextTypes: any,
+  //   contextType: any,
+  //   contextTypes: any,
+  //   defaultProps: any,
+  //   displayName: any,
+  //   getDefaultProps: any,
+  //   getDerivedStateFromError: any,
+  //   getDerivedStateFromProps: any,
+  //   mixins: any,
+  //   propTypes: any,
+  //   type: any,
+  //   ...
+  // };
+
+  // declare type MEMO_STATICS = {
+  //   $$typeof: any,
+  //   compare: any,
+  //   defaultProps: any,
+  //   displayName: any,
+  //   propTypes: any,
+  //   type: any,
+  //   ...
+  // };
+
+  // declare type FORWARD_REF_STATICS = {
+  //   $$typeof: any,
+  //   render: any,
+  //   defaultProps: any,
+  //   displayName: any,
+  //   propTypes: any,
+  //   ...
+  // };
+
+  declare type REACT_STATICS = {
+    // “Maybe” React statics
+    $$typeof?: any,
+    childContextTypes?: any,
+    compare?: any,
+    contextType?: any,
+    contextTypes?: any,
+    getDefaultProps?: any,
+    getDerivedStateFromError?: any,
+    getDerivedStateFromProps?: any,
+    mixins?: any,
+    render?: any,
+    type?: any,
+    // Common React statics
+    defaultProps: any,
+    displayName: any,
+    propTypes: any,
+    ...
+  };
+
+  declare type $HoistedStatics<S, C> = $Call<
+    & ((S, empty) => $Diff<S, REACT_STATICS>)
+    & ((S, any) => $Diff<S, $ObjMap<C, any> & REACT_STATICS>),
+    S,
+    C
+  >;
+
+  /*
+    TP - target component props
+    SP - source component props
+    T - target component statics
+    S - source component statics
+  */
+  declare function hoistNonReactStatics<TP, SP, T, S, C: { [key: string]: true, ... }>(
+    TargetComponent: React$ComponentType<TP> & T,
+    SourceComponent: React$ComponentType<SP> & S,
+    customStatics?: C
+  ): React$ComponentType<TP> & $HoistedStatics<S, C> & T;
+
+  declare module.exports: typeof hoistNonReactStatics;
+}
--- a/definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
+++ b/definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
@@ -80,5 +80,5 @@ declare module 'hoist-non-react-statics' {
     customStatics?: C
   ): React$ComponentType<TP> & $HoistedStatics<S, C> & T;
 
-  declare module.exports: typeof hoistNonReactStatics;
+  declare export default typeof hoistNonReactStatics;
 }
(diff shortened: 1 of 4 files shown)


definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
```diff
@@ -0,0 +1,84 @@
...
+    T - target component statics
+    S - source component statics
+  */
+  declare function hoistNonReactStatics<TP, T, S, C: { [key: string]: true, ... }>(
+    TargetComponent: React$ComponentType<TP> & T,
+    SourceComponent: React$ComponentType<any> & S,
+    customStatics?: C
+  ): React$ComponentType<TP> & $HoistedStatics<S, C> & T;
+
+  declare module.exports: typeof hoistNonReactStatics;
```
pascalduez: I would rather use an esm default export, to match the lib source.

definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
```diff
@@ -0,0 +1,84 @@
...
+    type?: any,
+    // Common React statics
+    defaultProps: any,
+    displayName: any,
+    propTypes: any,
+    ...
+  };
+
+  declare type $HoistedStatics<S, C> = $Call<
+    & (empty => $Diff<S, REACT_STATICS>)
```
jbrown215: Any chance you can use spreads instead of $Diff? Diff has a lot of problems (type-theory-wise) and causes bad error messages.

So instead of removing properties, add them with spread. More inline notes below should help explain what I mean

definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
```diff
@@ -0,0 +1,84 @@
...
+  >;
+
+  /*
+    TP - target component props
+    T - target component statics
+    S - source component statics
+  */
+  declare function hoistNonReactStatics<TP, T, S, C: { [key: string]: true, ... }>(
+    TargetComponent: React$ComponentType<TP> & T,
+    SourceComponent: React$ComponentType<any> & S,
```
jbrown215: Instead of `S` here, `{...S, ...REACT_STATICS}`

Now you don't need to diff them out!

definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
```diff
@@ -0,0 +1,84 @@
...
+    & (any => $Diff<S, $ObjMap<C, any> & REACT_STATICS>),
+    C
+  >;
+
+  /*
+    TP - target component props
+    T - target component statics
+    S - source component statics
+  */
+  declare function hoistNonReactStatics<TP, T, S, C: { [key: string]: true, ... }>(
```
jbrown215: From what I understand from HoistedStatics, you expect `S` to maybe have properties from REACT_STATICS. Instead of removing them, why not spread them and keep around "S without REACT_STATICS"

definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
d4wei: I'm not too familiar with esm default export, could you point me to an example of this? Thanks!

definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
d4wei: Thanks for the suggestion! This syntax does make things a lot cleaner, but I'm having trouble getting it to work 😭: https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AoctAGxQGc64BBOJADxiQDsATR5dDAB0AYVyQu3GAB4A3gB95AXwB8cWeThw6MFDGBo4KOAF4CKMpu279hgEamCdy1oDmSGEwAUASnVwiGABXKC5mIRRSOCVyGKoILh04ACFHLzAcMDoALnVFJT8TNS4g6moKZIjHfAsKoQczfGcKcgB6VrgYAE8wJBTaNABramAksw12rTtclC4ugBo2jqUWye7e-vRh0fhxpancnShgLldFyZXKdb7kge2xvKtp62PTxeUWniQaFCI4TCCXAwwASrA4UEEAGUbAY6NJISovFYRLkBBgACRicAJKQAFR6SGksy6agAZP4hJTIfM4JShLctiMkkpFj5cpCWmgEkkmJDHOwYBCMNC9LCvEwfBQuYl4Mk+WYBUKYCLbHQvMlJZRJgB1CClHhg3oYToACyQdD6MAgRjKrCgOCgcD18BgZq6cB4CXw8GySOlSRQAEZciUQHYkI6zLyIhRo2hY5ChO5PL4pdz4CgAEwhoJhiOOOUx8iF+PFxPJ7yaoA

jbrown215: hrm :\ that's disappointing. this may be due to an issue with polymorphism and type destructors. If it doesn't work, ignore me :)

d4wei: @jbrown215 Indeed 😞These extracted statics definitely don't behave like our daily object types: https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AoctAGxQGc64BBOJADxiQDsATR5dDAB0AYVyQu3GAB4A3gB95AXwB8cWeThw6MFDGBo4KOAF4CKMpu279hgEamCdy1oDmSGEwAUASnVwiGABXKC5mIRRSOCVyGPIeJBoUIjhMIK4MYAgw9hgoQQBlGwM6aQKVLysRAC5EYhgAEjFwbKkAFQBPMCRpFC4OtQAyOAKAGnIfWoKKKmydZgKARkdc-IwivRKvJh8ZpiWIin3FoTsjg7AcGAgYLqQhd09fGbQ5+BRF2q4gkDskKEcx0O5CBaHOJ0e3l2IIuVxudweHgAQs9KAkkik0hl9NlWBw1jANrY6AB1YAwAAWTDaAHdgDoyhUqrUBBgmuJWlwYJ1ur1+kMRuNJnAGgARYCYTBlUbqVQvN4LABMK3xhWKaFJ5KptPpMG20JhiuB+yNZ0NQkuN3h3URTwNry48xQiq+Pz+ALMJuNBSNYPNkNRXst11uNseKIN6NomPSmVxqzVmw1ZMpTC4Nwp-x1DPKlS0NTqgnZLUkXJ5PT6Azgw1kcCE9YK0SFU3ljvg+wAzCq8oniSmqenKVm6Tp9XsCh3vZOzZ2LXDQ-cA-aFSgO67fv9AROp0I-bOl+dJ8HrYvkaigA

definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
d4wei: Resolving this as we are running into trouble getting the suggested syntax to work.

definitions/npm/hoist-non-react-statics_v3.x.x/flow_v0.103.x-/hoist-non-react-statics_v3.x.x.js
d4wei: Resolving this as we are running into trouble getting the suggested syntax to work.

d4wei:
@pascalduez Updated! Could you take a look again? 🙏

PR #3570, opened 2019-09-25
[React-router_v.5.1.x] Add react hook

HsuTing:
- Links to documentation: https://github.com/ReactTraining/react-router/blob/0c058adc5b56c6d9b409af5e4d8330c645fac00c/packages/react-router/docs/api/hooks.md
- Link to GitHub or NPM: https://github.com/ReactTraining/react-router/

Other notes:
After `react-router@5.1.0` and `react-router-dom@5.1.0` were released, `react hook` has been added in those packages. Since `react-router_v5.x.x` and `react-router-dom_v5.x.x` exist. I am not sure how to update the new version to `5.1.x`.


AndrewSouthpaw:
I think you can leave move `5.x.x` to `5.0.x` and then create a `5.1.x` which is the more recent version with hooks. Give that a shot?

HsuTing:
> I think you can leave move `5.x.x` to `5.0.x` and then create a `5.1.x` which is the more recent version with hooks. Give that a shot?

OK! And another question. If `5.2.x` comes out ,should we add the new folder to `5.2.x`?
If we do not add `5.2.x`, and the users use `flow-typed install`. Which version will they get? `5.x.x` or `5.1.x`?

AndrewSouthpaw:
They would get `5.x.x`. We don't have a good story for indicating things like `^5.1.x`, unfortunately.

Or, we could roll everything into a `5.x.x` definition. It might not be technically correct for `5.0.x`, but I think since we're *adding* methods rather than changing anything, I'm not terribly concerned. It's fit to function. And it'll be way less maintenance than bumping the version each time the library gets bumped.

Others are welcome to contradict me if they disagree. 😄

HsuTing:
OK 👍 
I think rolling everything into a `5.x.x` definition is a good choice. I will modify `pr` later.

HsuTing:
@AndrewSouthpaw  I have added `react hook` to `v5.x.x` and removed `v5.1.x`.

PR #3577, opened 2019-09-27
Add moment-timezone types definition

Flex-Madao:
Found out we don't have moment-timezone's type definition.

- Link:  https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/moment-timezone
- Created by flow-gen.


definitions/npm/moment-timezone_v2.x.x/flow_v0.104.x-/moment-timezone_v2.x.x.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * Flowtype definitions for moment-timezone
+ * Generated by Flowgen from a Typescript Definition
+ * Flowgen v1.10.0
+ */
+
+import * as moment from "moment";
+declare module.exports: typeof moment;
```
pascalduez: Those two lines can't work with flow-typed.

definitions/npm/moment-timezone_v2.x.x/flow_v0.104.x-/moment-timezone_v2.x.x.js
```diff
@@ -0,0 +1,77 @@
...
+
+declare module "moment" {
```
pascalduez: Isn't this PR about "moment-timezone"?

definitions/npm/moment-timezone_v2.x.x/flow_v0.104.x-/moment-timezone_v2.x.x.js
pascalduez: If you need types from "moment" you'll need to copy them here. You won't be able to import them.

Flex-Madao: `moment-time` is not an independent library. it is based on `moment`, the usage will like `moment.tz()` or `moment('times-str').tz('TIMEZONE', boolean?)`

Flex-Madao:
@pascalduez will add some test cases later.

definitions/npm/moment-timezone_v2.x.x/flow_v0.104.x-/moment-timezone_v2.x.x.js
pascalduez: Please see https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#dont-import-types-from-other-libdefs

Flex-Madao:
@pascalduez  import already remove.

Flex-Madao:
@pascalduez update import problem now.

definitions/npm/moment-timezone_v2.x.x/flow_v0.104.x-/moment-timezone_v2.x.x.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * Flowtype definitions for moment-timezone
+ * Generated by Flowgen from a Typescript Definition
+ * Flowgen v1.10.0
+ */
+
+import * as moment from "moment-timezone";
```
pascalduez: That's quite confusing.

pascalduez:
Still missing tests.

Brianzchen:
If this definition is to some day get merged, the definition shouldn't be built on flow-gen I believe. It's just way too confusing. 

From what I understand `moment-timezone` is same as moment plus a few more features with `tz()`.

Also the definition should target package version 0.5.x not 2.x.x

PR #3578, opened 2019-09-28
[express] Remove $Subtype usage in express and cookie-parser

Ashoat:
Flow 0.89 and above have a lint that complains about this, and it has been considered "unclear" outside libdefs since 0.72.

### DIFF
--- a/definitions/npm/cookie-parser_v1.x.x/flow_v0.104.x-/cookie-parser_v1.x.x.js
+++ b/definitions/npm/cookie-parser_v1.x.x/flow_v0.104.x-/cookie-parser_v1.x.x.js
@@ -10,28 +10,10 @@ import type { Socket } from 'net';
 declare module 'cookie-parser' {
   /**
    * NOTE:
-   * The following block has all been copied from
-   * https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/express_v4.x.x/flow_v0.93.x-/express_v4.x.x.js#L103-L105
+   * The following block has all been copied from the express libdef
    */
 
   /* -------------------------- 8< ------------------------------------------------------------------ */
-
-  declare type express$RenderCallback = (
-    err: Error | null,
-    html?: string
-  ) => mixed;
-
-  declare type express$CookieOptions = {
-    domain?: string,
-    encode?: (value: string) => string,
-    expires?: Date,
-    httpOnly?: boolean,
-    maxAge?: number,
-    path?: string,
-    secure?: boolean,
-    signed?: boolean,
-    ...
-  };
 
   declare type express$RouterOptions = {
     caseSensitive?: boolean,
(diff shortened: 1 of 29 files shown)


Ashoat:
The CI was testing on `express_v4.x.x` since `express_v4.16.x` was updated. I went ahead and stripped the types from `express_v4.x.x` as well

pascalduez:
Ho @Ashoat,

please check #3337 as first attempt.  
Apparently it's not viable to just remove `$Subtype`.   

Would be nice if you could add the tests cases described in https://github.com/flow-typed/flow-typed/pull/3337#issuecomment-495603799

Ashoat:
I totally missed that whole discussion, thanks for the context. I've updated this PR with a more robust attempt, and detailed my approach [in that thread](https://github.com/flow-typed/flow-typed/pull/3337#issuecomment-537286813).

pascalduez:
Thanks a lot @Ashoat !

@mgreenw, @LoganBarnett I would appreciate your eyes here :)

mgreenw:
@Ashoat Thanks for the update! I think this approach works well. Your comment on #3337 is a solid explanation, and I agree that it makes sense to type the entire `express$Application` to avoid the Flow errors. As I said [here](https://github.com/flow-typed/flow-typed/pull/3337#issuecomment-495603799), Express was not designed with static typing in mind, so this is a good compromise at least to ensure that properties accessed are the correct type and exist theoretically on the `req`/`res`.

One "frustrating" thing about this approach (that I'm not sure is solvable at this point) is that any middleware, even the very first, could use the full extended `req` or `res` type. In the updated tests, the following is used: 

```js
customApp.use('/something', (req: test_express$CustomRequest, res: test_express$CustomResponse, next: express$NextFunction) => {
  req.foo;
  res.bar;
  // $ExpectError
  req.notHere;
  // $ExpectError
  res.notHere;
});
```

However, if this is the very first middleware that is hit, `req.foo` and `res.bar` will be undefined no matter what. In this case, unless I am mistaken, the explicit typing of `req` as `test_express$CustomRequest` is unsound.

Is there any way to mark the additional properties of any custom `req` and `res` to always use the `$Shape<T>` directive? That would ensure that every time they are used, they must first be checked. Otherwise, we are left in the awkward spot similar to how Flow types [array access](https://flow.org/en/docs/types/arrays/#toc-array-access-is-unsafe).


**Thanks again for your awesome work!**

mgreenw:
On a separate note, `$Supertype` and `$Subtype` were fully removed from Flow in `v.0.109.0`, so I fully support moving forward quickly on this to ensure support of up-to-date versions of Flow.

https://github.com/facebook/flow/releases/tag/v0.109.0

LoganBarnett:
This all looks great to me (though don't consider me the resident Express expert - I am but a dabbler). Thanks everyone for the efforts!

Ashoat:
@mgreenw - I like your idea, but I'm having trouble making it work. Normally I'd rely on type spreading to append parameters, but in this case because everything is defined as `declare class` and `declare interface` I don't think it's possible. My next idea was to use the '&' operator to combine `declare class` with a non-exact object, like this:

```js
declare type express$Middleware<
  ExtraReqProps: { ... } = { ... },
  ExtraResProps: { ... } = { ... },
> =
  ((
    req: express$Request & $Shape<ExtraReqProps>,
    res: express$Response & $Shape<ExtraResProps>,
    next: express$NextFunction,
  ) => mixed) |
  ((
    error: Error,
    req: express$Request & $Shape<ExtraReqProps>,
    res: express$Response & $Shape<ExtraResProps>,
    next: express$NextFunction,
  ) => mixed);
```

But that didn't work either. I got some strange error about `express$Router`'s function definition (ie. what happens when use `express.Router` as a function), that got triggered when `express$Router.use` was called in a test. Not sure what's going on there, but I know trying to use the `&` operator to combine properties on an object can be tricky in Flow.

The other thing is that even if I could get that approach to work, it would be a bit icky for the user, as they'd have to type their middlewares like `req: express$Request & $Shape<ExtraRequestProps>`. In an ideal world we could continue parameterizing on the custom request/response objects, but simply `$Diff` out the custom props, throw them into a `$Shape`, and then spread that into the "real" custom response/requests. But that seems like an even taller order than the above approach that I couldn't get to work.

You're certainly welcome to try and give it a go. Would be awesome if you could get it to work.

mgreenw:
@Ashoat Hm I don't know if it's possible within the constraints of Flow at this point.

 I don't have any additional ideas. I think it would be best to document the shortcomings of the current typing solution and suggest that users intentionally use `$Shape<>` themselves if they want safety, or don't use it if they are OK with that unsound behavior.

Where would such documentation live?

LoganBarnett:
@mgreenw I've been putting these comments in the libdef itself near relevant declarations, which is hopefully where people would go looking first when something goes awry. It would be great to see a kind of "caveats" a la Homebrew that appears when you install or upgrade a package, but I have not had the opportunity to champion that for `flow-typed`. I'm curious if there are better approaches.

mgreenw:
@LoganBarnett I think that's a solid approach for now. @Ashoat Would you be able to put a comment next to the typings with a mini guide on how to use these in a sound way? Maybe you could link to this PR?

Ashoat:
Sure, if you guys want me to add a comment to the libdef just let me know what it is and I’ll copy-paste it in. Or you can put up a PR after this one is merged.

mgreenw:
I think something like this could work. Thoughts?

```js
/*
  NOTE: Use caution when extending `express$Request` or `express$Response`. 
  When a request first hits server, it's `req` and `res` will not have any 
  additional properties, even if you explicitly type them with your custom subclass. 
  Subsequent middleware may assign these properties, but you must be cognizant
  of this ordering. One way to handle this is marking all properties as optional to
  force refinement every time a property is accessed. Therefore, we advise that 
  you always mark properties as _optional_ in `express$Request` and 
  `express$Response` subclasses. 

  You may decide not to do this, in which case the typings will be unsound and
  the behavior will be similar to how arrays work in Flow. See here for more
  information: https://flow.org/en/docs/types/arrays/#toc-array-access-is-unsafe

  See #3578 and #3337 for additional discussion. If you have ideas on how to
  improve these typings, please share them in #3578 or open a new issue.

  **BAD**
  declare class test_express$CustomRequest extends express$Request {
    foo: string;
  }

  **GOOD**
  declare class test_express$CustomRequest extends express$Request {
    foo: string | void;
  }
*/
```

Ashoat:
Perfecto! Added that in

mgreenw:
Lit! LGTM when you're ready.

Ashoat:
I think this is ready to go

pbondoer:
Happy to see someone else picked this up. Didn't have much time to work on it myself, so thank you @Ashoat! 🎉

pascalduez:
Awesome work @Ashoat, thanks @mgreenw, @LoganBarnett for the help.

mgreenw:
Heads up: if anyone gets an error like this after this PR, there is a solution:

**Error**
```js
Missing type annotation for Req2. Req2 is a type parameter declared in function type [1] and was implicitly instantiated
```

**Solution**
```
const router = new Router<>();
```

**Reason**
Flow [requires](https://flow.org/en/docs/types/generics/#toc-adding-defaults-to-parameterized-generics) that you put empty generic brackets to allow for defaults to work, just like you would put empty parenthesis on a function call with default parameters.

PR #3579, opened 2019-09-28
[react-color] Remove $Supertype usage

Ashoat:
Flow 0.89 and above have a lint that complains about this, and it has been considered "unclear" outside libdefs since 0.72.


Ashoat:
@pascalduez - friendly ping, figure this one may have slipped through the cracks :)

PR #3580, opened 2019-10-01
Add types for a few more webpack plugins

dmnd:
- Links to documentation: https://webpack.js.org/plugins/

Other notes:

Added types for some webpack plugins I'm using. I only tried to describe
types as I'm using these plugins; hopefully the way I'm calling these
plugins is common enough to be useful for others.


dmnd:
Oops, looks like I tested these wrong. Will fix.

definitions/npm/webpack_v4.x.x/flow_v0.104.x-/test_webpack_v4.x.x.js
```diff
@@ -0,0 +1,102 @@
...
+      moduleFilenameTemplate: 'file://[absolute-resource-path]',
+      fallbackModuleFilenameTemplate: 'file://[absolute-resource-path]?',
+      noSources: false,
+    }),
+    new webpack.HotModuleReplacementPlugin(),
+    new webpack.ContextReplacementPlugin(
+      /highlight\.js\/lib\/languages$/,
+      /^\.\/(javascript|typescript|xml)/
+    ),
+  ],
```
dmnd: Lines 46-68 are new. The rest are from the original test file.

dmnd:
Ok, CI is fixed. Sorry about that: this is now ready for review.

cli/src/commands/install.js
```diff
@@ -100,7 +100,7 @@ export function setup(yargs: Yargs) {
       alias: 'o',
       describe: 'Overwrite an existing libdef',
       type: 'string',
-      demand: false
+      demand: false,
```
pascalduez: I would prefer if you could revert this change, just to keep the history focused.

dmnd: No worries; dropped that commit.

PR #3584, opened 2019-10-02
Fixed wrong type in phoenix socket transport option

joonamarjakangas:
On Phoenix `Socket` initialization, option `transport` is expected by type definitions to be string. This is also found in Phoenix.js official documentation, but it's an error. What the constructor actually expects is either `window.WebSocket` or `Phoenix.LongPoll`. Giving any string as transport option to the constructor will throw an error when `socket.connect()` is called.

The source code for phoenix sockets can be found here: https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js
The initialization of Socket transport happens at row 737 and an error will be thrown at row 836 if the user erroneously gave a string to the constructor.

- Links to documentation: https://hexdocs.pm/phoenix/js/
- Link to GitHub or NPM: https://github.com/phoenixframework/phoenix
- Type of contribution: fix

Other notes:


villesau:
Nice catch @joonamarjakangas !

PR #3587, opened 2019-10-04
Add the supertest library

julienw:
- Links to documentation: https://github.com/visionmedia/supertest#readme and http://visionmedia.github.io/superagent/
- Link to GitHub or NPM:  https://github.com/visionmedia/supertest and https://github.com/visionmedia/superagent

Other notes:
I tried to be exhaustive for the main use cases, but I gave `agent` a miss on purpose, because it's not used a lot (at least not by me).
I reused some autogeneration from typescript but also wrote a lot by hand and reviewed what I copied.

I tried to leave comments on parts that I didn't completely type. For example the library exists for both the browser and node contexts, with slightly different APIs, so I decided to follow mostly the node version because I believe it won't be used much in browsers nowadays, in our fetch-era.
Also the "request" is also a stream, but I didn't know how to make it properly inherit from both a stream and a promise. Maybe with an intersection?

### DIFF
--- a/definitions/npm/supertest_v4.x.x/flow_v0.94.x-/supertest_v4.x.x.js
+++ b/definitions/npm/supertest_v4.x.x/flow_v0.94.x-/supertest_v4.x.x.js
@@ -1,0 +1,281 @@
+declare module 'supertest' {
+  import type { IncomingMessage, ServerResponse } from 'http';
+  import type { ReadStream as fs$ReadStream } from 'fs';
+
+  // ----- Types from superagent -----
+  declare type superagent$ResponseError = {|
+    status: number,
+    text: string,
+    method: string,
+    path: string,
+  |} & Error;
+
+  declare type superagent$Response = {|
+    text?: string, // Present for text/json/urlencoded answers.
+    body: any,
+    type: string,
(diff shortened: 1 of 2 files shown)


pascalduez:
> Also the "request" is also a stream, but I didn't know how to make it properly inherit from both a stream and a promise. Maybe with an intersection?

I'm also running a bit out of ideas, with interfaces we can extends several other interface at once, but not classes.  Looks like [they took the same path in TS](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/be146d853af634c1b3b44c39ae243898c10d630c/types/superagent/index.d.ts#L125)? Anyway, it's already really good, and deserve to be merged, can always be augmented later if needed.

julienw:
Thanks!

julienw:
>     Also the "request" is also a stream, but I didn't know how to make it properly inherit from both a stream and a promise. Maybe with an intersection?

I found a possible solution in flow's own libdef:
https://github.com/facebook/flow/blob/e3f290279943404b31dfb7a5626eb15a613e4645/lib/node.js#L1905

I haven't tried it yet :)

PR #3589, opened 2019-10-04
Feat/ramda/compose

narley:
Increase amount of arguments in the `compose` function


narley:
Hi  @LoganBarnett,
Could you please review this PR?
I had a scenario where I was composing more than 6 functions and flow would complaint. Decided to increase the amount of arguments `compose` would take.

Cheers.

LoganBarnett:
@narley I think this is fine. Would you be okay adding a minor test to it? if only so we can make sure a future change doesn't break the arity that you need to support.

narley:
@LoganBarnett I've added a couple of tests.

PR #3595, opened 2019-10-08
Fix and consolodate puppeteer type definitions

ganemone:
- Links to documentation: https://github.com/GoogleChrome/puppeteer
- Link to GitHub or NPM: https://github.com/GoogleChrome/puppeteer
- Type of contribution: fix

Other notes:

The flow type definitions for puppeteer were previously incorrect. There are not separate exports for functions like launch, connect, etc. These are all attached to a puppeteer instance on a default export. This consolidates the different definitions into a single definition and fixes the exports.


pascalduez:
I'm a bit worried about the files renaming, you treated `1.2.x`, and `1.12.x` the same and converted all to `1.20.x`, there should have been differences between those versions? 
Also if a user depend on a version lower than `1.20.x`, they will be not be able to update/install the definitions anymore.

ganemone:
> I'm a bit worried about the files renaming, you treated `1.2.x`, and `1.12.x` the same and converted all to `1.20.x`, there should have been differences between those versions?
> Also if a user depend on a version lower than `1.20.x`, they will be not be able to update/install the definitions anymore.

Yea I debated various options here. Ultimately, I do this this is the best move since the previous type annotations are incorrect for the versions they specify. So if someone does depend on these types and uses them with the puppeteer library, their code will be broken. You cannot destructure the puppeteer imports and call the functions individually, even on the older versions.

I thought it would be better to fix and consolidate the types rather than keeping around broken types. I'm happy to discuss this further however if you disagree.

PR #3599, opened 2019-10-13
Fix extra reselect resultFunc args

warp:
- Links to documentation: https://github.com/reduxjs/reselect#api
- Link to GitHub or NPM: https://github.com/reduxjs/reselect
- Type of contribution: fix

Other notes:
resultFunc does not receive State and Props these are only passed to input selectors

### DIFF
--- a/definitions/npm/reselect_v3.x.x/flow_v0.104.x-/test_reselect_3.x.x.js
+++ b/definitions/npm/reselect_v3.x.x/flow_v0.104.x-/test_reselect_3.x.x.js
@@ -241,8 +241,7 @@ createSelector(
     return x + y;
   }
 ).resultFunc(
-  { x: 100, y: 200 },
-  {x: 10},
-  10,
+  100,
+  200,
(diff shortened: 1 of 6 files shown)


definitions/npm/reselect_v3.x.x/flow_v0.104.x-/test_reselect_3.x.x.js
```diff
@@ -241,8 +241,7 @@ createSelector(
     return x + y;
   }
 ).resultFunc(
-  { x: 100, y: 200 },
-  {x: 10},
-  10,
+  100,
+  200,
```
AndrewSouthpaw: I don't understand reselect that well, can you explain why the test needed to be changed as well?

warp: resultFunc is called with the results of the input selectors not the state and props so the test has been updated to reflect this.

PR #3601, opened 2019-10-17
[styled-components v4]: Minor improvements to styled-components v4

omninonsense:
Mainly makes the defs a bit DRY-er. ~~I am not sure how well this work, but tests are passing at least, and if I'm reading the docs right, it should work.~~

I am not sure how I feel about exporting the "private" types. I guess it's fine, but the only alternative I could think of is creating a fake module called `styled-components/flow-typed` and importing stuff from there, but that seems shady.


omninonsense:
The failing tests are because of typedefs for older versions of styled-components.

cc @goodmind, if you want to have a look; the diff seems large, but there's not that many real changes. Probably because I re-ordered some things to make the code easier to follow

omninonsense:
Hi @AndrewSouthpaw, @GAntoine, @villesau.

Does this look OK to you?

definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
```diff
@@ -285,133 +302,21 @@ declare module 'styled-components' {
...
+    StyledComponent,
+    Interpolation,
+
+    // "private" types
+    TaggedTemplateLiteral,
+    StyledFactory,
+    StyledShorthandFactory,
+    ThemeProps,
+    PropsWithTheme,
+  } from 'styled-components';
```
AndrewSouthpaw: Did you double-check that importing within the module here does not lose typing? Types imported across libraries are implicitly converted to `any`, and while I don't think that's the case within a libdef, it'd be good to verify.

omninonsense: Argh! The tests were passing, so I assumed it works! Dang it. :sob:

You're right, it doesn't work the way I thought it does.

I am a bit confused though, because Flow did tell me I can only use `import type {…} from '…'` syntax in libdefs, to import from other libdefs.


I guess we could move the types to the top-level and have them as `npm$StyledComponents$…`, but I was never a fan of that approach :nauseated_face:; nevermind for now then.

omninonsense:
I'll probably fix a few more things. I checked some implementation details in styled-components, and noticed a few things are no longer correct (or stuff is missing).

definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
omninonsense: Scratch that. It _does_ work. I had a typo in the import; I was too occupied checking the type that I didn't notice it complained about the error during import. Lol. :woman_facepalming:

omninonsense: I'll mark this as resolved then

omninonsense:
@AndrewSouthpaw, @GAntoine should be ready for review now! :grinning:

omninonsense:
I think these should be ready for review now. :grin:

@AndrewSouthpaw @GAntoine

omninonsense:
Hmm… Still can't get this to work with [`Styles`] from [`polished`]. If I copy the recursive type, Flow has issues with pretty much any interpolation.

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/components/Switch.jsx:91:5

Could not decide which case to select, since case 1 [1] may work but if it doesn't case 3 [2] looks promising too. To
fix add a type annotation to return [3] or to props [4].

        src/components/Switch.jsx
        88│   top: 2.5px;
        89│   left: 5px;
        90│   position: absolute;
 [4][3] 91│   ${(props) => css`
        92│     background-color: ${transparentize(0.8, props.theme.foreground)};
        93│   `}
        94│
        95│   ${Input}:checked + ${Display} & {
        96│     background-color: ${(props) => transparentize(0.7, getColor(props))};

    [1]  9│     | ((executionContext: P) =>
        10│       | ((executionContext: P) => InterpolationBase)
        11│       | InterpolationBase
        12│     )
        13│     | Class<InterpolatableComponent<any>>
    [2] 14│     | InterpolationBase
```

the interpolation is defined as:

```ts
  declare type Styles = {[ruleOrSelector: string]: string | number | Styles, ...}

  declare export type InterpolationBase =
    | CSSRules
    | KeyFrames
    | string
    | number
    | false // falsy values are OK, true is the only one not allowed, because it renders as "true"
    | null
    | void
    | Styles // CSS-in-JS object returned by polished are also supported
```

---

I guess we can just leave leave it as is now, until we can think of something better. I am not sure how to fix this without breaking other stuff…

[`Styles`]: https://github.com/styled-components/polished/blob/b0762e9987d5e0d6c46fe8247b61d27ebce38b3f/src/types/style.js#L6-L8
[`polished`]: https://github.com/styled-components/polished/

definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
AndrewSouthpaw: Sounds good.

AndrewSouthpaw:
I'm gonna call this Good Enough®. I don't know styled-components super well, so if it messes stuff up then we can make fixes as they come up.

Good work @omninonsense .

jjenzz:
@omninonsense How do I use these types to specify prop types on my styled-components? I can't figure this out for the life of me 🙈

PR #3602, opened 2019-10-18
Updated octokit/rest to avoid vulnerability

StephenTG:
npm audit identified a vulnerability in https-proxy-agent which is included through octokit/rest. Updating octokit/rest will use a version that does not use the vulnerable https-proxy-agen

https://www.npmjs.com/advisories/1184

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -25,7 +25,7 @@
   },
   "dependencies": {
     "@babel/polyfill": "^7.0.0",
-    "@octokit/rest": "^15.12.1",
+    "@octokit/rest": "^16.33.1",
     "colors": "^1.3.2",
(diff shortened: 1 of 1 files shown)


AndrewSouthpaw:
Thanks!

andrewcashmore:
@AndrewSouthpaw any idea when the new version will be published?

AndrewSouthpaw:
@GAntoine is responsible for releases.

GAntoine:
AFAIK this is the only new change to the cli, so I'll push a patch tonight (or tomorrow at the latest).

andrewcashmore:
Thank you @AndrewSouthpaw / @GAntoine

PR #3604, opened 2019-10-21
[http-status-codes] Add lib-def of http-status-codes

kogai:
- Links to documentation: https://github.com/prettymuchbryce/http-status-codes#readme
- Link to GitHub or NPM:
    - https://github.com/prettymuchbryce/http-status-codes
    - https://www.npmjs.com/package/http-status-codes

Other notes: None


definitions/npm/http-status-codes_v1.x.x/test_http-status-codes_v1.x.x.js
```diff
@@ -0,0 +1,23 @@
+import { describe, it } from 'flow-typed-test';
+import React from 'react';
```
AndrewSouthpaw: Small nit, but why import react?

definitions/npm/http-status-codes_v1.x.x/test_http-status-codes_v1.x.x.js
```diff
@@ -0,0 +1,23 @@
+import { describe, it } from 'flow-typed-test';
```
AndrewSouthpaw: Please move the test file to be paired with the flow version. Sometimes tests need to vary based on implementation, so it's standard practice to keep the tests versioned as well.

kogai: Thanks to reviewing :pray:
And I've fixed it at f21726dd08297e2cd1e3535b1cf642b58ffecded

PR #3608, opened 2019-10-22
Add a lib def for @koa/router

julienw:
- Links to documentation: https://github.com/koajs/router/blob/master/API.md
- Link to GitHub or NPM: https://github.com/koajs/router

Other notes:
I copied most of it from `koa-router`; indeed this npm module has been forked and renamed because the original was unmaintained.
I only kept the version for "v0.104-" but adapted it to make it work with some earlier flow versions, which is simpler if we don't want to support versions of flow earlier than v0.84.
Also I added the newest additions (there's a new parameter in the `url` method).
Lastly I kept most tests but reorganized them with describe and it, as well as added new ones.

Tell me what you think!

### DIFF
--- a/definitions/npm/@koa/router_v8.0.x/flow_v0.84.x-/router_v8.0.x.js
+++ b/definitions/npm/@koa/router_v8.0.x/flow_v0.84.x-/router_v8.0.x.js
@@ -1,0 +1,119 @@
+declare module "koa-router" {
+  declare type Middleware = (
+    ctx: any,
+    next: () => Promise<void>
+  ) => Promise<void> | void;
+
+  declare type ParamMiddleware = (
+    param: string,
+    ctx: any,
+    next: () => Promise<void>
+  ) => Promise<void> | void;
+
+  declare class Router {
+    constructor(opts?: $Shape<{|
+      +prefix: string,
+      +sensitive: boolean,
+      +strict: boolean,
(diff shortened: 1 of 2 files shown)


definitions/npm/@koa/router_v8.0.x/flow_v0.84.x-/router_v8.0.x.js
```diff
@@ -0,0 +1,119 @@
+type KoaRouter$Middleware = (
```
pascalduez: Any reason to keep those outside of the module definition block?

julienw: No reason really, as I wrote in the description, I mostly copied a previously existing definition and didn't want to redo everything :)

julienw: Please tell me if you'd like that I revisit this to merge it.

pascalduez: I would personally just declare/export them inside the module scope, without the prefix.

julienw: Done!

PR #3609, opened 2019-10-22
Update styled-components_v4.x.x.js

migueloller:
These changes remove the coercion of props to exact and properly uses the component's instance.

Would love to hear from the type maintainers if this is a welcome change and what I can do to help make the types better!

### DIFF
--- a/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
+++ b/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
@@ -98,7 +98,7 @@ declare module 'styled-components' {
     value: StyleSheet,
   |}>
 
-  declare export class StylSheetManager extends React$Component<SCMProps> {
+  declare export class StyleSheetManager extends React$Component<SCMProps> {
     getContext(sheet: ?StyleSheet, target: ?HTMLElement): StyleSheet;
     render(): React$Element<StyleSheetProvider>
   }


goodmind:
Wouldn't this break some tests?

migueloller:
@goodmind, it looks like the second commit broke some tests but it might be worth checking if the tests are checking for the right think. As it is right now, if you have a component that uses `attrs` and the props object isn't exact, it will be made exact for you, which results in unintended side effects. For example, if you spread props to a component to pass things like `children` or `className`.

LoganBarnett:
@pascalduez I don't recall working on this libdef nor do I have recollection of using this library. I'm happy to help weigh in though! If you need that please direct me to something specific you'd like me to look at :)

pascalduez:
@LoganBarnett Oh sorry, I must have been confused with another library then. ;)

omninonsense:
The reason exact is used because of how spread in some inexact circumstances currently behaves (or "misbehaves").

There are some pretty cool and upcoming changes to how spreads will behave in Flow ([Search for `[spreads]` commits in the flow repo](https://github.com/facebook/flow/search?q=%5Bspreads%5D&type=Commits)).

Once that is released (probably 0.111?), I think think that most of this will need quite a bit of reworking because it will start generating a lot of errors. So, for the time being, I would keep the typo commit, and drop the other two commits. Once the spread changes are live, we can revisit these libdefs.

migueloller:
@omninonsense, that sounds good. I'll follow up once the spread semantics release is out! For now I'll drop the non-typo commits.

omninonsense:
LGTM! :heart:

AndrewSouthpaw:
I'm wondering about the build? Was it broken previously for this libdef @omninonsense ?

PR #3615, opened 2019-10-25
Add a libdef for the mozlog library

julienw:
- Links to documentation: https://github.com/mozilla/mozlog
- Link to GitHub or NPM: https://github.com/mozilla/mozlog

Other notes:
This is a mozilla tool that wraps `intel` with some specifics.

### DIFF
--- a/definitions/npm/mozlog_v2.2.x/flow_v0.84.x-/mozlog_v2.2.x.js
+++ b/definitions/npm/mozlog_v2.2.x/flow_v0.84.x-/mozlog_v2.2.x.js
@@ -1,0 +1,33 @@
+declare module 'mozlog' {
+  declare type LoggerFunction = (string, mixed) => void;
+
+  declare interface Logger {
+    trace: LoggerFunction;
+    verbose: LoggerFunction;
+    debug: LoggerFunction;
+    info: LoggerFunction;
+    warn: LoggerFunction;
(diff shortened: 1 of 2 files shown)


definitions/npm/mozlog_v2.2.x/flow_v0.84.x-/mozlog_v2.2.x.js
```diff
@@ -0,0 +1,33 @@
...
+
+    // aliases
+    warning: LoggerFunction;
+    o_O: LoggerFunction; // alias to warn
+    O_O: LoggerFunction; // alias to error
+  }
+
+  declare type LoggerCreator = (string) => Logger;
+
+  declare export default function mozlog(options: string | {|
```
julienw: To be honest I don't know if this is "legit" to use the es module style of exporting if the module itself is using the traditional `exports` object. I did that here so that the types don't leak outside of the libdef.

pascalduez: It's fine in most cases, and even preferred.   
The commonjs form can make certain types of API difficult if not impossible to model in Flow.

Also there's the argument that Flow users should most likely be using transpilation, so ES modules...

julienw:
Thanks for the quick review!

PR #3616, opened 2019-10-25
Add koa-logger libdef

julienw:
- Links to documentation: https://github.com/koajs/logger
- Link to GitHub or NPM: https://github.com/koajs/logger

### DIFF
--- a/definitions/npm/koa-logger_v3.2.x/flow_v0.53.x-/koa-logger_v3.2.x.js
+++ b/definitions/npm/koa-logger_v3.2.x/flow_v0.53.x-/koa-logger_v3.2.x.js
@@ -1,0 +1,19 @@
+declare module 'koa-logger' {
+  declare type TransporterArguments =
+    // This argument can take these 2 forms:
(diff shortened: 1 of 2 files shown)


definitions/npm/koa-logger_v3.2.x/flow_v0.53.x-/koa-logger_v3.2.x.js
```diff
@@ -0,0 +1,15 @@
+declare module 'koa-logger' {
+  declare type TransporterArguments =
+    // This argument can take these 2 forms:
+    // - Request: [format, method, url]
+    // - Response: [format, method, url, status, time, length]
+    // Typing it with only 1 tuple instead of a union makes it easier to work with.
```
julienw: or so I think... I don't use this myself so maybe future users will decide otherwise...

julienw:
please hold off before merging this. Actually I'll use the `module.exports` style here (maybe in addition to the export? can we do that?) because otherwise we can't `require` the module...

PR #3631, opened 2019-11-04
fix qs 6.5.x and add qs 6.9.x

LoganBarnett:
- Links to documentation: https://github.com/ljharb/qs
- Link to GitHub or NPM: https://github.com/ljharb/qs
- Type of contribution: new definition & fix

## Other notes:

The `qs` libdefs used the only `Function`, `Object`, and sometimes `any` types - some of which cause issues in newer versions of Flow and/or flow-typed. The old `qs` 6.5.x tests have had the bare minimum done to get the tests passing again. Most of the work was done against a new (copied) set of libdefs for 6.9.x. I did not know what versions added which arguments, or if certain arguments were omitted on the original 6.5.x, and perusing the documentation for `qs` did not reveal this in a timely manner. It seemed to make sense make a clean break to 6.9.x. The 6.9.x
version of `qs` should be entirely backwards compatible with 6.5.x, since only new, optional arguments have been added.

### DIFF
--- a/definitions/npm/qs_v6.5.x/flow_v0.104.x-/qs_v6.5.x.js
+++ b/definitions/npm/qs_v6.5.x/flow_v0.104.x-/qs_v6.5.x.js
@@ -6,7 +6,7 @@ declare module "qs" {
     delimiter?: string,
     depth?: number,
     parameterLimit?: number,
-    plainObjects?: boolean,
+    plainmixeds?: boolean,
     strictNullHandling?: boolean,
     ignoreQueryPrefix?: boolean,
     parseArrays?: boolean,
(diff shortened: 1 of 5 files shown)


LoganBarnett:
I am happy to appease the lintings we have re: `no-implicit-inexact`, but these lintings do not run locally which makes iterating on this both slow and tedious. Is there something I can do to see these errors in my local environment?

AndrewSouthpaw:
Hm. Does running `travis.sh` work for you?

LoganBarnett:
`./travis.sh` does work for me and gives me the same errors. I'll have a fix shortly. Thanks!

LoganBarnett:
@AndrewSouthpaw fixed!

PR #3648, opened 2019-11-18
Use readonly for lodash4 pickBy and omitBy

klaesra:
I was trying to use omitBy and pickBy on some properly defined $ReadOnly props, but was unable to due to the definitions of pickBy and omitBy.

Updated the definitions and it works. Figured I'd pass along my findings to the community :)

- Type of contribution: fix

Other notes: Please enlighten me if you find my PR lacking.


AndrewSouthpaw:
Great, thanks!

PR #3651, opened 2019-11-21
add overlayColor type

Preeternal:
- Links to documentation: https://reactnavigation.org/docs/en/drawer-navigator.html
- Link to GitHub or NPM: https://github.com/react-navigation/react-navigation
- Type of contribution:  fix 

Other notes:
Fix the issue https://github.com/react-navigation/react-navigation/issues/6465

### DIFF
--- a/definitions/npm/react-navigation_v1.x.x/flow_v0.104.x-/react-navigation_v1.x.x.js
+++ b/definitions/npm/react-navigation_v1.x.x/flow_v0.104.x-/react-navigation_v1.x.x.js
@@ -875,7 +875,6 @@ declare module 'react-navigation' {
     style?: ViewStyleProp,
     useNativeAnimations?: boolean,
     drawerBackgroundColor?: string,
-    overlayColor?: string,
     screenProps?: {...},
   |};
   declare type _DrawerNavigatorConfig = $Exact<{
--- a/definitions/npm/react-navigation_v1.x.x/flow_v0.104.x-/react-navigation_v1.x.x.js
+++ b/definitions/npm/react-navigation_v1.x.x/flow_v0.104.x-/react-navigation_v1.x.x.js
@@ -875,6 +875,7 @@ declare module 'react-navigation' {
     style?: ViewStyleProp,
     useNativeAnimations?: boolean,
     drawerBackgroundColor?: string,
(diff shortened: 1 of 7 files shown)


pascalduez:
I guess it would need to be added to the `@react-navigation/core` libdef as well? 

/cc @Ashoat

Preeternal:
Hi @Ashoat. What exactly I should update in @react-navigation/core and react-navigation-drawer? I already checked both I did not find any flow  libdefs there. In question about overlayColor, yes it present in these versions (1x, 2x, 3x)

Ashoat:
You'll need to update the libdefs here in flow-typed for both `@react-navigation/core` and `react-navigation-drawer`. To identify what needs to be changed, simply look at what you have changed here in `react-navigation`, and make sure to change the same things in the other libdefs.

Regarding React Navigation versions, I checked out the 1.x branch of `react-navigation` and ran `git grep overlayColor` and did not come up with anything. Can you please link the commit that introduces `overlayColor` so we can confirm which React Navigation versions your changes should be applied to?

Preeternal:
Here? Initially, I understood you, and I was looking for libdefs in the repositories of the main code.
So, we already use _overlayColor_ [here](https://github.com/flow-typed/flow-typed/blob/6d6fd081f3a87dda9e7ed6df06cd6f7621f8fb0c/definitions/npm/react-navigation-drawer_v2.x.x/flow_v0.104.x-/react-navigation-drawer_v2.x.x.js#L576) for react-navigation.
In _@react-navigation/core_  we do not use _DrawerViewOptionalProps_ or _DrawerViewConfig_. So we don't need to add _overlayColor_  there.
About v1, possible you are right. I mistaked. We use _overlayColor_ [here](https://github.com/flow-typed/flow-typed/blob/6d6fd081f3a87dda9e7ed6df06cd6f7621f8fb0c/definitions/npm/react-native-gesture-handler_v1.x.x/flow_v0.104.x-/react-native-gesture-handler_v1.x.x.js#L134) in _native-gesture-handler_v1.x.x_, not in _react-navigation_v1_, so I'll delete it for v1.

Preeternal:
https://github.com/flow-typed/flow-typed/pull/3698

PR #3657, opened 2019-12-01
Add type definitions for @skbkontur/react-ui

Eazymov:
- Links to documentation: https://tech.skbkontur.ru/react-ui/
- Link to GitHub or NPM: https://github.com/skbkontur/retail-ui/tree/master/packages/retail-ui


Eazymov:
We are removing flow typings from the @skbkontur/react-ui package and want to publish it via flow-typed

wKich:
@AndrewSouthpaw, we will remove flow typings in the next major release, so we need to publish it as flow-typed before.

PR #3667, opened 2019-12-09
Allow styled-components to return a function

### LINKED ISSUE #3666: [styled-components_v4.x.x] function is incompatible with array type
**Bug**

`Cannot call styled.div because function [1] is incompatible with array type [2].` error while using styled-components with prop function

Using the styled-components `ThemeProvider`, there is always a `theme` prop available in all styled components. Something, similar to the following example renders properly, but Flow check fails.

```js
const NewThing: React.AbstractComponent<{}> = styled.div(({ theme }) => {
  return css`
    background-color: ${theme.color.primary}
  `;
});
```

After running a flow-check, the follow error render

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/views/components/Foobar.jsx:22:58

Cannot call styled.div because function [1] is incompatible with array type [2].

     src/views/components/Foobar.jsx
      21│
 [1]  22│ const NewThing: React.AbstractComponent<{}> = styled.div(({ theme }) => {
      23│   return css`
      24│     background-color: ${theme.color.global.primary}
      25│   `;
      26│ });
      27│

 [2] 185│     [[call]]: <StyleProps, Theme>(string[], ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]) => StyledComponent<StyleProps, Theme, V>;
```

This should pass flow check because it is valid styled-components behavior.

- flow-bin `v0.109.0`
- styled-components `v4.3.2`

kyleknighted:
Fixes #3666

Allows a styled-component to return either an array of strings, or a function

### DIFF
--- a/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
+++ b/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
@@ -182,7 +182,8 @@ declare module 'styled-components' {
   |}
 
   declare export type StyledShorthandFactory<V> = {|
-    [[call]]: <StyleProps, Theme>(string[] | func, ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]) => StyledComponent<StyleProps, Theme, V>;
+    [[call]]: <StyleProps, Theme>(string[], ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]) => StyledComponent<StyleProps, Theme, V>;
+    [[call]]: <StyleProps, Theme>((props: PropsWithTheme<StyleProps, Theme>) => Interpolation<any>) => StyledComponent<StyleProps, Theme, V>;
     +attrs: <A: {...}, StyleProps = {||}, Theme = {||}>(((StyleProps) => A) | A) => TaggedTemplateLiteral<
       PropsWithTheme<{|...$Exact<StyleProps>, ...$Exact<A>|}, Theme>,
       StyledComponent<React$Config<{|...$Exact<StyleProps>, ...$Exact<A>|}, $Exact<A>>, Theme, V>
--- a/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
+++ b/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
@@ -182,7 +182,7 @@ declare module 'styled-components' {
   |}
 
   declare export type StyledShorthandFactory<V> = {|
-    [[call]]: <StyleProps, Theme>(string[], ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]) => StyledComponent<StyleProps, Theme, V>;
+    [[call]]: <StyleProps, Theme>(string[] | func, ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]) => StyledComponent<StyleProps, Theme, V>;
     +attrs: <A: {...}, StyleProps = {||}, Theme = {||}>(((StyleProps) => A) | A) => TaggedTemplateLiteral<
       PropsWithTheme<{|...$Exact<StyleProps>, ...$Exact<A>|}, Theme>,
       StyledComponent<React$Config<{|...$Exact<StyleProps>, ...$Exact<A>|}, $Exact<A>>, Theme, V>


kyleknighted:
CI is failing on a lot of unrelated tests, any advice is appreciated.

AndrewSouthpaw:
@omninonsense who has been doing a lot of work here. Ping me when it's been reviewed.

definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
```diff
@@ -182,7 +182,7 @@ declare module 'styled-components' {
   |}
 
   declare export type StyledShorthandFactory<V> = {|
-    [[call]]: <StyleProps, Theme>(string[], ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]) => StyledComponent<StyleProps, Theme, V>;
+    [[call]]: <StyleProps, Theme>(string[] | func, ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]) => StyledComponent<StyleProps, Theme, V>;
```
omninonsense: What is `func`, and where does it come from? Flow should error here, if it's not erroring it might be a bug in Flow.

omninonsense: Also, I would suggest separating/overloading the call signatures, since that is probably more in line with how the API behaves:

```suggestion
    [[call]]: <StyleProps, Theme>(string[], ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]) => StyledComponent<StyleProps, Theme, V>;
    [[call]]: <StyleProps, Theme>((props: PropsWithTheme<StyleProps, Theme>) => Interpolation<any>) => StyledComponent<StyleProps, Theme, V>;
```

:warning: **NOTE** I am not sure if the return in `(props: PropsWithTheme<StyleProps, Theme>) => Interpolation<any>` makes sense? What should be returned from this function? A JSS-like object, or something else?

omninonsense:
Hi @kyleknighted! Thanks for helping out with this. Can you take a look at the suggestions I added? :grin:

kyleknighted:
@omninonsense 
> What is func, and where does it come from? Flow should error here, if it's not erroring it might be a bug in Flow.

This was added to help #3666, the change works just fine in my local environment, but seems to not work well with React Native tests

After that, I have only used Flow for 2 days, so I am not sure I can process much more of your feedback.

Thank you for your suggestions, but if I remove `| func` it will fail flow-checks in my local instance, so I can't just remove it. Any other ideas?

omninonsense:
> @omninonsense
> 
> > What is func, and where does it come from? Flow should error here, if it's not erroring it might be a bug in Flow.
> 
> This was added to help #3666, the change works just fine in my local environment, but seems to not work well with React Native tests
> 
> After that, I have only used Flow for 2 days, so I am not sure I can process much more of your feedback.
> 
> Thank you for your suggestions, but if I remove `| func` it will fail flow-checks in my local instance, so I can't just remove it. Any other ideas?

I think what might be happening here is that `func` is being implicitly cast to `any`, we end up with `string[] | any(implicit)`, so whatever you pass it will type-check probably, but inference will at least be broken inside the function case. The suggested change in the review _should_ work fine (inference should work too, I think).

definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
kyleknighted: Okay, this change worked great in my local instance! Thank you!
Going to commit this and see how it does with the tests.

kyleknighted:
@omninonsense After committing your suggested changes, the same majority of tests continue to fail. 

Not sure why things like `styled-components_v2.x.x/flow_v0.104.x` are tested (and failing) from a seemingly unrelated change. Perhaps you have any idea on what is causing this? 

The change does appear to be working locally, however.

omninonsense:
@kyleknighted sorry for not responding for so long; totally missed the notification. The changes seem good to me now. Yeah, we can just ignore the failing tests for other libdefs for now (incl. ones for older versions of styled-components). I am still not sure what we should be returning from the "style function", but I think this is might be good enough for now?

ping/poke @AndrewSouthpaw 😁

kyleknighted:
Thanks for the response @omninonsense, no worries on delay!

omninonsense:
@kyleknighted Also probably makes sense to change the PR title and the commit message to "Allow styled-components to ~~return~~ **accept** a function" :smile:

PR #3672, opened 2019-12-11
fix: Make DataProxyReadQueryOptions generic in @apollo/react-hooks

CSNWEB:
DataProxyReadQueryOptions is used with a generic type in other places (i.e. readQuery in line 1023), so flow threw an error.

- Links to documentation: https://www.apollographql.com/docs/react/caching/cache-interaction/
- Link to GitHub or NPM: https://www.npmjs.com/package/@apollo/react-hooks
- Type of contribution: fix 

I think people who worked previously on this were: @ganemone and @pascalduez


CSNWEB:
I just realized that this causes more issues. Wondering if this should be generic and fixed everywhere down the line or if I should remove the generic type annotation from everything using DataProxyReadQueryOptions?

PR #3680, opened 2019-12-16
Refine chain definition, add test cases

quanlinc:
- Links to documentation: https://ramdajs.com/docs/#chain
- Link to GitHub or NPM:  https://github.com/ramda/ramda

Other notes:
Restrict input list to be read only, add type coverage for the case when second argument is a function


LoganBarnett:

PR #3681, opened 2019-12-16
Axios new options definition

quanlinc:
- Links to documentation: https://github.com/axios/axios
- Link to GitHub or NPM: https://github.com/axios/axios

Other notes:
Also did a little bit refactor to remove "any" type in definition and get tests passed

### DIFF
--- a/definitions/npm/axios_v0.19.x/flow_v0.80.x-/axios_v0.19.x.js
+++ b/definitions/npm/axios_v0.19.x/flow_v0.80.x-/axios_v0.19.x.js
@@ -4,7 +4,7 @@ declare module 'axios' {
 
   declare type AxiosTransformer<T> = (
     data: T,
-    headers?: {| [key: string]: mixed |},
+    headers?: { [key: string]: mixed, ...},
   ) => mixed;
 
   declare type ProxyConfig = {|
--- a/definitions/npm/axios_v0.19.x/flow_v0.80.x-/axios_v0.19.x.js
+++ b/definitions/npm/axios_v0.19.x/flow_v0.80.x-/axios_v0.19.x.js
@@ -4,8 +4,8 @@ declare module 'axios' {
 
   declare type AxiosTransformer<T> = (
     data: T,
-    headers?: {| [key: string]: any |},
-  ) => any;
+    headers?: {| [key: string]: mixed |},
+  ) => mixed;
 
   declare type ProxyConfig = {|
     host: string,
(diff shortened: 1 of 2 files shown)


LoganBarnett:

definitions/npm/axios_v0.19.x/flow_v0.80.x-/axios_v0.19.x.js
```diff
@@ -4,16 +4,16 @@ declare module 'axios' {
 
   declare type AxiosTransformer<T> = (
     data: T,
-    headers?: { [key: string]: any }
-  ) => any;
+    headers?: {| [key: string]: mixed |},
```
pascalduez: I don't think [an exact object works](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDyAdgQLxgDeAPmANoDWeWAXGAM4YBOAlsQOYBdNgFseADzwATMNQC+AblRgVqtStQw8GMBjyc2JclTBQ4cNgHIzcS2EVhUQA) here.

Later in the diff you used `{ [key: string]: mixed, ...}` for such case, any special thing here?

quanlinc: Good call, I think the original thought was flow will match up the type instead of literal key string, will this fixed and test covered

LoganBarnett: Maybe I misunderstood your comment, @quanlinc, but I'd like to clarify: We don't want to enumerate the possible keys here (they can be anything). I think we just need to make this inexact instead.

quanlinc: It could be just me not expressing myself clearly, I agree with what @pascalduez said, I had a wrong interpretation of using exact type, inexact is the way to go.

PresentProgrammer:
@quanlinc , with `"flow-bin": "0.81.0",`, the following error happens:
```
Error ---------------------------------------------------------------------------- flow-typed/npm/axios_v0.19.x.js:10:42

Unexpected token }

   10|     headers?: { [key: string]: mixed, ...},
                                                ^
```

quanlinc:
@PresentProgrammer Sorry for breaking your build, didn't expect someone is still using 0.81, when inexact type is not introduced.  I could fix this issue by splitting Axios libdef into two pieces like most other libraries do. But I also think it's consumers extra support burdens when using an older version of flow, or any tool. Which also reveals there's a gap in test between different flow-typed versions, which didn't catch your problem. I've done the work with the best of my knowledge by then and passed all tests. I'll leave the discussion to the community, at what extent do we need to support old version.

pascalduez:
Sorry that slipped through my review.   
We usually introduce inexact object notation from version `0.104.x`.  
I someone is willing to submit a PR to fix this, would be rad.

PresentProgrammer:
#3689 — related issue

pascalduez:
Should be fixed now.

PresentProgrammer:
Yes, issue with Flow 0.81 solved. Thank you!

PR #3682, opened 2019-12-16
Intersection readonly refactor

quanlinc:
- Links to documentation: https://ramdajs.com/docs/#intersection

Other notes:
Refactor intersection to take ready only input, $ReadOnlyArray is super type of Array, so it satisfies everything in Array, except it has fewer methods on it.


LoganBarnett:

PR #3683, opened 2019-12-23
let `lodash` take input object as `$ReadOnly`

BananaWanted:
- Links to documentation: https://lodash.com/docs/4.17.15
- Type of contribution: fix

Other notes:

As documented in `lodash` doc linked above, both `pickBy` and `omitBy` create new objects and leave input arguments untouched. So updated argument types to `$ReadOnly` accordingly.


pascalduez:
Would be nice if a kind soul could tackle fixing/updating lodash tests...

PR #3700, opened 2020-01-14
feat(cli): recommend matching commands

jamesgeorge007:
- Type of contribution: feature

`yargs` provides out of the box support to help out the users if they mistype a command.

### DIFF
--- a/cli/flow-typed/npm/yargs_v4.x.x.js
+++ b/cli/flow-typed/npm/yargs_v4.x.x.js
@@ -142,6 +142,8 @@ declare module "yargs" {
 
     pkgConf(key: string, cwd?: string): this;
 
+    recommendCommands(): this;
+
     // Alias of demand()!
     require(key: string, msg: string | boolean): this;
     require(count: number, max?: number, msg?: string | boolean): this;
--- a/cli/src/cli.js
+++ b/cli/src/cli.js
@@ -69,6 +69,7 @@ export function runCLI() {
     )
     .demand(1)
     .strict()
+    .recommendCommands()
     .help('h')
     .alias('h', 'help').argv;
 }


jamesgeorge007:
Not sure what leads to the CI failure :thinking: 

It works like a charm on my local machine:-
Screenshot from 2020-01-14 15-27-30

GAntoine:
Neat! What does that look like with `yargs`? Does it auto-guess the right command, or output `Did you mean xxxx?`

jamesgeorge007:
> Neat! What does that look like with `yargs`? Does it auto-guess the right command, or output `Did you mean xxxx?`
> 
> For reference, the error in Travis is `Cannot call commands.reduce(...).demand(...).strict().recommendCommands because property recommendCommands is missing in Yargs [1].`

https://github.com/flow-typed/flow-typed/pull/3700#issuecomment-574095802

valscion:
@jamesgeorge007 the error you're seeing is due to the `recommendedCommands` not being there in the `yargs` types :smile: https://github.com/flow-typed/flow-typed/blob/9facc3c967bb79251450f84e94746d7a71b526fd/definitions/npm/yargs_v4.x.x/flow_v0.104.x-/yargs_v4.x.x.js

Note that `flow-typed` CLI uses the types that are in this folder: https://github.com/flow-typed/flow-typed/tree/master/cli/flow-typed

GAntoine:
My bad, missed that the first time around! So I guess the types in the folder used by the CLI need to be updated as well, as part of this PR. I'm going to assume the current definition for `yargs` includes those changes(?), and can probably be copied in.

cli/flow-typed/npm/yargs_v4.x.x.js
```diff
@@ -142,6 +142,8 @@ declare module "yargs" {
 
     pkgConf(key: string, cwd?: string): this;
 
+    recommendCommands(): this;
```
pascalduez: The CLI [uses `yargs` version `^12.0.2`](https://github.com/flow-typed/flow-typed/blob/0b0eb4afe995b298c2b827beaaf5dc5b5982bd0d/cli/package.json#L43) So it would be good to update the lib-def here, since `recommendCommands` is part of the `10.x.x` def.

PR #3706, opened 2020-01-17
Update query-string 6.x definitions

### LINKED ISSUE #3693: query-string v6.x.x missing an option for Stringify() method
Hi !

According to [_query-string_ docs about `stringify()`](https://github.com/sindresorhus/query-string#stringifyobject-options), there is a new `boolean` option, called `skipNull` (added in version `6.9.0` in november 2019).

It would be great to have it introduced into the flow def.
Thx!

polosson:
Update `query-string_v6.x.x.js` definitions with latest options and methods that were missing.

- Type of contribution: update definition
- Links to documentation: https://github.com/sindresorhus/query-string#stringifyobject-options

I updated the tests as well as I could, hope that's sufficient.

closes #3693

### DIFF
--- a/definitions/npm/query-string_v6.x.x/flow_v0.104.x-/query-string_v6.x.x.js
+++ b/definitions/npm/query-string_v6.x.x/flow_v0.104.x-/query-string_v6.x.x.js
@@ -20,7 +20,7 @@ declare module 'query-string' {
 
   declare type ObjectParameters = $ReadOnly<{ [string]: ObjectParameter | $ReadOnlyArray<ObjectParameter>, ... }>
 
-  declare type QueryParameters = { [string]: string | Array<string> | null, ... }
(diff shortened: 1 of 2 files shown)


definitions/npm/query-string_v6.x.x/flow_v0.104.x-/query-string_v6.x.x.js
```diff
@@ -17,6 +22,8 @@ declare module 'query-string' {
 
   declare type QueryParameters = { [string]: string | Array<string> | null, ... }
 
+  declare type StringifyObjectParameter = { url: string, query: QueryParameters }
```
pascalduez: Choose between an exact or inexact object type. `{| |}` or `{ ... }`.

pascalduez:
Would be rad if someone could have a look at the failing tests, looks like there's a wrong assertion...

PR #3707, opened 2020-01-18
lodash: differenceBy/differenceWith update

lukeapage:
- Links to documentation: https://lodash.com/docs/4.17.15#differenceBy
- Type of contribution: fix

Other notes:

when using differenceBy and differenceWith the types of the "values" and array do not need to match e.g.

```
_.differenceBy([{ id: 1, a: 2 }], [{ id: 1 }], ({ id }) => id);
```

In fact if using the by or with alternatives I would argue they are more likely to not match.

### DIFF
--- a/definitions/npm/@testing-library/react_v8.x.x/flow_v0.104.x-/react_v8.x.x.js
+++ b/definitions/npm/@testing-library/react_v8.x.x/flow_v0.104.x-/react_v8.x.x.js
@@ -115,35 +115,28 @@ declare module '@testing-library/react' {
     eventProperties?: TInit
   ) => boolean;
 
-  declare type Queries = { ... };
-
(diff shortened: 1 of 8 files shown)


valscion:
This seems to break the lodash tests. Would you be so kind to debug why the tests fail and maybe add a new test case to demonstrate why this change is needed?

lukeapage:
@valscion the failures are on a part of the definition I didn't touch. I'm fairly sure its due to a new flow version that doesn't like part of the tests. I will try and send a pr to fix it and add a test to this one.

lukeapage:
I merged the fix here but i suggest you merge https://github.com/flow-typed/flow-typed/pull/3714 first

valscion:
Cool, thanks! This looks more ready to me now, and as the CI is green, this has a better chance of getting merged :relaxed:

PR #3709, opened 2020-01-20
Update react-redux_v7.x.x.js - shallowEqual

nosnickid:
### Links to documentation:
There is no explicit documentation for this function, just assorted references to it:

- https://react-redux.js.org/api/hooks#equality-comparisons-and-updates
- https://react-redux.js.org/api/hooks#recipe-useshallowequalselector
And a long discussion:
- https://github.com/reduxjs/react-redux/pull/1288

### Link to GitHub or NPM: 
https://github.com/reduxjs/react-redux/blob/master/src/utils/shallowEqual.js

### Other notes:
A fairly simple helper function that's commonly used to ease migration from `connect()` in react-redux to hooks.


nosnickid:
I'll take a look at the existing failures caused by the exact/inexact changes in newer versions of flow (can't remember the exact version causing it)

PR #3714, opened 2020-01-27
Fix lodash tests

lukeapage:
Lodash tests currently fail with the latest version of flow.

This should fix it.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/test_lodash-v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/test_lodash-v4.x.x.js
@@ -367,7 +367,7 @@ zip([{ x: 1 }], [{ x: 2, y: 1 }])[0][2];
 /**
  * _.zipWith
  */
-zipWith(["a", "b", "c"], [1, 2, 3], (str, num) => ({ [string]: num }));
+zipWith(["a", "b", "c"], [1, 2, 3], (str: string, num) => ({ [str]: num }));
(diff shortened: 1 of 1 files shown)


definitions/npm/lodash_v4.x.x/flow_v0.104.x-/test_lodash-v4.x.x.js
```diff
@@ -355,7 +355,7 @@ zip([{ x: 1 }], [{ x: 2, y: 1 }])[0][2];
 /**
  * _.zipWith
  */
-zipWith(["a", "b", "c"], [1, 2, 3], (str, num) => ({ [str]: num }));
+zipWith(["a", "b", "c"], [1, 2, 3], (str, num) => ({ [string]: num }));
```
pascalduez: Humm that's definitely not the same thing, here `str` is an argument, `string` is `undefined` as far as I can see.

lukeapage: You are right, I looked at it too quickly and thought it was a type. I’m not sure how you want to solve this error, I only want to allow my other Pr to go through

lukeapage:
Looks like this already went in:
https://github.com/flow-typed/flow-typed/commit/ae5262dc1fb7bf608cbdc940a57572fbe8bf6f0b

I did leave a comment on that PR asking that this be merged first.

lukeapage:
I've fixed the fix now and it passes.

PR #3716, opened 2020-01-29
move react-dom to flow-typed

jbrown215:
Other notes:

[This is verbatim the Flow definitions](https://github.com/facebook/flow/blob/90fcd8f83df4c791266f985689ad959a775f8667/lib/react-dom.js#L1-L117). I also moved the relevant tests to flow-typed.

We're keeping the Synthetic*Event types in Flow for now. We will need to keep the $JSXIntrinsics in Flow forever, since we use that type internally in our type checking.

These types will be moved out of Flow as early as 0.118, but might not make the cut until 0.119.

On the Flow side, I added a test where we have a `flow-typed/npm/react-dom_v16.x.x.js` with this exact definition and I ensure that we have access both to types in `react-dom.js` in Flow and `react-dom_v16.x.x.js` from flow-typed.


pascalduez:
Out of curiosity, what's motivating the move?

PR #3731, opened 2020-02-13
[styled-components v4] Various small improvements to the libdefs

omninonsense:
Changes:

 1. The return value of `StyleSheetManager.prototype.render`
 2. All styled components implicitly support `style`, `className`, and `children`, so reflect that.  
     People can still override them if they need to (e.g. wanting to narrow `children` with `children?: React.ChildrenArray<…>`) using `Props`.

**Note**: One big "breaking change" is that strict props are no longer "respected" on components. So, it won't catch typos any longer. This now more aligns with flow's new "only error if there's a runtime error" paradigm (unless I am wrong and this does cause runtime errors, but extra props shouldn't cause runtime errors).

Making it a draft while I still test some of this against our production codebase at work, too. The tests seem to be happy and this seems to work on one of our smaller projects. So, hopefully won't stay a draft for too long!

Also, this should probably be merged after #3667.

### DIFF
--- a/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
+++ b/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
@@ -21,7 +21,7 @@ declare module 'styled-components' {
     | false // falsy values are OK, true is the only one not allowed, because it renders as "true"
     | null
     | void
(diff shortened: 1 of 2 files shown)


omninonsense:
Fixed the CI error. Other CI errors are because of older styled-component libraries.

omninonsense:
Ok in our larger codebase there was some issues with `style` being defined on one component in a weird way, but that was the only problem. So I guess this is ready for review.

Issue #3127, opened 2019-02-05
[reselect] InputSelector is not an object error with Flow 0.92.1


jedwards1211:
Unfortunately in Flow 0.92.1, I get the following error:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ flow-typed/npm/reselect_v3.x.x.js:11:7

InputSelector [1] is not an object.

      8│     (state: TState, props: TProps, ...rest: any[]) => TResult
      9│
     10│   declare type OutputSelector<-TState, TProps, TResult> =
 [1] 11│     & InputSelector<TState, TProps, TResult>
     12│     & {
     13│       recomputations(): number,
     14│       resetRecomputations(): number,



Found 1 error
```

Tasemu:
+1

voronindenis:
+1

voronindenis:
This error occurs when a createStructuredSelector is wrapped in an unnamed function. For example:
`const mapStateToProps = (state, { something }) => createStructuredSelector({
    otherSomething: createGetSomethingSelector({ something }),
})`

Tasemu:
> This error occurs when a createStructuredSelector is wrapped in an unnamed function. For example:
> 
> ```js
> const mapStateToProps = (state, { something }) => createStructuredSelector({
>     otherSomething: createGetSomethingSelector({ something }),
> })
> ```

I'm getting this and am not using `createStructuredSelector` anywhere, nor am i using reselect...

augusto-altman:
Hello guys,

I had the same issue and I fixed it by redefining the type `OutputSelector` in the file `flow-typed/npm/reselect_v4.x.x.js` like this:
`  
declare type OutputSelector<-TState, TProps, TResult> = {|
    (state: TState, props: TProps, ...rest: any[]): TResult,
    recomputations(): number,
    resetRecomputations(): number,
    resultFunc(...args: any[]): TResult,
  |}
`
The previous definition was using type intersections and I replaced that with Flow's [calleable objects](https://flow.org/en/docs/types/functions/#toc-callable-objects). It works and as far as I understand it achieves a more accurate typing.
![image](https://user-images.githubusercontent.com/7969511/73792915-32f4d380-480a-11ea-86d3-c1a93cc213f5.png)

[...]

PR #3732, opened 2020-02-14
[reselect_v4.x.x.js] Redefine OutputSelector type as a callable object

augusto-altman:
- Links to documentation: https://github.com/reduxjs/reselect/blob/master/README.md
- Link to GitHub or NPM: https://github.com/reduxjs/reselect
- Type of contribution: fix

Other notes: This is a very simple change and is backward compatible so I didn't need to write any new tests. I am just redefining the `OutputSelector` type to a callable object since it achieves a more accurate typing (closer to the real api). The previous definition was an intersection between an `InputSelector` (function) and an object containing the rest of the api.
This also seems to fix [this issue](https://github.com/flow-typed/flow-typed/issues/3127), which happens to be really hard to reproduce. After applying the proposed change locally in one of the projects where I have encountered the issue it was properly fixed.


DmitryGonchar:
I think same fix should be applied to v3 (at least) as well?
https://github.com/flow-typed/flow-typed/issues/3127#issuecomment-593771559

Issue #3561, opened 2019-09-16
[ramda] last should be OK with getting a $ReadOnlyArray as argument


lensbart:
Title says it all!

This is the error that I get:
`Cannot call `last` because  read-only array type [1] is incompatible with  array type [2]`

PR #3733, opened 2020-02-15
[ramda] Cleanup definitions

moroine:
- Links to documentation: https://ramdajs.com/docs/
- Type of contribution:  fix & refactor

Other notes:
- Change function params to $ReadOnlyArray as ramda does not modify input
- Remove usage of `any` & `Object`

Should fix https://github.com/flow-typed/flow-typed/issues/3561


moroine:
@AndrewSouthpaw @LoganBarnett could you have a look at it?

LoganBarnett:
@moroine I think these are good changes. One thing I would stipulate is Ramda functions should accept `$ReadOnlyArray` and return `Array`. This allows the consumer to use either type as they like. `$ReadOnlyArray` is not actually a truly immutable array. Using the type is only a promise that you won’t mutate the array within the context of the type. I’m writing this from a phone, but it’s pretty easy to take a `Array` and pass it into something that takes a `$ReadOnlyArray`, and mutate the `Array` immediately afterwards. 

Another reason to type inputs as `$ReadOnlyArray` and output `Array` is so any other libraries or even your codebase itself can interop with Ramda. 

If you need more examples, we have a few open PRs for Ramda currently that are using this paradigm. I didn’t get a thorough look at the PR yet, but I’m pretty sure one of those covers `chain` already. I’ll take a look when I get back to my machine. 

Regarding the input and output types, does that sound reasonable to you?

LoganBarnett:
Oh. Looks you’re covering `Chain` instead of `chain`. And you are returning `Array`. That’s what I get for reading a diff from a tiny phone. Disregard my post.

moroine:
@LoganBarnett nice description of `$ReadOnlyArray` usage. To add more details on why we should always use ReadOnlyArray as param and Array as the return value is because ramda return a new instance.

Also as a side note, I tried many way to remove deprecated type `*`, especially in curry definitions but I don't think it's possible. I tried to propose a PR on flowJs which would make this possible but unfortunately I don't think it will be merged https://github.com/facebook/flow/pull/8201 😢

PR #3739, opened 2020-02-17
[@sentry/browser] Add Integrations and Transports, fix extras

USSRLivesOn:
In #3723, I added libdefs for @sentry/browser. I forgot to add types for two additional exports: Integrations and Transports. This PR adds them.

- Links to documentation:
https://docs.sentry.io/platforms/javascript/#sdk-integrations
- Link to GitHub or NPM:
https://github.com/getsentry/sentry-javascript/blob/5.12.3/packages/browser/src/index.ts

### DIFF
--- a/definitions/npm/@sentry/browser_v5.x.x/flow_v0.84.x-/browser_v5.x.x.js
+++ b/definitions/npm/@sentry/browser_v5.x.x/flow_v0.84.x-/browser_v5.x.x.js
@@ -356,9 +356,14 @@ declare module '@sentry/browser' {
 
         +blacklistUrls?: $ReadOnlyArray<string | RegExp>,
         +whitelistUrls?: $ReadOnlyArray<string | RegExp>,
-        +integrations?:
-            | $ReadOnlyArray<Integration<any>>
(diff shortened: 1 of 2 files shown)


USSRLivesOn:
My Flow-foo it at it's limits. If anyone has any suggestions on how to type this correctly to get CI to pass, I'd appreciate your input. It seems like typing the integrations and transports as a generic interface makes Flow throw an error, but not sure how to otherwise type this appropriately.

pascalduez:
> I'm not sure why Flow considers new Integrations.UserAgent() to be invalid in this case.

Try using a class instead of an interface? 
I doubt a constructor on an interface is a good thing, as well as trying to instantiate it.

USSRLivesOn:
@pascalduez I've fixed the problems in CI. The cause was that I was typing each of the integrations as an instance of the `Integration` class instead of the class itself, so `new Integration(...)` obviously didn't work. I've resolved it by using the `Class` utility from Flow. 

I've also added another commit to fix support for custom integrations and integrations from `@sentry/integrations`.

This is good to (re-)review.

PR #3740, opened 2020-02-17
Add hooks defs to reach router v1.1.x

saadshahd:
these hooks just got added to reach router v1.1.3

https://reach.tech/router/api/useLocation
https://reach.tech/router/api/useMatch
https://reach.tech/router/api/useNavigate
https://reach.tech/router/api/useParams

I've no idea what I'm doing. I'm new to flow but I felt like it's okay to start a PR


GAntoine:
Thanks for your contribution, you're off to a good start!

It looks like the tests aren't happy for the type for `Match`:
```
    Cannot use `Match` as a type. A name can be used as a type only if it refers to a type definition, an interface
    definition, or a class definition. To get the type of a non-class value, use `typeof`.
    
       138|   declare export function useRouteMatch(path?: string): Match;
                                                                    ^^^^^
```

You'll also need to write a couple tests to  cover the 4 new methods you added. You can find some examples in the existing files for reach router, or in the docs

pascalduez:
Stalled + mostly covered in #3937.

PR #3741, opened 2020-02-17
Add libdefs & tests for @sentry/integrations@5.x.x

USSRLivesOn:
- Links to documentation:
https://docs.sentry.io/platforms/javascript/#pluggable-integrations
- Link to GitHub or NPM:
https://github.com/getsentry/sentry-javascript/blob/5.12.3/packages/integrations/src/index.ts

The documentation is incomplete, so this is largely based on the TypeScript definitions linked above.

### DIFF
--- a/definitions/npm/@sentry/integrations_v5.x.x/flow_v0.84-/integrations_v5.x.x.js
+++ b/definitions/npm/@sentry/integrations_v5.x.x/flow_v0.84-/integrations_v5.x.x.js
@@ -1,0 +1,189 @@
+// @flow
+
+declare module '@sentry/integrations' {
+    declare export var Angular: Class<Integration<>>;
+    declare export var CaptureConsole: Class<Integration<{|
+        levels?: $ReadOnlyArray<$Values<Severity>>,
+    |}>>;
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw:
Thanks for the PR! Currently your build is failing, and we'll need that to pass to merge it in. You can check out the build in Travis, or read about running libdef tests locally in CONTRIBUTING.md.

USSRLivesOn:
@AndrewSouthpaw Thanks for the quick look! The CI failures here are for the most part the same issues I'm seeing in #3739, and I can't figure out how to type that appropriately. Not sure if what I'm trying to do is just a Flow limitation with generics or something else, but the errors I see don't make a lot of sense. See https://github.com/flow-typed/flow-typed/pull/3739#issuecomment-587232063.

USSRLivesOn:
@AndrewSouthpaw I've fixed the problems in CI. The cause was that I was typing each of the integrations as an instance of the `Integration` class instead of the class itself, so `new Integration(...)` obviously didn't work. I've resolved it by using the `Class` utility from Flow. 

This is good to (re-)review.

Issue #3726, opened 2020-02-06
flow-typed install react-dom failing


jbrown215:
**Do you want to request a *feature* or report a *bug*?**
Reporting a bug
**What is the current behavior?**
`flow-typed install --verbose --flowVersion 0.117.0 react-dom@16.12.0` outputs `• Searching for 1 libdefs...` and nothing else. Does not create a `flow-typed` folder with the libdef.

**If the current behavior is a bug, please provide the steps to reproduce.**
```
yarn global add flow-typed
mkdir test-project && cd test-project
npm init
touch .flowconfig
flow-typed install --verbose --flowVersion 0.117.0 react-dom@16.12.0
```

**What is the expected behavior?**
A libdef is installed at ./flow-typed/npm/
**If this is a feature request, what is motivation or use case for changing the behavior?**

**Local Environment Information**
https://gist.github.com/jbrown215/59653a36d85262fd77157ff60f5b2b4d

jbrown215:
Note that `flow-typed search react-dom` gives me:

```
Found definitions:
╔═══════════╤═════════════════╤══════════════╗
║ Name      │ Package Version │ Flow Version ║
╟───────────┼─────────────────┼──────────────╢
║ react-dom │ v16.x.x         │ >=v0.117.x   ║
╚═══════════╧═════════════════╧══════════════╝
```

jbrown215:
Caused by https://github.com/flow-typed/flow-typed/blob/090deac6453c58166559f733488a88c6636fb5ef/cli/src/commands/install.js#L217

will put up a PR

PR #3748, opened 2020-02-21
Remove react-dom from built-in flow libdefs

jbrown215:
- Type of contribution: fix

Other notes:
Fixes https://github.com/flow-typed/flow-typed/issues/3726. We are moving react-dom out of Flow. It is already in flow-typed!


jbrown215:
You rock, thanks!

Issue #3279, opened 2019-04-26
[styled-components] : How should we type styled component with props


TomPradat:
I'm not able to find a single working example to type styled components using props inside

```js
const Div = styled.div`
    color: ${({ color }) => color};
`;
```

I always get an error saying that color is missing in object type, etc ...

Does anyone have a working example ?

Also i've noticed that the libdef in this repo for `styled-components_v4.x.x.js` is in a subdir called `flow_v0.75.x-`. I'm using flow 0.97 so maybe i have a problem with this too, but flow-typed binary installs me this libdef

edahlseng:
@TomPradat #3228 should fix a lot of the issues you're seeing.

pascalduez:
Closing in hope #3228 solves the issue.

PR #3749, opened 2020-02-22
[styled-components v5] add libdefs

omninonsense:
# < :nail_care: >

Adds libdefs for v5 of styled-components. Mostly it's the same as libdefs for v4, but with the following non-breaking changes (copied from the [release notes]):

 - adds `useTheme` hook
 - adds `stylisPlugins` prop to `StyleSheetManager`
 - adds `disableVendorPrefixes` prop to `StyleSheetManager`
 - adds `disableCSSOMInjection` prop to `StyleSheetManager`

The *"subfunction" attrs syntax that was deprecated in v4* which is removed in v5 was never supported in v4 libdefs (because it was deprecated), so nothing to do here.

---

I also took the opportunity to fix some type issues I encountered in the v4 libdefs. Interpolations should be more reliable and error in fewer cases (still doesn't support Polished's `Styles`).

In some cases Flow would claim that `React$AbstractComponent` wasn't compatible with `StyledComponent`; I am not sure where that comes from since (maybe the intersection with `InterpolatableComponent` causes a bug). I couldn't manage to make a small reproducible case, and I couldn't see anything that triggered it specifically. Two other similar components didn't have this issue within the same codebase. I added this on a hunch and it fixes it, but it's not very versatile:

~~~js
    <
      Comp: StyledComponent<P, T, I>,
      Theme: ExtractTheme<Comp>,
      +Instance: ExtractInstance<Comp>,
      OwnProps = ExtractProps<Comp>
    >(
      Comp
    ): StyledFactory<
      {| ...$Exact<OwnProps> |},
      Theme,
      Instance
    >;
~~~

Additionally these libdefs are now styled with prettier.

[release notes]: https://styled-components.com/releases#v5.0.0

### DIFF
--- a/definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
+++ b/definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
@@ -1,0 +1,548 @@
+// @flow
+
+declare module 'styled-components' {
+  declare class InterpolatableComponent<P> extends React$Component<P> {
+    static +styledComponentId: string;
+  }
+
+  declare export type Styles = {
+    [ruleOrSelector: string]: string | number, // | Styles,
+    ...,
+  };
+
+  declare export type Interpolation<P> =
+    | ((
+        executionContext: P
+      ) => ((executionContext: P) => InterpolationBase) | InterpolationBase)
+    | InterpolationBase;
+
+  declare export type InterpolationBase =
+    | CSSRules
+    | KeyFrames
+    | string
+    | number
+    | false // falsy values are OK, true is the only one not allowed, because it renders as "true"
+    | null
+    | void
+    | Styles
+    | Class<InterpolatableComponent<any>>; // CSS-in-JS object returned by polished are also supported, partially
+
+  declare export type TaggedTemplateLiteral<I, R> = (
+    strings: string[],
+    ...interpolations: Interpolation<I>[]
+  ) => R;
+
+  // Should this be `mixed` perhaps?
+  declare export type CSSRules = Interpolation<any>[]; // eslint-disable-line flowtype/no-weak-types
+
+  declare export type CSSConstructor = TaggedTemplateLiteral<any, CSSRules>; // eslint-disable-line flowtype/no-weak-types
+  declare export type KeyFramesConstructor = TaggedTemplateLiteral<
+    any,
+    KeyFrames
+  >; // eslint-disable-line flowtype/no-weak-types
+  declare export type CreateGlobalStyleConstructor = TaggedTemplateLiteral<
+    any,
+    React$ComponentType<*>
+  >; // eslint-disable-line flowtype/no-weak-types
+
+  declare interface Tag<T> {
+    styleTag: HTMLStyleElement | null;
+    getIds(): string[];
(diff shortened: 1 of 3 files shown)


omninonsense:
The CI errors are for styled-components v2 and v3.

ianduvall:
Are there any examples showing how to specify your theme and props why calling `styled`? I saw #3279 asked the same question.

Also I'm seeing some Flow parse errors in `declare interface Styled`. You can replicate by copying the changes into Try Flow. Any suggestions?

omninonsense:
@ianduvall ah thanks for catching the error in the interface! I wonder why Flow didn't pick up on it during the rests. Makes me wonder if we our tests are flawed (or the way we run tests?) :thinking:

I have a fix for it. I was experimenting with something and thought it worked, but I think I gave up on it and forgot to remove this piece of code. Derp!

As to how to use them. There is a few ways. I will probably put a more detailed guide in the README. Also, I'll try to remember to ping you for when that happens.

```ts
type Props = {
  weight: number | string,
  ...
}

type Theme = {|
  accent: string,
  foreground: string,
  background: string,
|}

// You can be explicit and say HTMLButtonElement instead of using *
const FancyButton: StyledComponent<Props, Theme, *> = styled.button`
  font-weight: ${props => props.weight};
  background-color: ${props => props.theme.accent};
`

// or

// You can be explicit and say HTMLButtonElement instead of using _
const FancyButton = styled<Props, Theme, _>('button')`
  font-weight: ${props => props.weight};
  background-color: ${props => props.theme.accent};
`
```

to wrap other components (which I just broke lol) you could do it like this:

```ts
const SpacyButton = styled<typeof FancyButton, Theme>(FancyButton)`
  padding: 15px;
`
```

If you want to add style props in addition to the props the other component accepts, you can do this:

```ts
const SpacyButton = styled<typeof FancyButton, Theme, {space: number, ...}>(FancyButton)`
  padding: ${props => props.space}px;
`
```

Note that this second example is a bit... Fickle. The problem is that some components error on extra props, so we make them `$Exact<>` with a spread of Component's props and "own props" inside this exact object. But own props aren't removed or "consumed"m and they fall through to the underlying component at runtime (which can error).

There's have to either be more types to describe those interactions but we'd still need to trust the humans to not mistakes, or Flow would have to have some runtime knowledge (or hopefully with enough Flow coverage, Flow could maybe infer this, but I don't think we're there yet)

---

Also, check the tests for more examples, for now!

omninonsense:
Also note that `StyledComponent` is exported as a convenience for people who like to use the "styled-dot" notation rather than calling `styled()` with a string literal. Ideally, we'd pass all the types at call-time, instead of rely on Flow's inference (which is pretty damn good, but the problem is that it's difficult to describe some more stuff using the const-annotation approach, while super easy to do with the call position annotation). I use it quite often myself for simple components.

A quick "hack" regarding the theme... What I generally do is define a `flow-typed/types.js` in an project with the theme, if the project is small enough. If I'm writing a styleguide I import the theme explicitly, usually using something like 

```js
import { typeof default as Theme } from './theme';
```

I know it's a seems slightly cumbersome to import the theme every time, but I haven't found a better way yet.

ianduvall:
👏👏👏👏👏👏👏
Thanks this is great! Agreed, it would be great to move this into a README!

PR #3750, opened 2020-02-23
[WIP] Use custom flow-typed repository as source

moroine:

Other notes:

This PR adds the possibility to use a custom flow-typed repository as source in order to be able to fetch definitions from private repository.


moroine:
Draft PR due to missing tests

Brianzchen:
Hi @moroine how are plans to continue this PR? The idea is pretty cool and would provide an easy way to integrate flow into a large org where we may not have control over all repo's to add flow defs. I was thinking to build a custom installer solution for myself but this would solve that problem for others.

moroine:
@Brianzchen I didn't get time to continue on this PR 😢

Currently, I'm managing this using this kind of flowconfig

```
[ignore]
<PROJECT_ROOT>/node_modules/.*
!<PROJECT_ROOT>/node_modules/@nugit/.*

[declarations]
<PROJECT_ROOT>/node_modules/@nugit/.*

[options]
module.name_mapper='^@nugit/logger-service' -> '<PROJECT_ROOT>/node_modules/@nugit/logger-service/src/index.js'
module.name_mapper='^@nugit/env-service' -> '<PROJECT_ROOT>/node_modules/@nugit/env-service/src/index.js'
```

Brianzchen:
Hi @moroine this change actually looks like something that would change where flow-typed pulls it's flow-typed cache from, but that would mean this new source of truth would need to maintain every library. What if I want to just store private libraries and want to use it as a proxy like the issue I raised here #4142. Would that be more valuable?

And not sure if you plan to continue no this PR some day, but it looks like a really great change and lot of effort already put in!

moroine:
Yup, unfortunately, I didn't manage to finish this PR. I doesn't have the time currently 😬 .

I think the idea would be to be able to configure in a dotfile config which definition is pulled from which repo

PR #3752, opened 2020-02-24
[styled-components_v5] remove invalid types

omninonsense:
Removed the overload that never worked. It was a leftover from an unfinished experiment. :(

Added a few comments to silence the eslint noise.

### DIFF
--- a/definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
+++ b/definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
@@ -25,7 +25,7 @@ declare module 'styled-components' {
     | null
     | void
     | Styles
-    | Class<InterpolatableComponent<any>>; // CSS-in-JS object returned by polished are also supported, partially
+    | Class<InterpolatableComponent<any>>; // eslint-disable-line flowtype/no-weak-types
 
   declare export type TaggedTemplateLiteral<I, R> = (
     strings: string[],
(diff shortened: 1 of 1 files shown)


omninonsense:
CI errors unrelated to this version

definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
```diff
@@ -253,7 +253,7 @@ declare module 'styled-components' {
       ...Interpolation<PropsWithTheme<StyleProps, Theme>>[]
     ) => StyledComponent<StyleProps, Theme, V>,
     [[call]]: <StyleProps, Theme>(
-      (props: PropsWithTheme<StyleProps, Theme>) => Interpolation<any>
+      (props: PropsWithTheme<StyleProps, Theme>) => Interpolation<any> // eslint-disable-line flowtype/no-weak-types
     ) => StyledComponent<StyleProps, Theme, V>,
     +attrs: <A: { ... }, StyleProps = {||}, Theme = {||}>(
```
ianduvall: Should `StyleProps` and `Theme` use `: {||}` instead of `= {||}`?

definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
```diff
@@ -415,38 +415,13 @@ declare module 'styled-components' {
...
-  >;
-
   declare interface Styled {
-    <
-      Comp: StyledComponent<P, T, I>,
-      Theme: ExtractTheme<Comp>,
-      +Instance: ExtractInstance<Comp>,
-      OwnProps = ExtractProps<Comp>
-    >(
+    <Comp: React$ComponentType<P>, Theme, OwnProps = React$ElementConfig<Comp>>(
```
ianduvall: Same question here and line 516 (`OwnProps = React$ElementConfig<Comp>` to `OwnProps: React$ElementConfig<Comp>`)

ianduvall:
cc: @omninonsense

definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
omninonsense: No, if it used `: {||}` (adding a type, or "narrowing" the type) then it wouldn't accept any object apart from an empty object. The reason we have `= {||}` (default arg) is so people don't have to supply this themselves, since the default case is good enough most of the time. You can read more about it here: https://flow.org/en/docs/types/generics/#syntax-of-generics-

definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
omninonsense: See previous response about `:` vs `=`, in addition to that, I am not 100% sure how these "typed" or narrowed generics semantics work for the props-related types. It doesn't seem like there is any benefit of using them like that in this context (unless I was doing something wrong)

PR #3755, opened 2020-02-25
ES5 Core

RohanHart:
Based off Typescript's es5 definition and adjusted towards Flow's core libdef.
Requires the no_flowlib option.
Minimal tests unfortunately - pretty much just the ones that Flow's own core libdef has. Luckily the cli's tests seem to pick up these definitions over the built in ones.
Because this is not a real npm package I hope the cli can actually install it!


pascalduez:
Hit @RohanHart,

thanks for your contribution!

Can you explain the motivations for having those definitions in flow-type rather than in Flow's core?

RohanHart:
The Flow core libraries are targeting later versions of ES but I'm trying to use Flow with Rhino  (thanks NetSuite). There's a lot of additional methods and classes compared to ES5 and so code-completion and type checking don't give errors about those methods, etc., until runtime.

If Flow provided a way to specify which ES the code was targeting, as TypeScript does, then this should be part of Flow core.

pascalduez:
@RohanHart Could you please then include a Readme with all those explanations, or amend the def top comment. The clearer we are the better I guess.

pascalduez:
@RohanHart Thanks, you'd need to fix the test file naming.

```
• Unexpected file: core.es5_v5.1.0.js. This directory can only contain test files or a libdef file named `core.es5_v5.1.0.js`.
```

GAntoine:
Haha that's great, I think the presence of `.` in `core.es5` is messing with the cli. @pascalduez  The file is correctly named, but the cli is not happy with it.

RohanHart:
I didn't think to rerun validate-defs and that doesn't like the presence of the README. Moving it up one level fixes that

PR #3758, opened 2020-02-27
Fix definitions for @koa/router

julienw:
I realized that I got it wrong when I implemented it some months ago. Locally I was using an older version of the file that didn't have this problem.
Maybe this sort of issue could be found by automatic tests?


pascalduez:
>  Maybe this sort of issue could be found by automatic tests?

Yeah, sad that this passed unnoticed...

Thanks for the fix!

PR #3762, opened 2020-03-04
add support for styled-components/macro

penx:
add support for styled-components/macro, which exports the same as styled-components

- Links to documentation: https://styled-components.com/docs/tooling#babel-macro
- Link to GitHub or NPM: https://github.com/styled-components/styled-components/tree/master/packages/styled-components/src/macro

### DIFF
--- a/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
+++ b/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js
@@ -435,3 +435,7 @@ declare module 'styled-components/native' {
 
   declare export default Styled & ConvenientShorthands
 }
+
+declare module 'styled-components/macro' {
+  declare export * from 'styled-components';
+}
--- a/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/test_styled-components-macro.js
+++ b/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/test_styled-components-macro.js
@@ -1,0 +1,455 @@
+// @flow
+
+import { describe, it } from 'flow-typed-test'
+
+import * as React from 'react'
+import styled, {
+  createGlobalStyle,
+  css,
+  keyframes,
+  withTheme,
+  ThemeContext,
+
+  type CSSRules,
+  type KeyFrames,
+  type StyledComponent,
+} from 'styled-components/macro'
+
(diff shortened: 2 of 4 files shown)


penx:
the failing tests, as far as I can see, are for older versions of styled-components and unrelated to this change

AndrewSouthpaw:
Confirmed v4 and v5 all tests pass.

penx:
@AndrewSouthpaw I'm getting an odd issue with styled-components, I don't think it's related to this PR but someone want to check in case this needs to be reverted.

Unfortunately it's not something I can replicate in the unit tests as it requires exporting the component for it to come in to effect.

Specifically:

```js
import styled from 'styled-components';

const Red = styled.div`
  border: 1px solid red;
`;

export default Red;
```

or now:


```js
import styled from 'styled-components/macro';

const Red = styled.div`
  border: 1px solid red;
`;

export default Red;
```

I sometimes get this error:

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ flow-typed/npm/styled-components_v5.x.x.js:416:39

Missing type annotation for V. V is a type parameter declared in function type [1] and was implicitly instantiated at
$ObjMap [2].

     413│     ElementName
     414│   >;
     415│
 [2] 416│   declare type ConvenientShorthands = $ObjMap<
     417│     BuiltinElementInstances,
 [1] 418│     <V>(V) => StyledShorthandFactory<V>
     419│   >;
     420│
     421│   declare interface Styled {
     422│     <Comp: React$ComponentType<P>, Theme, OwnProps = React$ElementConfig<Comp>>(
```

The error seems to be erratic, in that sometimes Flow is reporting no errors.

Unfortunately I also cannot replicate this on https://flow.org/try due to this issue:

https://github.com/facebook/flow/issues/8313

cc @omninonsense

penx:
^ ignore that, sorry, it seems related to the fact that 'Red' isn't typed.

This works without error:
```js
// @flow
import type { StyledComponent } from 'styled-components';

import styled from 'styled-components/macro';

const Red: StyledComponent<{}, {}, HTMLDivElement> = styled.div`
  border: 1px solid red;
`;

export default Red;
```

PR #3697, opened 2020-01-10
Make history interface compatible with react-router

belka-ew:
- Links to documentation:
https://github.com/ReactTraining/history/tree/master/docs
https://github.com/ReactTraining/react-router/tree/master/packages/react-router/docs/api

- Type of contribution: fix

Other notes:

Should fix #1910.
Making `MemoryHistory.canGo` optional is actually wrong, but I don't see another way to work around flow bugs.
All `history/*` modules can be probably moved to `history`, it would reduce duplication. But I added `history` anyway becase normally everything is exported from `history`.
Properties I removed from `react-router-*` don't belong to the generic Location interface.

PR #3772, opened 2020-03-11
Add state and key back to the Location type of react-router

erogleva:
- Links to documentation: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/location.md
- Type of contribution: fix

I saw that in a previous PR these fields were removed; however, as visible from the documentation, these fields can appear in the react-router location and now whenever I try to access them, Flow shows an error about the missing properties. Would you consider adding them back?

### DIFF
--- a/definitions/npm/react-router-dom_v5.x.x/flow_v0.104.x-/react-router-dom_v5.x.x.js
+++ b/definitions/npm/react-router-dom_v5.x.x/flow_v0.104.x-/react-router-dom_v5.x.x.js
@@ -37,17 +37,20 @@ declare module "react-router-dom" {
 
   // NOTE: Below are duplicated from react-router. If updating these, please
   // update the react-router and react-router-native types as well.
-  declare export type Location = {
+  declare export type Location = $ReadOnly<{
     pathname: string,
     search: string,
     hash: string,
+    state?: any,
+    key?: string,
     ...
-  };
+  }>;
 
   declare export type LocationShape = {
     pathname?: string,
     search?: string,
     hash?: string,
+    state?: any,
     ...
   };
 
--- a/definitions/npm/react-router-native_v4.x.x/flow_v0.104.x-/react-router-native_v4.x.x.js
+++ b/definitions/npm/react-router-native_v4.x.x/flow_v0.104.x-/react-router-native_v4.x.x.js
@@ -21,17 +21,20 @@ declare module "react-router-native" {
 
   // NOTE: Below are duplicated from react-router. If updating these, please
   // update the react-router and react-router-dom types as well.
-  declare export type Location = {
+  declare export type Location = $ReadOnly<{
     pathname: string,
     search: string,
     hash: string,
+    state?: any,
+    key?: string,
     ...
-  };
+  }>;
 
   declare export type LocationShape = {
     pathname?: string,
     search?: string,
     hash?: string,
+    state?: any,
     ...
   };
 
--- a/definitions/npm/react-router_v5.x.x/flow_v0.104.x-/react-router_v5.x.x.js
+++ b/definitions/npm/react-router_v5.x.x/flow_v0.104.x-/react-router_v5.x.x.js
@@ -2,17 +2,20 @@ declare module "react-router" {
   // NOTE: many of these are re-exported by react-router-dom and
   // react-router-native, so when making changes, please be sure to update those
   // as well.
-  declare export type Location = {
+  declare export type Location = $ReadOnly<{
     pathname: string,
     search: string,
     hash: string,
+    state?: any,
+    key?: string,
     ...
-  };
+  }>;
 
   declare export type LocationShape = {
     pathname?: string,
     search?: string,
     hash?: string,
+    state?: any,
     ...
   };
 


pascalduez:
Those were removed in https://github.com/flow-typed/flow-typed/pull/3697

/cc @belka-ew Any insight?

belka-ew:
Yeah, the problem is that this definition isn't compatible with `MemoryLocation` (used in tests) and `HashLocation`.

```js
import type { Location } from "react-router"
import type { MemoryLocation } from "history/createMemoryHistory"

let loc: Location|null = null
let memoryLoc: MemoryLocation|null = null
loc = memoryLoc
```

> Cannot assign memoryLoc to loc because object type [1] is incompatible with undefined [2] in property state.
>
>     ...
>     28│
>     29│ let loc: Location|null = null
>     30│ let memoryLoc: MemoryLocation|null = null
>     31│ loc = memoryLoc
>     32│
>     33│
>     34│ axios.interceptors.response.use(R.identity,
>
>     flow-typed/npm/history_v4.x.x.js
> [1] 55│     state: {...},
>
>     flow-typed/npm/react-router_v5.x.x.js
> [2]  9│     state?: any,
>
>
>
> Found 1 error

erogleva:
I see. Would it then be correct to update the MemoryLocation and HashLocation definitions? According to the docs it seems that state is an optional field there too: https://github.com/ReactTraining/history/blob/3f69f9e07b0a739419704cffc3b3563133281548/docs/GettingStarted.md#listening

jedwards1211:
If `Location` were read-only you would be able to assign `MemoryLocation` to it.  I wonder if any of these shape types need to be mutable or inexact?  Though of course changing that might cause unexpected issues for people.

belka-ew:
`Location` is the generic interface for `MemoryLocation`, `HashLocation` and `BrowserLocation`. `BrowserLocation` and `MemoryLocation` have both, `key` and `state`; `HashLocation` don't have them. Therefore the documentation says "Locations **may** also have the following properties:", because they are not available in `HashLocation`, but they are always available in `MemoryLocation` and `BrowserLocation` (so not optional).

Now `history` itself doesn't distinguish between `MemoryLocation`, `HashLocation` and `BrowserLocation`, it always uses the same `createLocation()` function and just adds missing properties. I think this separation of Locations is coming from TS definitions, not from the `history` package itself.

Having said that, I'm not sure what is the proper solution.

maxmalov:
@belka-ew I wonder can we rethink `HashLocation` just a little to have `state` and `key` props which are explicitly undefined, i.e

```js
type HashLocation = {
    pathname: string,
    search: string,
    hash: string,
    state: void,
    key: void,
    ...
  }
```

jedwards1211:
@maxmalov @belka-ew have you considered using `$ReadOnly`?
```
type Location = $ReadOnly<{
  pathname: string,
  state?: any,
  ...
}>

type MemoryLocation = {
  pathname: string,
  state: {...},
  ...
}
          
let loc: Location|null = null
let memoryLoc: MemoryLocation|null = null
loc = memoryLoc // No errors!
```

maxmalov:
> Flow is erroring to prevent you from being able to set memoryLoc.state = null, so making Location read-only solves that error

@jedwards1211 Right now Flow is erroring that `BrowserLocation` has no state prop, because it was removed a few versions ago in `react-router` libdefs

belka-ew:
@jedwards1211 fine with me.

jedwards1211:
It might cause unexpected errors for people if they're mutating a `Location` for whatever reason.  Hopefully not?  It's more likely that they'd be mutating `LocationShape`

erogleva:
According to the docs, history.location is mutable https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/history.md#history-is-mutable but then, it's recommended not to access it from there

maxmalov:
@jedwards1211 Mutation is a little bit offtopic here. PR addresses the much simpler developer case since because of a single `HashLocation` (which I guess is not widely used) the rest location objects missing the field, which actually exists in the documentation and might be widely used across applications

pyrsv (commented):
it is good!
also need this in project

AndrewSouthpaw:
Would be happy to get to merge this in. Currently the discussion went stale, so I'm still waiting for some conclusion around that.

erogleva:
@belka-ew @jedwards1211 Would you consider the solution proposed by @maxmalov? As pointed out, it is already in the history libdefs and would solve the incompatibility issues.

> @belka-ew I wonder can we rethink `HashLocation` just a little to have `state` and `key` props which are explicitly undefined, i.e
> 
> ```js
> type HashLocation = {
>     pathname: string,
>     search: string,
>     hash: string,
>     state: void,
>     key: void,
>     ...
>   }
> ```
> 
> UPD: Looks like it is already in `history` libdefs https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/history_v4.10.x/flow_v0.104.x-/history_v4.10.x.js#L72

Anyway, I tried to look into this again but when I ran the tests again with the code, I got many warnings and some errors so I thought it would be better to fix these first: #3908

jedwards1211:
I guess, it seems like people wouldn't work around the issue that way if there were a cleaner solution...

jedwards1211:
> history.location is mutable
That was about the history object being mutable, not location objects

erogleva:
> > history.location is mutable
> 
> That was about the history object being mutable, not location objects. Though nothing prevents anyone from mutating a location object, it would be a bad idea and might should be flagged as an error by using a readonly flow type for it.

Ok, I guess I have misinterpreted what the docs say [here](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/location.md) about location. I will update the code so that Location is declared as a readonly type.

AndrewSouthpaw:
Ok it sounds like that resolves the conversation then. If it doesn't, feel free to chime in and I can either revert it for now or approve follow-up work.

PR #3763, opened 2020-03-05
Lodash - replace Array<> with $ReadOnlyArray<> in all remaining functions, which do not mutate arrays in their arguments

DmitryGonchar:
- Type of contribution: fix, refactor

This PR should close #1099 as it adds $ReadOnlyArray<> instead of Array<> to all remaining functions, which do not mutate input array.

Other notes:
There were many "partial fix" PRs before this one, as can be seen in #1099 . This PR fixes remaining functions and allows to pass readonly arrays to them.

Was checking with [lodash docs](https://lodash.com/docs/4.17.15) to verify if input array is mutated or not

Change is big, because I thought it is easier to redactor code a bit, especially have tests in sorted order. Therefore, I've split the independent work into separate commits. _It may be easier to check PR commit by commit_

PR #3777, opened 2020-03-17
Lodash - use $ReadOnly object for arguments

DmitryGonchar:
* Type of contribution: fix, refactor, tests

1. Lodash functions should accept $ReadOnly objects as arguments for functions which do not mutate them. Right now it does not.
This is also in line with guides in CONTRIBUTING doc - to prefer $ReadOnly objects over just `{}`;
This fix is similar to #3763 but it is for objects instead of arrays.

1.  This PR also adds some unit tests:
* test for code examples from docs for all Collections functions
* Reusing tests for functions which must have exactly same types according to docs (e.g. `each/eachRight/forEach/forEachRight`) . THis has helped to uncover some inconsistencies in them.
* tests for readonly objects - for functions with non-trivial changes. Can add to more.
* add testing of return types for all existing Collection functions. Adding some ExpectError tests - not to allow `any` as a return type.

Much more tests could have been added, but adding them often uncovers existing bugs in types (for example, each(), every(), flatMap() used to hide some errors). This PR includes some of the fixes, but adding more would increase the size of the code changes and risk of merging it.

3. Other (unplanned) fixes include:
* `T: Object` would allow T to be number, string, array, any etc. In functions where these values are not supported (at least there is no meaningful result returned from lodash). Such places were replaced with indexer object.
* `countBy`
    * support string in collection argument
* `each/eachRight/forEach/forEachRight`:
    * aligning to have same types (as they should per docs). Adding tests to enforce this. Is it possible to reuse a variable for their definition?
    * type iteratee arguments and return type
    * return the type of passed collection correctly now, previously e.g. readonly array would be changed to writeable array.
    * add support for opaque types keys (opaque values were already supported)
    * not to accept object as iteratee, 
* `filter/reject`:
    * aligning to have same types, adding tests to enforce this
* `every`
    * support string in collection argument
*`find/findLast`:
    * aligning to have same types, adding tests to enforce this
    * support `null` in collection argument
* `reduce/reduceRight`:
    * aligning to have same types, adding tests to enforce this
    * returns accumulator type instead of void if void or null passed to it as first argument. 
* `find`:
    * support null for iteratee and fromIndex args. This is useful e.g. to allow nullable objects to be passed to the function.
* `flatMap/flatMapDeep/flatMapDepth`:
    * do not accept object as iteratee argument anymore, same as in lodash docs 
    * add support for opaque keys

All changes have been checked in code sandbox that lodash actually does something meaningful for these news supported inputs.

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
@@ -168,10 +168,10 @@ declare module "lodash" {
   };
 
   declare type Key = string | number;
-  declare type IndexerObject<T, K = Key> = { [key: K]: T, ... };
-  declare type ReadOnlyIndexerObject<T, K = Key> = $ReadOnly<IndexerObject<T, K>>;
-  declare type NestedArray<T> = Array<Array<T>>;
-  declare type Collection<T> = $ReadOnlyArray<T> | ReadOnlyIndexerObject<T>;
+  declare type IndexerObject<V, K = Key> = { [key: K]: V, ... };
+  declare type ReadOnlyIndexerObject<V, K = Key> = $ReadOnly<IndexerObject<V, K>>;
+  declare type NestedArray<V> = Array<Array<V>>;
+  declare type Collection<V> = $ReadOnlyArray<V> | ReadOnlyIndexerObject<V>;
 
   declare type matchesIterateeShorthand = { [key: any]: any, ... };
   declare type matchesPropertyIterateeShorthand = [string, any];
@@ -553,9 +553,9 @@ declare module "lodash" {
     countBy<T>(array: void | null, iteratee?: ?ValueOnlyIteratee<T>): {...};
     countBy<T>(object: ReadOnlyIndexerObject<T>, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
     // alias of _.forEach
-    each<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<T> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
+    each<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<A> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
     // alias of _.forEachRight
-    eachRight<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<T> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
+    eachRight<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<A> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
     every<T>(array?: ?$ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): boolean;
     every<A, T: ReadOnlyIndexerObject<A>>(object: T, iteratee?: OIterateeWithResult<A, string, T, any>): boolean;
     filter<T>(array?: ?$ReadOnlyArray<T>, predicate?: ?Predicate<T>): Array<T>;
@@ -593,34 +593,34 @@ declare module "lodash" {
       predicate?: ?OPredicate<A, T>,
       fromIndex?: ?number
     ): V;
-    flatMap<A, K, U, T: $ReadOnlyArray<T>>(
+    flatMap<A, K, U, T: $ReadOnlyArray<A> = Array<A>>(
       array: T,
       iteratee?: ?AFlatMapIteratee<A, T, U>
     ): Array<U>;
-    flatMap<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<T>>(
+    flatMap<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<A>>(
       object: T,
       iteratee?: ?OFlatMapIteratee<A, K, T, U>
     ): Array<U>;
-    flatMapDeep<A, U, T: $ReadOnlyArray<T> = Array<T>>(
+    flatMapDeep<A, U, T: $ReadOnlyArray<A> = Array<A>>(
       array: T,
       iteratee?: ?AFlatMapIteratee<A, T, any>
     ): Array<U>;
-    flatMapDeep<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<T>>(
+    flatMapDeep<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<A>>(
       object: T,
       iteratee?: ?OFlatMapIteratee<A, K, T, any>
     ): Array<U>;
-    flatMapDepth<A, U, T: $ReadOnlyArray<T> = Array<T>>(
+    flatMapDepth<A, U, T: $ReadOnlyArray<A> = Array<A>>(
       array: T,
       iteratee?: ?AFlatMapIteratee<A, T, any>,
       depth?: ?number
     ): Array<U>;
-    flatMapDepth<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<T>>(
+    flatMapDepth<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<A>>(
       object: T,
       iteratee?: ?OFlatMapIteratee<A, K, T, any>,
       depth?: ?number
     ): Array<U>;
-    forEach<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<T> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
-    forEachRight<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<T> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
+    forEach<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<A> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
+    forEachRight<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<A> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
     groupBy<V, T>(
       array: $ReadOnlyArray<T>,
       iteratee?: ?ValueOnlyIteratee<T>
--- a/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
@@ -168,7 +168,8 @@ declare module "lodash" {
   };
 
   declare type Key = string | number;
-  declare type ReadOnlyIndexerObject<T, I = Key> = $ReadOnly<{ [id: I]: T, ... }>
+  declare type IndexerObject<T, I = Key> = { [id: I]: T, ... };
+  declare type ReadOnlyIndexerObject<T, I = Key> = $ReadOnly<IndexerObject<T, I>>;
   declare type NestedArray<T> = Array<Array<T>>;
   declare type Collection<T> = $ReadOnlyArray<T> | ReadOnlyIndexerObject<T>;
 
@@ -192,9 +193,9 @@ declare module "lodash" {
   declare type OIteratee<O> = OIterateeWithResult<any, string, O, any>;
 
   declare type AFlatMapIteratee<T, O, U> =
-    | ((item: T, index: number, array: O) => Array<U>)
-    | string;
-  declare type OFlatMapIteratee<V, T, U, K = string> = IterateeWithResult<V, K, T, Array<U>>;
+    | ((item: T, index: number, array: O) => Array<U> | U)
+    | string
+  declare type OFlatMapIteratee<V, K, T, U> = IterateeWithResult<V, K, T, Array<U> | U>;
 
   declare type Predicate<T> =
     | ((value: T, index: number, array: Array<T>) => any)
@@ -592,30 +593,30 @@ declare module "lodash" {
       predicate?: ?OPredicate<A, T>,
       fromIndex?: ?number
     ): V;
-    flatMap<A, U, T: ?$ReadOnlyArray<T> = Array<T>>(
+    flatMap<A, K, U, T: $ReadOnlyArray<T>>(
       array: T,
       iteratee?: ?AFlatMapIteratee<A, T, U>
     ): Array<U>;
-    flatMap<A, U, T: ?ReadOnlyIndexerObject<A> = ReadOnlyIndexerObject<T>>(
+    flatMap<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<T>>(
       object: T,
-      iteratee?: ?OFlatMapIteratee<A, T, U>
+      iteratee?: ?OFlatMapIteratee<A, K, T, U>
     ): Array<U>;
-    flatMapDeep<A, U, T: ?$ReadOnlyArray<T> = Array<T>>(
+    flatMapDeep<A, U, T: $ReadOnlyArray<T> = Array<T>>(
       array: T,
       iteratee?: ?AFlatMapIteratee<A, T, any>
     ): Array<U>;
-    flatMapDeep<A, U, T: ?ReadOnlyIndexerObject<A> = ReadOnlyIndexerObject<T>>(
+    flatMapDeep<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<T>>(
       object: T,
-      iteratee?: ?OFlatMapIteratee<A, T, any>
+      iteratee?: ?OFlatMapIteratee<A, K, T, any>
     ): Array<U>;
-    flatMapDepth<A, U, T: ?$ReadOnlyArray<T> = Array<T>>(
+    flatMapDepth<A, U, T: $ReadOnlyArray<T> = Array<T>>(
       array: T,
       iteratee?: ?AFlatMapIteratee<A, T, any>,
       depth?: ?number
     ): Array<U>;
-    flatMapDepth<A, U, T: ?ReadOnlyIndexerObject<A> = ReadOnlyIndexerObject<T>>(
+    flatMapDepth<A, K, U, T: ?ReadOnlyIndexerObject<A> | string = IndexerObject<T>>(
       object: T,
-      iteratee?: ?OFlatMapIteratee<A, T, any>,
+      iteratee?: ?OFlatMapIteratee<A, K, T, any>,
       depth?: ?number
     ): Array<U>;
     forEach<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<T> | string | void | null>(object: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
--- a/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
@@ -758,8 +758,8 @@ declare module "lodash" {
     ): Array<A>;
     sample<T>(collection: ?Collection<T>): T;
     sampleSize<T>(collection?: ?Collection<T>, n?: ?number): Array<T>;
-    shuffle<T>(array: ?Collection<T>): Array<T>;
-    size(collection: Collection<any> | string): number;
+    shuffle<T>(array?: ?Collection<T>): Array<T>;
+    size(collection?: ?Collection<any> | string): number;
     some<T>(array: void | null, predicate?: ?Predicate<T>): false;
     some<T>(array: ?$ReadOnlyArray<T>, predicate?: Predicate<T>): boolean;
     some<A, T: ReadOnlyIndexerObject<A>>(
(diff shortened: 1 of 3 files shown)


definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -179,12 +182,18 @@ declare module "lodash" {
...
 
+  declare type IterateeWithResult<V, O, R> =
+    | ((value: V, key: string, object: O) => R)
+    | string;
+
   declare type OIterateeWithResult<V, O, R> =
-    | Object
-    | string
-    | ((value: V, key: string, object: O) => R);
+    | ReadOnlyIndexerObject<V>
```
DmitryGonchar: `Object` is `any` in Flow, unfortunately. Therefore, it was replaced with more precise type

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -167,9 +167,12 @@ declare module "lodash" {
     ...
   };
 
+  declare type Key = string | number;
```
DmitryGonchar: It looks like object key cannot be anything else: http://www.ecma-international.org/ecma-262/#sec-object-initializer

Can add `| symbol` for Symbol, but it looks like Flow does not support it atm: https://github.com/facebook/flow/issues/3258.

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -200,10 +209,6 @@ declare module "lodash" {
     array: ?Array<T>
   ) => mixed;
   declare type Iteratee<T> = _Iteratee<T> | Object | string;
-  declare type FlatMapIteratee<T, U> =
```
DmitryGonchar: this was used for iteratees in functions which accept arrays, so it was renamed to `AFlatMapIteratee`, to make it more consistent with `OFlatMapIteratee` naming style

`| Object` was removed because
1. `Object` is `any`
1. according to docs, iteratee in all flatMap functions should be function, object is not mentioned, and also does not do anything useful if passed

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -544,19 +549,19 @@ declare module "lodash" {
...
     // alias of _.forEach
     each<T>(array: $ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): Array<T>;
     each<T: void | null>(array: T, iteratee?: ?Iteratee<any>): T;
-    each<T: Object>(object: T, iteratee?: ?OIteratee<T>): T;
+    each<A, T: ReadOnlyIndexerObject<A>>(object: T, iteratee?: ?IterateeWithResult<A, T, boolean | void>): T;
     // alias of _.forEachRight
     eachRight<T>(array: $ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): Array<T>;
     eachRight<T: void | null>(array: T, iteratee?: ?Iteratee<any>): T;
-    eachRight<T: Object>(object: T, iteratee?: OIteratee<T>): T;
+    eachRight<A, T: ReadOnlyIndexerObject<A>>(object: T, iteratee?: ?IterateeWithResult<A, T, boolean | void>): T;
```
DmitryGonchar: iteratee can be `null` now, to match `each/forEach` function definition

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -544,19 +549,19 @@ declare module "lodash" {
     // Collection
     countBy<T>(array: $ReadOnlyArray<T>, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
     countBy<T>(array: void | null, iteratee?: ?ValueOnlyIteratee<T>): {...};
-    countBy<T: Object>(object: T, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
+    countBy<T>(object: ReadOnlyIndexerObject<T>, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
     // alias of _.forEach
     each<T>(array: $ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): Array<T>;
     each<T: void | null>(array: T, iteratee?: ?Iteratee<any>): T;
-    each<T: Object>(object: T, iteratee?: ?OIteratee<T>): T;
+    each<A, T: ReadOnlyIndexerObject<A>>(object: T, iteratee?: ?IterateeWithResult<A, T, boolean | void>): T;
```
DmitryGonchar: this increases number of type parameters, which can require a change in the code, if they were specified explicitly. 

First option is to remove type T from parameters and specify `ReadOnlyIndexerObject<A>` instead of it in the type definition

Second option is to provide a default value, but that will make type parameter code a bit cumbersome `T: ReadOnlyIndexerObject<A> = ReadOnlyIndexerObject<A>`

But in the end, I've noticed that other existing function already use this style, so I've left it unchanged for now.

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -570,61 +575,67 @@ declare module "lodash" {
...
     forEachRight<T: void | null>(array: T, iteratee?: ?Iteratee<any>): T;
-    forEachRight<T: Object>(object: T, iteratee?: ?OIteratee<T>): T;
+    forEachRight<A, T: ReadOnlyIndexerObject<A>>(object: T, iteratee?: ?IterateeWithResult<A, T, boolean | void>): T;
     groupBy<V, T>(
       array: $ReadOnlyArray<T>,
       iteratee?: ?ValueOnlyIteratee<T>
     ): { [key: V]: Array<T>, ... };
     groupBy(array: void | null, iteratee?: ?ValueOnlyIteratee<any>): {...};
-    groupBy<V, A, T: { [id: any]: A, ... }>(
+    groupBy<V, A, T: ReadOnlyIndexerObject<A>>(
```
DmitryGonchar: Btw, I'm not sure why A is used for value type, I'd use V for it. But was following current naming convention for consistency

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/test_lodash-v4.x.x.js
```diff
@@ -202,56 +231,128 @@ describe('Collection', () => {
...
+    countBy(["one", "two", "three"], (v: boolean) => v);
+    // $ExpectError
+    countBy(["one", "two", "three"], (v, arg2: number) => arg2);
   });
 
+  /**
+   * Using such functions allows to:
+   * 1) Ensure all tests are run and up to date for each function
+   * 2) Enforce that type definitions for all related functions (e.g. each, eachRight, forEach, forEachRight) are the same, as they should be by the docs.
+   */
```
DmitryGonchar: I've wanted to addd this as a PR comment initially, but then thought that the purpose will be clear for more people if it's put in code

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -570,61 +575,67 @@ declare module "lodash" {
...
+    find<V, A, T: ReadOnlyIndexerObject<A>>(
       object: T,
-      predicate?: OPredicate<A, T>,
-      fromIndex?: number
+      predicate?: ?OPredicate<A, T>,
+      fromIndex?: ?number
     ): V;
     findLast<T>(
-      array: ?$ReadOnlyArray<T>,
+      array: $ReadOnlyArray<T>,
```
DmitryGonchar: this case is handled now in the `array: void | null`  case below - same as for find()

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -570,61 +575,67 @@ declare module "lodash" {
...
+    findLast<T>(
+      array: void | null,
+      predicate?: ?Predicate<T>,
+      fromIndex?: ?number
+    ): void;
+    findLast<V, A, T: ReadOnlyIndexerObject<A>>(
       object: T,
-      predicate?: ?OPredicate<A, T>
+      predicate?: ?OPredicate<A, T>,
+      fromIndex?: ?number
```
DmitryGonchar: reusing tests has helped to find that this argument was missing from findLast() types

DmitryGonchar:
I see some Flow errors in our repo with these changes (which are legit, it's just that Object does not hide real errors anymore). I will need to improve some types further (and add tests for new improvements)

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -544,19 +551,15 @@ declare module "lodash" {
     // Collection
     countBy<T>(array: $ReadOnlyArray<T>, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
     countBy<T>(array: void | null, iteratee?: ?ValueOnlyIteratee<T>): {...};
-    countBy<T: Object>(object: T, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
+    countBy<T>(object: ReadOnlyIndexerObject<T>, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
     // alias of _.forEach
-    each<T>(array: $ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): Array<T>;
-    each<T: void | null>(array: T, iteratee?: ?Iteratee<any>): T;
-    each<T: Object>(object: T, iteratee?: ?OIteratee<T>): T;
+    each<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<T> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
```
DmitryGonchar: 3 type defs look unneeded in this case so they were merged into one.
Another reason was to make newly added tests pass. E.g. previously each() for array would return writeable array even if it was passed a readonly collection
https://github.com/flow-typed/flow-typed/pull/3777/commits/0fffa7196438089079113bbd2aa8d9792b3573ce

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -544,19 +551,15 @@ declare module "lodash" {
...
     // alias of _.forEachRight
-    eachRight<T>(array: $ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): Array<T>;
-    eachRight<T: void | null>(array: T, iteratee?: ?Iteratee<any>): T;
-    eachRight<T: Object>(object: T, iteratee?: OIteratee<T>): T;
+    eachRight<A, K, T: ReadOnlyIndexerObject<A> | $ReadOnlyArray<T> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
     every<T>(array?: ?$ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): boolean;
-    every<T: Object>(object: T, iteratee?: OIteratee<T>): boolean;
+    every<A, T: ReadOnlyIndexerObject<A>>(object: T, iteratee?: OIterateeWithResult<A, string, T, any>): boolean;
```
DmitryGonchar: return type is `any` because docs say `returns falsey`. Which can be anything. I don't think it can be typed better, apart from making it a generic type parameter. But since it is not the type returned from the function itself, I think any is fine here

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/test_lodash-v4.x.x.js
```diff
@@ -59,11 +76,46 @@ import xorBy from "lodash/xorBy";
...
+type ObjectOrNull = IndexerObject | null;
+const objectOrNull: ObjectOrNull = readOnlyArray[0] > 100 ? indexerObject : null;
+
+type ObjectOrVoid = IndexerObject | void;
+const objectOrVoid: ObjectOrVoid = readOnlyArray[0] > 100 ? indexerObject : undefined;
+
+opaque type OpaqueKey = number;
+const toOpaqueKey = (v: number): OpaqueKey => v;
+type ObjectWithOpaqueKey = { [OpaqueKey]: string, ... };
+const objectWithOpaqueKey: ObjectWithOpaqueKey  = { [toOpaqueKey(1)]: 'a', [toOpaqueKey(2)]: 'b' };
```
DmitryGonchar: Unfortunately, I cannot test errors for opaque types properly, because these types are opaque only "outside of the file in which they are defined". 
I've tried to create a sibling file, but it says "module is not found" for 
`import {type OpaqueKey } from './test-opaque-types-v4.x.x'


So, testing for example that `forEach(opaqueObject, (v, k: string) => !k);`

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/test_lodash-v4.x.x.js
```diff
@@ -59,11 +76,46 @@ import xorBy from "lodash/xorBy";
...
+
+// test objects
+type IndexerObject = { [string]: number, ... };
+const indexerObject: IndexerObject = { a: 1, b: 2, c: 3 };
+
+type ExactObject = {| a: number, b: number, c: number; |};
+const exactObject: ExactObject = { a: 1, b: 2, c: 3 };
+
+type ExactHeterogeneousObject = {| a: number, b: string, c: boolean; |};
+const exactHeterogeneousObject: ExactHeterogeneousObject = { a: 1, b: 'abc', c: true };
```
DmitryGonchar: I'm thinking, maybe I should just make `exactObject` values to be of different types and use one test case instead of 2?

DmitryGonchar:
@GAntoine this PR is ready to be reviewed :)

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -544,19 +552,17 @@ declare module "lodash" {
     // Collection
     countBy<T>(array: $ReadOnlyArray<T>, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
     countBy<T>(array: void | null, iteratee?: ?ValueOnlyIteratee<T>): {...};
-    countBy<T: Object>(object: T, iteratee?: ?ValueOnlyIteratee<T>): { [string]: number, ... };
+    countBy(string: string, iteratee?: ?ValueOnlyIteratee<string>): { [string]: number, ... };
```
GAntoine: I don't think `string` is a valid type for the `collection` argument.

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
```diff
@@ -544,19 +552,17 @@ declare module "lodash" {
...
-    each<T: Object>(object: T, iteratee?: ?OIteratee<T>): T;
+    each<A, K, T: ReadOnlyIndexerObject<A, K> | $ReadOnlyArray<A> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
     // alias of _.forEachRight
-    eachRight<T>(array: $ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): Array<T>;
-    eachRight<T: void | null>(array: T, iteratee?: ?Iteratee<any>): T;
-    eachRight<T: Object>(object: T, iteratee?: OIteratee<T>): T;
+    eachRight<A, K, T: ReadOnlyIndexerObject<A, K> | $ReadOnlyArray<A> | string | void | null>(collection: T, iteratee?: ?IterateeWithResult<A, K, T, boolean | void>): T;
     every<T>(array?: ?$ReadOnlyArray<T>, iteratee?: ?Iteratee<T>): boolean;
-    every<T: Object>(object: T, iteratee?: OIteratee<T>): boolean;
+    every(str: string, iteratee?: ?Iteratee<string>): boolean;
```
GAntoine: Same here, though let me know if I'm missing something...

lukeapage:
> @lukeapage if you have some time, could one of you take a look at this?

I tried running it on our codebase - 130 new errors.. mostly
1. flatmap changes in provided typed
2. now requiring types to be provided to disambiguate
3. correctly working out typed for map where as before it must have ended up as `any` and been silently untyped

I skimmed the problems and didn't see any obvious library ones - mostly ones in our code. A few extra annoyances e.g. we have a call like

```
_.compact([ condition && thingA]);
```
and now compact doesn't take booleans - it expects everything to be `thingA` or null/undefined.
Not sure how to solve that - I tried editing the library to make `false` be allowed but flow isn't clever enough to work out the only boolean value being passed is `false`. Probably in this case I'll refactor or suppress the error.

Looks good, though not looking forward to fixing the errors its found.

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
lukeapage: You can always go to lodash.com and run lodash in the console to try it out.
It does work
e.g.
```
_.every('abc', (a) => (console.log(a), true));
```
As to whether its luck or intetentional, you'd have to ask Dalton
I agree its not documented.
Probably luck...

GAntoine: Ah ok, well if the library doesn't complain then I'm happy with this change.

definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
DmitryGonchar: Yes, it's supported by lodash. My guess is that every collection and probably array method supports string (as array of chars). I just did not want to add it in this PR, not to bloat its scope.

In docs page, you can click "Try in REPL" https://lodash.com/docs/4.17.15#every and test any code
`_.every('aaa', (c) => c === 'a');` returns true
`_.every('aab', (c) => c === 'a');` returns false

DmitryGonchar:
@lukeapage Was not able to Reply to your comment, so answering in a separate one:

+1, we've had same existing code errors uncovered. That's why I've initially closed PR - to check if they're caused by lib def changes or are legit errors. They were legit errors, plus some places were passing string to function which has not been supported before and was not complaining because of ` | Object` clause.

> 2. now requiring types to be provided to disambiguate
Can you please specify for which method and in which case? I've tried to avoid that and fixed a couple of methods not to have this errors. E.g. when both array and object defs where nullable `x?: ?T`, Flow complained it cannot choose between the two when passed object was nullable. I've made only one def to be nullable instead of two and the problem went away.

> _.compact([ condition && thingA]);
As you can see in this PR changes, I did not touch this method. So this problem is most likely caused by some other merged PR.
To fix it - I think you should either specify `count<myType | false>` when using the function, or change it to ternary `condition ? thingA : undefined`. Ideally, flow should auto-detect and support first solution without specifying type explicitly. But again, I have not looked into this function, so I don't know why. Feel free to add your test case to tests and fix the lib def :) 


@GAntoine You've requested changes, but I'm not able to find - which ones? I just see 2 resolved comments

DmitryGonchar:
@GAntoine How about merging this one?

DmitryGonchar:
@AndrewSouthpaw ^ There has been no comments for more than a week now

GAntoine:
Sorry about the wait, got busy with work. Lets keep an eye on the issues over the next few weeks, in case someone has trouble adapting to the type changes.

DmitryGonchar:
Sure. Happy to help fixing those in the libdef

draperunner:
@DmitryGonchar I see this `K` appeared in your commit. It seems unused for the array version of flatMap. What type should I use here? Just an underscore?

```
flatMap<A, K, U, T: $ReadOnlyArray<A> = Array<A>>(
     array: T,
     iteratee?: ?AFlatMapIteratee<A, T, U>
   ): Array<U>;
```

DmitryGonchar:
Thanks! 

Yes, you can use `_`. If it does not work because Flow cannot infer types - use anything (like `string`), since this type is not used anywhere.

I'll push a fix for it today. Checked other functions - this is the only one where `K` is unused.

draperunner:
Yes, I was able to fix this by just removing the generics and let Flow infer the types. Thanks for answering and fixing this so quickly, though!

lukeapage:
I'm trying the upgrade for real and I'm stuck here (example simplified so it might not make sense):

```
export const getX = (
    input: $ReadOnlyArray<{ id: string, ... }>,
) => {
    const projection = ({ id }: { id: string, ... }): $ReadOnlyArray<string> =>[id];

    const fields: $ReadOnlyArray<string> = _.flatMap<
        { id: string, ... },
        string,
        $ReadOnlyArray<{ id: string, ... }>,
    >(input, projection);
```

I get:

```
Error ------------------------------------------------------ 
Cannot call _.flatMap because read-only array type [1] is incompatible with IndexerObject [2].                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           test.js                                                                                                                                                                                                                                          [1]  89\|     input: $ReadOnlyArray<{ id: string, ... }>,                                                                                                                                                                                                                                     :                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              101\|                                                                                                                                                                                                                                                                                              102\|     const fields: $ReadOnlyArray = _.flatMap<                                                                                                                                                                                                                  103\|         { id: string, ... },                                                                                                                                                                                                                                                                 104\|         string,                                                                                                                                                                                                                                                                              105\|         $ReadOnlyArray<{ id: string, ... }>,                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   flow-typed/npm/lodash_v4.x.x.js                                                                                                                                                                                                                                                               [2] 606\|     flatMap \| string = IndexerObject>(
--
```

It seems like flow cannot work out if its an array or an object and instead of disregarding the object case, it complains its invalid

lukeapage:
@GAntoine I tracked this down to the above PR: https://github.com/flow-typed/flow-typed/pull/3823

PR #3785, opened 2020-03-21
fs-extra: inherit fs library types

pirroman:
Right now fs-extra types cover only the new methods that were added to fs-extra. I have added fs library types to cover fs methods that are also available in fs-extra.

### DIFF
--- a/definitions/npm/fs-extra_v7.x.x/flow_v0.104.x-/fs-extra_v7.x.x.js
+++ b/definitions/npm/fs-extra_v7.x.x/flow_v0.104.x-/fs-extra_v7.x.x.js
@@ -1,547 +1,535 @@
 declare module "fs-extra" {
   import type { Stats, ReadStream, WriteStream } from "fs";
+  import typeof fsTypes from "fs";
 
-  declare export type SymlinkType = "dir" | "file";
-  declare export type FsSymlinkType = "dir" | "file" | "junction";
+  declare type SymlinkType = "dir" | "file";
+  declare type FsSymlinkType = "dir" | "file" | "junction";
 
-  declare export type CopyFilterSync = (src: string, dest: string) => boolean;
-  declare export type CopyFilterAsync = (
+  declare type CopyFilterSync = (src: string, dest: string) => boolean;
+  declare type CopyFilterAsync = (
     src: string,
     dest: string
   ) => Promise<boolean>;
 
-  declare export type CopyOptions = {
+  declare type CopyOptions = {
     dereference?: boolean,
     overwrite?: boolean,
     preserveTimestamps?: boolean,
     errorOnExist?: boolean,
-    recursive?: boolean,
-    ...
+    recursive?: boolean
   };
 
-  declare export type CopyOptionsAync = CopyOptions & { filter?: CopyFilterSync | CopyFilterAsync, ... };
+  declare type CopyOptionsAync = CopyOptions & {
+    filter?: CopyFilterSync | CopyFilterAsync
+  };
 
-  declare export type CopyOptionsSync = CopyOptions & { filter?: CopyFilterSync, ... };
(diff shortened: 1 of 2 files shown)


definitions/npm/fs-extra_v7.x.x/flow_v0.104.x-/fs-extra_v7.x.x.js
```diff
@@ -1,773 +1,832 @@
...
 
-  declare export function mkdtemp(prefix: string): Promise<string>;
-  declare export function mkdtemp(
+  declare function mkdtemp(prefix: string): Promise<string>;
+  declare function mkdtemp(
     prefix: string,
     callback: (err: ErrnoError, folder: string) => void
   ): void;
+
+  declare module.exports: {
```
pascalduez: By removing all the `export` and using `module.exports` you change this libdef from ES modules to commonjs. It should ba avoided when possible.

definitions/npm/fs-extra_v7.x.x/flow_v0.104.x-/fs-extra_v7.x.x.js
```diff
@@ -1,773 +1,832 @@
...
 
-  declare export type CopyOptions = {
+  declare type CopyOptions = {
     dereference?: boolean,
     overwrite?: boolean,
     preserveTimestamps?: boolean,
     errorOnExist?: boolean,
-    recursive?: boolean,
-    ...
+    recursive?: boolean
```
pascalduez: Please don't remove all those inexact notations `...`, they are intended. Or convert those type to exact.

pirroman: Thank you! I fixed it locally at my place a long time ago, but there was no time to make a pr. I was sure that I had the relevant version of the source and that nobody had had time to make any commits here yet. Last commit 1 August 2019. I fixed it about September 2019, but apparently I already had an out-of-date version. I rolled it back.

definitions/npm/fs-extra_v7.x.x/flow_v0.104.x-/fs-extra_v7.x.x.js
pirroman: I totally agree with you! But I have no idea how to combine type spreading and ES Module.

pascalduez: Untested: `declare export default` ?

pascalduez:
```js
declare export default $ReadOnly<{|
  ...fsTypes;
  access: typeof access;
  appendFile: typeof appendFile;
  chmod: typeof chmod;
  // [...]
|}
```

pirroman: I remembered that the problem was with this notation approach I get errors for the next code:
```javascript
if( ... ) {
    const fs = require('fs-extra');
    const data = fs.readFileSync(argsFileName, { encoding: 'utf8' });
...
```
`// Error: Cannot call fs.readFileSync because property readFileSync is missing in  module 'fs-extra' [1].`

I would also add that `fs-extra` itself uses commonjs  export format.

definitions/npm/fs-extra_v7.x.x/flow_v0.104.x-/fs-extra_v7.x.x.js
pascalduez: Okay yes, that's a bit weird to force ES modules for a node lib like this one.

Thanks for the PR!

PR #3786, opened 2020-03-21
fix(getInstalledNpmLibDefs): fix <<STUB>> parsing

jedwards1211:
partial fix for #3363

- Type of contribution: fix

Other notes:

We should still probably make getInstalledNpmLibDefs tolerate any bad version strings.

### DIFF
--- a/cli/src/lib/npm/npmLibDefs.js
+++ b/cli/src/lib/npm/npmLibDefs.js
@@ -359,90 +359,128 @@ export async function findNpmLibDef(
 type InstalledNpmLibDef =
   | {|kind: 'LibDef', libDef: NpmLibDef|}
   | {|kind: 'Stub', name: string|};
+
+type ParsedSignedCodeVersion =
+  | {|
+      kind: 'LibDef',
+      libDef: $Diff<NpmLibDef, {|path: string|}>,
+    |}
+  | {|kind: 'Stub', name: string|};
+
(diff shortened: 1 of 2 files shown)


cli/src/lib/npm/npmLibDefs.js
```diff
@@ -359,90 +359,128 @@ export async function findNpmLibDef(
...
+): ?ParsedSignedCodeVersion {
+  if (signedCodeVer === null) {
+    return null;
+  }
+  if (signedCodeVer.startsWith('<<STUB>>/')) {
+    return {
+      kind: 'Stub',
+      name: signedCodeVer.substring('<<STUB>>/'.length),
+    };
+  }
```
GAntoine: So is moving up this `STUB` check the biggest change here? This diff is kinda hard to read, let me know if there are other changes in this file. Nice refactor by the way. 👌

jedwards1211: yeah it's the biggest change.  The extracted `getInstalledNpmLibDef` I thought I would need for testing before I realized I only needed to test the `parseSignedCodeVersion` function for now.

GAntoine: Cool. With everything else being the same, I'll pull this change locally and test it tomorrow. If all looks good, we'll push out a patch to npm in the short term.

PR #3787, opened 2020-03-21
build: fix eslint/prettier integration

### LINKED ISSUE #1187: Prettier toolchain causing unexpected errors
https://github.com/flowtype/flow-typed/pull/1143 is causing a lot of headache when working with definitions. Everything seems to work fine after committing. Then, under the scenes, the files are rewritten in ways that cause errors. Specifically troublesome for tests that have `// $ExpectError` on lines that then get word wrapped.

jedwards1211:
fix #1187

- Type of contribution: fix

### DIFF
--- a/cli/.eslintrc
+++ b/cli/.eslintrc
@@ -3,28 +3,22 @@
   "env": {
     "es6": true
   },
-  "plugins": [
-    "flowtype",
-    "prettier"
-  ],
+  "plugins": ["flowtype"],
   "extends": [
-    "plugin:flowtype/recommended"
+    "plugin:flowtype/recommended",
+    "plugin:prettier/recommended",
+    "prettier/flowtype"
   ],
-  "globals": {
-  },
+  "globals": {},
   "rules": {
-    "prettier/prettier": [
-      "error",
-      { "singleQuote": true,
-        "bracketSpacing": false,
-        "trailingComma": "all",
-      }
-    ],
     "eqeqeq": [2, "smart"],
     "semi": 2,
-    "no-unused-vars": ["error", {
-      "varsIgnorePattern": "^_.?.?$",
-      "argsIgnorePattern": "^_.*$",
-    }],
+    "no-unused-vars": [
+      "error",
+      {
+        "varsIgnorePattern": "^_.?.?$",
+        "argsIgnorePattern": "^_.*$"
+      }
+    ]
   }
 }
--- a/cli/package.json
+++ b/cli/package.json
@@ -52,6 +52,7 @@
     "babel-eslint": "^10.0.3",
     "babel-jest": "^25.1.0",
     "eslint": "^6.8.0",
+    "eslint-config-prettier": "^6.10.1",
     "eslint-plugin-flowtype": "^4.6.0",
     "eslint-plugin-prettier": "^3.1.2",
     "flow-bin": "^0.118.0",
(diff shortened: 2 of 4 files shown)


cli/package.json
```diff
@@ -52,6 +52,7 @@
     "babel-eslint": "^10.0.3",
     "babel-jest": "^25.1.0",
     "eslint": "^6.8.0",
+    "eslint-config-prettier": "^6.10.0",
     "eslint-plugin-flowtype": "^4.6.0",
     "eslint-plugin-prettier": "^3.1.2",
```
GAntoine: Should this be removed now that `prettier` is no longer in `plugins`?

jedwards1211: yeah we could, `eslint-plugin-prettier`'s recommended config does extend it

GAntoine: Oops, sorry I meant remove `eslint-plugin-prettier` but keep `eslint-config-prettier`.

cli/.eslintrc
```diff
@@ -3,28 +3,18 @@
...
-    "flowtype",
-    "prettier"
-  ],
-  "extends": [
-    "plugin:flowtype/recommended"
-  ],
-  "globals": {
-  },
+  "plugins": ["flowtype"],
+  "extends": ["plugin:flowtype/recommended", "plugin:prettier/recommended"],
```
pascalduez: They [seem to recommand](https://github.com/prettier/eslint-plugin-prettier#recommended-configuration) extending the `"prettier/react"` config afterward.

cli/.eslintrc
```diff
@@ -3,28 +3,18 @@
...
+  "globals": {},
   "rules": {
-    "prettier/prettier": [
```
pascalduez: So basically this disable any ESLint linting about style?

cli/.eslintrc
```diff
@@ -3,28 +3,18 @@
...
-  "plugins": [
-    "flowtype",
-    "prettier"
-  ],
-  "extends": [
-    "plugin:flowtype/recommended"
-  ],
-  "globals": {
-  },
+  "plugins": ["flowtype"],
```
pascalduez: Tests are failing, probably needs to keep `"prettier"` in plugins?

jedwards1211: According to the docs, `plugin:prettier/recommended` does three things:

- Enables `eslint-plugin-prettier`.
- Sets the `prettier/prettier` rule to `"error"`.
- Extends the `eslint-config-prettier` configuration.

I'll have to investigate why the tests are failing

jedwards1211: okay apparently we did need to keep `eslint-config-prettier` in the dev deps

cli/.eslintrc
jedwards1211: oh right on

cli/.eslintrc
jedwards1211: `plugin:prettier/recommended` sets the `prettier/prettier` rule to "error"

cli/package.json
jedwards1211: no we definitely need both if you want to check formatting via `eslint`

jedwards1211: I kinda blame `eslint` for introducing formatting lint rules without being a true code formatter, if they hadn't done that this would be way less confusing.

PR #3790, opened 2020-03-23
Fix/3761 enzyme use Element instead of Node

### LINKED ISSUE #3761: [enzyme] get(...).props shows `props is missing in $Iterable, Boolean, Number ...` error

Bug with enzyme Flow types.

Using `wrapper.find(Foo).get(0).props` as per [enzyme docs](https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/get.html) produces error (at least in Flow v0.119.1)

```
Cannot get anchor.get(...).props because:
 • property props is missing in $Iterable [1].
 • property props is missing in Boolean [1].
 • property props is missing in Number [1].
 • property props is missing in React.Portal [1].
 • property props is missing in String [1].
 • property props is missing in null [1].

 expect(anchor.get(0).props.href).toEqual(operation.url);

     flow-typed/npm/enzyme_v3.x.x.js
 [1] 50│     get(index: number): React$Node,
```

Not aware of a sandbox which does both flow checking and importing from npm lib

We use
```
flow-bin: v0.119.1
enzyme: v3.9.0

# should not matter
react: v16.9.0
jest: v24.9.0
```

Code which shows error should be simple, same as in docs

```
it('my test', () => {
  const Foo = () => <div>some text</div>
  const MyComponent = () => <Foo foo="bar"/>

  const wrapper = shallow(<MyComponent />);
  expect(wrapper.find(Foo).get(0).props.foo).to.equal('bar');
});
```

No Flow error.

It happens because React$Node can be string, number etc., not necessarily React component with props.

At the same time, both enzyme docs 
https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/get.html

and Typescript types (for v3)
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/enzyme/index.d.ts#L159

use ReactElement type, analog of React$Element in flow.

Actually, they do not use ReactNode type (analog of React$Node in Flow) *anywhere at all*, it is not even imported.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/enzyme/index.d.ts#L26

That's why I think type should be fixed/changed in flow-typed as well.

Happy to create a PR for it.

DmitryGonchar:
* Type of contribution: fix, refactor tests
- Links to documentation: https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/get.html

Other notes:

This should close #3761 . When doing the change for `get()`, I've noticed that both enzyme docs and [its Typescript definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/enzyme/index.d.ts) do not use Node at all. That's why I've decided to update all other methods.
I've tried to update render() as well, but existing `opaqueNode` test was failing, so I've decided to leave definition as is.

To make reviewing easier, you can skip "no actual changes" commits, they are grouping tests or auto-sorting methods alphabetically (to be in the same order as in docs).


definitions/npm/enzyme_v3.x.x/flow_v0.104.x-/enzyme_v3.x.x.js
```diff
@@ -96,10 +96,9 @@ declare module "enzyme" {
       root: any,
       options?: ?Object
     ): ShallowWrapper<T>,
-    equals(node: React$Node): boolean,
```
DmitryGonchar: It has been already defined in a Wrapper

definitions/npm/enzyme_v3.x.x/flow_v0.104.x-/enzyme_v3.x.x.js
```diff
@@ -11,63 +11,53 @@ declare module "enzyme" {
...
-    find(selector: UntypedSelector): this,
-    find<T: React$ElementType>(selector: T): ReactWrapper<T>,
-    findWhere(predicate: PredicateFunction<this>): this,
+    at(index: number): this,
+    childAt(index: number): this,
+    children(selector?: UntypedSelector): this,
+    children<T: React$ElementType>(selector: T): ReactWrapper<T>,
+    closest(selector: UntypedSelector): this,
+    closest<T: React$ElementType>(selector: T): ReactWrapper<T>,
+    contains(nodes: React$Element | $ReadOnlyArray<React$Element>): boolean,
```
pascalduez: `React$Element` requires a type argument, so wondering why this all pass... 

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAlgWwA5wE4AuyYAhgM5gBKApiQMZFR5yZgDketDbqqdcAOzJEC1YQC4qXAgBIAojGqZqAogF4wAHgAm6AG4A+UcM3Bdh3tup0YJTmCgBXAQ3SCwxggApqMSTXpZBSUVAgBKST04dG1UIA

pascalduez:
Generally looks good to me, the change from Node to Element makes sense.  

I wouldn't mind other reviews or opinions though.

definitions/npm/enzyme_v3.x.x/flow_v0.104.x-/enzyme_v3.x.x.js
DmitryGonchar: I can specify `<any>` explicitly. I think for input arguments it is ok, but for return types - we should make it optionally typeable by the user. So I think I'll need to specify, for example:
`get<T = any>(...): Element<T>`. What do you think?

pascalduez: Sounds good. What baffles me is that the tests pass currently without it...

DmitryGonchar: Done. I don't know why, looks like a Flow bug to me.

PR #3805, opened 2020-04-08
add deep merge v3

Brianzchen:
- Links to documentation:
- Link to GitHub or NPM: https://www.npmjs.com/package/deepmerge

Other notes:


Brianzchen:
@GAntoine I've made the changes as you suggested and the build is still breaking do you have any suggestions? Also how come for this library 0.25-0.103 can't run tests?

Fairly new to flow so trying to learn all I can 😄

GAntoine:
It seems to be an error with v1, so I'm gonna merge this in with the assumption that v3 works as expected.

Brianzchen:
Thanks, probably v1 needs to be updated to follow v2/v3 test format

PR #3806, opened 2020-04-09
Add missing exports @testing library/react

samtsai:
- Links to documentation: https://testing-library.com/docs/dom-testing-library/api-queries#screen
- Link to GitHub or NPM:
[`screen`](https://github.com/testing-library/dom-testing-library/blob/master/src/screen.js)
see [Typescript `screen.d.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/testing-library__dom/screen.d.ts)
[`waitForElementToBeRemoved`](https://testing-library.com/docs/dom-testing-library/api-async#waitforelementtoberemoved)
see [Typescript `wait-for-element-to-be-removed.d.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/testing-library__dom/wait-for-element-to-be-removed.d.ts)
- Type of contribution: addition | fix

Other notes:
Backported fixes to previous versions for accuracy and passing tests.

### DIFF
--- a/definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
+++ b/definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
@@ -202,13 +202,31 @@ declare module '@testing-library/react' {
     eventProperties?: TInit
   ) => boolean;
 
+  declare type Screen<Queries = GetsAndQueries> = {
+    ...Queries,
+    debug: (
+      baseElement?:
+        | HTMLElement
+        | DocumentFragment
+        | Array<HTMLElement | DocumentFragment>,
+      maxLength?: number
+    ) => void,
+    ...
+  };
+
   declare type RenderResult<Queries = GetsAndQueries> = {
     ...Queries,
-    container: HTMLDivElement,
-    unmount: () => void,
+    container: HTMLElement,
+    unmount: () => boolean,
     baseElement: HTMLElement,
     asFragment: () => DocumentFragment,
-    debug: (baseElement?: HTMLElement) => void,
+    debug: (
+      baseElement?:
+        | HTMLElement
+        | DocumentFragment
+        | Array<HTMLElement | DocumentFragment>,
+      maxLength?: number
+    ) => void,
     rerender: (ui: React$Element<*>) => void,
     ...
   };
@@ -242,10 +260,10 @@ declare module '@testing-library/react' {
   ): RenderResult<CustomQueries>;
 
   declare export var act: ReactDOMTestUtilsAct;
-  declare export function cleanup(): void;
+  declare export function cleanup(): Promise<void>;
 
   declare export function waitFor<T>(
-    callback?: () => T,
+    callback: () => T,
     options?: {|
       container?: HTMLElement,
       timeout?: number,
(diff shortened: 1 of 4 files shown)


definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
```diff
@@ -202,13 +202,31 @@ declare module '@testing-library/react' {
     eventProperties?: TInit
   ) => boolean;
 
+  declare type Screen<Queries = GetsAndQueries> = {
+    ...Queries,
```
pascalduez, line 206: Spreading in an inexact object doesn't behave a intended.

definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
```diff
@@ -254,18 +272,8 @@ declare module '@testing-library/react' {
...
-      container?: HTMLElement,
-      timeout?: number,
-      interval?: number,
-      mutationObserverOptions?: MutationObserverInit,
-    |}
-  ): Promise<HTMLElement>;
-
   declare export function waitForElementToBeRemoved<T>(
-    callback?: () => T,
+    callback: (() => T) | T,
```
pascalduez, line 276: Humm okay, so if you pass an HTML Element as `callback` it's properly inferred? 
https://testing-library.com/docs/dom-testing-library/api-async#waitforelementtoberemoved

definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/test_react_v10.x.x.js
```diff
@@ -104,8 +105,6 @@ describe('waitForElement', () => {
 
 describe('waitForElementToBeRemoved', () => {
   it('should fail on invalid inputs', () => {
-    // $ExpectError
-    waitForElementToBeRemoved(1);
```
pascalduez, line 108: That's the problem, it's now too loose and let you use anything.

samtsai: I don't disagree, what's an alternative way to type this?

definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
samtsai: What would be the more appropriate approach? I was trying to follow previous discussions and carry over the latest

definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/test_react_v10.x.x.js
pascalduez: Like it was before, two separate declarations.

definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
pascalduez: Try to use exact as much as possible, if not possible intersection?

PR #3810, opened 2020-04-20
Add lib def for Segment Analytics

michaelmcchen:
- Links to documentation: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/
- Link to GitHub or NPM: https://github.com/segmentio/analytics.js, https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/segment-analytics

Other notes:
- Didn't notice the `describe`/`it` request - will add that momentarily. Would like feedback on my question that I point out inline though!


definitions/npm/@segment/analytics_v3.11.2/flow_v0.103.x/analytics_v3.11.2.js
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// Adding type to Window Global as that is how we access Segment functions (via snippet load, see: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/quickstart/)
+declare type Window = {
+  analytics: SegmentAnalytics$AnalyticsJS,
+  [key: any]: any, // Keep rest of elements untyped (default)
+  ...
+}
+
+declare var window: Window;
```
michaelmcchen, line 189: Curious to hear your thoughts on doing this (enforcing the type via Window global), since Segment recommends loading their lib via web snippet (and thus functions are accessible via `window.analytics...` vs. a NPM module

michaelmcchen:
@GAntoine Hi! Tagging you since you're one of the active maintainers for the repo. Could I get a quick review? I appreciate it! 🙏

definitions/npm/@segment/analytics_v3.11.2/flow_v0.103.x/analytics_v3.11.2.js
GAntoine: You're right about the way it's loaded on window. In our codebase (TS unfortunately), we manually set `analytics` on `Window` like so, so maybe this isn't necessary:

```
interface Window {
    analytics: SegmentAnalytics.AnalyticsJS;
}
```

GAntoine: Yeah let's start with that, and if it's an issue for people we can always add it in the definition

GAntoine:
Ah hm, without Window the tests won't work. And since it doesn't export anything this might be hard to test.... 🤔

definitions/npm/@segment/analytics_v3.11.2/flow_v0.103.x-/test_analytics_v3.11.2.js
```diff
@@ -0,0 +1,19 @@
+window.analytics.load("apiKey");
```
AndrewSouthpaw, line 1: It might work if you just kind of hack the situation? Something like...

`const window: { analytics: MyType } = { analytics: ({}: any) }`

If you wrap the tests in a `describe` then the statements are never actually run in a JS context so that might work out.

Give that a shot?

PR #3827, opened 2020-05-04
Add uuid v8

TrySound:
All is the same. Only deprecated direct imports dropped and added native
node esm support.

- Link to GitHub or NPM: https://github.com/uuidjs/uuid
- Type of contribution: new definition | addition


TrySound:
Looks like old tests failed with new flow. Not really related to my change, right?

Issue #3816, opened 2020-04-23
@testing-library/react v10: href property throws error


Brianzchen:
**Do you want to request a *feature* or report a *bug*?**
Bug in `@testing-library/react`

**What is the current behavior?**
Elements that are returned from `getByTestId` are always `HTMLElement` so when I try to access `element.href` it throws flow error.

**If the current behavior is a bug, please provide the steps to reproduce.**
```
  const { getByTestId } = render(
      <Footer
        {...props}
      />,
    );

    expect(getByTestId('customer-service-link').href).toBe('tel:131 802');
```

Have tried to cast it but I'm told 
```
Cannot assign `getByTestId(...)` to `linkElement` because  `HTMLElement` [1] is incompatible with  `HTMLLinkElement`
```

```
  const { getByTestId } = render(
      <Footer
        {...props}
      />,
    );

    const linkElement: HTMLLinkElement = getByTestId('customer-service-link')

    expect(linkElement.href).toBe('tel:131 802');
```

**What is the expected behavior?**
`href` should be an optionally available property or I should be able to optionally pass the type I expect back such as `getByTestId<HTMLLinkElement>('customer-service-link')`

**Local Environment Information**
Flow v. 0.122.0, latest `@testing-library/react` v10 flow types.

Brianzchen:
Solved this by first checking the `instanceof` the element before I access the values. But all my tests will need to assert the amount of expects prior

```
expect.assertions(1);

const { getByTestId } = render(
  <Footer
    {...props}
  />,
);

const linkElement = getByTestId('customer-service-link')

if (linkElement instanceof HTMLLinkElement) {
  expect(linkElement.href).toBe('tel:131 802');
}
```

pascalduez:
Hi @Brianzchen,

there've been a previous attempt at this in #3404. 

I proposed to try with a default generic. 

The libdef changed a bit since, but out of the blue it could look like:
```js
declare type GetByBoundAttribute<T: HTMLElement = HTMLElement> = (
  text: TextMatch,
  options?: TextMatchOptions
) => T;
```
Want to give it a try?

Brianzchen:
Thanks @pascalduez that's a really helpful start. But I've come across a few issues over a few test scenario's.

When I try to simply implement your change, the types declaration for functions such as `getByTestId` throws errors saying `Cannot use GetByBoundAttribute [1] without 0-1 type arguments.`

Next I try to add generics to the `GetByBoundAttribute` caller with
```
[...]

PR #3837, opened 2020-05-14
[@testing-library/react] update GetByBoundAttribute to accept generic

Brianzchen:
- Links to documentation: https://testing-library.com/docs/dom-testing-library/api-queries
- Link to GitHub or NPM: https://www.npmjs.com/package/@testing-library/react
- Type of contribution: fix

Other notes:
Update queries to generics such as `getByTestId` which will remove query from returning `HTMLElement` until brackets are used.

Previous behaviour:
```
const element = getByTestId('test'); // element is HTMLElement
const element: HTMLSelectElement = getByTestId('test'); // errors
```

New behaviour
```
const element = getByTestId('test'); // element is any
const element: HTMLSelectElement = getByTestId('test'); // element is HTMLSelectElement
const element = getByTestId<>('test'); // element is HTMLElement
const element = getByTestId<HTMLSelectElement>('test'); // element is HTMLSelectElement
```

Seems like this is probably better behaviour and allows better type casting

### DIFF
--- a/definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
+++ b/definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
@@ -51,10 +51,10 @@ declare module '@testing-library/react' {
     waitForElementOptions?: WaitForElementOptions
   ) => Promise<HTMLElement[]>;
 
-  declare type GetByBoundAttribute = (
+  declare type GetByBoundAttribute = <T: HTMLElement = HTMLElement>(
     text: TextMatch,
     options?: TextMatchOptions
-  ) => HTMLElement;
+  ) => T;
 
   declare type FindByBoundAttribute = (
     text: TextMatch,


Brianzchen:
https://github.com/flow-typed/flow-typed/issues/3816

Brianzchen:
@pascalduez @GAntoine Is this PR ok?

pascalduez:
Not a big fan of this change, all users will loose their typing without even noticing it.

Brianzchen:
@pascalduez I agree that's a tradeoff. Maybe it's my lack of knowledge but previously the only other way to check an input's value would have to be like

```
expect.assertions(1); // This is necessary otherwise test would pass because no expects failed

const inputElement = getByTestId('input');

if (inputElement instanceof HTMLInputElement) {
  expect(inputElement.value).toBe('test');
}
```

I think the above is an extreme amount of code just to get around flow for unit testing. I imagine many devs would have not bothered flow typing their spec files or just use `$FlowFixMe` as an escape hatch which I think discourages the usefulness of flow

pascalduez:
The other "solution" would be:

```
const inputElement: HTMLInputElement = (getByTestId('input'): any);

expect(inputElement.value).toBe('test');
```

@Brianzchen maybe come to the Flow Discord chan so we can ask for directions?

PR #3838, opened 2020-05-19
Jest: improve jest.requireActual type

kayhadrin:
- Links to documentation: https://jestjs.io/docs/en/jest-object#jestrequireactualmodulename
- Link to GitHub or NPM: https://github.com/facebook/jest/blob/master/e2e/__tests__/jestRequireActual.test.ts

W.r.t. `requireActual`, here's an example of how to use it with the new generic type argument:

```js
// Assuming you have:
import typeof MyModule from 'MyModule';
// where MyModule is a type like `{}`

// Before
const myModule = requireActual('MyModule');

// Casting myModule to arbitrary types
(myModule: MyModule) // OK, but we're actually not sure
(myModule: 'do not trust me') // Flow says it's ok, but it's not!

// After
const myModule = requireActual<MyModule>('MyModule');

// Casting myModule to arbitrary types
(myModule: MyModule) // OK
(myModule: 'do not trust me') // Flow error as expected
```

### DIFF
--- a/definitions/npm/jest_v21.x.x/flow_v0.104.x-/jest_v21.x.x.js
+++ b/definitions/npm/jest_v21.x.x/flow_v0.104.x-/jest_v21.x.x.js
@@ -361,7 +361,7 @@ type JestObjectType = {
   * Returns the actual module instead of a mock, bypassing all checks on
   * whether the module should receive a mock implementation or not.
   */
- requireActual(moduleName: string): any,
+ requireActual<T>(m: $Flow$ModuleRef<T> | string): T,
  /**
   * Returns a mock module instead of the actual module, bypassing all checks
   * on whether the module should be required normally or not.
--- a/definitions/npm/jest_v21.x.x/flow_v0.104.x-/test_jest-v21.x.x.js
+++ b/definitions/npm/jest_v21.x.x/flow_v0.104.x-/test_jest-v21.x.x.js
@@ -290,3 +290,12 @@ expect(wrapper).toMatchSelector("span");
 expect(wrapper).toMatchSelector();
 // $ExpectError
 expect(wrapper).toMatchSelector(true);
+
+{
+  // Type hint becomes the returned type
+  const FooModule1: string = jest.requireActual<string>('FooModule');
+  // $ExpectError Type hint becomes the returned type
+  const FooModule2: number = jest.requireActual<string>('FooModule');
+  // No type hint means we return any
+  const FooModule3: boolean = jest.requireActual('FooModule');
+}
--- a/definitions/npm/jest_v21.x.x/flow_v0.25.x-v0.38.x/jest_v21.x.x.js
+++ b/definitions/npm/jest_v21.x.x/flow_v0.25.x-v0.38.x/jest_v21.x.x.js
@@ -349,7 +349,7 @@ type JestObjectType = {
    * Returns the actual module instead of a mock, bypassing all checks on
    * whether the module should receive a mock implementation or not.
    */
-  requireActual(moduleName: string): any,
+  requireActual<T>(m: $Flow$ModuleRef<T> | string): T,
   /**
    * Returns a mock module instead of the actual module, bypassing all checks
    * on whether the module should be required normally or not.
--- a/definitions/npm/jest_v21.x.x/flow_v0.25.x-v0.38.x/test_jest-v21.x.x.js
+++ b/definitions/npm/jest_v21.x.x/flow_v0.25.x-v0.38.x/test_jest-v21.x.x.js
@@ -207,3 +207,12 @@ expect(wrapper).toMatchSelector("span");
 expect(wrapper).toMatchSelector();
 // $ExpectError
 expect(wrapper).toMatchSelector(true);
+
+{
+  // Type hint becomes the returned type
+  const FooModule1: string = jest.requireActual<string>('FooModule');
+  // $ExpectError Type hint becomes the returned type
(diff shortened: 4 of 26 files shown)


kayhadrin:
🏓

AndrewSouthpaw:
I think @pascalduez has been handling most PRs to this libdef... pinging him. :-)

definitions/npm/jest_v21.x.x/flow_v0.104.x-/jest_v21.x.x.js
```diff
@@ -361,7 +361,7 @@ type JestObjectType = {
   * Returns the actual module instead of a mock, bypassing all checks on
   * whether the module should receive a mock implementation or not.
   */
- requireActual(moduleName: string): any,
+ requireActual<T>(m: $Flow$ModuleRef<T> | string): T,
```
pascalduez, line 364: Could you explain a bit more in details how is this working, and what it's improving.

pascalduez:
> I think @pascalduez has been handling most PRs to this libdef... pinging him. :-)

@AndrewSouthpaw  it's more the `@testing-libray/*` things I'm worried about.

definitions/npm/jest_v21.x.x/flow_v0.104.x-/jest_v21.x.x.js
kayhadrin: I don't think I could explain it better than the [Flow commit that introduced `$Flow$ModuleRef<>`](https://github.com/facebook/flow/commit/296ed9ce7896790798b9b2fcb3ba8af7c43f7dbe) :-)

----

> Summary:
> This diff introduces a configurable string prefix that will upgrade any string literal in any position into a `$Flow$ModuleRef<$Exports<'TheModule'>>`. This lets currently hard-coded APIs like `requireLazy` be typed in userland, instead of being special-cased in the typechecker.
> 
> An example usage is a function like `ifRequired('m#Foo', Foo ==> { /* callback */ });`. This function talks to the underlying `require` internals to actually get the module exports and this new API would let this function be typed like so:
> 
> ```
> declare function ifRequired<TModule, TYes>(
>   id: $Flow$ModuleRef<TModule>,
>   cb: (module: TModule) => TYes,
> );
> ```
> 
> Flow would then understand that the first argument to the callback is in fact the same as requiring `Foo` directly. This reduces boilerplate and `typeof` in some of the magic infrastructural APIs.

----

W.r.t. `requireActual`, I've added an example of how it improves things in the PR summary.

kayhadrin: 🏓

AndrewSouthpaw:
@pascalduez do you want to give it one more once-over?

kayhadrin:
Ping @pascalduez :-)

kayhadrin:
🏓

pascalduez:
@kayhadrin Sorry for the delay... Could you please rebase?

kayhadrin:
> @kayhadrin Sorry for the delay... Could you please rebase?

Done!

definitions/npm/jest_v21.x.x/flow_v0.104.x-/jest_v21.x.x.js
gwing33: Looks like this has broken my build.
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ flow-typed/npm/jest_v23.x.x.js:827:23

Cannot resolve name $Flow$ModuleRef.

     824│    * Returns the actual module instead of a mock, bypassing all checks on
     825│    * whether the module should receive a mock implementation or not.
     826│    */
     827│   requireActual<T>(m: $Flow$ModuleRef<T> | string): T,
     828│   /**
     829│    * Returns a mock module instead of the actual module, bypassing all checks
     830│    * on whether the module should be required normally or not.
```

I realize my flow version is sufficiently behind as I'm on `0.79.1`, we are looking to upgrade it soon.

pascalduez: @kayhadrin @gwing33 Any taker for a fix PR?

kayhadrin: So if I understand correctly, you'd like to revert part of this PR so that only Flow v0.86.0+ is affected?

pascalduez: @kayhadrin Yes, would be nice, scoping using file name versions.

corneliug: @kayhadrin @pascalduez Hi, is there a PR for reverting the changes for just Flow v0.86.0+? I have a build that's failing on v0.72.0 because of that change and I am looking to track it.

gwing33: I started to look at creating a PR, then got stuck with the `./build_and_test_cli.sh` command, I may take another stab tomorrow. A quick workaround I added in was just to `declare var $Flow$ModuleRef: any;`. We are in the process of upgrading flow, so it's short term for us.

I also just wanted to give a shout out to @pascalduez and @kayhadrin, thank you for your work and for what you're contributing to the community!

PR #3839, opened 2020-05-20
feat: add [input_v1.x.x]

Jabster28:
Added definitions for [input](https://npmjs.com/package/input)


Jabster28:
closing because it seems there's no way to put optional parameters before required ones in flow

Issue #3809, opened 2020-04-16
"Invalid Version: undefined" semver on version 3.x flow-typed


ghtjiann4321:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
after running yarn flow-typed install
```
UNCAUGHT ERROR: TypeError: Invalid Version: undefined
    at new SemVer (/path/project/node_modules/flow-typed/node_modules/semver/classes/semver.js:22:13)
error Command failed with exit code 1.
```

**If the current behavior is a bug, please provide the steps to reproduce.**

- Install Fresh CRA
- install flow-bin flow-typed
- run yarn flow init
- run yarn flow-typed install

**What is the expected behavior?**
- should generate the flow-typed directory

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Local Environment Information**
If this issue is specific to the CLI, provide the results of running `npx envinfo` from the command line in a GitHub gist.

kubuntu: 19.04
node: 10.18.1
flow-typed: 3.1.0
flow-bin: 0.122.0

dstaley:
This is almost certainly due to [this bug](https://github.com/npm/node-semver/issues/323).

bradvogel:
Is there a workaround?

pzrq:
> Is there a workaround?

So far I've seen downgrading flow-typed (if one can) to v2.6.2 or manually pinning semver to a known good version in `yarn.lock` such as:

```
semver@^7.1.3:
  version "7.1.3"
  resolved "https://npm.shootsta.com/semver/-/semver-7.1.3.tgz#e4345ce73071c53f336445cfc19efb1c311df2a6"
  integrity sha512-ekM0zfiA9SCBlsKa2X1hxyxiI4L3B6EbVJkkdgQXnSEEaHlGdvyodMruTiulSRWMMB4NeIuYNMC9rTKTz97GxA==
```

GAntoine:
I won't get around to getting. this fixed until this weekend, so if someone is willing to open a PR that fixes the issue in the code, that would be great! If not, I'll try my hand at it, or end up pinning `semver` to a lower version.

randallb:
You can also pin using yarn resolutions in package.json.

https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-how-to-use-it

xthule:
It looks like the "bug" from semver isn't going to be fixed.

linonetwo:
[...]

PR #3842, opened 2020-05-30
Fix: Add workaround for semver throwing errors for versions of 'v0.x.x'

xthule:
Workaround for #3809 

Semver has updated and throws an error with versions of 'v0.x.x'. I'm not sure if this is the best place for this, but it avoids changing the `libDef` definition and the cache.


mxzinke:
Merge?

PR #3843, opened 2020-06-02
[react-navigation] Libdefs for React Nav 5

Ashoat:
React Nav 5 is a complete rewrite, so I went ahead and completely rewrote the types. A lot more stuff can now be typechecked!

Note that we now have seven different packages users can import from. To make contributing to the libdefs easier I've simplified them into just two sections. This means the libdefs are longer than they need to be sometimes, as Section 1 may have some types not necessarily for a given libdef.


Ashoat:
For anybody landing here because they're confused about how to use the new types, give this [Medium post](https://medium.com/@ashoat_t/flow-libdefs-for-react-navigation-5-7dcf1f9da8b7) a read.

PR #3844, opened 2020-06-05
Jest: add the done(Error) functionality

julienw:
Since Jest v21 it's possible to call `done` with an Error (implemented in https://github.com/facebook/jest/pull/4016). Before one had to do either `done()` or `done.fail(Error)`.

- Links to documentation: https://jestjs.io/docs/en/asynchronous#callbacks
- Type of contribution: addition & fix

Other notes:
* This is existing since Jest v21 but it was easier to do it only since Jest v24. If others need it they'll be able to change it by following the example.
* I also realized that the libdefs weren't working for some older flow versions, so I changed the directory names accordingly. Because that's from very old flow versions I didn't bother fixing that.
* `done.fail` isn't documented but I noticed it's supported by test cases there, so I decided to leave it. Also I didn't want to bring new errors to users for things that are working even if obscure.

### DIFF
--- a/definitions/npm/jest_v24.x.x/flow_v0.61.x-v0.103.x/jest_v24.x.x.js
+++ b/definitions/npm/jest_v24.x.x/flow_v0.61.x-v0.103.x/jest_v24.x.x.js
@@ -590,7 +590,6 @@ type SnapshotDiffType = {
     |},
     testName?: string
   ): void,
-  ...
 }
 
 interface JestExpectType {
@@ -965,7 +964,7 @@ type JestSpyType = {
 };
 
 type JestDoneFn = {|
-  (): void,
+  (error?: Error): void,
   fail: (error: Error) => void,
 |};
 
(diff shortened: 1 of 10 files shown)


definitions/npm/jest_v24.x.x/flow_v0.61.x-v0.103.x/jest_v24.x.x.js
```diff
@@ -590,7 +590,6 @@ type SnapshotDiffType = {
     |},
     testName?: string
   ): void,
-  ...
```
julienw, line 593: This is incorrect in older flow versions and was causing an error, so I removed it. Of course I left it in the libdef for the newer versions.

definitions/npm/jest_v25.x.x/flow_v0.104.x-/jest_v25.x.x.js
```diff
@@ -927,8 +927,8 @@ type JestObjectType = {
 type JestSpyType = { calls: JestCallsType, ... };
 
 type JestDoneFn = {|
- (): void,
- fail: (error: Error) => void,
+  (error?: Error): void,
+  fail: (error: Error) => void,
```
julienw, line 931: There was an indentation error here as well.

julienw:
Thanks for the quick review!

Issue #3847, opened 2020-06-16
@react-navigation/stack - unable to set headerBackTitle into 'null' without flow error


Preeternal:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Got error:
```
Cannot create `Stack.Screen` element because  null [1] is incompatible with  string [2] 
in property `headerBackTitle` of property `options`.Flow(InferError)
index.js(172, 28): [1] null
stack_v5.x.x.js(1231, 23): [2] string
```

**If the current behavior is a bug, please provide the steps to reproduce.**
set 'headerBackTitle' into 'null'
**What is the expected behavior?**

`Title string used by the back button on iOS, or null to disable label. Defaults to the previous scene's headerTitle.`

PR #3848, opened 2020-06-16
@react-navigation/stack - unable to set headerBackTitle into 'null' w…

Preeternal:
Fix for the [issue](https://github.com/flow-typed/flow-typed/issues/3847)


Preeternal:
Sure. @Ashoat Could you please to check? Done.

PR #3854, opened 2020-06-24
Fixed bluebird's Promise.all and Promise.resolve types

jps327:
1. Bluebird's `Promise.all` annotation was not inferring the correct type for tuples of promises. Switched to use `$TupleMap` so that the types are correctly inferred whether we use a tuple or an array as the argument. The type annotation is based off of JavaScript's builtin Promise.all annotation: https://github.com/facebook/flow/blob/v0.111.3/lib/core.js#L689
2. Bluebird's Promise.resolve was inferring `empty` as the generic type. Fixed by adding separate annotations for when `resolve` is called without any arguments.

Passes all tests with `quick_run_def_tests.sh`

Links to documentation: http://bluebirdjs.com/docs/api-reference.html
Link to GitHub or NPM: https://www.npmjs.com/package/bluebird
Type of contribution: fix

### DIFF
--- a/definitions/npm/bluebird_v3.x.x/flow_v0.104.x-/test_bluebird_v3.x.x.js
+++ b/definitions/npm/bluebird_v3.x.x/flow_v0.104.x-/test_bluebird_v3.x.x.js
@@ -13,6 +13,8 @@ promise.reflect().then(inspection => {
 // $ExpectError
 new Bluebird();
 
+const emptyPromise: Promise<void> = Promise.resolve();
+
 Bluebird.all([
   new Bluebird(() => {}),
 ]);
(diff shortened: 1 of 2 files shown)


definitions/npm/bluebird_v3.x.x/flow_v0.104.x-/test_bluebird_v3.x.x.js
```diff
@@ -13,6 +13,8 @@ promise.reflect().then(inspection => {
 // $ExpectError
 new Bluebird();
 
+const emptyPromise: Promise<void> = Promise.resolve();
```
pascalduez, line 16: This is not testing Bluebird but core Promise, right?

PR #3855, opened 2020-06-25
Support latest flow versions

julienw:
This started as a quick change to support latest versions of flow. Indeed since 0.125.0 there's an error when `suppress_comment` is specified in `.flowconfig`.

But doing this I realized that the test running function was then broken: indeed flow-typed runs the tests in "groups", that is the different directories in each libdef directories. And the mechanism was sharing the flowconfig file when running several flow versions, so that they can run in parallel.
So this patch includes a small algorithm to partition a list of flow versions into a list of list of flow versions that are "flowconfig-compatible".
I hope the code is understandable enough, it was kind of hard to come up with good variable names.

Other notes:

My strategy to update libdefs has been to run this command:
> git grep -Fl '$ExpectError' | xargs sed -i 's/$ExpectError/$FlowExpectedError/g'

Also I tested the patch doing things like:
> node cli/dist/cli.js run-tests --numberOfFlowVersions 100 idb

As a conclusion I'd like to add that I'm amazed that making flow-typed work with latest Flow versions isn't part of a version release process. For me this is a lack of regard to the community still working with Flow. Should I recall that 0.125.0 is more than 1 month old?


ammanvedi:
yes please

julienw:
Hey, anything I can do to help reviewing this?
There are a lot of file changes but this was done automatically.

Sadly the more we're waiting the more this will bitrot. Especially I'll have to double check that some landed patches didn't include new expectations... Actually I wonder how these new patches can be verified at all without this patch?

PR #3857, opened 2020-06-26
Add libdefs for idb v5

julienw:
- Links to documentation: https://github.com/jakearchibald/idb
- Link to GitHub or NPM: https://github.com/jakearchibald/idb

Other notes:
~~This PR contains the commits from #3855 so that my use of `$FlowExpectedError` works across all flow versions supported by this libdef.~~ <= Now that #3855 has been merged I rebased on top of latest master and this isn't in this PR anymore.
I started by copying the libdef for idb_v2 but then I heavily rewrote the libdef and also added a lot of new tests. (so there are 2 commits fot these 2 steps)

### DIFF
--- a/definitions/npm/idb_v5.x.x/flow_v0.83.x-/idb_v5.x.x.js
+++ b/definitions/npm/idb_v5.x.x/flow_v0.83.x-/idb_v5.x.x.js
@@ -1,0 +1,427 @@
+// Derived from TS typings at https://github.com/jakearchibald/idb/blob/c8bab898f01bacab96097f87c1e42a0c19e01425/lib/idb.d.ts
+
+type IDBArrayKey = Array<IDBValidKey>;
+type IDBValidKey = number | string | Date | IDBArrayKey;
+
+// TODO: upstream this to Flow DOM definitions
+declare class DOMStringList {
+  +length: number;
+  // Comment syntax here as @@iterator is invalid syntax for eslint and babylon
+  /*:: @@iterator(): Iterator<string>; */
+  contains(str: string): boolean;
+  item(index: number): string | null;
+  [index: number]: string;
+}
+
+declare module 'idb' {
+  /** This is your entry point to the API. It's exposed to the global scope unless you're using a module system such as browserify, in which case it's the exported object. */
+  declare module.exports: IDBStatic;
+
+  /**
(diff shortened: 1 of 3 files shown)


definitions/npm/idb_v5.x.x/flow_v0.83.x-/idb_v5.x.x.js
```diff
@@ -1,427 +1,643 @@
 // Derived from TS typings at https://github.com/jakearchibald/idb/blob/c8bab898f01bacab96097f87c1e42a0c19e01425/lib/idb.d.ts
+// Updated for changes between v2 and v5.
 
-type IDBArrayKey = Array<IDBValidKey>;
-type IDBValidKey = number | string | Date | IDBArrayKey;
+declare type $idb$IDBArrayKey = $ReadOnlyArray<$idb$IDBValidKey>;
```
julienw, line 4: All types are defined outside a module definition so that they can be reused. Indeed this library has 2 very similar entries, that differ only by a few methods.

julienw:
I rebased on top of latest master, now this contains only the libdef changes.

definitions/npm/idb_v5.x.x/flow_v0.83.x-/idb_v5.x.x.js
GAntoine: Is there any concern of "pollution" from having these outside the module definition? Might be a silly question, I haven't worked with Flow in a little while...

julienw: Yeah, that will pollute the global scope, that's why I prefixed the names with some prefix.
This is pretty common in the libdefs because we were _always_ polluting the global scope when using commonjs style (with `declare module.exports: XXX`). Only with the module style (`declare export XXX`) the global scope isn't polluted anymore.

See for example https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/algoliasearch_v3.x.x/flow_v0.71.x-v0.103.x/algoliasearch_v3.x.x.js.

I'd be happy to make the prefix more "difficult" to collide with if you think that's needed!

julienw: And I thought it was better (and especially more future-proof) to pollute the global scope than copy-pasting everything.

Issue #3863, opened 2020-07-10
Export DataProxy type in @apollo/react-hooks


ammanvedi:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
 we have this declaration in `@apollo/react-hooks` for all flow versions

```
  declare interface DataProxy {
    readQuery<QueryType, TVariables>(
      options: DataProxyReadQueryOptions<TVariables>,
      optimistic?: boolean
    ): QueryType | null;
   ....
  }
```

**If this is a feature request, what is motivation or use case for changing the behavior?**

In my use case i would like to abstract the behaviour of the `update` config property of the mutation config into a separate function

```
const update = ( proxy: ???, data: FetchResult<SomeReturnType> ) => {...}

myMutation({ variables: {...}, update })
```

at the moment i cannot pull the dataProxy type out, so i would like to propose

```
  declare export interface DataProxy {
    readQuery<QueryType, TVariables>(
      options: DataProxyReadQueryOptions<TVariables>,
      optimistic?: boolean
    ): QueryType | null;
   ....
  }
```

so i can import this type and use it to type my function

I am happy to make a pr for this but am also wondering if this is by design, perhaps other contributors have an idea

@jedwards1211 @PhilSwanson @ganemone

PR #3864, opened 2020-07-10
[@apollo/react-hooks] export DataProxy

ammanvedi:

Other notes:

I have raised https://github.com/flow-typed/flow-typed/issues/3863 and this is a tentative pr to solve that issue. 

Not sure if adding an export needs to be tested so any feedback in that regard would be appreciated


ammanvedi:
@ganemone @jedwards1211 @budde377

PR #3869, opened 2020-07-22
Update @sentry/browser flow types for 5.18

comp615:
Per the change below, Sentry has renamed their props to allowUrls and denyUrls to use inclusive language. Update flow typedef to respect that:

- Links to documentation: https://docs.sentry.io/platforms/javascript/
- Link to GitHub or NPM: https://github.com/getsentry/sentry-javascript/commit/1204302906bbe74edc8a18f8ded047b3041059d7
- Type of contribution: fix

Other notes:
Thanks for the ongoing work :)

### DIFF
--- a/definitions/npm/@sentry/browser_v5.x.x/flow_v0.90.x-/browser_v5.x.x.js
+++ b/definitions/npm/@sentry/browser_v5.x.x/flow_v0.90.x-/browser_v5.x.x.js
@@ -394,8 +394,8 @@ declare module '@sentry/browser' {
             hint?: EventHint,
         ) => Promise<SentryEvent | null> | SentryEvent | null,
 
-        +blacklistUrls?: $ReadOnlyArray<string | RegExp>,
-        +whitelistUrls?: $ReadOnlyArray<string | RegExp>,
+        +denyUrls?: $ReadOnlyArray<string | RegExp>,
+        +allowUrls?: $ReadOnlyArray<string | RegExp>,
         // This really should be typed as:
         //    | $ReadOnlyArray<Integration<any>>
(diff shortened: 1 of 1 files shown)


comp615:
Error appears to be a transient travis-infra failure

gwing33:
> Prior to version 5.17.0, allowUrls and denyUrls were called whitelistUrls and blacklistUrls respectively

While I'm not sure why a breaking change wouldn't bump sentry to v6, I will say this change should be limited to `v5.17.x`.

I'm on `v5.6.2` and I am now getting:
```
Cannot call Sentry.init with object literal bound to the first parameter because property
blacklistUrls is missing in Options [1] but exists in object literal [2]. [prop-missing]
```

comp615:
That makes sense, I should probably make a new version then, but I had seen
other props added late as we were upgrading. So I wasn't sure if they were
just tracking head or what.

Currently both are valid but the old ones are marked as deprecated.

gwing33:
Makes sense, I trust you know more about what's going on than I, flow-typed libraries keepin' me informed 😅.

If it's as you say, then could safely include both sets, even for older `v5.x` versions. Let me know if I can help, I appreciate your efforts here.

comp615:
It's helpful context already to learn how specific people like to get with the versioning of flow types. I don't have any context either really, so if there's a more general guideline, would be helpful. To recap:

- whitelistUrls exists since 5.0.0, and should continue to function 5.x.x
- allowUrls added 5.18.0 and should continue to function 5.x.x

As it seems in this format I can't do ranges like `^5.18.x`, I think adding both makes sense! My hope was to just remove the deprecated aliases and assume people would do the minor version upgrade of the library alongside the flow upgrade. But since it is still valid as a prop in 5.x.x (albeit deprecated) you're right it probably should still exist. I'll re-add in #3870 3870

PR #3873, opened 2020-08-04
Add 'cache' prop for 'react-select_v1.x.x'

aidin36:
- Links to documentation: https://github.com/JedWatson/react-select/blob/v1.3.0/README.md#L147
- Link to GitHub or NPM: https://github.com/JedWatson/react-select/blob/v1.3.0/src/Async.js#L9

### DIFF
--- a/definitions/npm/react-select_v1.x.x/flow_v0.104.x-/react-select_v1.x.x.js
+++ b/definitions/npm/react-select_v1.x.x/flow_v0.104.x-/react-select_v1.x.x.js
@@ -168,7 +168,9 @@ declare module 'react-select' {
     valueRenderer?: (option: OptionType, idx?: number) => React$Node,
     // optional style to apply to the component wrapper
     wrapperStyle?: {...},
-    isSearchable?: boolean
+    isSearchable?: boolean,
+    // whether to cache the search results
+    cache?: boolean
   |};
 
   declare type AsyncProps = {|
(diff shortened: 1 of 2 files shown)


definitions/npm/react-select_v1.x.x/flow_v0.104.x-/react-select_v1.x.x.js
```diff
@@ -168,7 +168,9 @@ declare module 'react-select' {
     valueRenderer?: (option: OptionType, idx?: number) => React$Node,
     // optional style to apply to the component wrapper
     wrapperStyle?: {...},
-    isSearchable?: boolean
+    isSearchable?: boolean,
+    // whether to cache the search results
+    cache?: boolean
```
pascalduez, line 173: Shouldn't it be `void | false` instead?

aidin36: Sorry for the delay.

I'm relatively new to Flow, so I'm not sure.
I test it, and `void | false` cause it to not accept a `boolean` variable. If we want to have the `void` there, should it be `void | boolean`?

pascalduez: The thing is from react-select doc it can only be false, if I'm not mistaken. 
But I'm probably nitpicking here.

PR #3875, opened 2020-08-12
Fix the type of 'cache' prop for 'react-select_v1.x.x'

aidin36:
I added the `cache` prop to `react-select` in another PR https://github.com/flow-typed/flow-typed/pull/3873 . However, the type was wrong. According to the documents, `cache` can be any object, or `false` to disable it.

Sorry for the previous mistake.

- Links to documentation: https://github.com/JedWatson/react-select/blob/v1.3.0/README.md?rgh-link-date=2020-08-04T04%3A11%3A08Z#L147
- Link to GitHub or NPM: https://github.com/JedWatson/react-select/blob/v1.3.0/src/Async.js?rgh-link-date=2020-08-04T04%3A11%3A08Z#L9
- Type of contribution: fix

### DIFF
--- a/definitions/npm/react-select_v1.x.x/flow_v0.104.x-/react-select_v1.x.x.js
+++ b/definitions/npm/react-select_v1.x.x/flow_v0.104.x-/react-select_v1.x.x.js
@@ -170,7 +170,7 @@ declare module 'react-select' {
     wrapperStyle?: {...},
     isSearchable?: boolean,
     // whether to cache the search results
-    cache?: boolean
+    cache?: {...} | false
   |};
 
   declare type AsyncProps = {|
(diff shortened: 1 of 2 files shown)


definitions/npm/react-select_v1.x.x/flow_v0.104.x-/react-select_v1.x.x.js
```diff
@@ -170,7 +170,7 @@ declare module 'react-select' {
     wrapperStyle?: {...},
     isSearchable?: boolean,
     // whether to cache the search results
-    cache?: boolean
+    cache?: {...} | false
```
gusdecool, line 173: This is an invalid syntax and causing issue when run linter.

```
ESLint: 
Parsing error: Unexpected token

  176 |     isSearchable?: boolean,
  177 |     // whether to cache the search results
> 178 |     cache?: {...} | false
      |                 ^
  179 |   |};
  180 | 
  181 |   declare type AsyncProps = {|
```

I think the value should be `{} | false`

pascalduez: Hello, two things here:

1. You shouldn't lint your `flow-typed` folder. 
2. It's a valid syntax, just make sure you have the proper ESLint and sub dependencies versions.

PR #3882, opened 2020-08-27
Move type import inside of declaration

meandmax:
Updating to flow 0.132 results in this being an error. I think moving it inside of the module and make it a type import would fix it right?

### DIFF
--- a/definitions/npm/redux-observable_v0.14.x/flow_v0.104.x-/redux-observable_v0.14.x.js
+++ b/definitions/npm/redux-observable_v0.14.x/flow_v0.104.x-/redux-observable_v0.14.x.js
@@ -1,6 +1,5 @@
-import { Observable } from 'rxjs';
-
 declare module 'redux-observable' {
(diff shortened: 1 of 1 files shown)


pascalduez:
I don't think this work, it should resolve to any. There's not resolution mechanism between lib defs sadly.     
https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#dont-import-types-from-other-libdefs

Brianzchen:
Could be fixed if we have a solution to https://github.com/flow-typed/flow-typed/issues/1857

Brianzchen:
Hey @meandmax so this is now a possibility and I've attempted to actually fix this for you. But it turns out redux-observable types in its current state are broken beyond repair unless they're rewritten at this stage.

Though given the sad news happening at work I doubt you'll want this PR anymore so I'm going to close it. Feel free to raise a new one if you want to work on it again.

Issue #3880, opened 2020-08-25
styled-components - destructuring a prop as a function within a styled(Component) is no longer valid with v5


linuspahl:
**What is the current behavior?**
After updating the styled-component interface definition from [v4](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js) to [v5](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/styled-components_v4.x.x/flow_v0.104.x-/styled-components_v4.x.x.js) it is no longer possible to destructure a prop as a function within a `styled(Component)`.
I am receiving the flow error described here https://github.com/flow-typed/flow-typed/issues/3778 which got fixed for v4 with https://github.com/flow-typed/flow-typed/issues/3779

It looks like the code which got added for v4 interface definition with https://github.com/flow-typed/flow-typed/issues/3779 is missing in the v5 interface definition.

I will create a PR to add the missing part for the v5 interface definition.

**If the current behavior is a bug, please provide the steps to reproduce.**
Please have a look at the mentioned issue, it includes an example.

**What is the expected behavior?**
Should not throw an error, because it is a valid approach.

PR #3779, opened 2020-03-17
Allow function within styled(..)

kyleknighted:
Fixes #3778 

Allows functions to be used and does proper type checking within.

_Thanks go to @dennisoelkers for helping troubleshoot and find a solution for this._

PR #3883, opened 2020-08-27
Allow function within styled(..) for styled-components v5

linuspahl:
As described in https://github.com/flow-typed/flow-typed/issues/3880 it looks like the code which got added by @kyleknighted for the styled-components v4 interface (https://github.com/flow-typed/flow-typed/pull/3779) is not part of the v5 interface definition.

This PR reimplements the code for the v5 interface definition and adds a test for the related use case.
I also tried to add a test which includes `// $FlowExpectedError` but didn't find a suitable one.

There are still some other tests which fail when running `flow-typed run-tests styled-components_v5`, but they are not related to this change. I made sure the test is failing with the previous styled-component v5 interface definition.

Fixes: https://github.com/flow-typed/flow-typed/issues/3880

### DIFF
--- a/definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
+++ b/definitions/npm/styled-components_v5.x.x/flow_v0.104.x-/styled-components_v5.x.x.js
@@ -27,10 +27,11 @@ declare module 'styled-components' {
     | Styles
     | Class<InterpolatableComponent<any>>; // eslint-disable-line flowtype/no-weak-types
 
-  declare export type TaggedTemplateLiteral<I, R> = (
-    strings: string[],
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw (commented):
It appears that the many existing build errors are already existing, and were not caused by this change. Can you confirm that?

linuspahl:
@AndrewSouthpaw I can confirm that, while writing the test it looked like this change leads to less test failures, but there are still many existing.

linuspahl:
Hey @AndrewSouthpaw, thanks again for having a look! If you are still fine with these changes can you merge them?

AndrewSouthpaw:
Hey! sorry about that. I dropped the ball on this one, thanks for the ping.

PR #3889, opened 2020-09-07
Provide a libdef for react-splitter-layout v4

julienw:
- Links to documentation: https://github.com/zesik/react-splitter-layout
- Link to GitHub or NPM:  https://github.com/zesik/react-splitter-layout

Other notes:
I copied the libdef for the version 3 and just added the new import, which is empty because this is a CSS file.

### DIFF
--- a/definitions/npm/react-splitter-layout_v4.x.x/flow_v0.104.x-/react-splitter-layout_v4.x.x.js
+++ b/definitions/npm/react-splitter-layout_v4.x.x/flow_v0.104.x-/react-splitter-layout_v4.x.x.js
@@ -1,0 +1,19 @@
+declare type $$reactsplitterlayout$$Props = {|
+  +customClassName?: string,
+  +vertical?: boolean,
+  +percentage?: boolean,
+  +primaryIndex?: 0 | 1,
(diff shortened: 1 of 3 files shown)


AndrewSouthpaw (approved):
LGTM, thanks for putting this together!

AndrewSouthpaw:
I'm noticing a lot of build errors associated with the project. Assuming you just copy-pasted, I'm imagining these are extant errors; but if you feel like tackling them at some point that'd be grand. 🙏

julienw:
> I'm noticing a lot of build errors associated with the libdef in Travis. Assuming you just copy-pasted, I'm imagining these are extant errors; but if you feel like tackling them at some point that'd be grand. pray

Ah, that's interesting, I'm pretty sure I ran the tests locally (and for more Flow versions).

Oh I see, the errors are with the v3, not the v4. Not sure why the CI script decided to run the tests for the v3 libdef...

Indeed I fixed some of the errors for newer Flow versions in the v4, simply by moving around the `FlowExpectedError` lines, because they were reported at different locations.
Probably I can fix it by copying the test file back to the v3.

PR #3028, opened 2018-12-21
WIP: Add cypress

lukeapage:
Mostly just needs more tests

PR #3894, opened 2020-09-09
Add cypress v5 definitions

Brianzchen:
- Links to documentation: https://docs.cypress.io/guides/overview/why-cypress.html
- Link to GitHub or NPM: https://github.com/cypress-io/cypress

Other notes: Thanks to @lukeapage for #3028 which was a great starting base

### DIFF
--- a/definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
+++ b/definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
@@ -417,10 +417,12 @@ declare interface Cypress$Chainable {
   ): any,
 
   /**
+   * Returns the window object
+   *
    * @see https://docs.cypress.io/api/commands/reload.html
    */
-  reload(options?: Cypress$LoggableTimeoutable): Window,
-  reload(forceReload: boolean): Window,
+  reload(options?: Cypress$LoggableTimeoutable): any,
+  reload(forceReload: boolean): any,
 
   /**
    * @see https://docs.cypress.io/api/commands/request.html
@@ -548,6 +550,7 @@ declare interface Cypress$Chainable {
    * @see https://docs.cypress.io/api/commands/uncheck.html
    */
   uncheck(options?: Cypress$CheckOptions): Cypress$Global,
+  uncheck(value: string): Cypress$Global,
   uncheck(values: Array<string>): Cypress$Global,
 
   /**
@@ -629,7 +632,7 @@ declare interface Cypress$Timeoutable {
 declare type Cypress$LoggableTimeoutable = { ... } & Cypress$Loggable & Cypress$Timeoutable
 
 declare type Cypress$ClockObject = {|
-  tick: (milliseconds: number) => Cypress$ClockObject,
+  tick: (milliseconds: number) => Promise<Cypress$ClockObject>,
   restore: () => void,
 |};
 
--- a/definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
+++ b/definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
@@ -477,15 +477,13 @@ declare interface Cypress$Chainable {
   /**
    * @see https://docs.cypress.io/api/commands/select.html
    */
-  select(text: string, options?: Cypress$SelectOptions): Cypress$Global,
   select(value: string, options?: Cypress$SelectOptions): Cypress$Global,
-  select(texts: Array<string>, options?: Cypress$SelectOptions): Cypress$Global,
   select(values: Array<string>, options?: Cypress$SelectOptions): Cypress$Global,
 
   /**
    * @see https://docs.cypress.io/api/commands/server.html
    */
-  server(options?: Cypress$ServerOptions): Cypress$Global,
+  server(options?: Cypress$ServerOptions): void,
 
   /**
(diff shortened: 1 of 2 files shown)


Brianzchen:
I still need to add more tests. Just opening PR now to get early feedback

definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
```diff
@@ -0,0 +1,803 @@
...
+declare type Cypress$Env = { [key: string]: string, ... };
+
+declare type Cypress$Window = any;
+
+declare interface Cypress$Core {
+  /**
+   *
+   * @see https://docs.cypress.io/api/cypress-api/custom-commands.html
+   */
+  addChildCommand(name: string, fn: Function): void,
```
pascalduez, line 65: `Function` has been aliased to `any` from Flow 0.88.0.

**suggestion:** replace with a real function type?

definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
```diff
@@ -0,0 +1,803 @@
...
+  config(key: string, value: any): void,
+  config(object: { ... }): void,
+
+  /**
+   * @see https://docs.cypress.io/api/cypress-api/env.html
+   */
+  env(): Cypress$Env,
+  env(name: string): string,
+  env(name: string, value: string): void,
+  env(object: Cypress$Env, ...rest: Array<void>): void,
```
pascalduez, line 84: I guess typing rest params is more like `...rest: Array<any>`.

definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
```diff
@@ -0,0 +1,803 @@
...
+    isElement: (element: any) => boolean,
+    isFocusable: (element: any) => boolean,
+    isFocused: (element: any) => boolean,
+    isHidden: (element: any) => boolean,
+    isJquery: (element: any) => boolean,
+    isScrollable: (element: any) => boolean,
+    isVisible: (element: any) => boolean,
+    isWindow: (element: any) => boolean,
+    unwrap: (element: any) => Array<any>,
+    wrap: (element: any) => any,
```
pascalduez, line 163: **suggestion:** can't all those `any` be replaced with something like `Element` or `HTMLElement`?

definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
```diff
@@ -0,0 +1,803 @@
...
+  body?: Cypress$RequestBody,
+  failOnStatusCode?: boolean,
+  followRedirect?: boolean,
+  form?: boolean,
+  gzip?: boolean,
+  headers?: Object,
+  method?: Cypress$HttpMethod,
+  qs?: string,
+  url: string
+|} & Cypress$Loggable & Cypress$Timeoutable
```
pascalduez, line 693: **suggestion:** prefer spread to intersection for exact objects.

pascalduez (changes requested):
Thanks for this!

A bunch of `Function` that needs to be replaced by: ideally a real type, otherwise `any`.

definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
Brianzchen: Based on how this usage is documented,
```
cy.get('img').then(($el) => {
  Cypress.dom.isVisible($el) // true
})
```

I tried the following in my code
```
cy.get(selectors.loginLink).then((ele) => {
  console.log(ele);
});
```
Which gave me
Screen Shot 2020-09-25 at 8 32 56 am

From what I can tell cypress has their own wrapped objects for all these things. Even in their docs when it says it should return a dom element, they actually mean it's an instance of `Cypress$Global` which annoyed me so much when typing it

definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
Brianzchen: Wow! I didn't know that, this is so misleading. I found the docs for it so I'll make the changes to use `(...args: Array<any>) => any` instead

https://flow.org/en/docs/types/functions/#toc-function-type

definitions/npm/cypress_v5.x.x/flow_v0.25.x-/cypress_v5.x.x.js
Brianzchen: Do you mean to say, change it to this instead?

For some reason when I do this, method overloading doesn't work anymore for the functions that use this type as an option, any idea how I could fix this?
```
declare type Cypress$RequestOptions = {|
  ...Cypress$Loggable,
  ...Cypress$Timeoutable,
  auth?: Object,
  body?: Cypress$RequestBody,
  failOnStatusCode?: boolean,
  followRedirect?: boolean,
  form?: boolean,
  gzip?: boolean,
  headers?: Object,
  method?: Cypress$HttpMethod,
  qs?: string,
  url: string
|}
```

pascalduez (approved):

pascalduez:
Okay let's move forward and improve in later PRs if needed.

Thanks!

PR #3895, opened 2020-09-11
Add definitions for Yargs v16

julienw:
- Links to documentation: https://github.com/yargs/yargs
- Link to GitHub or NPM:  https://github.com/yargs/yargs
- Type of contribution: new definition and fix

Other notes:
I mostly copied over the libdef from the v15 without checking more. From yargs' documentation nothing more happened from the API side. Only a function was removed from the API but it wasn't in the libdef because not in the official API.

I fixed the test and realized that we had some other tests with `$ExpectError` instad of `$FlowExpectedError` so I changed them in a subsequent commit. I didn't run the tests for that ones though.

### DIFF
--- a/definitions/npm/yargs_v16.x.x/flow_v0.104.x-/yargs_v16.x.x.js
+++ b/definitions/npm/yargs_v16.x.x/flow_v0.104.x-/yargs_v16.x.x.js
@@ -1,0 +1,298 @@
+declare module "yargs" {
+  declare type Argv = {
+    [key: string]: any,
+    _: Array<string>,
+    $0: string,
+    ...
+  };
+
+  declare type Options = $Shape<{
+    alias: string | Array<string>,
+    array: boolean,
+    boolean: boolean,
+    choices: Array<mixed>,
+    coerce: (arg: {[key: string]: any, ...} | any) => mixed,
+    config: boolean,
+    configParser: (configPath: string) => { [key: string]: mixed, ... },
+    conflicts: string | Array<string> | { [key: string]: string, ... },
+    count: boolean,
+    default: mixed,
+    defaultDescription: string,
+    demandOption: boolean | string,
+    desc: string,
+    describe: string,
+    description: string,
+    global: boolean,
+    group: string,
+    implies: string | { [key: string]: string, ... },
+    nargs: number,
+    normalize: boolean,
+    number: boolean,
+    required: boolean,
+    requiresArg: boolean,
+    skipValidation: boolean,
+    string: boolean,
+    type: "array" | "boolean" | "count" | "number" | "string",
+    ...
+  }>;
+
+  declare type CommonModuleObject = {|
+    command?: string | Array<string>,
+    aliases?: Array<string> | string,
+    builder?: { [key: string]: Options, ... } | ((yargsInstance: Yargs) => mixed),
+    handler?: ((argv: Argv) => void) | ((argv: Argv) => Promise<void>)
+  |};
+
+  declare type ModuleObjectDesc = {|
+    ...CommonModuleObject,
+    desc?: string | false
+  |};
+
+  declare type ModuleObjectDescribe = {|
+    ...CommonModuleObject,
+    describe?: string | false
+  |};
+
+  declare type ModuleObjectDescription = {|
+    ...CommonModuleObject,
+    description?: string | false
+  |};
+
+  declare type ModuleObject =
+    | ModuleObjectDesc
+    | ModuleObjectDescribe
+    | ModuleObjectDescription;
+
+  declare type MiddleWareCallback =
+    | (argv: Argv, yargsInstance?: Yargs) => void
+    | (argv: Argv, yargsInstance?: Yargs) => Promise<void>;
+
+  declare type Middleware = MiddleWareCallback | Array<MiddleWareCallback>;
+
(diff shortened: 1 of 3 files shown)


pascalduez:
Thanks @julienw !

The tests seems completely out of the blue :\ testing a bunch of other lib-defs. 
Could you try re-triggering them? I don't seem to have access to the the "re-run" button in Travis.

julienw:
Aaaah I won't fix the world in this patch :-)

There are warnings in older yargs libdefs that I didn't touch, just because `FlowExpectedError` expects a code now.
There are errors in older libdefs:
* bluebird is failing but I'll definitely not fix that. I just used the right `$FlowExpectedError` in a search-and-replace operation.
* jest errors seem to be expected... I can look whether I've put the FlowExpectedError line in the right location. There are errors in older versions that I didn't touch though. Sometimes Flow changes the location of errors and that's a bit painful to fix in these tests :/
* Then it's react-redux but I definitely don't want to look at these libdefs 😅. My guess is that the error location changed in recent versions, because I see a bunch of "unused error suppression" warnings.

I just did a search-and-replace on old `$ExpectError` pragmas to use the new ones. I think they were there because the PR that added these lines was started before my patch where I changed all this some weeks ago landed (it took some time to land if you remember well...). I can try to run these libdefs on master so that we can see if there's a difference. What do you think?

definitions/npm/bluebird_v3.x.x/flow_v0.104.x-/test_bluebird_v3.x.x.js
```diff
@@ -51,7 +51,7 @@ Bluebird.all([1, Bluebird.resolve(1), Promise.resolve(1)]).then(function(r: Arra
 Bluebird.all(['hello', Bluebird.resolve('world'), Promise.resolve('!')]).then(function(r: Array<string>) { });
 Bluebird.all(['hello', Bluebird.resolve('world'), Promise.resolve('!')]).then(function(r: [string, string, string]) { });
 
-// $ExpectError Wrong tuple type
+// $FlowExpectederror Wrong tuple type
```
Brianzchen, line 54: Should these be `$FlowExpectedError` instead of `$FlowExpectederror`?

julienw: Oh wow you're totally right. I'll look at that.

julienw:
I'm splitting this PR in 2 parts, so that it will be easier to review. I keep here only the changes to yargs, and will be putting the other changes in separate PRs.

julienw:
Now that's only warnings on the libdefs for yargs versions I didn't touch.

julienw:
I had a look at the jest libdefs: as I supposed the error suppression line is at the wrong location, and that changed in flow v0.132. So there's no way to make a test file that works for both newer and older flow versions. The obvious solution is to make 2 subdirectories so that the test files can be different...
Also there's a ton of warnings because $FlowExpectedError expect the failed check code.
I don't have the time to fix this now, sorry. So it will stay like this for now...

There are also a bunch of errors in react-redux but I definitely don't want to look at those.

definitions/npm/yargs_v16.x.x/flow_v0.104.x-/yargs_v16.x.x.js
```diff
@@ -0,0 +1,296 @@
...
+      desc: string | false,
+      module: ModuleObject
+    ): this;
+
+    command(module: ModuleObject): this;
+
+    commandDir(
+      directory: string,
+      options?: {
+        exclude?: string | Function,
```
pascalduez, line 105: `Function` -> `any`, if you could type it a bit more precisely?

pascalduez:
Test are failing but only for older untouched versions, so fine.

Only a few `Function` that could get replaced and we are good to go :sweat_smile:

definitions/npm/yargs_v16.x.x/flow_v0.104.x-/yargs_v16.x.x.js
julienw: TBH I just copied the libdef from v15 ahah. I don't know this lib at all.
Do you really want me to look at this before landing? Obviously I didn't have the time in 2 weeks, so I'm not sure I'll be able to in the future either.

Brianzchen: He just means don't use `Function` because it alias' to `any`. I got the same comment here :)

julienw: Yeah I understand the comment. What I'm saying is that I just copied this over from v15. So this isn't a regression from what we had before, and still providing some type fo the users of yargs v16 even if not perfect.

pascalduez: Okay, not forcing you to update those ;)   
But now we need a rebase... sorry @julienw.

julienw:
@pascalduez I rebased and reported the changes from #3911 in the new libdef too. Also I ran the tests locally on all supported flow versions, everything passes successfully \o/

pascalduez (approved):
Thanks @julienw !

PR #3899, opened 2020-09-21
[enzyme_v3.x.x.js] Added Wrapper.props() return type, and fixed test warnings

eidheim:
- Links to documentation: https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/props.html

Other notes:
This change makes it possible to infer `props()` type, while previous uses should mostly work as before. If not, one can specify type through a generic argument through `props<T>(): React$ElementProps<T>`.

### DIFF
--- a/definitions/npm/enzyme_v2.3.x/flow_v0.104.x-/test_enzyme-v2.3.js
+++ b/definitions/npm/enzyme_v2.3.x/flow_v0.104.x-/test_enzyme-v2.3.js
@@ -13,7 +13,7 @@ const B: boolean = render(<div />, { context: { foo: true } })
   .equals(<div />);
 
 // Test against chaining returning `any`
-// $FlowExpectedError
+// $FlowExpectedError[incompatible-cast]
 (shallow(<div />).children(): boolean);
 
 (shallow(<div />).children(): ShallowWrapper);
@@ -30,7 +30,7 @@ const StatelessComponent = () => <div />;
 shallow(<div />).find(ClassComponent);
 shallow(<div />).find(StatelessComponent);
 shallow(<div />).find({ a: 1 });
-// $FlowExpectedError
+// $FlowExpectedError[incompatible-cast]
 shallow(<div />).find(true);
 
 shallowWrapper.instance();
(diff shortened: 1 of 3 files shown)


eidheim:
Note that `props<T>(): React$ElementProps<T>` could possibly be `props<T>(): React$ElementProps<Class<T>>`. I'll leave that up to the reviewer.

definitions/npm/enzyme_v2.3.x/flow_v0.104.x-/test_enzyme-v2.3.js
```diff
@@ -30,7 +30,7 @@ const StatelessComponent = () => <div />;
 shallow(<div />).find(ClassComponent);
 shallow(<div />).find(StatelessComponent);
 shallow(<div />).find({ a: 1 });
-// $FlowExpectedError
+// $FlowExpectedError[incompatible-cast]
```
Brianzchen, line 33: Looks like you've used the wrong error code here
```suggestion
// $FlowExpectedError[incompatible-call]
```

Brianzchen:
Can you add a test to validate this works as intended? I added your change to my project and it throws a bunch of errors, mainly because I wrap everything in `React.AbstractComponent` that returns `mixed`. So I tried adding `expect(wrapper.find('input').props<typeof props>().id).toBe(props.id);` but it doesn't overload to the second definition properly.

eidheim:
I think your usage of the second definition makes more sense than what I had in mind. That is the generic argument should be prop type, not component/element type as it is now. I'll have a look at this in the coming days, and if I can make it work, I'll create tests as you suggest.

eidheim:
Right now, you would have to write:
```js
expect(wrapper.find('input').props<'input'>().id).toBe(props.id)
```
And if you have a React component called `Input` with props `{type: string}` you can simply write:
```js
expect(wrapper.find(Input).props().type).toBe('text')
```
and `type` would be inferred.

But a second option is to change the definitions to:
```js
    props(): Object,
    props<T = React$ElementProps<RootComponent>>(): T,
```
Then the first definition would be untouched, but one could still write:
```js
expect(wrapper.find('input').props<typeof props>().id).toBe(props.id)
```
And if you have a React component called `Input` with props `{type: string}` you can write:
```js
expect(wrapper.find(Input).props<>().type).toBe('text')
```
and `type` would be inferred. 

Maybe the second option is better?

Note that the real issue, is that `find()` and similar functions makes inferring impossible since the argument is not necessarily a JSX intrinsic string or component.

eidheim:
I'm closing this due to difficulties inferring correct type for all use cases.

PR #3906, opened 2020-09-24
Add type and test for reduxjs toolkit

enjikaka:
- Links to documentation: https://redux-toolkit.js.org/api/createReducer https://redux-toolkit.js.org/api/createAction
- Link to GitHub or NPM: https://github.com/reduxjs/redux-toolkit

Other notes: Only typed createAction and createReducer as that are they only ones we have use for, doing the rest is quite a pain. Based of the typescript defs. Impossible to get 100 % right as Immer (which Toolkit uses for state/reducers) cannot be typed in Flow since Flow does not support making ReadOnly-obj writeable - which is how Immer solves immutability (it acts as mutable). This is at least enough to get 100 % coverage in actions.js and reducers.js files. Note that, just like the TypeScript defs, types only cover the whole reducer if the builder callback is used - the object map cannot covered in the same way.

### DIFF
--- a/definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/test_toolkit_v1.x.x.js
+++ b/definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/test_toolkit_v1.x.x.js
@@ -1,0 +1,49 @@
+// @flow
+import { describe, it } from 'flow-typed-test';
+import { createAction, createReducer } from '@reduxjs/toolkit';
+
+describe('@redux/toolkit', () => {
+  describe('createAction', () => {
+    it('supports being called with an action type string', () => {
+      createAction('increment');
+    });
+
+    it('supports being called with an action type string and a payload type', () => {
+      type Payload = $ReadOnly<{|
+        someData: 'The FitnessGram Pacer Test is a multistage aerobic capacity test that progressively gets...',
+      |}>;
+
+      createAction<Payload>('injectData');
+    });
+  });
+
+  describe('createReducer', () => {
+    it('supports being called with an object map', () => {
+      const state = {
+        name: 'Pelle',
+      };
+
+      const SET_NAME = 'SET_NAME';
+
+      createReducer(state, {
+        [SET_NAME]: (state, action) => {
+          state.name = action.payload.name;
+        }
+      });
+    });
+
+    it('supports being called with an builder callback', () => {
+      const state = {
+        name: 'Pelle',
+      };
+
+      const setName = createAction<string>('SET_NAME');
+
+      createReducer(state, builder => {
+        builder.addCase(setName, (state, action) => {
+          state.name = action.payload.name;
+        });
+      });
+    });
+  });
+});
--- a/definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/toolkit_v1.x.x.js
+++ b/definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/toolkit_v1.x.x.js
@@ -1,5 +1,14 @@
 declare module '@reduxjs/toolkit' {
-  import type { AnyAction, Action, Reducer } from 'redux';
+  // These types are copied directly from the redux libdef.
+  // Importing them in this libdef causes a loss in type coverage.
+  // import type { Action, Reducer } from 'redux';
+
+  declare type Action<T> = {
+    type: T,
+    ...
+  };
+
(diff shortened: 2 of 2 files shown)


definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/toolkit_v1.x.x.js
```diff
@@ -0,0 +1,194 @@
+declare module '@reduxjs/toolkit' {
+  import type { AnyAction, Action, Reducer } from 'redux';
```
pascalduez, line 2: Sadly that can't work, those types will all resolve to `any`. 
See https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#dont-import-types-from-other-libdefs 

**suggestion:** copy paste those types here in the lifdef...

definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/toolkit_v1.x.x.js
```diff
@@ -0,0 +1,194 @@
...
+
+
+  /**
+   * Basic type for all action creators.
+   *
+   * @inheritdoc {redux#ActionCreator}
+   */
+  declare interface BaseActionCreator<P, T = string> {
+    type: T;
+    match: Action<unknown> => boolean;
```
pascalduez, line 108: AFAIK there's no `unknown` type in Flow. Use `mixed` instead?

definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/test_toolkit_v1.x.x.js
```diff
@@ -0,0 +1,49 @@
...
+      const setName = createAction<string>('SET_NAME');
+
+      createReducer(state, builder => {
+        builder.addCase(setName, (state, action) => {
+          state.name = action.payload.name;
+        });
+      });
+    });
+  });
+});
```
pascalduez, line 49: **suggestion:** could you add a few error assertions as well?

definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/toolkit_v1.x.x.js
pascalduez: Maybe using [this workaround](https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#dont-import-types-from-other-libdefs), so we have some sort of "convention".

definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/test_toolkit_v1.x.x.js
enjikaka: The test suite does not properly run those it seems so I skipped them. I just get `Unused suppression comment` on all `// $FlowExpectedError`s, Even though the type is correct and should give error. It spams that form other libdefs too.

definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/toolkit_v1.x.x.js
enjikaka: Left over from TS defs. Why didn't Flow say it doesn't exist...? I'll change it to `mixed`.

definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/toolkit_v1.x.x.js
enjikaka: Inlined them now.

definitions/npm/@reduxjs/toolkit_v1.x.x/flow_v0.104.x-/test_toolkit_v1.x.x.js
Brianzchen: Can you give an example of it not working or the issues you're getting?

pascalduez (approved):
Okay, let's move forward, we want this to happen :)   
There will be plenty of time to improve.

enjikaka:
Sorry for not hearing back! I abandoned this experiment at work, don't remember why now: I think we just concluded time would be better spent getting started with TypeScript. Hopefully this is a good start for you to improve upon though. :)

PR #3907, opened 2020-09-26
Storybook react v6

Brianzchen:
- Links to documentation: https://www.npmjs.com/package/@storybook/react
- Link to GitHub or NPM: https://www.npmjs.com/package/@storybook/react

Other notes:
`Context` type is not well documented and was derived from what was printed in the console
Screen Shot 2020-09-26 at 4 54 13 pm


Brianzchen:
@pascalduez I was hoping to remove the describe/it blocks to move `module` to top level but I'm still getting errors. Any recommendations?

AndrewSouthpaw:
> I was hoping to remove the describe/it blocks to move module to top level but I'm still getting errors. Any recommendations?

Hmm I would imagine that wouldn't work because it's a restriction of ES6 syntax. Were the tests working before? I'm wondering if they just never worked...

I haven't worked with storybook, but could you bundle it as an imported constant rather than referring explicitly to `module`?

Brianzchen:
> > I was hoping to remove the describe/it blocks to move module to top level but I'm still getting errors. Any recommendations?
> 
> Hmm I would imagine that wouldn't work because it's a restriction of ES6 syntax. Were the tests working before? I'm wondering if they just never worked...
> 
> I haven't worked with storybook, but could you bundle it as an imported constant rather than referring explicitly to `module`?

I don't think it ever worked, I got errors as soon as I started working on it. I'm now thinking maybe I'll just declare a var for a fake module and use that instead

AndrewSouthpaw (changes requested):
Nice, I'm glad that worked. Would you be able to reinstate the `describe/it` blocks?

Brianzchen:
@AndrewSouthpaw all done!

pascalduez:
I swear this worked perfectly back in time, maybe something changed at the Flow level.   
But if we can't test with real world usage how can we assert this is even working, something doesn't click here, I'm not to keen on the f`akeModule` thing. Sorry for not bringing much help here, I would need to delve in to better understand...

Brianzchen:
> I swear this worked perfectly back in time, maybe something changed at the Flow level.
> But if we can't test with real world usage how can we assert this is even working, something doesn't click here, I'm not to keen on the f`akeModule` thing. `module` is a Node global. Sorry for not bringing much help here, I would need to delve in to better understand...

Can I leave this with you to figure out? I've done my best and tried multiple ways. You can check my previous test runs to see the behaviours that occurred.

pascalduez:
> Can I leave this with you to figure out?

Will try, but current personal situation is quite tense, probably won't be able to invest much time in it soon.

Brianzchen:
Yeah I don't mind keeping this around. Given that storybook v6 came out recently it'd be great to supply users with defs but v5 still gets the job done at least

AndrewSouthpaw:
Alright. Sounds like the decision for now is to leave this PR open until @pascalduez has time later to investigate.

Brianzchen:
@pascalduez any news on this one?

definitions/npm/@storybook/react_v6.x.x/flow_v0.25.x-/react_v6.x.x.js
```diff
@@ -0,0 +1,64 @@
...
+    kind: string;
+    filename: string;
+    stories: Array<StoryObject>;
+  }
+
+  declare function addDecorator(decorator: StoryDecorator): void;
+  declare function addParameters(parameters: DecoratorParameters): void;
+  declare function clearDecorators(): void;
+  declare function configure(fn: () => void, module: NodeModule): void;
+  declare function setAddon(addon: Object): void;
```
pascalduez, line 58: `Object` to be replaced.

definitions/npm/@storybook/react_v6.x.x/test_react_v6.x.x.js
```diff
@@ -0,0 +1,205 @@
+// @flow
+
+import { describe, it } from 'flow-typed-test';
+import React from 'react';
```
pascalduez, line 4:
```suggestion
import * as React from 'react';
```

PR #3887, opened 2020-08-31
Add a definition for react-modal-v3.11.x

knu:
- Links to documentation: http://reactcommunity.org/react-modal/
- Link to GitHub or NPM: https://github.com/reactjs/react-modal
- Type of contribution: new definition for a new version

PR #3912, opened 2020-09-28
Minor improvements to react-modal-v3.11.x

Brianzchen:
- Links to documentation: http://reactcommunity.org/react-modal/
- Link to GitHub or NPM: https://github.com/reactjs/react-modal
- Type of contribution: fix

Other notes:
Improvements to #3887 

@knu @AndrewSouthpaw


definitions/npm/react-modal_v3.11.x/flow_v0.104.x-/react-modal_v3.11.x.js
```diff
@@ -1,6 +1,6 @@
 declare module 'react-modal' {
   declare type ModalProps = {
-    isOpen?: boolean,
+    isOpen: boolean,
```
pascalduez, line 3: It's a `defaultProp`: https://github.com/reactjs/react-modal/blob/master/src/components/Modal.js#L79

Brianzchen: True, I was matching it up with https://github.com/reactjs/react-modal/blob/master/src/components/Modal.js#L31 and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-modal/index.d.ts#L56 which both specify as required.

It's probably not necessary then since user doesn't get runtime proptype errors if using `.defaultProps`.

I'm thinking this PR probably isn't necessary and I may close, thoughts?

PR #3915, opened 2020-10-05
Include libdef for supertest v5 and fix the type for the retry function

julienw:
- Links to documentation: https://github.com/visionmedia/supertest and https://visionmedia.github.io/superagent/#retrying-requests
- Link to GitHub or NPM: https://github.com/visionmedia/supertest and https://github.com/visionmedia/superagent
- Type of contribution: new definition and fix

Other notes:
There's no API change between v4 and v5, so I first fixed the issue in v4, and then copied everything over to a new v5 libdef.


julienw:
Thanks for the quick review!

PR #3920, opened 2020-10-27
react-navigation fix: Replace NavigationNavigateAction with NavigationAction

Taggotty:
- Links to documentation: https://reactnavigation.org/docs/4.x/navigation-actions
- Type of contribution: fix

### DIFF
--- a/definitions/npm/@react-navigation/core_v3.x.x/flow_v0.104.x-/core_v3.x.x.js
+++ b/definitions/npm/@react-navigation/core_v3.x.x/flow_v0.104.x-/core_v3.x.x.js
@@ -385,7 +385,7 @@ declare module '@react-navigation/core' {
             ...
           },
       params?: NavigationParams,
-      action?: NavigationNavigateAction
+      action?: NavigationAction
     ) => boolean,
     setParams: (newParams: NavigationParams) => boolean,
     ...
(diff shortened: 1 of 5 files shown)


AndrewSouthpaw (approved):
LGTM, seems like the travis failures are from existing issues with the test files.

Ashoat:
This PR is wrong. The types it purports to fix are incorrect, see [here](https://github.com/react-navigation/react-navigation/blob/4.x/packages/core/src/routers/getNavigationActionCreators.ts#L22).

@AndrewSouthpaw - In the future, could you please avoid merging these without asking me to review first?

In this case the fix is for an outdated version of React Navigation, so I'm not really motivated to try and fix them. If somebody could revert that would be great, if not I guess we can leave the types broken.

AndrewSouthpaw:
Hey @Ashoat, sorry about that! I forgot this was the library that you provide reviews on, and it slipped my mind to check the `CONTRIBUTING.md` file. 🤦 

Thanks for reverting it, @pascalduez.

Ashoat:
No worries... and after rereading my comment it was unnecessarily aggressive, sorry about that. I must've had a bad Sunday morning 😅

Taggotty:
Also sorry that I didn't check this up thoroughly enough. Somehow I came across [this code](https://github.com/react-navigation/react-navigation/blob/8d0e0f48adb8c40301fdea2a88c30dd15ec5c353/packages/react-navigation/typescript/react-navigation.d.ts#L582) and came to the wrong conclusions. Have a nice day.

PR #3932, opened 2020-11-12
Add last arg to spy call

AGawrys:
- Type of contribution: new definition | addition | **fix** | refactor

Other notes: Fixing my previous PR where lastArg was a function, it's a property. Thank you


pascalduez (approved):

pascalduez:
Thanks! 

So the test failed to catch this somehow... :disappointed:

PR #3934, opened 2020-11-13
Add last arg to spy call -[FIX]

AGawrys:
- Links to documentation: [sinon docs](https://sinonjs.org/releases/latest/spy-call/)
- Type of contribution: new definition | addition | **fix** | refactor

Other notes: LastArg should actually be mixed since it's based on the object it's called on 😥  Fix for parcel


AGawrys:
@pascalduez Sorry about the previous PRs, this should be the final fix 😳

pascalduez (approved):

pascalduez:
@AGawrys No worry, I should also have caught it at the review... :relaxed:   

Side note: would be nice to update the tests with suppression comment codes.

PR #3936, opened 2020-11-14
[luxon_v1.9.x] add toRelative and toRelativeCalendar

benthemonkey:
- Links to documentation:
      toRelative: https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-toRelative
      toRelativeCalendar: https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-toRelativeCalendar
- Link to GitHub or NPM: https://github.com/moment/luxon

Other notes:
The methods I'm adding were not added to the library until [v1.9.0](https://github.com/moment/luxon/blob/master/CHANGELOG.md#190) . I wasn't sure whether there's a remedy for that.

### DIFF
--- a/definitions/npm/luxon_v0.2.x/flow_v0.104.x-/test_luxon.js
+++ b/definitions/npm/luxon_v0.2.x/flow_v0.104.x-/test_luxon.js
@@ -33,7 +33,7 @@ Settings.defaultZone = new CustomZone();
 
 (Info.eras("long"): Array<string>);
 (Info.eras("long", { locale: "de-DE" }): Array<string>);
-// $FlowExpectedError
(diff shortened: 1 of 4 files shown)


pascalduez:
>  The methods I'm adding were not added to the library until v1.9.0 . I wasn't sure whether there's a remedy for that.

To be safe, we should clone that libdef and scope it with the proper version, in this case `luxon_v1.9.x`.

That would be a good occasion to try to fix/update the tests?

benthemonkey:
@pascalduez  let me know if there's anything else I need to do. I think it's ready to go.

pascalduez (approved):
Thanks!

PR #3945, opened 2020-11-20
Add libdef files for @entur/sdk

draperunner:
- Links to documentation: https://sdk.entur.org
- Link to GitHub or NPM: https://github.com/entur/sdk

These library definitions have been (and currently are) part of the package's repository, but will be removed from there in the future.


definitions/npm/@entur/sdk_v1.x.x/flow_v0.25.x-/sdk_v1.x.x.js
```diff
@@ -0,0 +1,1032 @@
...
+        geocoder?: string,
+        nsr?: string,
+        scooters?: string,
+    |},
+    headers?: {|
+        [string]: string
+    |},
+    fetch?: (
+        url: string,
+        init?: Object,
```
pascalduez, line 97: `Object` is deprecated and now an alias to `any`, could it be type more precisely?

definitions/npm/@entur/sdk_v1.x.x/flow_v0.25.x-/sdk_v1.x.x.js
```diff
@@ -0,0 +1,1032 @@
+type $entur$sdk$TransportMode =
```
pascalduez, line 1: The inexact object syntax `...` was added in Flow 0.83.x, so it should be scoped to at least that version. 
We usually use 0.104.x, for when we run the codemod on the whole repo.

definitions/npm/@entur/sdk_v1.x.x/test_sdk_v1.x.x.js
```diff
@@ -0,0 +1,17 @@
+import { describe, it } from 'flow-typed-test';
+import createEnturService from '@entur/sdk'
+
+describe('#someFunction', () => {
```
pascalduez, line 4:
```suggestion
describe('#createEnturService', () => {
```

definitions/npm/@entur/sdk_v2.x.x/test_sdk_v2.x.x.js
```diff
@@ -0,0 +1,17 @@
...
+    })
+  });
+
+  it('fail if using wrong typefor service', () => {
+    // $FlowExpectedError[incompatible-type] wrong return type
+    const numberService: number = createEnturService({
+      clientName: 'flow-typed'
+    })
+  });
+})
```
pascalduez, line 17: I guess it could use a few more tests?

draperunner: I agree. I'll add some.

definitions/npm/@entur/sdk_v1.x.x/flow_v0.25.x-/sdk_v1.x.x.js
draperunner: I bumped the range to 0.83-

definitions/npm/@entur/sdk_v1.x.x/flow_v0.25.x-/sdk_v1.x.x.js
draperunner: In this specific case, I would prefer to keep the type not too specific. Users of this library can substitute the default `fetch` implementation with their own, for advanced use cases. I know that `node-fetch` uses some really specific union types for this `init` parameter, while the browser `fetch` uses some other type. If I restrict this `init` parameter too much, I think that can lead to type conflicts for users that are not really helpful.

There are some other usages of Object in this file though, I'll see if I can improve those.

pascalduez: Then please replace them with `any`, at least it's explicit.

pascalduez (approved):
Thanks!

PR #3951, opened 2020-11-25
Cleanup travis

### LINKED ISSUE #3949: Move to github actions from travis?
New builds now take forever because of an apparently recent change to travis that stifles open source projects.
related: https://www.theregister.com/2020/11/02/travis_ci_pricng/

Was recommended that we should switch to github actions. Which should be reliable and has unlimited runs

Create a PR in flow-typed and you'll be waiting for the build for over an hour

It should run and finish within 4 minutes

Builds are slow

The build machine

Brianzchen:
Closes #3949

Remove references to Travis and replace badge with github actions


.github/workflows/main.yml
```diff
@@ -31,5 +31,10 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
 
+      - name: Before Install
+        run: |
+          git config --global user.name "Travis CI"
```
pascalduez, line 36: "Cleanup travis" ?  :wink:

Brianzchen: 🤦 Wonder why this never went through my head. Maybe you guys wanna just update this name directly in master? 😄

PR #3954, opened 2020-11-26
Docs: update new libdefs minimum Flow version

pascalduez:
We want to explicitly use either exact or inexact objects for all new definitions. 
Since the inexact object syntax was added in Flow 0.83.x, it does not make sense to keep scoping new libdef to 0.25.x?
Anyway, we only test the last 15 versions...


GAntoine (approved):

AndrewSouthpaw (approved):

AndrewSouthpaw:
Thanks @pascalduez!

PR #3955, opened 2020-11-26
Fix Aphrodite type import resolution

Brianzchen:
- Links to documentation: https://www.npmjs.com/package/aphrodite
- Link to GitHub or NPM:  https://www.npmjs.com/package/aphrodite
- Type of contribution: fix

Other notes:
Rename types `StyleSheetTestUtils` and `StyleSheetServer` as I realised that even if you don't have the `export` keyword it is still able to be imported which causes a conflict with the real named export.


pascalduez (approved):

pascalduez:
> I realised that even if you don't have the export keyword it is still able to be imported

That's because you used a commonjs module declaration.   
ES modules preferred when possible.

Issue #3965, opened 2020-12-04
install-test is flaky in CI


jedwards1211:
Getting this error randomly in some of my PRs ([example](https://github.com/flow-typed/flow-typed/runs/1501427959?check_suite_focus=true)):

```
$ jest && yarn lint
FAIL src/commands/__tests__/install-test.js
  ● install (command) › end-to-end tests › overwrites tweaked libdefs when --overwrite is passed

    expect(received).toBe(expected) // Object.is equality

    Expected: "// flow-typed signature: 7a7fa7619340bfebb825752ae5efb893
    // flow-typed version: 0beab4f35a/foo_v1.x.x/flow_>=v0.38.x·
    "
    Received: ""

      631 | 
      632 |         // Verify that the tweaked libdef file wasn't overwritten
    > 633 |         expect(await (await fs.readFile(libdefFilePath)).toString()).toBe(
          |                                                                      ^
      634 |           libdefFileContent,
      635 |         );
      636 |       });

      at toBe (src/commands/__tests__/install-test.js:633:70)
      at src/commands/__tests__/install-test.js:294:11
      at testProject (src/lib/TEST_UTILS.js:31:3)
      at fakeProjectEnv (src/commands/__tests__/install-test.js:271:14)
```

jedwards1211:
What's with `await (await fs.readFile(libdefFilePath)).toString()`?  `Buffer.toString()` is not async

AndrewSouthpaw:
Is this still an issue for you @jedwards1211?

jedwards1211:
Probably but I don't know yet, my sequelize PR is the first one I've made since reporting this

PR #3966, opened 2020-12-04
refactor: clean up fs.readFile usage

jedwards1211:
The way `fs.readFile` was being used looked like a code smell, and sure enough, the way it was being mocked would cause surprise errors for anyone who adds idiomatic `fs-extra` calls in new or updated tests.  Pass encoding (`'utf8'`, the default) to get `readFile` to resolve to a string instead of a buffer, and use `readJson` where applicable.  Update fs mocks to behave a bit more like `fs-extra` in this regard (though I'm not sure it would be possible to support all options, but we can at least support encoding string as second argument).

### DIFF
--- a/cli/src/commands/__tests__/install-test.js
+++ b/cli/src/commands/__tests__/install-test.js
@@ -42,7 +42,7 @@ async function touchFile(filePath) {
 }
 
 async function writePkgJson(filePath, pkgJson) {
-  await fs.writeFile(filePath, JSON.stringify(pkgJson));
+  await fs.writeJson(filePath, pkgJson);
 }
 
 describe('install (command)', () => {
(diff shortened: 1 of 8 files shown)


jedwards1211:
argh, nevermind, now I see that `readFile` isn't faithfully mocked

AndrewSouthpaw:
@GAntoine or @villesau if either of you have poked at the CLI lately?

AndrewSouthpaw:
is this related to #3965 ?

jedwards1211:
@AndrewSouthpaw it doesn't fix that issue AFAIK.  I just noticed the weird fs calls while reporting that issue.

pascalduez (approved):

AndrewSouthpaw (approved):

AndrewSouthpaw:
Thanks!

PR #3979, opened 2020-12-28
[koa] add render function to ctx

Brianzchen:
Add `render` function to the `ctx` object. This is only there if the `koa-views` middleware is added but without it it's impossible to call this function without adding a suppression comment.

- Links to documentation: https://www.npmjs.com/package/koa-views
- Link to GitHub or NPM: https://www.npmjs.com/package/koa-views

Other notes: This follows the way jest works with it's middleware for enzyme since we cannot extend lib defs


definitions/npm/koa_v2.x.x/flow_v0.104.x-/koa_v2.x.x.js
```diff
@@ -215,6 +215,8 @@ declare module 'koa' {
     accept: $PropertyType<Request, 'accept'>,
     app: Application,
     cookies: Cookies,
+    // koa-views props
+    render: (root: string, opts?: { [key: string]: any, ... }) => Promise<void>,
```
pascalduez, line 219: Shouldn't it be marked as optional if it's only available trough a plugin?

Brianzchen: I don't think so. At least that's not what was done in jest. I know it's more sound to do a check, but it will definitely exist if the middleware is applied. From what I can see I think this is the general convention until we can figure out a way to extend flow types depending on other dependencies existing.

You probably know better than me, do we have any ideas or plans for that?

pascalduez (approved):

PR #3981, opened 2020-12-31
feat: create sequelize 5/6 defs

jedwards1211:
- Links to documentation: https://sequelize.org/v5/manual/upgrade-to-v5.html#breaking-changes, https://sequelize.org/master/manual/upgrade-to-v6.html
- Link to GitHub or NPM: https://github.com/sequelize/sequelize
Other notes:
I copied over the flow_v0.42.x-v0.103.x folders and applied the same changes to them, but I honestly haven't tested them.  (Is there a way to tell run_tests what versions of flow to use?)

### DIFF
--- a/definitions/npm/sequelize_v5.x.x/CONTRIBUTING.md
+++ b/definitions/npm/sequelize_v5.x.x/CONTRIBUTING.md
@@ -1,0 +1,5 @@
+# Sequelize
+
+## Common reviewers
+
+- jedwards1211 (original author)
--- a/definitions/npm/sequelize_v5.x.x/flow_v0.104.x-/sequelize_v5.x.x.js
+++ b/definitions/npm/sequelize_v5.x.x/flow_v0.104.x-/sequelize_v5.x.x.js
@@ -1,0 +1,7552 @@
+// @flow
+
+declare module 'sequelize' {
+  /**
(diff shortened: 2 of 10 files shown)


AndrewSouthpaw:
> Is there a way to tell run_tests what versions of flow to use?

Not to my knowledge, short of just hacking the CLI code.

AndrewSouthpaw:
@jedwards1211 since this is mostly a copy-pasta job, I'm fine with merging if you've poked it enough to be satisfied.

jedwards1211:
Actually let's wait.  I'm changing a bunch of stuff over to exact objects and spreads.

AndrewSouthpaw:
Sounds good! Let me know when you're ready.

Brianzchen:
@jedwards1211 any news on this?

dannyjlaurence-exp:
We too are in need of these - anything we can do to help @jedwards1211? Even a rough outline of the exact objects or spread changes you were thinking about would point us in the right direction.

Brianzchen:
@dannyjlaurence-exp As far as I can tell these are new definitions which doesn't break anything existing. There's also tests in place to an extent. If you can give us some peace of mind that these defs are useful to you, I think we can merge and improve later

PR #3984, opened 2021-01-05
[lodash.padstart] add def

Brianzchen:
- Links to documentation: https://lodash.com/docs/4.17.15#padStart
- Link to GitHub or NPM: https://www.npmjs.com/package/lodash.padstart

Other notes:


TrySound:
Per package lodash utilities are deprecated and will be removed in v5. Better use lodash/padStart or native `'string'.padStart(10)`

Brianzchen:
Makes sense thanks! Will close this in that case

PR #3985, opened 2021-01-05
Add flow-typed definitions for match-sorter_v5.x.x

eliot-stripe:
- Links to documentation: https://github.com/kentcdodds/match-sorter#usage
- Link to GitHub or NPM: https://github.com/kentcdodds/match-sorter

Other notes:

I based this on match-sorter_v2.3.x and then updated the definitions for the newer version. In particular, the tests are nearly identical.

### DIFF
--- a/definitions/npm/match-sorter_v5.x.x/flow_v0.104.x-/match-sorter_v5.x.x.js
+++ b/definitions/npm/match-sorter_v5.x.x/flow_v0.104.x-/match-sorter_v5.x.x.js
@@ -1,25 +1,26 @@
-type $npm$matchSorter$KeyFn<T> = (item: T) => string | Array<String>;
-type $npm$matchSorter$KeyObj = {|
-  key: string,
-  minRanking?: number,
-  maxRanking?: number,
-  threshold?: number,
-|};
-
-type $npm$matchSorter$Rankings = {|
-  CASE_SENSITIVE_EQUAL: number,
-  EQUAL: number,
-  STARTS_WITH: number,
-  WORD_STARTS_WITH: number,
-  CONTAINS: number,
-  ACRONYM: number,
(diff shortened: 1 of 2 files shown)


definitions/npm/match-sorter_v5.x.x/flow_v0.104.x-/match-sorter_v5.x.x.js
```diff
@@ -0,0 +1,35 @@
+type $npm$matchSorter$KeyFn<T> = (item: T) => string | Array<String>;
+type $npm$matchSorter$KeyObj = {|
+  key: string,
+  minRanking?: number,
+  maxRanking?: number,
+  threshold?: number,
+|};
+
+type $npm$matchSorter$Rankings = {|
```
pascalduez, line 9: Is there any benefit to keep those type globally scoped as is, that used to be the "default" in old lidbefs but should be questioned IMO.

eliot-stripe: Nope; I'm pretty new to writing libdefs, so I wasn't aware of the current best practices. I'll change these to not be globally scoped.

eliot-stripe: I updated the code to not use global types.

I wasn't sure whether to export these intermediate types, so I decided not to initially. Let me know if you disagree.

pascalduez: It's fine, It'll be just a PR way if someone needs them.

pascalduez (approved):
Thanks!

PR #3987, opened 2021-01-07
[aphrodite] Improve types of css properties

Brianzchen:
Went through all css properties that were previously just `string | number` and matched them up with what's in the TS types file, also added anything that was missing from the TS types. This should make this library much more reliable for flow users

- Links to documentation: https://github.com/Khan/aphrodite/blob/master/typings/css-properties.d.ts
- Link to GitHub or NPM: https://www.npmjs.com/package/aphrodite

Other notes:


Brianzchen:
@AndrewSouthpaw any chance you or someone else could have a look at this. I understand it's huge but I'm pretty sure I'm the only person who uses this def since I added it fairly recently

AndrewSouthpaw (approved):
LGTM, thanks for the ping.

PR #3994, opened 2021-01-21
[testing-library/react] fix missing types error

Brianzchen:
Because of flow 0.143.0, lib defs now better throw errors. These are the errors caught from react testing library for types that don't exist.

- Links to documentation: https://github.com/testing-library/react-testing-library
- Link to GitHub or NPM: https://github.com/testing-library/react-testing-library
- Type of contribution: fix

Other notes:

### DIFF
--- a/definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
+++ b/definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
@@ -134,7 +134,7 @@ declare module '@testing-library/react' {
   ) => Promise<HTMLElement>;
 
   declare type FindAllByRole = (
-    role: Matcher,
+    role: any,
     options?: ByRoleOptions,
     waitForElementOptions?: WaitForElementOptions
   ) => Promise<HTMLElement[]>;
@@ -235,7 +235,7 @@ declare module '@testing-library/react' {
     container?: HTMLElement,
     baseElement?: HTMLElement,
     hydrate?: boolean,
-    wrapper?: React.ComponentType,
+    wrapper?: React$ComponentType<any>,
   |};
 
   declare export type RenderOptionsWithCustomQueries<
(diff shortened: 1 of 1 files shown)


definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
```diff
@@ -134,7 +134,7 @@ declare module '@testing-library/react' {
   ) => Promise<HTMLElement>;
 
   declare type FindAllByRole = (
-    role: Matcher,
+    role: any,
```
pascalduez, line 137: I don't get this.

definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
```diff
@@ -245,17 +245,17 @@ declare module '@testing-library/react' {
     container?: HTMLElement,
     baseElement?: HTMLElement,
     hydrate?: boolean,
-    wrapper?: React.ComponentType,
+    wrapper?: React$ComponentType<any>,
   |};
 
   declare export function render(
-    ui: React.ReactElement<any>,
+    ui: React$Node,
```
pascalduez, line 252: That's definitely not the same thing, a Node is really permissive compared to an Element.

pascalduez: Looks like a poor copy/pasta from TS. 
It should be `ui: React$Element<any>`

definitions/npm/@testing-library/react_v10.x.x/flow_v0.104.x-/react_v10.x.x.js
Brianzchen: Definitely agree, should be improved, but as a first pass fixing libraries that blocked people from upgrading to latest flow version I would say is more important.

Basically flow already has a massive learning curve, needing most people to maintain lib defs because the repo that holds them are broken isn't a good thing to have in the community, I would expect that if something broke between versions I should be able to run `flow-typed install` immediately to fix it.

PR #3995, opened 2021-01-21
[webpack] Fix import errors

Brianzchen:
After updating to 0.143.0 new errors in lib defs are being highlighted. We should not be importing outside of a module declaration.

- Links to documentation: https://github.com/webpack/webpack
- Type of contribution: fix

Other notes: `OutgoingMessage` doesn't exist in `http` module. Should probably be added in flow but that's probably a different problem to solve. This at least doesn't break the type and keeps it as it was before which would have exposed `any`


Brianzchen:
Errors are randomly caused by the cli tests which I haven't run

AndrewSouthpaw (approved):

AndrewSouthpaw:
Thanks @Brianzchen

PR #4001, opened 2021-01-22
Fix lowercase MiddlewareCallback to MiddleWareCallback

raedle:
The `yargs_v15.x.x.js` typedef had a typo for the `MiddleWareCallback`. Changed the `w` from lowercase to upperwase `W`.

Test plan:

Run `./build_and_test_cli.sh`
```
Test Suites: 9 passed, 9 total
Tests:       112 passed, 112 total
Snapshots:   0 total
Time:        8.836s
Ran all test suites.
$ eslint .
Done in 14.49s.
Done in 31.68s.
All libdefs are named and structured correctly. (Found 1787)
```

- Link to GitHub or NPM: https://github.com/yargs/yargs/blob/7ffb5e8ae4f4f94f94ffd10a3aa8b410b2ed2fe4/lib/middleware.ts
- Type of contribution: fix

Other notes:

While the source project uses `MiddlewareCallback`, I stuck to the notion of `yargs_v16.x.x.js`, which was using `MiddleWareCallback` throughout.

### DIFF
--- a/definitions/npm/yargs_v15.x.x/flow_v0.118.x-/yargs_v15.x.x.js
+++ b/definitions/npm/yargs_v15.x.x/flow_v0.118.x-/yargs_v15.x.x.js
@@ -67,7 +67,7 @@ declare module "yargs" {
     | (argv: Argv, yargsInstance?: Yargs) => void
     | (argv: Argv, yargsInstance?: Yargs) => Promise<void>;
 
-  declare type Middleware = MiddlewareCallback | Array<MiddleWareCallback>;
+  declare type Middleware = MiddleWareCallback | Array<MiddleWareCallback>;
 
   declare class Yargs {
     (args: Array<string>): Yargs;


pascalduez (approved):

pascalduez:
The way I'm seeing it is it's a typo and `MiddlewareCallback` does not exist so resolving to `any`? 
Happy to revert if it's causing more harm than good.

jedwards1211:
This commit gets rid of a flow error by making a reference match the typo in `MiddleWareCallback`.  I think it would be better to fix by changing all occurrences with uppercase W to lowercase w, and then users of yargs 16+ can search/replace in their code.  Leaving a typo in the code sets a bad precedent

jedwards1211:
It's unfortunate that type defs don't have some kind of release notes though to notify users about changes like that

jedwards1211:
Maybe we could engineer a way to show the git commits that have occurred to a type def since the prior one the user had installed

PR #3996, opened 2021-01-21
[testing-library/react] Fix v11 of the library

Brianzchen:
Because of flow 0.143.0, lib defs now better throw errors. These are the errors caught from react testing library for types that don't exist.

Links to documentation: https://github.com/testing-library/react-testing-library
Link to GitHub or NPM: https://github.com/testing-library/react-testing-library
Type of contribution: fix

Other notes:

PR #4002, opened 2021-01-23
[@testing-library/react_v11.x.x] Improve by role queries

pascalduez:
- Links to documentation: https://github.com/testing-library/dom-testing-library/tree/master/types
- Link to GitHub or NPM: https://github.com/testing-library/dom-testing-library

Other notes: 
Following #3996, improve "by role" queries (only allow actual roles, and not random strings or regex), fix and update matchers and options types.

/cc @Brianzchen


Brianzchen (approved):
LGTM. Though Another use case of needing to solve cross lib defs 😞 https://github.com/flow-typed/flow-typed/issues/1857

pascalduez:
There's indeed also an issue with Github CI tests. There's an error with this libdef tests (on master) which doesn't shows up here...

PR #4005, opened 2021-01-26
Adapted cy.request to return Cypress$Global rather than a promise

debel27:
- Links to documentation: https://docs.cypress.io/api/commands/request.html
- Link to GitHub or NPM: /
- Type of contribution: fix

Other notes: `cy.request` does not actually return a Promise (though it is `then`able). Such as stated in the [examples section](https://docs.cypress.io/api/commands/request.html#Examples), you can chain the result with other Cypress commands

```
cy.request('/admin').its('body');
cy.request('https://jsonplaceholder.cypress.io/comments').as('comments')
```

So it makes more sense for `cy.request` to return `Cypress$Global`, as many other commands. 

Unfortunately, by returning `Cypress$Global` we lose the typing information of what the command actually yields (`Cypress$RequestResponse`). In the future, maybe  `Cypress$Global` should be parametrized ?

@Brianzchen FYI

### DIFF
--- a/definitions/npm/cypress_v6.x.x/flow_v0.25.x-/cypress_v6.x.x.js
+++ b/definitions/npm/cypress_v6.x.x/flow_v0.25.x-/cypress_v6.x.x.js
@@ -446,15 +446,15 @@ declare interface Cypress$Chainable {
   /**
    * @see https://docs.cypress.io/api/commands/request.html
    */
-  request(url: string): Promise<Cypress$RequestResponse>,
-  request(url: string, body: Cypress$RequestBody): Promise<Cypress$RequestResponse>,
-  request(method: Cypress$HttpMethod, url: string): Promise<Cypress$RequestResponse>,
+  request(url: string): Cypress$Global,
+  request(url: string, body: Cypress$RequestBody): Cypress$Global,
+  request(method: Cypress$HttpMethod, url: string): Cypress$Global,
   request(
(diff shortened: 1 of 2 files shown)


Brianzchen:
This seems legit to me, it's more important to fix a broken type than the typing information. If you want to improve cypress more in the future be my guest, I don't use the tool enough to know how best to proceed with it. 

Though can I suggest we cleanup `Cypress$RequestResponse` since it's not used anymore and also update the tests to validate against the new expected behaviour?

debel27:
I'd be fine removing the `Cypress$RequestResponse` type, but this type is declared globally. By removing it, we might break projects that use it. But if it is not considered a serious concern for libdefs, I can remove it for sure.

debel27:
I've updated the tests to reflect the type changes.

Brianzchen:
True.. Should we change the thenable tests to be something like this?

```
cy.request('test').then(({ body }: Cypress$RequestResponse) => {
});
```

debel27:
Good idea. Done.

Brianzchen (approved):
LGTM

pascalduez (approved):

debel27:
It appears only the first commit has been merged into master. Would you mind merging the other two ?

pascalduez:
> It appears only the first commit has been merged into master. Would you mind merging the other two ?

Everything is merged, but squashed.

Issue #3565, opened 2019-09-23
Relax 'flow-bin' dependency check


mrtnzlml:
**Do you want to request a *feature* or report a *bug*?**

I want to request a _feature_.

**What is the current behavior?**

Flow-typed currently throws this error in our setup:

```
UNCAUGHT ERROR: Error: Failed to find a flow-bin dependency in package.json.
Please install flow-bin: `npm install --save-dev flow-bin`
```

**What is the expected behavior?**

It should not throw.

**If this is a feature request, what is motivation or use case for changing the behavior?**

We have custom `flow-bin` implementation which uses some additional features on top of the regular bin. So we distribute our own package (which still depends on `flow-bin` but eventually doesn' have to). As a consequence, we do not have `flow-bin` in package.json directly but seems like `flow-typed` requires it.

It would be nice to relax this requirement and allow more advanced project setups. Somehow related (but different) issue: https://github.com/flow-typed/flow-typed/issues/1282

julienw:
I got the same problem when I tried to configure a project with a multi-repo structure.

Brianzchen:
But how would flow-typed know what version of lib defs to provide you? Always install latest as a fallback maybe?

julienw:
Maybe with running `node_modules/.bin/flow --version` ? Or a form of `node_modules/.bin/flow --json` which outputs the flow version in a json format, but still checks everything.

mrtnzlml:
This should be workaroundable via `--flowVersion` option, see https://github.com/flow-typed/flow-typed/pull/4006 for more details.

Issue #1282, opened 2017-09-26
Improve Flow Version Detection


aikar:
```UNCAUGHT ERROR: Error: Failed to find a flow-bin dependency in package.json.
Please install flow-bin: `npm install --save-dev flow-bin```

I do not have flow-bin in my package.json as flow is part of a build-common dev-dependency of ours that we share over multiple modules.

flow-typed should instead do a require.resolve('flow-bin') to find the flow-bin folder and then load its package.json to detect the current version.

this would remove the need for all the complex flow version number checks in this area of the code, and would allow my workflow to work with flow-typed.

Currently ,the tool really only lets me use flow-typed inside of my common repo, and i'm unable to use it in any per-module repo.

jedwards1211:
I've recently been trying to do this for a lot of projects in my orgs, and I've noticed that unfortunately it seems like a common assumption across a lot of tools that things will be in `devDependencies`.  For example with `prettier` as a transitive dependency, VSCode format on save was failing to work... Hopefully people will become more aware of the need for build-common dev dependencies.

PR #4006, opened 2021-01-26
[cli] Check flow version in node_modules

Brianzchen:
Fixes https://github.com/flow-typed/flow-typed/issues/3565

When checking for flow version it will still check package.json but will fallback to check `node_modules/flow-bin/package.json` as a fallback.

- Links to documentation: N/A
- Link to GitHub or NPM: N/A

Other notes:

cc: @mrtnzlml @julienw Hope this satisfies your requirements?


cli/src/lib/npm/npmProjectUtils.js
```diff
@@ -74,6 +74,22 @@ export async function findPackageJsonPath(pathStr: string): Promise<string> {
   return path.join(pkgJsonPathStr, 'package.json');
 }
 
+async function getInstalledFlowPackageJsonData(pathStr: string): Promise<PkgJson | null> {
+  const flowBinPackagePath = 'node_modules/flow-bin/package.json';
```
julienw, line 78: I see there was a good suggestion in #1282 to find the package.json.

julienw:
> cc: @mrtnzlml @julienw Hope this satisfies your requirements?

TBH I don't have the project where I tried to use it, because I did it differently then, so I can't really try. But this looks reasonable!

cli/src/lib/npm/npmProjectUtils.js
julienw: Actually I believe that if it's a monorepo using `require.resolve` might be the only solution?

julienw:
ping also @aikar who might have some good feedback as the author of #1282.

cli/src/lib/npm/npmProjectUtils.js
Brianzchen: Let me try that, my only experience with mono-repos is yarn workspace which pulls node_modules to the top. Is this a case when node_modules would be embedded inside a nested dir?

aikar:
Not familiar with the projects code, but on a fundamental level looks related.

cli/src/lib/npm/npmProjectUtils.js
jedwards1211: I remember awhile back I had some issues with `require.resolve` and started using the `resolve` package instead in a lot of my codemod tools.  I think the issue may have been that `require.resolve` didn't work in code running inside of VSCode extensions, but `resolve` did?

Brianzchen: I don't think we can use `require()` or `require.resolve()` on a `flow-bin`. The way the code currently looks up package.json uses an absolute path with `cwd`. There is one area that does a require with a relative [path](https://github.com/flow-typed/flow-typed/blob/master/cli/src/lib/npm/npmProjectUtils.js#L125) and this is relative to where the script is run, I don't think it has the ability to import/require dependencies from another project, especially since it's generally installed globally. That's probably why this script generally leverages `fs.read`

If I changed this to require or require.resolve it would run inside the `flow-typed` project not your project right?

jedwards1211: It seems to work:

julienw: @Brianzchen I don't have much experience with mono-repos really, so please do what you think is best.

Brianzchen:
Thanks for the feedback guys. I think I'd consider this PR stalled for now. I won't close it unless we have to but in retrospect but I think I didn't have a clear understanding of what the requirements are and until @mrtnzlml provides some input this PR probably isn't helping anyone. Mono-repos can be done in a number of ways and it depends what the advanced setup comprises of to make the change useful.

mrtnzlml:
I think a nice solution would be to have an option to specify `--bin` manually. For example, Flow itself has a [CLI tool](https://github.com/facebook/flow/blob/096cbb737e01032d552ee61de820daae8e537603/tool) for migrating large codebases and it does it in the following way:

```text
./tool add-comments --all --comment "…" --bin /Users/XYZ/Work/node_modules/.bin/flow …
```

This way, you don't have to deal with the path resolving (which might be different in every repo as you suggested) but you rely on the actual binary path which everyone can change. What do you think?

Brianzchen:
@mrtnzlml Yes that shouldn't be very hard to do, but when I started digging into passing options I found that there's a `flowVersion` option you can just use if you know the explicit version you need. Would we still want the `bin` option then for flow-typed to do a lookup?

```
flow-typed install --flowVersion 0.100.0
```

I guess if this satisfies the issue the problem is mainly just that it's not documented anywhere so you wouldn't know unless you read source code.

mrtnzlml:
I think that would solve the issue as long as specifying `--flowVersion` disables the check which was throwing `Failed to find a flow-bin dependency in package.json.`.

Brianzchen:
Yes it does. It runs through a different code path to determine the flow version

Found out it's already documented here https://github.com/flow-typed/flow-typed/wiki/CLI-Commands-and-Flags#flags-1

mrtnzlml:
Cool, thanks! 🙂 It would be nice if these options were part of `flow-typed --help` output. Is it worth a new issue?

PR #4011, opened 2021-02-02
Fix write empty file

HsuTing:
Sometime, `flow-typed install` would write an empty file.

- Links to documentation: None
- Link to GitHub or NPM: None
- Type of contribution: Fix

Other notes:

Should I need to rename the function name?


AndrewSouthpaw:
Nice! Can you provide repro steps for when it used to create an empty file, so we can verify this PR fixes it?

HsuTing:
> Nice! Can you provide repro steps for when it used to create an empty file, so we can verify this PR fixes it?

I am not sure how to reproduce this bug. If I try to install some existing types, "jest" for example. Sometime this would create an empty file.

circle log: [link](https://circleci.com/api/v1.1/project/github/mikojs/core/4218/output/108/0?file=true&allocation-id=60079209d0647123549f87cb-0-build%2F7A0C24F1)

cli/package.json
```diff
@@ -39,7 +39,6 @@
     "rimraf": "^3.0.2",
     "semver": "^7.3.2",
     "table": "^5.4.6",
-    "through": "^2.3.8",
```
Brianzchen, line 42: Can you also remove `through_vx.x.x.js` if you're removing this dep?

HsuTing: Sure. I had removed `through_vx.x.x.js`.

Brianzchen:
@AndrewSouthpaw I'm not familiar with this part of the codebase but I can definitely vouch that this has happened to me. Sometimes when I run `flow-typed install` an existing lib def in my `flow-typed/npm` dir would be rewritten to an empty file, it's no biggie because I revert that before I commit into my project but it's just not a smooth process.

@HsuTing I think it would be worth explaining how you know this is the issue and then why your change fixes it. Given that this is a very random issue that happens time to time, we can't easily verify this and would simply need to trust that this works

HsuTing:
Well, I think this is an issue because my job in circleci sometimes fails. When using ssh to check my job, there is always an empty file in `flow-typed/npm`. Those failure jobs bothered me a lot. I'm not sure whether there is a bug in my code or just `flow-typed install` fails again.

To be honest, I am not sure if this pr fixes this issue directly. I could not find the exact code which cause this bug. However, I used patch-package with this [patch](https://github.com/mikojs/core/blob/main/patches/flow-typed%2B3.2.1.patch) and there won't be any empty file in circleci again

AndrewSouthpaw (changes requested):
Pending a question in thread :)

cli/package.json
Brianzchen: I recommended that @HsuTing remove the lib def because `through` is not used in the cli codebase anymore. So it makes sense to also remove `through` as a lib def in `cli/flow-typed/npm` This is unrelated to lib defs in `definitions`

HsuTing: I had remove `through_vx.x.x.js` again. Because updating branch, `through` appears again.

AndrewSouthpaw: Ah, I wasn't paying attention to the path name, thanks for the catch!

AndrewSouthpaw (approved):

PR #4012, opened 2021-02-05
fix imports for morgan lib def

atabel:
Newer versions of flow-bin don't allow using import outside declare module blocks.
This PR fixes imports for morgan lib type definitions.


AndrewSouthpaw (approved):
Thanks!

PR #4013, opened 2021-02-05
[mocha] Fix tests that break in 0.144.0

Brianzchen:
- Links to documentation: https://www.npmjs.com/package/mocha
- Link to GitHub or NPM: https://www.npmjs.com/package/mocha
- Type of contribution: fix

Other notes:
Also add mocha versions 7 and 8. Looking at the release logs they are just breaking changes that deprecate older node versions with cli fixes
https://github.com/mochajs/mocha/releases


AndrewSouthpaw (approved):
Wow, impressive work! Thanks for going through the trouble to add the new types. I skimmed through it and it generally looks fine, so calling it 🚢  for now, and as issues crop up we can tackle them. 😄

PR #4014, opened 2021-02-06
[cli] Update cli flow version to 0.144.0

Brianzchen:
I thought it would make sense that the tool that maintains our type defs  aren't running on a flow version that's just a year old at this point.

- Links to documentation: N/A
- Link to GitHub or NPM: N/A

Other notes:


cli/src/commands/create-stub.js
```diff
@@ -8,7 +8,7 @@ import {findFlowRoot} from '../lib/flowProjectUtils.js';
 import {getPackageJsonData, loadPnpResolver} from '../lib/npm/npmProjectUtils';
 import {path} from '../lib/node';
 
-export function setup(yargs: Object) {
+export function setup(yargs: Object): any {
```
pascalduez, line 11: If we could also kill those `Object`, would be nice.

cli/src/lib/__mocks__/node.js
```diff
@@ -13,7 +13,7 @@ export const fs = {
   // This is a custom function that our tests can use during setup to specify
   // what the files on the "mock" filesystem should look like when any of the
   // `fs` APIs are used.
-  mockFiles: {},
+  mockFiles: ({ ...null }: { [key: string]: any }),
```
pascalduez, line 16: Looks like this breaks the CI, also could you add a comment, that `{...null}` is not straightforward...

cli/src/lib/libDefs.js
```diff
@@ -5,7 +5,7 @@ import semver from 'semver';
 import {cloneInto, rebaseRepoMaster} from './git.js';
 import {mkdirp} from './fileUtils.js';
 import {fs, path, os} from './node.js';
-import {versionToString} from './semver.js';
+import {versionToString, type VersionRange, type Version} from './semver.js';
```
pascalduez, line 8: CI break, `VersionRange` unused.

cli/src/lib/__mocks__/node.js
Brianzchen: actually don't need `...null` anymore. Needed it before when it was untyped and was exported an as unsealed object

cli/src/commands/create-stub.js
Brianzchen: I'll look into it

GAntoine (approved):

Issue #3925, opened 2020-11-04
Multiple errors when running tests on react-redux


NicolasBellec:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
running tests on react-redux flow-types returns multiple errors

**If the current behavior is a bug, please provide the steps to reproduce.**

Just clone the repository and launch "npx flow-typed run-tests react-redux"

**What is the expected behavior?**
The tests should pass

**If this is a feature request, what is motivation or use case for changing the behavior?**

**Local Environment Information**
If this issue is specific to the CLI, provide the results of running `npx envinfo` from the command line in a GitHub gist.

Brianzchen:
It's because of suppression codes that were mandated somewhere in one of the versions in 2020. This causes every previous test to throw a bunch of errors if tests are run against them. As we maintain libraries we would also add changes to fix up all the suppression codes but there are so many test files and we haven't found a way to automate it yet so there are a lot of tests that would have this issue currently.

I suppose `react-redux` is a pretty important and useful lib. I'll see if I can put some effort to clean that one up.

PR #4015, opened 2021-02-07
[react-redux] Fix all errors

Brianzchen:
Closes https://github.com/flow-typed/flow-typed/issues/3925

- Links to documentation: https://www.npmjs.com/package/react-redux
- Link to GitHub or NPM: https://www.npmjs.com/package/react-redux
- Type of contribution: fix

Other notes:


Brianzchen:
Fixed all suppression codes but has few errors that will take time to look into. Opening earlier so that I can easily see a diff as I progress

Brianzchen:
So not sure if this is expected behaviour but I had to break up the tests into `0.142.x-` because 142 highlighted new soundness errors that didn't show up before which gave me a conflict of newer versions saying please add suppression codes and old versions telling me they're not needed. But as a result tests don't run against two separate flow version suites. The tests pass now and works for newer versions but older ones I don't know how to test?

AndrewSouthpaw:
> But as a result tests don't run against two separate flow version suites. 

Hmm I don't think I'm following, can you try explaining again?

Brianzchen:
@AndrewSouthpaw Previously react-redux defs were split into `104-` and because of recent changes that improved the type system, flow now highlights additional errors from `142` onwards. This causes issues with the tests because if we run the same test files against last 15 versions, some would require additional error suppressions whereas older versions would not. So it was not possible to both make all tests pass and keep all versions running on the same `104-` test files.

So I broke them up, to `104-141` and `142-`. But after I did this and made the `142-` tests pass. I realised that if you check the test logs it only runs tests that are from `142-` and onwards, and so far I'm have no idea how to make the lower versions tests run so I can't validate if those work. Though I haven't actually changed the actual libdef so I haven't necessarily broke anything. But it also means that we won't be able to easily maintain react-redux defs < flow 142
 
Screen Shot 2021-02-13 at 9 42 40 am

AndrewSouthpaw:
Ah! Thanks for explaining. I'm not too worried about it, as you said as long as you weren't making any changes to that version we're good. 

FWIW I believe you can use the `run-tests` without the `--onlyChanged` flag to test all versions of a libdef.

fcosta-godaddy:
Hi folks, first of all, thank you for this PR and thank you for all the work maintaining flow-types!

Before and after this PR, I've been having trouble annotating exported `connected` components with `types-first` and couldn't find help anywhere else. 
After running the flow types-first annotation codemod (see https://flow.org/en/docs/cli/annotate-exports/ ) on `connected` components flow annotates the export like the following:

Before the codemod:
```js
export default connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(MoveSectionOptions);
``

After the codemod:
```js
export default (connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(MoveSectionOptions): Class<
  ConnectedComponent<OwnProps, typeof MoveSectionOptions>
> &
  typeof MoveSectionOptions);
```

But `ConnectedComponent` is not really exported by the type definition module and that throws an error when I run flow.
Is there an alternative to properly using react-redux with types-first or should `ConnectedComponent` be exported?

I believe more people are having this same issue and are having trouble knowing what to do :/
Any help would be greatly appreciated, thank you!

Brianzchen:
@fcosta-godaddy I don't actually use flow + connect so my advice might be wrong. But how does this work for you? I'm exporting as an abstract component instead of whatever definition that react-redux provides because at it's core anything that isn't a simple react component is an abstract component. 

```
// @flow
import { connect } from 'react-redux';

type OwnProps = {
  b: boolean,
};

type Props = {
  ...OwnProps,
  a: number,
  ...
};

const Comp = (props: Props) => <div {...props} />;

const mapStateToProps = (state) => ({
  a: state.test.a,
});

export default (connect<Props, OwnProps, _, _, _, _>(mapStateToProps)(Comp): React.AbstractComponent<OwnProps>);

```

fcosta-godaddy:
Ah, this makes perfect sense, thank you so much!
I'll try that out!

Issue #3988, opened 2021-01-09
Latest version of flow-typed cli can't run "flow-typed install"


Brianzchen:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I'm currently running an older version of flow-typed and whenever I install a new dep I would run `flow-typed install` and it would update all my defs for me.

But for my colleagues who are just getting started with flow, who install the latest version of flow-typed aren't able to run this command, they can run `flow-typed install dep@version` though. Below is a screen grab of errors they get

<img width="1792" alt="Screen Shot 2021-01-09 at 1 25 24 pm" src="https://user-images.githubusercontent.com/12436524/104083929-bdee1180-5296-11eb-9fdc-ea113c70eb85.png">

**If the current behavior is a bug, please provide the steps to reproduce.**

```
yarn global add flow-typed
# go into a project
flow-typed install
```

**What is the expected behavior?**
It should scan package.json and install new/update lib defs

**Local Environment Information**
```
System:
  OS: macOS 11.1
  CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  Memory: 378.20 MB / 16.00 GB
  Shell: 5.8 - /bin/zsh
 Binaries:
  Node: 14.15.1 - /usr/local/bin/node
  Yarn: 1.22.10 - /usr/local/bin/yarn
  npm: 6.14.10 - /usr/local/bin/npm
 Managers:
  Homebrew: 2.7.0 - /usr/local/bin/brew
  pip3: 19.2.3 - /usr/bin/pip3
  RubyGems: 3.0.3 - /usr/bin/gem
 Utilities:
  Make: 3.81 - /usr/bin/make
  GCC: 4.2.1 - /usr/bin/gcc
  Git: 2.23.0 - /usr/local/bin/git
  Clang: 1200.0.32.28 - /usr/bin/clang
 Servers:
  Apache: 2.4.46 - /usr/sbin/apachectl
 SDKs:
  iOS SDK:
   Platforms: iOS 14.3, DriverKit 20.2, macOS 11.1, tvOS 14.3, watchOS 7.2
 IDEs:
  Nano: 2.0.6 - /usr/bin/nano
  VSCode: 1.52.1 - /usr/local/bin/code
  Vim: 8.2 - /usr/bin/vim
  Xcode: 12.3/12C33 - /usr/bin/xcodebuild
 Languages:
[...]

PR #4018, opened 2021-02-10
[cli] Lock semver to fix install command

Brianzchen:
Fixes https://github.com/flow-typed/flow-typed/issues/3988

After deep digging the issue is actually related to the semver package. Versions `7.3.3` and `7.3.4` which are currently latest both break our tests in the following screenshot. We use loose `^` but have a lock file which only tests against current lock which is version `7.3.2`. I've now locked the version but I think we should consider removing the lock file completely as this is a published package and we cannot guarantee the version that will be installed on a users machine matches what's in our yarn.lock (would appreciate feedback on this matter). 

Screen Shot 2021-02-10 at 5 59 42 pm

- Links to documentation: N/A
- Link to GitHub or NPM: N/A
- Type of contribution: fix

Other notes: @GAntoine please make sure this is merged before you do release, it's one of the biggest blockers for new people installing flow-typed


AndrewSouthpaw (approved):
Seems reasonable to me. I don't have input on the "should we remove the package lock file", but I haven't done a lot of research into patterns of published packages. I would *imagine* it makes sense to keep the lock file because it helps ensure this package works correctly when installed on someone's machine, but maybe I'm not understanding it correctly.

@Brianzchen if you want you could open a separate issue to drive the discussion. I'll call it out of scope for this PR and merge in your improvement.

AndrewSouthpaw:
cc @GAntoine to do a release after this builds.

josemigallas:
Is there an estimate on when this will be released?

Brianzchen:
@josemigallas Just been released, check `v3.3.0`

PR #4020, opened 2021-02-13
[next_v9.x.x] Add useRouter definition

cybemos:
`useRouter` is missing from next/router definition.

https://github.com/vercel/next.js/blob/canary/docs/api-reference/next/router.md#userouter

### DIFF
--- a/definitions/npm/next_v2.x.x/flow_v0.104.x-/test_next_v2.x.x.js
+++ b/definitions/npm/next_v2.x.x/flow_v0.104.x-/test_next_v2.x.x.js
@@ -16,13 +16,13 @@ const { createServer } = require('http')
 const { parse } = require('url')
 
 // server
-// $FlowExpectedError
+// $FlowExpectedError[incompatible-call]
 next({ dev: 1 });
-// $FlowExpectedError
+// $FlowExpectedError[incompatible-call]
 next({ dir: false });
-// $FlowExpectedError
+// $FlowExpectedError[incompatible-call]
 next({ quiet: 'derp' });
-// $FlowExpectedError
+// $FlowExpectedError[incompatible-call]
 next({ staticMarkup: 42 });
 
 const app = next({ dev: true, dir: ".", quiet: false});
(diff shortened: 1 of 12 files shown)


definitions/npm/next_v9.x.x/flow_v0.104.x-/test_next_v9.x.x.js
```diff
@@ -127,12 +127,12 @@ export default class TestDocument extends Document {
 }
 
 export class TestApp extends App {
-  static async getInitialProps({ Component, router, ctx }: AppInitialProps) {
+  static async getInitialProps({ Component, router, ctx }: AppInitialProps): Promise<*> {
```
Brianzchen, line 130: Should we change these all to use `any` instead of `*` since it's deprecated? ref: https://flow.org/en/docs/types/utilities/#toc-existential-type
```suggestion
  static async getInitialProps({ Component, router, ctx }: AppInitialProps): Promise<any> {
```

definitions/npm/next_v2.x.x/flow_v0.104.x-/test_next_v2.x.x.js
```diff
@@ -88,12 +88,12 @@ const p: string = Router.pathname;
...
 
 export default class TestDocument extends Document {
-  static async getInitialProps (ctx: Context) {
+  static async getInitialProps (ctx: Context): Promise<any> {
     const props = await Document.getInitialProps(ctx)
     return { ...props, customValue: 'hi there!' }
   }
 
-  render () {
+  render(): React$Node {
```
pascalduez, line 96: Is that really required?   
Also should rather be `React$Element<'html'>`.

cybemos: According to [flow documentation for react](https://flow.org/en/docs/react/types/#toc-react-node), it seems that `render()` should return a Node.

I'm not sure if we should set the node as optional.

cybemos: And without specifying it, CI tests won't pass

pascalduez: I think those docs samples are a bit too generic, but anyway, it doesn't make a big difference here I guess.

pascalduez (approved):

Issue #4022, opened 2021-02-17
CLI regression


lhorie:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

CLI throws an error `runCLI is not a function`

```
TypeError: runCLI is not a function
    at Object.<anonymous> (/Users/lhorie/Documents/test/regression/node_modules/flow-typed/dist/cli.js:73:3)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)
    at startup (internal/bootstrap/node.js:283:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3)
```

**If the current behavior is a bug, please provide the steps to reproduce.**

```bash
#!/bin/bash
mkdir regression && cd regression
echo '{"scripts": {"flow-typed": "flow-typed"}}' > package.json
npm install flow-typed
npm run flow-typed
```

It looks like https://github.com/flow-typed/flow-typed/commit/bdba696f6d045aebf9b65a1a8362c193e3f82782#diff-876a81aa99b34ac61a7059067328e4bbe5ac6ed01064d3bf6fdfa4d7f0222832R20 removed `export` from runCLI, which causes the logic at https://github.com/flow-typed/flow-typed/blob/master/cli/src/cli.js#L91 to fail 

**What is the expected behavior?**

Running CLI should not throw error. The simplest solution would be to add the `export` back.

**If this is a feature request, what is motivation or use case for changing the behavior?**

N/A

**Local Environment Information**
If this issue is specific to the CLI, provide the results of running `npx envinfo` from the command line in a GitHub gist.

```
> npx envinfo
Environment:
  OS:  macOS 10.15.7
  Node:  10.21.0
  Yarn:  1.22.10
  npm:  6.14.4
  Watchman:  Not Found
  Xcode:  Xcode 11.0 Build version 11A420a
  Android Studio:  Not Found
```

Brianzchen:
:O my bad, didn't realise this was used since no tests broke

GAntoine:
No worries, just published 3.3.1 to fix this.

PR #4023, opened 2021-02-17
Unbreak CLI

lhorie:
Fixes https://github.com/flow-typed/flow-typed/issues/4022


GAntoine (approved):
Thanks for the quick fix!

PR #4029, opened 2021-02-23
[styled-components] Fix tests

Brianzchen:
- Links to documentation: https://styled-components.com/
- Link to GitHub or NPM: https://github.com/styled-components/styled-components
- Type of contribution: fix

Other notes:


Brianzchen:
So far only v2/3 are failing but they look completely broken to me

Brianzchen:
@GAntoine @pascalduez @AndrewSouthpaw I haven't really touched this PR for a while but the comment above is still true, v2/3 complicated to fix and so the tests break. But this should probably get reviewed so that issues like https://github.com/flow-typed/flow-typed/issues/4068 and many other issues related to `styled-components` can continue. Would like your thoughts on the matter

AndrewSouthpaw:
Hey @Brianzchen ! Sorry for the radio silence, this month was a pretty full one for me. This looks good.

AndrewSouthpaw (approved):

Brianzchen:
I doubt anyone would maintain v2/3 anymore though we shouldn't delete old lib defs. I'm wondering if you have any idea of how we could proceed? Maybe we delete breaking tests from these versions so that don't break tests when maintaining styled-components?

PR #4030, opened 2021-02-23
[redux] Fix suppression codes

Brianzchen:
- Links to documentation: https://www.npmjs.com/package/redux
- Link to GitHub or NPM: https://www.npmjs.com/package/redux
- Type of contribution: fix

Other notes:
In both v3 and v4, https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/redux_v4.x.x/flow_v0.104.x-/test_createStore.js#L15 this test has different error codes for <134 and >=134. Older versions need `[incompatible-return]` whereas newer versions need `[incompatible-call]`. So I've had to break them into separate flow versions


AndrewSouthpaw (approved):
LGTM, thanks!

PR #3935, opened 2020-11-14
[lodash_v4.x.x] improve castArray

nifgraup:

Issue #3919, opened 2020-10-26
Update all tests with suppression codes


Brianzchen:
Can we use the internal flow too to update all suppression codes in one go so that there's less friction with new flow changes?

```
sainati Today at 10:38 AM
yes
you can use the tool script that comes with the repo
so ./tool add-comments <SUPPRESSION TEXT> will add suppression comments with the correct codes and that comment
and ./tool remove-comments will remove unused suppressions
this is what we do to upgrade our codebases internally
```

pascalduez:
Cool!

This is exactly what I wanted to ask them :) 

I hope we will be able to make it work, flow-typed tests are not a standard codebase, so dunno how the tool/flow will be able to make sense of them.

Brianzchen:
I've run a test on this, you basically need to have both the `flow` and `flow-typed` repo's cloned. Then you'll need to install `flow-bin` in the definitions folder (which can be removed later prior to merging) and have flow on the same version.

Then in the flow repo, run, `./tool add-comments ../flow-typed/definitions` where I encountered a few issues,
- Because the imports are a bit random, it threw lots of module not found errors
- Currently flow can't handle adding suppression codes to existing comments, but the flow team will see if that ability can be open sourced

jedwards1211:
I'm trying to write test cases and when I add suppression codes it just warns about unused suppression, is it just be flow being buggy as usual?

jedwards1211:
Even though the Flow docs explicitly state that you can suppress multiple error codes like this:

```js
  // $FlowFixMe[incompatible-call]
  // $FlowFixMe[prop-missing]
```
I get an unused suppression warning on the first comment.
Getting really tired of dealing with bugs like this.

jedwards1211:
Oh, I guess some of those buggy warnings were happening in 0.125-0.126 and then fixed in later versions

pascalduez:
> Oh, I guess some of those buggy warnings were happening in 0.125-0.126 and then fixed in later versions

Yes, I experienced the same thing. Ended up scoping some libdef to 0.127.x-

Brianzchen:
Closing, don't think there's a clear way to automate this

PR #4032, opened 2021-02-28
Annotate return types for various lodash array functions with Maybe

stackptr:
- Links to documentation: https://lodash.com/docs/4.17.15 (inadequate, see below)
- Link to GitHub or NPM: https://github.com/lodash/lodash
- Type of contribution: fix

Functions in `lodash` that operate on `Array<T>` to return `T` should actually use a `?T` for the return type to account for `undefined`, which is given with an empty array as input.

Grepping in the latest source[1] for `: undefined` and narrowing that to `return` sites for exposed functions (tagged as `@memberOf _`) shows the following functions:

- `head`
- `last`
- `nth`
- `max`
- `maxBy`
- `min`
- `minBy`

In addition to the above, `first` is an alias of `head`.

Open questions:

- These functions may not be comprehensive. Unfortunately, the documentation only makes clear that some of these return `undefined`, while omitting this fact for others.
- It's likely that earlier versions of `flow` and previous versions of `lodash` should be updated as well, but as this is the interface I am most exposed to I have started here in case I would need further context to do that.

1: https://github.com/lodash/lodash/blob/4.17.15/lodash.js

### DIFF
--- a/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
+++ b/definitions/npm/lodash_v4.x.x/flow_v0.104.x-/lodash_v4.x.x.js
@@ -275,12 +275,12 @@ declare module "lodash" {
       fromIndex?: ?number
     ): -1;
     // alias of _.head
-    first<T>(array: ?$ReadOnlyArray<T>): T;
+    first<T>(array: ?$ReadOnlyArray<T>): ?T;
     flatten<T, X>(array?: ?$ReadOnlyArray<$ReadOnlyArray<T> | X>): Array<T | X>;
     flattenDeep<T>(array?: ?$ReadOnlyArray<any>): Array<T>;
     flattenDepth(array?: ?$ReadOnlyArray<any>, depth?: ?number): Array<any>;
     fromPairs<A, B>(pairs?: ?$ReadOnlyArray<[A, B]>): {| [key: A]: B |};
-    head<T>(array: ?$ReadOnlyArray<T>): T;
+    head<T>(array: ?$ReadOnlyArray<T>): ?T;
     indexOf<T>(array: $ReadOnlyArray<T>, value: T, fromIndex?: number): number;
     indexOf<T>(array: void | null, value?: ?T, fromIndex?: ?number): -1;
     initial<T>(array: ?$ReadOnlyArray<T>): Array<T>;
@@ -333,10 +333,10 @@ declare module "lodash" {
     ): Array<T>;
     join<T>(array: $ReadOnlyArray<T>, separator?: ?string): string;
     join<T>(array: void | null, separator?: ?string): "";
-    last<T>(array: ?$ReadOnlyArray<T>): T;
+    last<T>(array: ?$ReadOnlyArray<T>): ?T;
     lastIndexOf<T>(array: $ReadOnlyArray<T>, value?: ?T, fromIndex?: ?number): number;
     lastIndexOf<T>(array: void | null, value?: ?T, fromIndex?: ?number): -1;
-    nth<T>(array: $ReadOnlyArray<T>, n?: ?number): T;
+    nth<T>(array: $ReadOnlyArray<T>, n?: ?number): ?T;
     nth(array: void | null, n?: ?number): void;
     pull<T>(array: Array<T>, ...values?: $ReadOnlyArray<?T>): Array<T>;
     pull<T: void | null>(array: T, ...values?: $ReadOnlyArray<?any>): T;
@@ -949,12 +949,12 @@ declare module "lodash" {
     ceil(number: number, precision?: number): number;
     divide(dividend: number, divisor: number): number;
     floor(number: number, precision?: number): number;
-    max<T>(array: ?$ReadOnlyArray<T>): T;
-    maxBy<T>(array: ?$ReadOnlyArray<T>, iteratee?: Iteratee<T>): T;
+    max<T>(array: ?$ReadOnlyArray<T>): ?T;
+    maxBy<T>(array: ?$ReadOnlyArray<T>, iteratee?: Iteratee<T>): ?T;
     mean(array: $ReadOnlyArray<*>): number;
     meanBy<T>(array: $ReadOnlyArray<T>, iteratee?: Iteratee<T>): number;
-    min<T>(array: ?$ReadOnlyArray<T>): T;
-    minBy<T>(array: ?$ReadOnlyArray<T>, iteratee?: Iteratee<T>): T;
+    min<T>(array: ?$ReadOnlyArray<T>): ?T;
+    minBy<T>(array: ?$ReadOnlyArray<T>, iteratee?: Iteratee<T>): ?T;
     multiply(multiplier: number, multiplicand: number): number;
     round(number: number, precision?: number): number;
     subtract(minuend: number, subtrahend: number): number;
@@ -1779,14 +1779,14 @@ declare module "lodash/fp" {
       array: $ReadOnlyArray<T>
     ): number;
     // alias of _.head
-    first<T>(array: $ReadOnlyArray<T>): T;
+    first<T>(array: $ReadOnlyArray<T>): ?T;
     flatten<T, X>(array: $ReadOnlyArray<$ReadOnlyArray<T> | X>): Array<T | X>;
     unnest<T, X>(array: $ReadOnlyArray<$ReadOnlyArray<T> | X>): Array<T | X>;
     flattenDeep<T>(array: $ReadOnlyArray<any>): Array<T>;
     flattenDepth(depth: number): (array: $ReadOnlyArray<any>) => Array<any>;
     flattenDepth(depth: number, array: $ReadOnlyArray<any>): Array<any>;
     fromPairs<A, B>(pairs: $ReadOnlyArray<[A, B]>): {| [key: A]: B |};
-    head<T>(array: $ReadOnlyArray<T>): T;
+    head<T>(array: $ReadOnlyArray<T>): ?T;
     indexOf<T>(value: T): (array: $ReadOnlyArray<T>) => number;
     indexOf<T>(value: T, array: $ReadOnlyArray<T>): number;
     indexOfFrom<T>(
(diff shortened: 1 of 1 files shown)


stackptr:
I see that CI is failing for this PR, but when running the tests locally from a clean `master` I see what seem to be many existing errors:

```
> cd cli && node dist/cli.js run-tests lodash_v4

<snip>

    
    ... 54 more warnings (only 50 out of 7 errors and 97 warnings displayed)
    To see all errors, re-run Flow with --show-all-errors
    
```

stackptr:
The last PR working in this area, #3935, also notes these extant issues, mentions #3919 but has not moved much in a few months from what I can see.

I think a reasonable course here is to note the quantity of extant issues without my changes (which I've done in my previous comment), and then validate that this quantity is unchanged after my changes. Unfortunately this presents a fairly unergonomic workflow in attempting to add what I see as necessary tests for these `fn([]) => undefined` cases. If that path is proves difficult I will open a separate PR to address extant issues following the approach in #3919.

khyveasna11111908 (approved):
```

```

AndrewSouthpaw:
@stackptr Thanks for the contribution!

The changes you're proposing are likely breaking for many users who rely on this library not being "exactly right" with type access from the array.

There's been many discussions about array access in Flow and whether they should return `Maybe` types.

- https://github.com/facebook/flow/pull/6823
- https://github.com/facebook/flow/issues/5972
- https://github.com/facebook/flow/issues/3702

...just to name a few.

You're right, the types are technically unsound, and also even the flow team has not reached an accord about [how they want to treat basic array access](https://flow.org/try/#0MYewdgzgLgBARgQwE4C4bSQSzAcwNoC6MAvDIQFCiSwBmIIaG2OJ8yAdAA4icAUAlOSA).

That said, these are **community defined** lib defs. If consumers of the lodash types really want these functions to have `Maybe` types and to type-check everything that comes out of them, you all can certain choose that path.

(Personally, I wouldn't want to maybe-check all my uses of such functions, but also I don't use lodash so it's not my concern.)

But, I'd want to see some more lively discussion and assent from other folks involved with the lodash types. Perhaps could you start by tagging other people who have contributed to this lib def in the past and see if we can get their input?

stackptr:
Thanks for all the context @AndrewSouthpaw. Unfortunately I just learned about this unsafe access this morning, after working with `flow` for the past couple years. In my opinion, losing type safety for convenience is beyond problematic and should be something that at the very least can be opted-out of, but that is beyond the scope of this PR.

> But, I'd want to see some more lively discussion and assent from other folks involved with the lodash types. Perhaps could you start by tagging other people who have contributed to this lib def in the past and see if we can get their input?

I definitely agree about getting buy-in from the community here and was actually going to raise that in a different issue. As a consumer of `flow-typed` definitions it is unclear to me how changes are determined in this project as valid. Tagging people who have contributed is a good start!

These are people who have authored PRs changing `definitions/npm/lodash_v4.x.x` in the past two years: @anu365, @beaucollins, @bobrippling, @cakoose, @DmitryGonchar, @edoardo-bluframe, @fbontin, @Jasyer, @jbrown215, @JoelMagee, @julienw, @klaesra, @kubijo, @lukeapage , @lyleunderwood, @maxmalov, @mwiencek, @nifgraup, @nvie, @oiuhg, @pascalduez, @SBoudrias

(I don't know if there is a better way of finding people to tag, so apologies if I've missed someone or have tagged someone who doesn't want anything to do with this project anymore!)

**Here is my case for making this change**:

- Internally, `lodash` is returning `undefined` explicitly, rather than relying on the default unsafe behavior of accessing an element of an array. For example, `head` first checks for a non-zero `length` (it relies on treating `0` as false) before accessing the element. Because of this, I don't think there is a case to be made for consistency with the current paradigm of unsafe array access:

```js
// https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L7423-L7425
function head(array) {
  return (array && array.length) ? array[0] : undefined;
}
```

- I wouldn't expect this to be catastrophic for current consumers of this interface because as far as I can tell the recommendation for working with third-party libdefs is to check them in to the project itself, in order to have control over when updates are applied. The [blog post explaining how to work with this repo](https://flow.org/blog/2016/10/13/Flow-Typed/) goes so far as to bold this recommendation and calls out that types may become more precise through improvement:

> Once libdefs have been installed, **we recommend that you check them in to your project’s repo**. Libdefs in the flow-typed repository may be improved over time (fixing a bug, more precise types, etc). If this happens for a libdef that you depend on, you’ll want to have control over when that update is applied to your project. Periodically you can run `flow-typed update` to download any libdef updates, verify that your project still typechecks, and the commit the updates.

To expand on the above mentality, for my own project that consumes these interfaces, it is assumed that any time an interface is upgraded it will potentially produce type errors that need to be addressed, but that the effort to do so will lead to code less prone to runtime errors.

I'm happy to hear countervailing opinions on all of the above!

cakoose:
With something like `first(...)`, there are two sources of a null/undefined response:
1. The array being empty.  This is similar to an out-of-bounds access and would be inclined to stick with Flow's precedent of not capturing that in the types.
2. The array argument being null/undefined.

Maybe case 2 can be improved with a more precise overload:

```ts
/* @flow */

declare function first<T>(array: $ReadOnlyArray<T>): T;
declare function first<T>(array: ?$ReadOnlyArray<T>): ?T;

function test1(x: ?Array<string>): ?string {
  return first(x);
}

function test2(x: Array<string>): string {
  return first(x);
}

function test3(x: ?Array<string>): string {
  return first(x); // error flagged (good!)
}
```

[try.flow.org link](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVATApgYxgQwCcswoBXAOxwBcBLOC02wgZ2oB4AVAPgAojC+AJ4AuMABIASlnwYA8hRhCAgoUFCu3AJRjOAbky4CxUpRr1GUZm0381wsQH4pM+YpX2NPHWEf705FR0DGDUWGwAjLwAHmKq6uxshLQUAObcYAA+YBRkMDA+jkkpqWAA3qhgYMTUZISW1tQxWgYAvgFmwYxhbABMMXGeidTJadpixWnlldVYtfVMrE3RLajtqIHmIT3UAMwDYPHCw6PpWTl5BRMjJdNVNXUNS816YMDAYFhqcISkBKmpLAYMC8VJwOAYACEWjWQA)

(I'm not sure sure if Flow allows overloads everywhere.  If not, maybe an intersection would work?)

This doesn't distinguish between null and undefined, which may be a problem.  Could maybe add more overloads...

maxmalov:
Hey, @stackptr, type safety is cool for sure, but the Flow itself won't warn you about index-out-of-bound cases. Consider the [following example](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzCjjgC4xsBXAWwCMBTAJwG0BdMAXjGYEYAaMACYBAZlYBudMTi8ADPNYA6fHADK+RgEtsAcwAUASnFA). Same also true for `{| [string]: T |}` types. So my 5 cents on this topic is to be consistent with Flow behavior and avoid `?T` in return values.

lyleunderwood:
I'll preface my comment by stating that I personally lean more toward the opinion that flow _should_ error on out-of-bounds, but that is not really what we're talking about apparently.

In the abstract, if you have a function that is basically just a collection access utility, it makes the most sense to try to be consistent with flow's standard array access behavior. Developers must be aware of this behavior in order to successfully write flow code, and having some utility functions mixed in that do use maybe types could setup a false expectation. They're gonna run into this reality eventually, the sooner the better as far as I'm concerned.

stackptr:
Thanks for the feedback so far!

> Flow itself won't warn you about index-out-of-bound cases

It might be pedantic, but what I am talking about is the general case of array access, of which index-out-of-bound is a special case. I think that special case is very hard if not impossible to implement given various constraints of JavaScript, specifically mutability of arrays and impurity in functions.

> Developers must be aware of this behavior in order to successfully write flow code

Very good point. As I mentioned in an earlier comment, I've been professionally developing a project using `flow` for the last few years without absorbing this behavior, mostly as a consequence of the project having an established pattern to use a utilities library for array access (the previous library typed these as `?T`, then we switched to `lodash` and here I am).

On the one hand, I agree that there are conceptual decisions that make sense to internalize when working with `flow` (or any project) and that changes and proposals for adjacent projects should fit within those decisions. However, as best I can tell, the rationale for unsafe array access seems to have historically been to make it more convenient for codebases to adopt `flow` without having to fix myriad issues. As flow becomes adopted by more codebases, the trade-off between convenience and permitting unsafe behaviors can be flipped— to paraphrase the overview for [Flow strict](https://flow.org/en/docs/strict/), which I think makes a good point in identifying how project goals can shift.

I have opened an issue (https://github.com/facebook/flow/issues/8603) to build further context on the decision to have unsafe Array access. If contributors in `flow` have a preference in not addressing this, and there is a preference here for following the preferences of that project, then that is definitely sufficient to answer whether this PR represents a step forward. More likely, I suspect there is a split of preference (which would make sense if there has been multiple discussions of this in the past). If the issue is not conceptual but practical — that is to say, if safe array access makes sense as a goal but is too hard to do for the main type-checker project — I think _this_ project should not be guided by practical concessions that may not exist here.

lyleunderwood:
> as best I can tell, the rationale for unsafe array access seems to have historically been to make it more convenient for codebases to adopt `flow` without having to fix myriad issues. 

This line of thinking was the unfortunate genesis of many of flow's more problematic features, like unsealed objects.

cakoose:
> > as best I can tell, the rationale for unsafe array access seems to have historically been to make it more convenient for codebases to adopt `flow` without having to fix myriad issues.
> 
> This line of thinking was the unfortunate genesis of many of flow's more problematic features, like unsealed objects.

I don't think it's fair to Flow to put array bounds checking in that bucket.  It's a much harder problem that nobody has a great solution for.

There _are_ solutions that people use in more limited domains, ones where correctness is much more important.  In those cases people are willing to spend a ton of extra effort to prove things like "this program will never allocate more memory than available" or "this program will never infinite loop".

But there isn't an even _mildly_ popular general-purpose language that does it.  Most languages avoid unsoundness by raising a runtime error.  Ideally, JS engines would agree on some switch to enable runtime errors for out-of-bounds array accesses, but I'm not holding my breath.

Flow could have inserted a runtime bounds check before all array accesses, but that would have been in strong conflict with the original philosophy of never changing the runtime semantics.  But maybe that's an option now?

(Not sure what the performance hit would be.  Java VMs are sort of good at eliminating redundant bounds checks; not sure how JS engines do on that front.)

lyleunderwood:
Oh for sure, making array access unsafe was a pretty reasonable choice, it's just also one with flaws.

cakoose:
Yeah, I'm just saying that the lack of static bounds checking doesn't fit in the "make it more convenient for codebases to adopt `flow`" category.  It's more in the "nobody knows how to do this" category.

PR #4039, opened 2021-03-03
[jest-useragent-mock] Add def

Brianzchen:
- Links to documentation: https://github.com/ariesjia/jest-useragent-mock/blob/master/index.d.ts
- Link to GitHub or NPM: https://www.npmjs.com/package/jest-useragent-mock

Other notes: I also added an extra space to `- Links to documentation: ` line because there's always a space after `- Link to GitHub or NPM: ` in the PR template and it's always kind of annoyed me, hope this is fine.

### DIFF
--- a/definitions/npm/jest-useragent-mock_v0.1.x/flow_v0.83.x-/jest-useragent-mock_v0.1.x.js
+++ b/definitions/npm/jest-useragent-mock_v0.1.x/flow_v0.83.x-/jest-useragent-mock_v0.1.x.js
@@ -1,5 +1,5 @@
 declare module 'jest-useragent-mock' {
-  export declare function mockUserAgent(userAgent?: string): void;
+  declare export function mockUserAgent(userAgent?: string): void;
 
-  export declare function clear(): void;
+  declare export function clear(): void;
 }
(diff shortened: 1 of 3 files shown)


definitions/npm/jest-useragent-mock_v0.1.x/flow_v0.83.x-/jest-useragent-mock_v0.1.x.js
```diff
@@ -0,0 +1,5 @@
+declare module 'jest-useragent-mock' {
+  declare export function mockUserAgent(userAgent?: string): void;
```
pascalduez, line 2: It's `string | null` in the TS definition, any reason to diverge?

Brianzchen: Because when I read into the [source code](https://github.com/ariesjia/jest-useragent-mock/blob/master/src/index.js#L15). Allowing null has no real value. User agent should always be a string, allowing empty value is just a means of a fallback but extending it to support `null` feels wrong. In fact you can probably pass in number or really anything and it still works but ideally it should only be a string.

implicitly leaving out the value seems ok, explicitly passing in an empty value such as `null` isn't useful and ideally shouldn't be allowed. Although `fakeUserAgent` is initialised as `null` it does still default the return value as `''`.

pascalduez: https://github.com/ariesjia/jest-useragent-mock/issues/5

I understand your reasoning, but diverging from the TS typings (if any) makes maintaining the flow typed one much tedious and confusing.

PR #4045, opened 2021-03-19
Add react-native-contacts type definitions for v5.x.x

katieisnell:
- Links to documentation: 
- Link to GitHub or NPM: 

Other notes:
I struggled to write tests for the library `react-native-contacts`, so please can someone help me to understand how to do this?

### DIFF
--- a/definitions/npm/react-native-contacts_v5.x.x/flow_v0.83.x-/react-native-contacts_v5.x.x.js
+++ b/definitions/npm/react-native-contacts_v5.x.x/flow_v0.83.x-/react-native-contacts_v5.x.x.js
@@ -1,0 +1,132 @@
+declare module 'react-native-contacts' {
+  declare interface EmailAddress {
+    label: string;
+    email: string;
+  }
+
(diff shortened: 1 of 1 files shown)


Brianzchen:
@katieisnell for testing, follow some of the examples in the project. Maybe this one for app-root-path that I did recently. 
https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/app-root-path_v3.x.x/flow_v0.142.x-/test_app-root-path_v3.x.x.js

Group your tests in describe/it blocks and then write tests that validate your library definitions so that if someone were to accidentally change it to something unexpected it would fail. Remember you are testing just the definitions not the library itself so test if you can use a function in a particular way you expect.

Also don't forget to add some cases you expect to fail and then use suppression codes to validate they fail as expected, this will be helpful if you want to check that function A definitely returns void or something.

Once you've written some tests you can run it with `./quick_run_def_tests.sh` if you are on linux/macos or `quick_run_def_tests.cmd` if you are on windows.

PR #4046, opened 2021-03-22
react-query_v3.x.x updates

ianduvall:
- Links to documentation: react-query.tanstack.com
- Link to GitHub or NPM: https://github.com/tannerlinsley/react-query
- Type of contribution: fix

Other notes:
This fix a few issues.
- Infer data type from query function instead of reverse. See https://github.com/tannerlinsley/react-query/pull/1470
- Default generics to `mixed` like TypeScript does with `unknown`
- Use `$ReadOnlyArray` for query keys. See https://github.com/tannerlinsley/react-query/pull/1854
- Remove impossible null states (`void | T` instead of `?T`)
- Attempt to fix infinite query types
- Export all types


benthemonkey (approved):
looks great, thanks!

PR #4048, opened 2021-03-22
[redux-thunk] Add definition

Brianzchen:
- Links to documentation: https://www.npmjs.com/package/redux-thunk#injecting-a-custom-argument
- Link to GitHub or NPM: https://www.npmjs.com/package/redux-thunk

Other notes: At long last it exists

### DIFF
--- a/definitions/npm/redux-thunk_v2.x.x/flow_v0.83.x-/redux-thunk_v2.x.x.js
+++ b/definitions/npm/redux-thunk_v2.x.x/flow_v0.83.x-/redux-thunk_v2.x.x.js
@@ -1,23 +1,25 @@
 declare module 'redux-thunk' {
-  // Types from redux, should use imports in the near future once we figure out
-  // cross dependency imports in flow-typed
+  // Types from redux with generics removed to be compatible
+  // with redux & redux-mock-store,
+  // should use imports in the not too distant future once we
+  // figure out cross dependency imports in flow-typed
   declare export type DispatchAPI<A> = (action: A) => A;
 
-  declare export type Dispatch<A: { type: any, ... }> = DispatchAPI<A>;
+  declare export type Dispatch = DispatchAPI<{ type: any, ... }>;
 
-  declare export type MiddlewareAPI<S, A, D = Dispatch<A>> = {
-    dispatch: D,
-    getState(): S,
+  declare export type MiddlewareAPI = {
+    dispatch: Dispatch,
+    getState(): any,
     ...
   };
 
-  declare export type Middleware<S, A, D = Dispatch<A>> = (
-    api: MiddlewareAPI<S, A, D>
-  ) => (next: D) => D;
+  declare export type Middleware = (
+    api: MiddlewareAPI
+  ) => (next: any) => any;
   // --- End of types from redux
 
-  declare export type Thunk = Middleware<S, A> & {|
-    withExtraArgument(arg: any): Middleware<S, A>,
+  declare export type Thunk = Middleware & {|
+    withExtraArgument(arg: any): Middleware,
   |};
 
   declare module.exports: Thunk;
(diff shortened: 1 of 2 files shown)


definitions/npm/redux-thunk_v2.x.x/flow_v0.83.x-/redux-thunk_v2.x.x.js
```diff
@@ -0,0 +1,26 @@
+declare module 'redux-thunk' {
+  // Types from redux with generics removed to be compatible
+  // with redux & redux-mock-store,
+  // should use imports in the not too distant future once we
```
pascalduez, line 4: We should favor the `@@` "solution" for cross libs, not the fanciest but at least we have a common thing to rely on.

So here 
```
declare module '@@redux' { ... }

declare module 'redux-thunk' {
  import type { Dispatch } from '@@redux'
}
```

Brianzchen: That works I'll make the change. I only wonder is there docs around this to help others in future and also, given that redux is a core lib for many, would multiple `@@redux` eventually cause conflict? For example I use both redux-saga and redux-thunk

pascalduez: Your right, I haven't tested but it's probably an issue if we have several diverging `@@redux` spread in libdefs... :face_with_head_bandage:

definitions/npm/redux-thunk_v2.x.x/flow_v0.83.x-/redux-thunk_v2.x.x.js
```diff
@@ -0,0 +1,29 @@
+// Types from redux with generics removed to be compatible
+// with redux & redux-mock-store,
```
pascalduez, line 2:
```suggestion
// with redux-thunk & redux-mock-store,
```

?

pascalduez: That's were we have a problem, why would the redux types shouldn't be "compatible"?

Brianzchen: What I mean here is that because `Middleware` is typed from `@@redux` it should have compatibility when passed into redux or redux-mock-store. Because these other libs have some type requirements when passing into them.

I guess maybe these comments could straight up be removed? Is that ok with you?

definitions/npm/redux-thunk_v2.x.x/flow_v0.83.x-/redux-thunk_v2.x.x.js
Brianzchen: Should I rename this to `'@@redux-redux-thunk'`? 🤷

definitions/npm/redux-thunk_v2.x.x/flow_v0.83.x-/redux-thunk_v2.x.x.js
pascalduez: Sorry but I still don't get it :sweat_smile: 
Why wouldn't we copy/paste the required types from the "real" redux libdef?

Brianzchen: I've updated the comments. Let me know if it's more clear, it could do without comments but I felt it's better to write something so that we're aware of future state

pascalduez: Actually my question is why do the types here differs from the ones in [the redux libdef](https://github.com/flow-typed/flow-typed/blob/c53229884877b9116a985a202c6cc966149f0e0e/definitions/npm/redux_v4.x.x/flow_v0.134.x-/redux_v4.x.x.js#L10-L20)? Mostly removed generics.

Brianzchen: I guess I was just lazy. But I get your point. Made an update anyways now so that it's matching redux!

pascalduez (approved):
Thanks!

PR #4059, opened 2021-04-05
[mongo_v3.6.x]: Improve types

moroine:
- Start to type Collection (I give up for today, and end up using a lot of `any` but at least interface shape is defined.
- Missing tests, but this would be too much of work


definitions/npm/mongodb_v3.6.x/flow_v0.114.x-/mongodb_v3.6.x.js
```diff
@@ -18,11 +18,84 @@ declare module 'mongodb' {
   // http://mongodb.github.io/node-mongodb-native/3.5/api/BSONRegExp.html
   declare export var BSONRegExp: any; // TODO
 
+  // TODO: type all possible operations
+  declare export type FilterQuery<Doc: { ... }> = $ReadOnly<$Shape<Doc> | { ... }>;
```
moroine, line 22: MongoDb filters can be either shape of a document or using special operators...

pascalduez (approved):

PR #4062, opened 2021-04-13
Jest v26: add missing `createMockFromModule`

mrtnzlml:
See: https://jestjs.io/docs/jest-object#jestcreatemockfrommodulemodulename

### DIFF
--- a/definitions/npm/jest_v26.x.x/flow_v0.134.x-/jest_v26.x.x.js
+++ b/definitions/npm/jest_v26.x.x/flow_v0.134.x-/jest_v26.x.x.js
@@ -828,10 +828,14 @@ type JestObjectType = {
    */
   isMockFunction(fn: Function): boolean,
   /**
+   * Alias of `createMockFromModule`.
+   */
+  genMockFromModule(moduleName: string): any,
+  /**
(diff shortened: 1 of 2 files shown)


definitions/npm/jest_v26.x.x/flow_v0.134.x-/jest_v26.x.x.js
```diff
@@ -832,6 +832,7 @@ type JestObjectType = {
    * mocked version of the module for you.
    */
   genMockFromModule(moduleName: string): any,
+  createMockFromModule(moduleName: string): any,
```
Brianzchen, line 835: Looks like we generally add JS doc comments which help in the editor. Do you want to add that also? 

Could also add a test to validate the type like I did for the recent change to the file like this https://github.com/mrtnzlml/flow-typed/blob/jest_v26_createMockFromModule/definitions/npm/jest_v26.x.x/flow_v0.134.x-/test_jest-v26.x.x.js#L309

mrtnzlml: both fixed

Brianzchen (approved):

PR #4063, opened 2021-04-14
Export a bunch of types from @apollo/react-{common,components,hooks}, fix a few other issues

jedwards1211:
This is for Apollo client v3.  I'm still not using/don't have defs for v4.

Note: changes to `@apollo/react-hoc` are missing from this...hopefully not many people are using `@apollo/react-hoc` at this point, HoCs are gross

Anyone using `eslint-plugin-flow` will need a lot of these types to satisfy missing function argument warnings, etc.

I also fixed a few other issues I was having.

I know these are working on Flow 0.131.0, we'll see about newer versions...

### DIFF
--- a/definitions/npm/@apollo/react-common_v3.x.x/flow_v0.104.x-/react-common_v3.x.x.js
+++ b/definitions/npm/@apollo/react-common_v3.x.x/flow_v0.104.x-/react-common_v3.x.x.js
@@ -2,22 +2,22 @@ declare module '@apollo/react-common' {
   import type { Node } from 'react';
 
   /* start utils */
-  declare type Record<T, U> = {
+  declare export type Record<T, U> = {
     [key: $Keys<T>]: U,
-    ...
+    ...,
   };
   /* end utils */
 
   /* start graphql */
-  declare type DocumentNode = any;
-  declare type ExecutionResult<T> = {
+  declare export type DocumentNode = any;
+  declare export type ExecutionResult<T> = {
     data?: T,
     extensions?: { [string]: any, ... },
     errors?: any[],
     ...
   };
-  declare type GraphQLError = any;
-  declare type VariableDefinitionNode = any;
+  declare export type GraphQLError = any;
+  declare export type VariableDefinitionNode = any;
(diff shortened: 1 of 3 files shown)


definitions/npm/@apollo/react-common_v3.x.x/flow_v0.104.x-/react-common_v3.x.x.js
```diff
@@ -74,14 +74,14 @@ declare module '@apollo/react-common' {
   /* end context */
 
   /* start parser */
-  declare export type DocumentType = 'Query' | 'Mutation' | 'Subscription'
+  declare export type DocumentType = 'Query' | 'Mutation' | 'Subscription';
 
   declare export type IDocumentDefinition = {
     type: DocumentType,
     name: string,
     variables: $ReadonlyArray<VariableDefinitionNode>,
```
Brianzchen, line 82: Would be helpful to fix this too
```suggestion
    variables: $ReadOnlyArray <VariableDefinitionNode>,
```

Brianzchen (commented):
Found one thing you can easily fix, but there are a few errors with undefined references. This probably doesn't effect your current version of flow but from 0.143.0 lib defs now correctly throw errors, do you want to clean those up too?
https://github.com/facebook/flow/releases/tag/v0.143.0

definitions/npm/@apollo/react-common_v3.x.x/flow_v0.104.x-/react-common_v3.x.x.js
jedwards1211: It's weirding me out that Flow didn't catch this in the type defs in my project...

Brianzchen (approved):
Personally this change is good for me. Overall you're syncing your changes of adding exports back into the repo which is always a good thing. There are errors regarding suppression comments and undefined vars which have always been there so it's not like you caused it.

If maintainers are happy with an incremental change we should merge and someone in the future can fix the other errors.

AndrewSouthpaw (approved):

PR #4066, opened 2021-04-21
[react-modal_v3.x] Update definitions

knu:
- Links to documentation: https://github.com/reactjs/react-modal/blob/master/CHANGELOG.md#3131---tue-13-apr-2021-141946-utc
- Link to GitHub or NPM: https://github.com/reactjs/react-modal

Other notes:

- Unify v3.*.x definitions so you don't need an update per minor release
- Update setAppElement()


Brianzchen:
Can I jus check, why do we keep adding new definitions for each minor bump in `react-modal`? Would it be easier to have a `3.x.x` def and just maintain that? Or does react-modal have some release pattern where minor bumps add dramatic changes? 

I think supporting just major versions would make everything easier with maintenance

knu:
@Brianzchen I agree.  I have been adding minor versions just because there were already separate directories for v3.1.x and v3.6.x when I started working on this.  I've merged v3.* into v3.x.

Brianzchen (approved):
Nice. I've also done this for a few other libs because it was getting painful whenever a minor version was released.

definitions/npm/react-modal_v3.x.x/test_react-modal_v3.x.x.js
```diff
@@ -9,7 +9,6 @@ describe('Modal.setAppElement', () => {
     ReactModal.setAppElement('#foo');
     ReactModal.setAppElement(document.body);
 
-    // $FlowExpectedError[incompatible-call]
```
Brianzchen, line 12: I think you need to add this back, number is still incompatible

knu:
OK, tests now pass and it's all set!

Brianzchen (approved):

pascalduez (approved):

PR #4070, opened 2021-04-30
feat(ioredis): add v4 defs

jedwards1211:
- Links to documentation: https://github.com/luin/ioredis/releases/tag/v4.0.0
- Link to GitHub or NPM: https://github.com/luin/ioredis

Other notes:

none of the breaking changes in v4 affect the existing type signatures,
so I copied the v3 defs verbatim


AndrewSouthpaw:
Looks like there's a bunch of suppression errors, and I'm guessing those are all from v3, so I'm calling this "a step in the right direction". Thanks @jedwards1211!

veedeo:
@jedwards1211 @AndrewSouthpaw ioredis v4 supports 'zpopmin' command but current libdef doesn't contain it

Brianzchen:
> @jedwards1211 @AndrewSouthpaw ioredis v4 supports 'zpopmin' command but current libdef doesn't contain it

@veedeo would be interested in adding this property yourself?

PR #4071, opened 2021-04-30
feat(pg): add v8 type defs

jedwards1211:
- Links to documentation: https://node-postgres.com/api
- Link to GitHub or NPM: 

Other notes:

- docs show that `notice` event arg is now `Error`, but from the ts type defs they're working on it
  looks to be a `NoticeMessage` class that shares a lot of properties with `DatabaseError` but
  doesn't actually extend error.
- I exported the `DatabaseError` class but I'm not sure if any promise rejections or error events
  are 100% guaranteed to be an instance of `DatabaseError` and not some other `Error`.
- I made the event emitter classes extend from a `TypedEventEmitter` class instead of `events$EventEmitter`
  for better typing (type errors occur if you try to make `TypedEventEmitter` with `events$EventEmitter`.
  trying to strongly type events properly in Flow makes me want to hang my head and cry)

### DIFF
--- a/definitions/npm/pg_v8.x.x/flow_v0.104.x-/pg_v8.x.x.js
+++ b/definitions/npm/pg_v8.x.x/flow_v0.104.x-/pg_v8.x.x.js
@@ -1,0 +1,409 @@
+declare module pg {
+  // Note: Currently There are some issues in Function overloading.
+  // https://github.com/facebook/flow/issues/2423
+  // So i temporarily remove the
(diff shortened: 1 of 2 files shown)


GAntoine:
It looks like the failures are mostly caused by $FlowFixMe error names. Is that all that's needed to get this merged in, or are any of the "other notes" blockers?

Brianzchen:
pgv8 has been added https://github.com/flow-typed/flow-typed/pull/4207. Additional changes can be made in a new PR

Issue #4051, opened 2021-03-24
Should we accept purposely incomplete lib defs of huge libraries?


Brianzchen:
# Problem

There are huge libraries in the JS ecosystem that are not flow-typed. Namely `material-ui` which many have asked for and `firebase` which I'm starting to get into now. 

I'm slowly typing out my stub for `firebase` which works well but it's a huge effort for myself or anyone for that matter if the first submission of the type def had to be fully complete because I doubt I'll use all the features to be able to ensure it works for everyone.

I did previously add the definition for cypress, but even that was a huge effort that took me days/weeks but I had the motivation because someone else had previously attempted a PR I could base the definitions off. I don't use all the functions from cypress so I couldn't validate that they're all accurate but from the previous attempt I was confident enough to make a submission.

Not having things like `material-ui` I think really hurts the flow community's growth because although the avid flow users don't use material-ui since I personally see it as a more hobbyist library, a lot of new comers would come from that and would find the ecosystem not worth it without their most depended on library typed in comparison to TS.

# Proposal

What I'm wondering is if the general community here would accept a partially typed definition, so that something can at least be started even if it doesn't work for everyone initially. If not the case, I doubt I would ever feel confident enough to submit my stub back to the community.

Of course whatever submitted would include thorough testing against the lib def of what's there so that quality is kept, just that complete definitions won't be there.

I would also note that I my thoughts are that incomplete objects are typed as `{| |}` instead of `{ ... }` and more object properties are simply added over time.

---

What's your thoughts? Tagging some active contributors to get the conversation going
@pascalduez @AndrewSouthpaw @villesau @GAntoine @jedwards1211 @Ashoat @atabel

GAntoine:
I need to mull this over a little, but my gut instinct says that incomplete libdefs are better then no libdefs, as long as the untyped sections are stubbed, so they don't throw errors when running `flow`.

pascalduez:
Like you said, as long as the added  types are properly tested, and maybe a few comments added at the top of the libdef, it's all fine. We already accepted several partial libdefs. 
I also have the same issues, I locally have (I can't count how much) started branches which stalled because it quickly became overwhelming.

jedwards1211:
For me at least, writing type tests is the biggest time suck when making large libdefs.

[...]

PR #4073, opened 2021-05-02
[semver_v7.x.x] add definitions

aminya:
- Links to documentation: https://github.com/npm/node-semver
- Link to GitHub or NPM: https://github.com/npm/node-semver
- Type of contribution: new definition | addition 

Other notes:


Brianzchen:
You need to modify this change in a few ways to actually work with flow.
- You need to have a `declare module '' { ... }` declaration so that if you import `semver` is actually knows where to resolve to
- Everything should be in a single file for the definitions itself, you can look to other flow-typed defs for inspiration on that
- To provide compatibility with the most users the docs recommend starting your def on version 0.83.0 and then working your way up, unless there's a good reason such as being impossible to type util later versions
- Your tests look fine, but you'll want to add suppression codes so that your tests at least all pass when you get this PR reviewed

aminya:
> You need to modify this change in a few ways to actually work with flow.
> 
> * You need to have a `declare module '' { ... }` declaration so that if you import `semver` is actually knows where to resolve to
> * Everything should be in a single file for the definitions itself, you can look to other flow-typed defs for inspiration on that
> * To provide compatibility with the most users the docs recommend starting your def on version 0.83.0 and then working your way up, unless there's a good reason such as being impossible to type util later versions
> * Your tests look fine, but you'll want to add suppression codes so that your tests at least all pass when you get this PR reviewed

> You need to modify this change in a few ways to actually work with flow.
> 
> * You need to have a `declare module '' { ... }` declaration so that if you import `semver` is actually knows where to resolve to
> * Everything should be in a single file for the definitions themselves, you can look to other flow-typed defs for inspiration on that

But we want to be able to import functions from a specific file and not the top-level entry:
```js
const semverSatisfies = require('semver/functions/satisfies')
```

This looks like a very large limitation to me. I can probably bundle the code using a bundler, and then also wrap the individual files inside `declare module`.

mischnic:
1. The path/name of these files doesn't matter
2. You need to wrap the code inside `declare module ".." { }` no matter if it's in one big file or in multiple files
3. The decision to put everything into one file is just a convention of flow-typed, not a limitation of flow (the language).

aminya:
So, what should be the module? Should I write this for each individual file?
```js
declare module 'semver/functions/satisfies' {
	// types for semver/functions/satisfies
}
```

If that's the case, I can probably make a script to do so. It can be integrated into Flowgen itself so people don't have to do all this manual work.

or is it like this for each file:
```js
declare module 'semver' {
	// types for semver/functions/satisfies
}
```

aminya:
There is a related open issue that has not been fixed yet. 
https://github.com/joarwilk/flowgen/issues/27

mischnic:
As I've said:

You need to do this:
```js
declare module 'semver/functions/satisfies' {
	// types for semver/functions/satisfies
}
```

```js
declare module 'semver/functions/somethingelse' {
	// types for semver/functions/somethingelse
}
```

Flow doesn't care if these two are in a single file or not, but flow-typed wants you to put them into a single file:

> Everything should be in a single file for the definitions itself, you can look to other flow-typed defs for inspiration on that

aminya:
What should happen for the top-level file? Can I import them or should I use a bundler?
```js
declare module 'semver' {
	// types for semver entry file
}
```

mischnic:
Here are the types for a package with many of these submodules : https://github.com/flow-typed/flow-typed/blob/6a61eceaa15de58e7335228e0d4027036d028210/definitions/npm/%40material-ui/core_v1.x.x/flow_v0.58.x-v0.88.x/core_v1.x.x.js#L4022

aminya:
Just a clarification. I thought this PR was going to be straightforward. I actually don't need to use FlowTyped. I just want to pass the tests in this PR. I used Flowgen to generate types from @types/semver. But it removed `import/exports` from the files. I added them back in manually, but this doesn't seem to be enough. So, if there is a way I can use Flow itself that doesn't have these limitations, I would be happy with that.

https://github.com/parcel-bundler/parcel/pull/6218

Brianzchen:
Understood, my recommendation would be that you take the existing [semver_v6.2.x](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/semver_v6.2.x/flow_v0.104.x-/semver_v6.2.x.js) and paste it as `semver_v7.x.x` and fix up the names as necessary. Looks like there's been no breaking changes [here](https://github.com/npm/node-semver/blob/master/CHANGELOG.md). Though this may be incomplete as in missing some new functions it should be acceptable as an initial commit as explained [here](https://github.com/flow-typed/flow-typed/issues/4051) though you can of course add the functions you need to be compatible with your PR.

definitions/scripts/index.js
```diff
@@ -0,0 +1,67 @@
...
+const { readFile, writeFile } = promises
+const { join, relative, resolve, dirname } = require("path")
+
+/**
+ * Wrap the files generated by flowgen in declare module
+ *
+ * @param {string} packageName the name of the package
+ * @param {string} packageDirGiven the path of the package
+ */
+async function wrapDeclarePackage(packageName, packageDirGiven) {
```
aminya, line 12: I added a script to bundle the files that are generated by flowgen.

aminya (commented):
I wrote a package called flowgen-package that handles everything for the generation of a FlowTyped compatible file. Using my package we can automatically generate types for all the `@types/*` packages.

I have not still registered it because this can be part of flowgen itself.

https://github.com/aminya/flowgen-package

definitions/npm/semver_v7.x.x/flow_v0.104.x-/semver_v7.x.x.js
```diff
@@ -0,0 +1,689 @@
...
+  import type semverMajor from './functions/major';
+  import type semverMinor from './functions/minor';
+  import type semverPatch from './functions/patch';
+  import type semverPrerelease from './functions/prerelease';
+  import type semverCompare from './functions/compare';
+  import type semverRcompare from './functions/rcompare';
+  import type semverCompareLoose from './functions/compare-loose';
+  import type semverCompareBuild from './functions/compare-build';
+  import type semverSort from './functions/sort';
+  import type semverRsort from './functions/rsort';
```
pascalduez, line 16: All those does not work. 
* The files doesn't exist for a start...
* You can't import from other files in libdefs
* Types must start with a cap (convention)
* Named exports needs wrapping curlies

definitions/npm/semver_v7.x.x/flow_v0.104.x-/semver_v7.x.x.js
```diff
@@ -0,0 +1,689 @@
...
+  declare export var compareIdentifiers: typeof identifiers.compareIdentifiers;declare export var rcompareIdentifiers: typeof identifiers.rcompareIdentifiers;declare export var SEMVER_SPEC_VERSION: "2.0.0";export type ReleaseType = "major"
+  | "premajor"
+  | "minor"
+  | "preminor"
+  | "patch"
+  | "prepatch"
+  | "prerelease";export interface Options {
+  loose?: boolean,
+  includePrerelease?: boolean,
+  } export type CoerceOptions = {
```
pascalduez, line 62: Could you make en effort on formatting, run prettier or at least properly separate blocks. This is really harsh to read and not matching our (very minimal) standards.

pascalduez (changes requested):
Can't be merged as is, will need quite some work.

definitions/npm/semver_v7.x.x/flow_v0.104.x-/semver_v7.x.x.js
pascalduez: Ok, I understand a bit better what your trying to achieve.
It should be more in this vein:

```js
declare module "semver" {
  import typeof { foo } from 'semver/foo';
}

declare module "semver/foo" {
  declare export function foo();
}
```

aminya:
I updated the flowgen-package. Now, it converts the relative import types to `packageName/...`. Could you take a look?
https://github.com/aminya/flowgen-package

@pascalduez

definitions/npm/semver_v7.x.x/flow_v0.104.x-/semver_v7.x.x.js
```diff
@@ -0,0 +1,836 @@
...
+    semverMinVersion as minVersion,
+    semverValidRange as validRange,
+    semverOutside as outside,
+    semverGtr as gtr,
+    semverLtr as ltr,
+    semverIntersects as intersects,
+    simplify as simplifyRange,
+    rangeSubset as subset,
+  };
+  import type SemVer from "semver/classes/semver"
```
pascalduez, line 79: Does not work, type imports needs to be wrapper in curlies.

definitions/npm/semver_v7.x.x/flow_v0.104.x-/semver_v7.x.x.js
```diff
@@ -0,0 +1,836 @@
...
+  import type Range from "semver/classes/range"
+  import type Comparator from "semver/classes/comparator"
+  declare export { SemVer, Range, Comparator };
```
pascalduez, line 82: That does not work either, your mixing normal exports and types.

definitions/npm/semver_v7.x.x/flow_v0.104.x-/semver_v7.x.x.js
```diff
@@ -0,0 +1,836 @@
...
+// Generated from @types/semver/preload using github.com/aminya/flowgen-package
+declare module "semver/preload" {
+  import type semver from "semver/"
+  declare module.exports: typeof semver;
+
+}
+
+// Generated from @types/semver/classes/comparator using github.com/aminya/flowgen-package
+declare module "semver/classes/comparator" {
+  import type sermver from "semver"
```
pascalduez, line 133: Typo

AndrewSouthpaw:
Thanks for the reviewing effort you're doing @pascalduez and @Brianzchen!

definitions/npm/semver_v7.x.x/flow_v0.104.x-/semver_v7.x.x.js
```diff
@@ -0,0 +1,836 @@
+// Generated from @types/semver using github.com/aminya/flowgen-package
+declare module "semver" {
+  import type * as semverParse from "semver/functions/parse"
```
pascalduez, line 3: That's not supported by Flow, check the CI tests.

aminya: What's the flow-typed compatible replacement?

aminya: > What's the flow-typed compatible replacement?

I can't fix this if Flow doesn't support an alternative for this syntax.

Brianzchen: Consider examples like the following for imports of types and vars
import types: https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/redux-thunk_v2.x.x/flow_v0.83.x-/redux-thunk_v2.x.x.js#L23
import vars: https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/webpack_v5.x.x/flow_v0.104.x-/webpack_v5.x.x.js#L3

I believe the overall issue here is that you cannot import all named exports into a flow module, you need to specify all individual vars, and I think you didn't mean to `import type` because looking at your exports you're trying to export vars. Looking at what `semver/functions/parse` does I think you want something like

```
import typeof semverParse from 'semver/functions/parse';
```

definitions/npm/semver_v7.x.x/flow_v0.104.x-/semver_v7.x.x.js
Brianzchen: The main issue here is that you shouldn't be importing `semver/classes/semver` as a type, `import typeof SemVer from 'semver/classes/semver';` should be enough based on the type def of `semver/classes/semver`

PR #4080, opened 2021-05-14
[url-template] Add definition

Brianzchen:
- Links to documentation: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/url-template/index.d.ts
- Link to GitHub or NPM: https://github.com/bramstein/url-template

Other notes: Also clean up suppression errors from #4070


pascalduez (approved):
Thanks!
Best would have been two separate PR's though.

PR #4087, opened 2021-06-06
Fix non existing ramda generics

belka-ew:
- Type of contribution: fix

Other notes:

A few of the ramda definitions seem to use Generics that aren't defined. I' m not sure why the available tests don't fail, I've got an error as soon as I installed the definition.


AndrewSouthpaw (approved):
Interesting! Nice catch.

PR #4096, opened 2021-06-16
Use more interface types for Typescript

mischnic:
- Links to documentation: 
- Link to GitHub or NPM: 
- Type of contribution: fix/refactor

This changes some object types in the typescript defs to be interfaces instead. Flow 0.153.0 requires this to
- be able to pass subclasses of `ParseConfigHost` to funtions that have `ParseConfigHost` as a parameter
- to be able to say `class Something implements ParseConfigHost`, `ParseConfigHost` needs to be an interface

### DIFF
--- a/definitions/npm/typescript_v3.3.x/flow_v0.104.x-/typescript_v3.3.x.js
+++ b/definitions/npm/typescript_v3.3.x/flow_v0.104.x-/typescript_v3.3.x.js
@@ -3297,7 +3297,7 @@ declare module "typescript" {
     ...
   };
 
-  declare type ParseConfigHost = {
+  declare interface ParseConfigHost {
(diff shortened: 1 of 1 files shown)


GAntoine (approved):

pascalduez:
LGTM. But shouldn't we scope those changes to `flow_v0.153.x-`?  which means copying the libdef, and scoping the previous one to `flow_v0.104.x-v0.152.x`

mischnic:
I can do that if you want. Just wasn't sure if that mechanism is rather for "more" breaking changes (like syntax)

mischnic:
@GAntoine  @pascalduez What do you think?

pascalduez (approved):

pascalduez:
Let's try as is, we'll see if we get complains ;)

PR #4099, opened 2021-06-23
[lodash] Compatibility with flow 0.153.0 WIP

Brianzchen:
- Links to documentation: https://www.npmjs.com/package/lodash
- Link to GitHub or NPM: https://www.npmjs.com/package/lodash
- Type of contribution: fix

Other notes: Update lodash def from being a `class` to being a `type`.


Brianzchen:
Actually this is a WIP, previously as a class it supported method overloading, now we need to convert all these overload be intersection property defs

AndrewSouthpaw (approved):
Amazing work @Brianzchen! I've done some spot checking of the PR, thanks for breaking out the commits into logical pieces. I appreciate how you improved several of the type defs along the way, e.g. with `isNumber` etc.

PR #4107, opened 2021-07-03
[koa] Add body into request object

Brianzchen:
When using `koa` with `koa-body` you can access the request body through the ctx, in koa-body docs you can look for `ctx.request.body` for reference.

- Links to documentation: https://www.npmjs.com/package/koa-body
- Link to GitHub or NPM: https://www.npmjs.com/package/koa

Other notes:

### DIFF
--- a/definitions/npm/koa_v2.0.x/flow_v0.104.x-v0.152.0/koa_v2.0.x.js
+++ b/definitions/npm/koa_v2.0.x/flow_v0.104.x-v0.152.0/koa_v2.0.x.js
@@ -24,6 +24,7 @@ declare module 'koa' {
     // props added by middlewares.
     [key: string]: mixed,
     app: Application,
+    body: { [key: string]: any },
     req: http$IncomingMessage<>,
     res: http$ServerResponse,
     ctx: Context,
--- a/definitions/npm/koa_v2.0.x/flow_v0.104.x-v0.152.0/koa_v2.0.x.js
+++ b/definitions/npm/koa_v2.0.x/flow_v0.104.x-v0.152.0/koa_v2.0.x.js
@@ -24,7 +24,7 @@ declare module 'koa' {
     // props added by middlewares.
     [key: string]: mixed,
     app: Application,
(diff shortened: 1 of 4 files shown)


definitions/npm/koa_v2.0.x/flow_v0.104.x-v0.152.0/koa_v2.0.x.js
```diff
@@ -24,6 +24,7 @@ declare module 'koa' {
     // props added by middlewares.
     [key: string]: mixed,
     app: Application,
+    body: { [key: string]: any },
```
pascalduez, line 27: Mark it as exact or inexact.  

Once upon a time we were linting about this...

Brianzchen: Yeah I've tended to stop doing that in my own codebase, because from what I've found even if you add exact or inexact to indexer objects it doesn't seem to actually do anything. Flow doesn't even throw a `ambiguous-object-type=error` error for indexer objects but does as soon as you remove the indexer.

What do you recommend here @pascalduez? I'm ok to add it in for consistency but want to just highlight that I think it doesn't actually do anything.

pascalduez: It indeed doesn't do anything, I just try to enforce it at work, so that devs have to make a choice, and explicitly mark their objects, rather than relying on weak defaults. This is also to favor exact objects whenever possible.   
Moreover to ease the ?future Flow plan to move to exact by default.   

So I would stick with explicitly marking our objects, that was the path we took here anyway, when we ran the 0.104 codemods.

Brianzchen: Fair enough, and all done 😄

pascalduez (approved):

PR #3461, opened 2019-07-16
Add react-redux hooks, `useDispatch` & `useSelector`

ssorallen:
Resolves #3452

- Links to documentation: https://react-redux.js.org/api/hooks
- Link to GitHub or NPM: https://github.com/reduxjs/react-redux
- Type of contribution: addition

Other notes:

Adds implementations for the hooks added in [react-redux v7.1.0](https://github.com/reduxjs/react-redux/releases/tag/v7.1.0).

PR #4109, opened 2021-07-10
[react-redux] Improve useDispatch typing

Brianzchen:
When `useSelector` and `useDispatch` was added here #3461, the typings for `useDispatch` were super basic and basically provided no type checking unless a generic was passed in but gets very tedious adding this type for every single instance of `useDispatch` throughout an application.

In my team, we're already typing the actions themselves so `dispatch` should have the capability to figure out what the return types are without needing additional typing in the calling component. We also use a lot of thunks to run reusable and async code and it's not helpful when the typed thunks provide zero type analysis to the calling component.

This change keeps the current spec that was originally built, but if thunks (action creators that are functions that return functions) are passed in it will appropriately flow the return type back to the caller.

- Links to documentation: https://www.npmjs.com/package/react-redux
- Link to GitHub or NPM: https://www.npmjs.com/package/react-redux
- Type of contribution: fix

Other notes: CC original contributors to this spec, @ssorallen @AndrewSouthpaw


Brianzchen:
Not sure if we should even have that generic at all, I would prefer to remove it but I've kept current impl because of backwards compat for those that actually use it. If we were to remove it we could more strongly type this by preventing `dispatch` from being passed no args, but the generic `D` will always end up as the resolved type if nothing matches.

Brianzchen:
@pascalduez any thoughts on this PR

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/test_useDispatch.js
```diff
@@ -40,4 +42,66 @@ describe('useDispatch', () => {
...
+
+  it('handles action creator passed in and is typed', () => {
+    function Com() {
+      const dispatch = useDispatch();
+      return (
+        <button
+          onClick={function() {
+            const action = dispatch({ anything: 'string' });
+            (action.anything: string);
+            // $FlowExpectedError[incompatible-cast] will be number because of passed action
```
pascalduez, line 54:
```suggestion
            // $FlowExpectedError[incompatible-cast] will be string because of passed action
```

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/test_useDispatch.js
```diff
@@ -16,7 +15,10 @@ describe('useDispatch', () => {
       return (
         <button
           onClick={function() {
-            dispatch({ type: 'action' });
+            const action = dispatch({ type: 'action' });
+            (action.type: string);
+            // $FlowExpectedError[incompatible-cast] will be number because of passed generic
```
pascalduez, line 20:
```suggestion
            // $FlowExpectedError[incompatible-cast] will be string because of passed generic
```

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/react-redux_v7.x.x.js
```diff
@@ -200,7 +200,15 @@ declare module "react-redux" {
...
 
-  declare export function useDispatch<D>(): D;
+  declare export function useDispatch<D>(): (
+    & (<T: { [key: string]: any }>(T) => T)
+    // Supports thunks at their various lengths and use cases depending if user has typed them as tuple vs array
+    & (<T>((...args: [any]) => T) => T)
+    & (<T>((...args: [any, any]) => T) => T)
+    & (<T>((...args: [any, any, any]) => T) => T)
+    & (<T>((...args: Array<any>) => T) => T)
+    & D
```
pascalduez, line 210: I'm surprised an intersection works here, isn't it an intersection instead? Or I could use a bit of explanations TBH.

pascalduez:
> @pascalduez any thoughts on this PR

Added a few comments. I'm not feeling too confident here, I would prefer someone else review. Could you ask on Discord if some of the Flow members are willing to?

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/react-redux_v7.x.x.js
Brianzchen: So `useDispatch` is a function that returns a function, the returned function needs to be an intersection so it can accept a wide variety of api contracts depending on its usage and have them all be valid.
It's not a union type because if it was, there would need to be a type refinement before usage of the function

Brianzchen:
@pascalduez seems it's really hard to get reviewers but this is a really game changing type change for my team. I'm wondering what makes you not so confident? I think I've already added solid test coverage but I can add more if it helps.

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/test_useDispatch.js
```diff
@@ -40,4 +42,66 @@ describe('useDispatch', () => {
...
+          onClick={() => {
+            const a = dispatch(thunkAction());
+            a.toString();
+            // $FlowExpectedError[incompatible-cast] it will return a string, not any which could be cast to number
+            (a: number)
+          }}
+        >
+          Dispatch time
+        </div>
+      );
```
AndrewSouthpaw, line 81: I'm curious why we need to wrap it in a div with onClick? Could we just run those things in the test? It would be less obfuscated perhaps.

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/test_useDispatch.js
```diff
@@ -40,4 +42,66 @@ describe('useDispatch', () => {
...
+    }
+  });
+
+  it('handles thunks defined as a promise to keep their type', (done) => {
```
AndrewSouthpaw, line 85: Do we actually need it async for flow to work?... I'm confused.

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/react-redux_v7.x.x.js
```diff
@@ -200,7 +200,15 @@ declare module "react-redux" {
...
   // ------------------------------------------------------------
 
-  declare export function useDispatch<D>(): D;
+  declare export function useDispatch<D>(): (
+    & (<T: { [key: string]: any }>(T) => T)
+    // Supports thunks at their various lengths and use cases depending if user has typed them as tuple vs array
+    & (<T>((...args: [any]) => T) => T)
+    & (<T>((...args: [any, any]) => T) => T)
+    & (<T>((...args: [any, any, any]) => T) => T)
+    & (<T>((...args: Array<any>) => T) => T)
```
AndrewSouthpaw, line 209: I'm curious about the use case here, could you provide a couple examples? It seems like line 209 is a catch-all, does 206-208 provide value then?

AndrewSouthpaw (changes requested):
Nice work! Just needing some clarifications, but I think this is mostly ready to go.

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/test_useDispatch.js
Brianzchen: Good idea, I just followed previous tests but you're right it makes it easier

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/test_useDispatch.js
Brianzchen: Updated to not be async and just verify it's a promise of given type

definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/react-redux_v7.x.x.js
Brianzchen: Added some more tests samples of usages, since action thunks that are defined as spread array vs spread tuple are non-compatible they need to be catered separately. Tuples are generally more strongly typed since you can type each arg, but if someone was lazy to use array version I didn't want to block the ability to have return types defined which is real value of this change

Brianzchen:
@AndrewSouthpaw any chance of this morning forward?

AndrewSouthpaw (approved):
So sorry for the delay! Just changed jobs and it's been a bit chaotic. Trying to find the new normal.

Brianzchen:
> So sorry for the delay! Just changed jobs and it's been a bit chaotic. Trying to find the new normal.

All good! Thanks to you and other maintainers for putting in the effort to maintain where you can

PR #4111, opened 2021-07-20
[linked-list_3.0.x] add linked-list defs

judysocute:
- Links to documentation: https://www.npmjs.com/package/linked-list
- Link to GitHub or NPM: https://www.npmjs.com/package/linked-list

### DIFF
--- a/definitions/npm/linked-list_v3.0.x/flow_v0.104.x-/linked-list_v3.0.x.js
+++ b/definitions/npm/linked-list_v3.0.x/flow_v0.104.x-/linked-list_v3.0.x.js
@@ -1,0 +1,36 @@
+/*
+ * Type def from from source code of linked-list.
+ * this: https://github.com/wooorm/linked-list
+**/
+declare module 'linked-list' {
+  declare class Item {
+    list: ?List<Item>;
+    next: ?Item;
+    prev: ?Item;
+    head: ?Item;
+    tail: ?Item;
+    size: number;
+
+    constructor(): Item;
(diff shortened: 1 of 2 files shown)


definitions/npm/linked-list_v3.0.x/flow_v0.104.x-/linked-list_v3.0.x.js
```diff
@@ -0,0 +1,36 @@
+/*
+ * Type def from from source code of linked-list.
+ * this: https://github.com/wooorm/linked-list
+**/
+declare module 'linked-list' {
+  declare class Item {
```
Brianzchen, line 6: I'm surprised, you don't need to write `declare export class Item {` to be able to import?

Brianzchen:
Can we make this `3.x.x` instead of `3.0.x` so that future feature versions don't go untyped by default?

judysocute:
> Can we make this `3.x.x` instead of `3.0.x` so that future feature versions don't go untyped by default?

sure we can~

definitions/npm/linked-list_v3.0.x/flow_v0.104.x-/linked-list_v3.0.x.js
judysocute: oh...
let me check it again

Brianzchen: I'm just super curious because I thought you do. But your tests don't seem to die 😅

PR #4116, opened 2021-08-01
Create base lib definitions

### LINKED ISSUE #1857: Is it currently possible to have dependencies between library definitions?

Brianzchen:
Closes #1857

This change aims to create support for sharing type definitions across lib defs, by having base def files that can be defined in a dependency in any library-version-flow version.

Base dependencies are defined in the following way:
- Added under the `definitions/base` directory as a file
- Named in the format of `base-[module]_v[major].[minor].[patch]-[optional end version].js`. Though version naming is just an indicator and has no limitation under it's version installation (rationale of limitation explained [below](https://github.com/flow-typed/flow-typed/pull/4116#issuecomment-890484288))
- lib defs that want to use a base def will create a `package.json` under `[def]/[flow version/` in a property `baseDependencies` which takes an array of base definitions
- Base definitions declare modules in any way but is recommended to follow the format of `@flow-typed/base-[module]` as to avoid conflict with real npm packages

---

This PR in the end will aim to accomplish 3 things:
- [x] When a change is made to a regular lib def, any dependencies referenced in the package.json will be loaded into the test suite before testing starts
- [x] If a base def is modified, any lib def that uses this base def will have itself written to the test suite with their tests run as a testing platform
- [x] When a user installs a lib def, the base defs referenced in package.json should also be installed under `flow-typed/base`

Other notes:
- If/once this merges we need to release this as a major breaking change, and anyone on flow-typed < v4 will need to upgrade otherwise all type defs that have migrated to using base defs will be unusable.


cli/src/lib/libDefs.js
```diff
@@ -271,6 +275,19 @@ async function parseLibDefsFromPkgDir(
...
 
+          // If there is a package.json populate dependencies with full path
+          // to the base file
+          if (flowDirItem === 'package.json') {
+            const pkg = JSON.parse(fs.readFileSync(path.join(flowDirPath, flowDirItem), 'utf8'));
+            dependenciesPaths.push(...pkg.baseDependencies.map((dep) => {
+              // find the real path with flow version
+              console.log(dep);
+              // return dep;
+              return '/Users/brianchen/projects/flow-typed/definitions/base/base-redux/flow_v0.83.x-/base-redux.js';
```
Brianzchen, line 286: Need to figure out which base def to load exactly for a given test group

Brianzchen:
From my current understanding, testing against each flow version writes a `.flowconfig` file into a test suite. But this write is per test group, not per version (test group meaning flow version ranges such as `flow_v0.55.x-v0.88.x` or `flow_v0.134.x-` are each a single test group.).

This would mean that if `redux` needed to depend on something like `base-redux` it would need to load a definition that worked on exactly the same test group versions. 

If dep A had a test group 140-150 and dep B had a test group 130 - 145, and both required from base def C then def C would always need to have a definition that matched 130-150 which may run into issues. If we imagine that on version 140 a new breaking syntax change was available that removed an older syntax, that would basically force dep B in this case to be broken up into 130 - 139 and 140 - 145, though this is not necessarily a bad thing since it would have caused a fragment aways though makes one def can cause many defs to break on one release which is hard to maintain which again would have needed to be maintained anyways if people used it.

Seems like at this point, the simplest thing is that base defs are defined explicitly with name and supported flow versions

Brianzchen:
When a lib def is detected to be changed. It should probably just run every lib def that has a package.json supported otherwise we'll need to detect what may have been deleted or changed as opposed to a direct association.

Brianzchen:
After my final change I've attempted to add tests to the test suite runner but it crashes, in particular when it tries to load definitions, I guess because it can't parse lib defs as plain JS. Though I've tested this myself and it has been proven that you can see the tests from this PR run `redux` and has `base-redux` loaded correctly otherwise tests would fail.

Open to feedback on this case, but otherwise I've tested the install process which was impossible to validate otherwise.

pascalduez:
I'm personally not in favor of this solution.
As I wrote in the issue, feels like adding more complexity and more importantly confusion and weakness.  
Libdefs resolves just fine between them once installed in a project, so it should be just our tests system to be fixed/reviewed. 
Impacting final users to deal with even more fragmentation is not a good path for me.

Possible solutions (without giving much thoughts on it, sorry): a libdef could "declare" dependencies (a "fake" package.json?) and a sandboxed environment created with the libdef and dependencies) before launching the tests.

Brianzchen:
> Possible solutions (without giving much thoughts on it, sorry): a libdef could "declare" dependencies (a "fake" package.json?) and a sandboxed environment created with the libdef and dependencies) before launching the tests.

Thanks for your feedback as always @pascalduez. I took your package.json idea and implemented it for the base defs. I agree your solution is much simpler and maybe we can do that instead but I did have one concern that I'll just post here too. Do you think this would become an issue?

>If I wanted to install react-redux, and it had types defined in redux. Which version of redux is installed? Of course we know that flow versions would be aligned, but should redux@4.x.x or redux@3.x.x be installed? You would think redux@4.x.x because it's the latest that but would cause conflicts for anyone who is on redux@3.x.x, the wrong lib def would be installed for a user and they would have duplicated redux modules. Furthermore, if we simply ignore the 3rd point above, redux@3.x.x defs might not be aligned/maintained to expose the right types for react-redux to handle which would cause an issue for flow-typed consumers to handle on their own, needing to check the repo to find out what the right version of redux lib def is necessary for their react-redux to work

ref: https://github.com/flow-typed/flow-typed/issues/1857#issuecomment-886245309

Otherwise in the package.json I can have react-redux rely on redux@4 and make an assumption that others use this too, then redux can be broken down to match parity with react-redux. Though I'd still like to have the feature of installing depended lib defs automatically which in this case would always try to install redux@4

Brianzchen:
May need to tweak and ensure that each base def is installing with an appropriate code signature.

Brianzchen:
Closing for #4331

PR #4121, opened 2021-08-17
typescript: unbreak unbinding checks for LanguageServiceHost

wbinnssmith:
The definitions for the typescript package don't seem to have any tests, so this broke silently with Flow's addition of unbinding checks.

From https://github.com/parcel-bundler/parcel/pull/6718#discussion_r688793140:

> The "methods" for this interface are a mix of methods and properties whose values are arrow functions. Methods cannot be unbound, but properties can, making them incompatible — unfortunately this means that our subclass can't use entirely methods nor properties.
> 
> I think the reasoning for this is the interfaces these inherit from have optional properties, something that can't be done with methods, and since they're incompatible, they must stay consistent. That's the case with getCurrentDirectory, which is a property in the parent interface but was a method here.

@gkz @jbrown215 How can we best express "optional" methods in an interface? 

cc @matthewwithanm @mischnic @devongovett

Test Plan: Added a very simple test only importing `LanguageServiceHost` and verified it failed before applying this change.

### DIFF
--- a/definitions/npm/typescript_v3.3.x/test_typescript_v3.3.x.js
+++ b/definitions/npm/typescript_v3.3.x/test_typescript_v3.3.x.js
@@ -1,0 +1,3 @@
+// @flow
+
+import type { LanguageServiceHost } from 'typescript';
--- a/definitions/npm/typescript_v3.3.x/flow_v0.104.x-/typescript_v3.3.x.js
+++ b/definitions/npm/typescript_v3.3.x/flow_v0.104.x-/typescript_v3.3.x.js
@@ -8441,7 +8441,7 @@ declare module "typescript" {
     +getProjectReferences?: () => $ReadOnlyArray<ProjectReference> | void,
     +getLocalizedDiagnosticMessages?: () => any,
     +getCancellationToken?: () => HostCancellationToken,
-    getCurrentDirectory(): string,
+    +getCurrentDirectory: () => string,
     getDefaultLibFileName(options: CompilerOptions): string,
     +log?: (s: string) => void,
     +trace?: (s: string) => void,


definitions/npm/typescript_v3.3.x/test_typescript_v3.3.x.js
```diff
@@ -0,0 +1,3 @@
+// @flow
+
+import type { LanguageServiceHost } from 'typescript';
```
wbinnssmith, line 3: Previously, this would fail with the following under Flow 0.154.0 and later:

```
Cannot extend `GetEffectiveTypeRootsHost` [1] with `LanguageServiceHost` because function type [2] cannot be unbound
    from the context [2] where it was defined in the `this` parameter of property `getCurrentDirectory`. [method-unbinding]
```

jbrown215:
> How can we best express "optional" methods in an interface?

I'm not sure, tagging in @dsainati1

dsainati1:
Instead of trying to express "optional" interface methods, I would instead ask how best to model the type relationship you are trying to express here. 

Given the code in your linked issue, 

```
interface Super {
  /* "optional" */ run(): string;
}

// should pass
class Super2 implements Super {
}

const s1 = new Super1();

// should fail
s1.run
```

It seems as though you want the `run` method to not be unbindable, but also not necessarily present on all classes that implement `Super`. The way I would suggest expressing this is not with an optional method, but rather by having two interfaces rather than one: a `Super` interface like you have that expresses the required methods a class needs to be a `Super`, and then a `Runnable` interface that gives the type of this `run` method. This way, classes like `Super2` that are instances of `Super` but do not have a `run` method can be written as only implementing `Super`, but a hypothetical class `Super3` that is a  `Super` with a `run` method can be written as:

```
class Super3 implements Super, Runnable {
    run() : string { /* implmentation goes here */}
}
```

This expresses the relationship between these types more accurately than an optional method would, and allows Flow to make more guarantees about your code.

wbinnssmith:
Thanks @dsainati1 and @jbrown215 — that makes a lot of sense 🙇. For the maintainers here, this particular case comes from TypeScript's own definitions for the typescript api. If we were to take this approach across this entire libdef (which would clean up the types quite a bit I imagine), we would be trading away the value in matching typescript's own definitions for these advantages. I'm not sure which is preferred, and I likely don't have the time to holistically address this myself.

@Brianzchen @pascalduez I'm happy merging this as is, but it may be worthwhile deciding which direction should be taken with this particular libdef (and others that are derived from TypeScript sources/libdefs).

pascalduez:
Given that this particular libdef was script generated and not intended to be merged first hand, I guess this fix is welcome. As for making further improvements plans, open to anyone :)

Issue #4042, opened 2021-03-11
Support --create-pr flag


FezVrasta:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Right now, to send a PR with new types, I need to manually create a PR, which involves forking the repository, create a branch, make the changes, commit, push, and send the PR.

**If the current behavior is a bug, please provide the steps to reproduce.**

Following an approach similar to `patch-package --create-issue` flag, we could have a `create--pr` flag that will automatically create a PR with the type definitions that the user created or modified locally.

**What is the expected behavior?**
```
$ flow-typed create-pr <package>@<version>
Please insert a valid GitHub access token with repo:write permissions:
*************
A new Pull Request with your contribution has been created and it's available at https://github.com/flow-typed/flow-typed/pulls/xxxx
```

**If this is a feature request, what is motivation or use case for changing the behavior?**
The current approach can limit the amount of contributions to this project.

**Local Environment Information**
If this issue is specific to the CLI, provide the results of running `npx envinfo` from the command line in a GitHub gist.

pascalduez:
Hi @FezVrasta,

could be nice.
How would that work with tests files? :sweat_smile:

FezVrasta:
One can add them to the same local directory and the script would match them (prepend test_ to the file name)?

Brianzchen:
Didn't know these things existed, is there any project that does something similar to this?

Updating existing defs is simple, but creating a new one need some finesse to make all the types with the correct naming scheme. Even if we don't do what you're suggesting, some cli command that automates initial file creation would be great.

AndrewSouthpaw:
Yeah never heard of `patch-package` and it'd be cool to see something like that in the CLI.

Brianzchen:
@FezVrasta Wondering do you have ideas of any additional features you'd like here or is the `./create-def.sh` enough which has been good enough for my usage.

FezVrasta:
@Brianzchen I didn't have time to test it yet, I'm sorry.

PR #4125, opened 2021-08-23
Create def script for easier lib def creation

Brianzchen:
Create a handy script inspired from #4042 that will create a basic structure for new type def configured to an appropriate lowest supported version as recommended in contribution guide.

Comes built with a lib def file plus a test file with base structure

Usage: `./create_def.sh lib-name 1.x.x`

### Skeleton File/Folder Structure
*sample when running `./create_def.sh test 1.x.x`*
Screen Shot 2021-08-28 at 4 31 14 pm

### Script Error Handling
Screen Shot 2021-08-23 at 11 55 58 pm

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,6 +49,8 @@ version 4.1.0. **[You've followed all of the
 best practices for writing high quality definitions,](https://github.com/flow-typed/flow-typed/issues/13#issuecomment-214892914)**
 and now you'd like to contribute it:
 
+> Flow-typed comes with a handy script to simplify the file creation process `./create_def.sh` which takes two args, the library name and the version you'd like to type. In the :point_down: example that would be `./create_def.sh left-pad 4.x.x`.
+
 #### 1) Create a new directory called `definitions/npm/left-pad_v4.1.0/`.
 
 ***We call this the "package version directory".***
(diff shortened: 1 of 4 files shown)


Brianzchen:
@FezVrasta not exactly what you were asking for but what do you think? Should remove a lot of small mistakes with repetition when creating a new lib def

GAntoine:
Could/should this be a command instead of a bash file? That way we can add it to the docs and help menu.

Brianzchen:
> Could/should this be a command instead of a bash file? That way we can add it to the docs and help menu.

Sorry @GAntoine what do you mean? Do you want this to be a node command that lives in `cli/package.json` and then is exposed via a bash script as opposed to being a bash script itself? I was basically copying the way `./quick_run_def_tests.sh` gets used.

GAntoine:
Yeah exactly! In the same way you'd run `flow-typed create-stub lib-name`, you could run `flow-typed create-def lib-name 1.x.x`. Then we could add it to the `help` menu and the wiki here on Github.

Brianzchen:
@GAntoine Made a change, let me know if it's what you're looking for. Learnt more about the cli so that was good experience :)

CONTRIBUTING.md
```diff
@@ -49,6 +49,8 @@ version 4.1.0. **[You've followed all of the
 best practices for writing high quality definitions,](https://github.com/flow-typed/flow-typed/issues/13#issuecomment-214892914)**
 and now you'd like to contribute it:
 
+> Flow-typed comes with a handy script to simplify the file creation process `./create_def.sh` which takes two args, the library name and the version you'd like to type. In the :point_down: example that would be `./create_def.sh left-pad 4.x.x`.
```
AndrewSouthpaw, line 52: This'll need to be updated :)

AndrewSouthpaw (approved):
Looks great. @Brianzchen looks like maybe you didn't push the latest changes turning it into a CLI command?

CONTRIBUTING.md
Brianzchen: Oh yep, let me fix this later today

Brianzchen: Actually I think it's fine as is. It just proxies to call `node cli/dist/cli.js create-def left-pad 4.x.x` so docs with how to contribute are still valid

PR #4129, opened 2021-08-28
rxjs: Fix incorrect type for retryWhen.

djcsdy:
- Links to documentation: 
- Link to GitHub or NPM: 
- Type of contribution:  fix

Other notes:

Fix inconsistent/incorrect type for `retryWhen`.

I'm not sure why but flow treats `(rxjs$Observable<T>) => rxjs$Observable<T>` as a different type to `rxjs$MonoTypeOperatorFunction<T>` although they look like they should be compatible to me. As a consequence without this change it is a type error to use `retryWhen` in a `.pipe(...)`, which is how it is intended to be used. For this reason, this change fixes three failing tests.


Brianzchen (approved):
Tested locally, rxjs errors goes from `12 errors and 80 warnings` to `11 errors and 80 warnings`

AndrewSouthpaw (approved):

PR #4130, opened 2021-08-28
rxjs: Fix faulty test for retryWhen.

djcsdy:
- Links to documentation: 
- Link to GitHub or NPM: 
- Type of contribution: fix

Other notes:

This fixes a failing test.

It looks like someone typoed a duplicate `retryWhen` nested inside another `retryWhen`, but it's also possible they did it deliberately for some reason. Nevertheless it causes the test to fail.


AndrewSouthpaw (approved):
Makes sense, lgtm! Thanks!

PR #4133, opened 2021-09-03
[cli] .ignore file support

### LINKED ISSUE #4127: flow-typed ignore file
Feature

When I run `flow-typed install` it goes through package.json to install all deps I may use. But there some libs that I can't imagine I would ever get typed such as `eslint-x` `@babel/preset-x` `@babel/plugin-x`. As a result running the install command always updates stubs which generates more noise in my PR than is desirable.

Second problem: I sometimes run a mono repo application where a library then wouldn't have any flow files (I use them uncompiled), so when I run `flow-typed install` it would install stubs. I then have to delete it, my alternative is to create a `stub.js.flow` file in the lib so that flow-typed won't try to install stubs that override whatever types are defined in my uncompiled js code.

## Solution
I would like flow-typed to support an ignore file `.flowtypedignore` (not fussed on the name) that live in the root dir where I can specify some exact matches to library names that I don't want to have flow-typed try to install.

Future thinking we could expand this ignore file to work on particular scopes and wildcards, but the simplest use case would work well for now.

Brianzchen:
Closes #4127.

This PR supports a new `.ignore` file that can be added to the root `flow-typed` dir of a project where your lib defs currently live. ie: `./flow-typed/.ignore`

Here you can add lines of exclusions separated by new lines supporting explicit dependencies or scopes
eg the file may look like
```
@babel
@babel/
@testing-library/react
eslint

```

Other notes: Tested locally and works pretty well

### DIFF
--- a/cli/src/commands/__tests__/install-test.js
+++ b/cli/src/commands/__tests__/install-test.js
@@ -796,5 +796,101 @@ describe('install (command)', () => {
         ).toEqual(true);
       });
     });
+
+    it("doesn't install definitions that are ignored", () => {
+      return fakeProjectEnv(async FLOWPROJ_DIR => {
+        // Create some dependencies
+        await Promise.all([
+          mkdirp(path.join(FLOWPROJ_DIR, 'src')),
+          writePkgJson(path.join(FLOWPROJ_DIR, 'package.json'), {
+            name: 'test',
+            devDependencies: {
+              'flow-bin': '^0.43.0',
+            },
+            dependencies: {
+              foo: '1.2.3',
+            },
+          }),
+          mkdirp(path.join(FLOWPROJ_DIR, 'node_modules', 'foo')),
+          mkdirp(path.join(FLOWPROJ_DIR, 'node_modules', 'flow-bin')),
+        ]);
+
+        await touchFile(path.join(FLOWPROJ_DIR, 'src', '.flowconfig'));
+        await mkdirp(path.join(FLOWPROJ_DIR, 'src', 'flow-typed'));
+        await touchFile(
+          path.join(FLOWPROJ_DIR, 'src', 'flow-typed', '.ignore'),
+        );
+        await fs.writeJson(
+          path.join(FLOWPROJ_DIR, 'src', 'flow-typed', '.ignore'),
+          'foo',
+        );
+
+        // Run the install command
+        await run({
+          overwrite: false,
+          verbose: false,
+          skip: false,
+          rootDir: path.join(FLOWPROJ_DIR, 'src'),
+          explicitLibDefs: [],
+        });
+
+        // Installs libdef
+        expect(
+          await fs.exists(
+            path.join(
+              FLOWPROJ_DIR,
+              'src',
+              'flow-typed',
+              'npm',
+              'foo_v1.x.x.js',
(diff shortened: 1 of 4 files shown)


Brianzchen:
@AndrewSouthpaw @GAntoine

cli/src/commands/__tests__/install-test.js
```diff
@@ -796,5 +796,178 @@ describe('install (command)', () => {
...
+              'src',
+              'flow-typed',
+              'npm',
+              '@scoped',
+              'package_vx.x.x.js',
+            ),
+          ),
+        ).toEqual(false);
+      });
+    });
```
AndrewSouthpaw, line 971: Love the tests, great work.

cli/src/commands/install.js
```diff
@@ -284,8 +284,22 @@ async function installNpmLibDefs({
     }
     console.log(`• Searching for ${libdefsToSearchFor.size} libdefs...`);
   } else {
+    let ignoreDefs;
+    try {
+      ignoreDefs = fs
+        .readFileSync(path.join(cwd, libdefDir, '.ignore'), 'utf-8')
+        .replace(/"/g, '')
+        .split('\n');
+    } catch (e) {
```
AndrewSouthpaw, line 293: Can we scope the catch to only handle the "couldn't find the file" error and otherwise throw? I'm always leery of just silencing all errors. (JS is much worse about being able to do this, which is a bummer...)

cli/src/lib/npm/npmProjectUtils.js
```diff
@@ -90,6 +97,18 @@ export function getPackageJsonDependencies(
         if (deps[pkgName]) {
           console.warn(`Found ${pkgName} listed twice in package.json!`);
         }
+        const pkgIgnored = ignoreDefs.reduce((acc, cur) => {
+          if (acc) return acc;
+          const ignoreDef = cur.trim();
+          if (ignoreDef === '') return acc;
+          // if we are looking to ignore a scope dir
+          if (ignoreDef.charAt(0) === '@' && ignoreDef.indexOf('/') === -1) {
```
AndrewSouthpaw, line 105: Does this mean `@foo/` would not ignore all subfolders? I think this would surprise some people... maybe we could make that more resilient and add a test for it?

cli/src/lib/npm/npmProjectUtils.js
```diff
@@ -90,6 +97,18 @@ export function getPackageJsonDependencies(
         if (deps[pkgName]) {
           console.warn(`Found ${pkgName} listed twice in package.json!`);
         }
+        const pkgIgnored = ignoreDefs.reduce((acc, cur) => {
```
AndrewSouthpaw, line 100: I think you could use `.some` instead of `.reduce` and it'll both fail early and read more easily.

AndrewSouthpaw (changes requested):
Marvelous work! Excited to see this land in the system.

Brianzchen:
@AndrewSouthpaw Thanks for suggestions all done!

moroine:
Any idea when this will be fixed into a new flow-typed version?

GAntoine:
I'll publish a release this weekend.

PR #4135, opened 2021-09-07
[@testing-library/react] Returned union

### LINKED ISSUE #4117: [@testing-library/react] matcher return type improvement
feature

Currently with react testing library, when using the various `getByTestId`, `screen.getByTestId`, `render.getByTestId`, `findAllByTestId` and etc always return as `HTMLElement`. This is not very ergonomic and I've had a problem with this for over a year shown by my attempted fix https://github.com/flow-typed/flow-typed/pull/3837.

We currently need to write code in the form of
```
const element: HTMLAnchorElement = (screen.getByTestsId('test'): any);
expect(element.href).toBe('blah');
```
Or without type casting but produces more code
```
expect.hasAssertions();
const element = screen.getByTestsId('test');
if (element instanceof HTMLAnchorElement) {
  expect(element.href).toBe('blah');
}
```
When people without flow would write it as 
```
expect(screen.getByTestsId('test').href).toBe('blah');
```

I'd like to change the defined type and related types to be like the following
```
declare type ReturnElements = HTMLElement & HTMLButtonElement & HTMLInputElement & ...;

declare type GetByBoundAttribute = (
  text: TextMatch,
  options?: TextMatchOptions
) => ReturnElements;
```
In this way, `getByTestId` is never returned as `any` but always has the advantage of easily casting into a specific HTMLElement or just using as is just like in the without flow example above

cc: @pascalduez would appreciate your feedback here!

Brianzchen:
Closes #4117.

- Links to documentation: https://testing-library.com/docs/react-testing-library/intro/
- Link to GitHub or NPM: https://www.npmjs.com/package/@testing-library/react

Other notes: There are two types created to support this feature. `IntersectionHTMLElement` and `UnionHTMLElement` which are necessary for what is returned from RTL and then accepted into a function from RTL or any custom func someone has created.

@pascalduez @AndrewSouthpaw

### DIFF
--- a/definitions/npm/@testing-library/react_v11.x.x/flow_v0.104.x-/react_v11.x.x.js
+++ b/definitions/npm/@testing-library/react_v11.x.x/flow_v0.104.x-/react_v11.x.x.js
@@ -2,7 +2,7 @@
  * A local copy from:
  * https://github.com/A11yance/aria-query/blob/2e6a3011a0d8987655f3a14853934fe3df38a8d8/flow/aria.js
  */
-declare module '@@aria-query' {
+ declare module '@@aria-query' {
   declare export type ARIAAbstractRole =
     | 'command'
     | 'composite'
@@ -490,7 +490,7 @@ declare module '@testing-library/react' {
   declare export function waitFor<T>(
     callback: () => T | Promise<T>,
     options?: {|
-      container?: HTMLElement,
+      container?: UnionHTMLElement,
       timeout?: number,
       interval?: number,
       mutationObserverOptions?: MutationObserverInit,
@@ -500,7 +500,7 @@ declare module '@testing-library/react' {
   declare export function waitForElementToBeRemoved<T>(
     callback: (() => T) | T,
     options?: {|
-      container?: HTMLElement,
+      container?: UnionHTMLElement,
       timeout?: number,
       interval?: number,
       mutationObserverOptions?: MutationObserverInit,
@@ -512,7 +512,7 @@ declare module '@testing-library/react' {
    * In most cases you should be able to find/replace `wait` with `waitFor`.
    * Learn more: https://testing-library.com/docs/dom-testing-library/api-async#waitfor.
    */
-  declare export function wait(
+   declare export function wait(
     callback?: () => void,
     options?: {
       timeout?: number,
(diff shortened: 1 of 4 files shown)


pascalduez:
Like I said in https://github.com/flow-typed/flow-typed/issues/4117#issuecomment-892519150 I'm not keen of this solution myself. 
It would be great to have several users feedback here instead of the usual suspects before deciding on merging it or not.

MonsieurLanza:
I agree with @pascalduez. You may want to simplify your code, others may want precise typing. 

Besides this, HTML has way more than 5 elements (an informal list here : https://www.codingexercises.com/a-complete-list-of-all-htmlelement-interface-subclasses), some with some exotic attributes we still cannot test here without cast.

Brianzchen:
Thanks for the feedback, I would love this change but I agree this is a community project so the community should decide, I'm happy to wait until we get enough feed whether that leads to this being merged or not 😃 

> Besides this, HTML has way more than 5 elements (an informal list here : https://www.codingexercises.com/a-complete-list-of-all-htmlelement-interface-subclasses), some with some exotic attributes we still cannot test here without cast.

Thanks @MonsieurLanza, appreciate the list you've given. I mostly just added enough for my own usecase but it can always be enhanced as others need

AndrewSouthpaw:
My $0.02, which is reflected in the Ramda types primarily: I prefer "useful" over "correct". We've seen that bias play out a couple times within the core flow team as well, and it's one that aligns for me. 

Particularly in this case, where we're adding types for _tests_, which are self-validating anyway. If types in tests aren't useful and requires me to add additional casting and such, I just turn them off because it's more trouble than it's worth. They're more like guidelines of catching major mistakes in logic of what type it is, or mistyping a property.

AndrewSouthpaw:
@Brianzchen have you been using this in the wild? How has the DX been with typing for the library?

Brianzchen:
> @Brianzchen have you been using this in the wild? How has the DX been with typing for the library?

I've copied the changes into one of my projects months ago and so far haven't heard any complaints from my team. Feels like a very seamless change to those that don't use the extra functionality

AndrewSouthpaw (approved):
Since we haven't heard from anyone else, and Brian's team hasn't had any issues, I think this is reasonable to move forward.

PR #4136, opened 2021-09-14
[redux-saga] Add done to runSaga

Brianzchen:
Adding a property `done` to the `Task` interface used in `runSaga`. After flow typing some more of my codebase I found this threw an error. The strange thing is there is hardly any documentation on this property and also doesn't appear in TS types but it's 100% usable, docs attached and I use it in my test like so,

```
  it('clears unauth route if it was previously populated', async () => {
    // ... arrange

    await runSaga(
      // ... implementation detail
    ).done;

    // ... assertion 
  });
```

- Links to documentation: http://cef62.github.io/redux-saga/docs/advanced/UsingRunSaga.html
- Link to GitHub or NPM: https://www.npmjs.com/package/redux-saga

Other notes:


Brianzchen:
Actually nvm 😂

PR #4140, opened 2021-09-22
feat: install cmd supports monorepo

moroine:
`install` command supports mono-repo

### DIFF
--- a/cli/src/commands/install.js
+++ b/cli/src/commands/install.js
@@ -25,6 +25,7 @@ import {
 
 import {
   findFlowSpecificVer,
+  findWorkspacesPackages,
   getPackageJsonData,
   getPackageJsonDependencies,
   loadPnpResolver,
@@ -265,7 +266,17 @@ async function installNpmLibDefs({
       const termMatches = term.match(/(@[^@\/]+\/)?([^@]+)@(.+)/);
       if (termMatches == null) {
         const pkgJsonData = await getPackageJsonData(cwd);
-        const pkgJsonDeps = getPackageJsonDependencies(pkgJsonData, []);
+        const workspacesPckJsonData = await findWorkspacesPackages(
+          pkgJsonData,
+          cwd,
+        );
+        const pkgJsonDeps = workspacesPckJsonData.reduce((acc, pckData) => {
+          return {
+            ...acc,
+            // TODO: warn duplicates with different versions
+            ...getPackageJsonDependencies(pckData, []),
+          };
+        }, getPackageJsonDependencies(pkgJsonData, []));
         const packageVersion = pkgJsonDeps[term];
         if (packageVersion) {
           libdefsToSearchFor.set(term, packageVersion);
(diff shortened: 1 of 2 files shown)


Brianzchen:
Any tests we can add to verify this?

cli/src/lib/npm/npmProjectUtils.js
```diff
@@ -74,6 +78,51 @@ export async function findPackageJsonPath(pathStr: string): Promise<string> {
   return path.join(pkgJsonPathStr, 'package.json');
 }
 
+export async function findWorkspacesPackagePaths(
+  pkgJson: PkgJson,
+  cwd: string,
+): Promise<string[]> {
+  if (
+    pkgJson.content.private !== true ||
+    !Array.isArray(pkgJson.content.workspaces)
```
Brianzchen, line 87: It seems that yarn has updated their api over time, but my projects seem to use `workspaces.packages` can you make it support that also so that flow-typed has the best compatibility with the end user?

cli/src/commands/install.js
```diff
@@ -265,7 +266,17 @@ async function installNpmLibDefs({
       const termMatches = term.match(/(@[^@\/]+\/)?([^@]+)@(.+)/);
       if (termMatches == null) {
         const pkgJsonData = await getPackageJsonData(cwd);
-        const pkgJsonDeps = getPackageJsonDependencies(pkgJsonData, []);
+        const workspacesPckJsonData = await findWorkspacesPackages(
```
Brianzchen, line 269: Nitpick but suggest we rename this var to be consistent with how it's used in rest of the project
```suggestion
        const workspacesPkgJsonData = await findWorkspacesPackages(
```

Brianzchen (commented):
Functionally LGTM!

cli/src/lib/npm/npmProjectUtils.js
moroine: Sorry, I don't really understand your question. 

`workspaces` field seems to be `string[]` according to:
- npm: https://docs.npmjs.com/cli/v7/using-npm/workspaces
- yarn: https://classic.yarnpkg.com/lang/en/docs/workspaces

Do you mean you have, `workspaces.packages` field? 

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#workspaces

moroine:
@Brianzchen I've added some tests

cli/src/lib/npm/npmProjectUtils.js
Brianzchen: Yeah I think docs have since updated but they still support the old way. I implemented workspaces way back when it was first released and never really touched it since, my package.json looks like this
```
  "workspaces": {
    "packages": [
      "packages/*"
    ]
  },
```
Do you think we should continue to support this so that if I want to make use of this I also don't need to change the way my project is configured?

Brianzchen (approved):
This is a great start. I hope this will get merged and we can improve incrementally later if need be

moroine:
@Brianzchen just added support for your workspaces

Brianzchen:
@moroine can you fix the conflicts

moroine:
@Brianzchen updated

Brianzchen (approved):

moroine:
@GAntoine any update on this?

Brianzchen:
@moroine until the changes merge I've published them here https://www.npmjs.com/package/ftyped

Brianzchen:
@moroine do you mind pulling from master, your test will have a single flow error that you can suppress that's caused by new method-unbinding

moroine:
@Brianzchen updated

GAntoine (approved):
So after discussing it with @Brianzchen, I'm happy with this change. Couple things to cleanup:

* Could you fix the tests so they pass on both node versions?
* Can we print warning messages for duplicate versions?

Once those are addressed, I'd want to release this as an RC, so people can try it in their existing yarn monorepos.

moroine:
@GAntoine I've updated the PR

Brianzchen (approved):

PR #4143, opened 2021-10-03
[cli] install will also install lib defs of dependencies that ship with flow type

### LINKED ISSUE #4137: flow-typed install doesn't install deps deps
bug

When I have libraries that ship with `flow.js`, `flow-typed install` does not check those libs for their dependencies packages to add to the flow-typed/npm dir.

This is frustrating because flow will throw errors for module not found. Instead I need to check node_modules find the library and see what version of the missing libdef and then manually install each of them.

When I run `flow-typed install` there is already logic to not install stubs or defs for `flow.js` dependencies, maybe we can also check out dependencies to see internally if they have `flow.js` and then install lib defs for that dependency's dependencies. 

**Example:**
I have a dependency called `a@1.2.3`, it has `flow.js` so that's great I don't need lib defs and I can get type checking out of the box. But `a@1.2.3` depends on `b@1.1.1`, this does NOT have `flow.js` files and as package `a` was being built it used `flow-typed`. When I install `a@1.2.3` and then run `flow-typed install` I should expect to see `b@1.x.x` installed into my flow-typed

Brianzchen:
Closes #4137

- Links to documentation: 
- Link to GitHub or NPM: 
- Type of contribution: fix

### Edit

Example use case:
In my company we flowtype our internal libraries that we then share across codebases. When those libs are built they may have dependencies such as `uuid` or `redux`. While building them, they would need to depend on flow-typed.

When these libs are shipped (lets call one of them `@company/component`) it transpiles into regular js with corresponding `flow.js` files.

Now in my project I want to install `@company/component` and when flow analyses the codebase it would throw uuid module not found errors. This is because `@company/component` needs the definitions for `uuid` but I previously never had it in my `flow-typed/npm` dir.

Ok no worries, I will run `flow-typed install` and everything will be solved. Except because I don't use `uuid` but my dependency does, type defs for uuid will never be installed. I could manually install it with `flow-typed install uuid` but that throws and says I need to specify a version, so now I need to dig into node_modules find the dependency and know that I need to run `flow-typed install uuid@8`.

This is a pretty confusing and not noobie friendly. So with this change, as we parse through all npm libraries that are internally typed and aren't typed through flow-typed we will search their inner dependencies and if it's not already a dependency of the root project it will install their dependencies lib defs too.


Brianzchen:
@GAntoine Does this make sense?

cli/src/commands/__tests__/install-test.js
```diff
@@ -1067,6 +1067,198 @@ describe('install (command)', () => {
...
+              FLOWPROJ_DIR,
+              'src',
+              'flow-typed',
+              'npm',
+              'foo_v1.x.x.js',
+            ),
+          ),
+        ).toEqual(false);
+      });
+    });
```
AndrewSouthpaw, line 1261: Thanks for the tests, that helped me understand the proposed changes in behavior 🎉

AndrewSouthpaw (approved):
Wow! I just assumed that's what the CLI did already. Thanks for this improvement!

AndrewSouthpaw:
@GAntoine This will need a CLI minor version bump, I think.

Issue #4151, opened 2021-10-21
Please upgrade the dependency `table` to a more recent version because of the security alert for ansi-regex


julienw:
Here is the dependency chain: flow-typed > table > string-width > strip-ansi
Here is the CVE: https://github.com/advisories/GHSA-93q8-gq69-wqmw

Without this upgrade, projects depending on flow-typed will get the alert (even if they're possibly not vulnerable, I haven't studied so closely here).

Thanks!

PR #4152, opened 2021-10-21
Bump a number of dependencies (October 2021)

GAntoine:


cli/package.json
```diff
@@ -38,25 +38,25 @@
...
-    "@babel/preset-flow": "^7.8.3",
+    "@babel/cli": "^7.15.7",
+    "@babel/core": "^7.15.7",
+    "@babel/preset-env": "^7.15.7",
+    "@babel/preset-flow": "^7.14.5",
     "babel-eslint": "^10.0.3",
-    "babel-jest": "^25.1.0",
-    "eslint": "^6.8.0",
+    "babel-jest": "^27.3.1",
+    "eslint": "^7.32.0",
```
Brianzchen, line 53: With this upgrade we may need to migrate to @babel/eslint-parser instead of `babel-eslint` should we do this in separate PR?

cli/package.json
```diff
@@ -38,25 +38,25 @@
     "prettier": "^1.19.1",
     "rimraf": "^3.0.2",
     "semver": "7.3.2",
-    "table": "^5.4.6",
-    "unzipper": "^0.10.8",
+    "table": "^6.7.2",
+    "unzipper": "^0.10.11",
     "which": "^2.0.2",
-    "yargs": "^15.1.0"
+    "yargs": "^17.2.1"
```
Brianzchen, line 44: This is a large bump are we aware of the impact here or should it also be a separate PR?
Just trawling through changelogs I'm seeing a wall of text that I don't necessarily understand
https://github.com/yargs/yargs/blob/main/CHANGELOG.md#1600-2020-09-09

Brianzchen (commented):
I've left some comments on some that I think may need additional work or testing, have you verified stuff still works?

Overall I think we should maintain our dependencies and this is a good chance for it. But maybe lets have one just for the simple ones + `table` which was mentioned in #4151 and that can be shipped as a simple patch release.

PR #4153, opened 2021-10-22
[cli] Update flow-bin version to latest

Brianzchen:
Keep flow-bin up to date and update cli flow-typed defs with up to date from repo while making use of new `.ignore` file


Brianzchen:
@pascalduez looks all good now!

PR #4155, opened 2021-11-05
Fix history.listen to have an object as a parameter

mikeldking:
- Links to documentation: 
https://github.com/remix-run/history/blob/8bef6f4d50548f46ab7c97e171b3d8634093e7a7/packages/history/index.ts#L131
- Link to GitHub or NPM: 
- Type of contribution: fix

Other notes:

### DIFF
--- a/definitions/npm/history_v5.x.x/flow_v0.104.x-/history_v5.x.x.js
+++ b/definitions/npm/history_v5.x.x/flow_v0.104.x-/history_v5.x.x.js
@@ -22,7 +22,7 @@ declare module 'history' {
     go(n: number): void,
     goBack(): void,
     goForward(): void,
-    listen(({ location: HistoryLocation, action: Action }) => void): Unregister,
+    listen(({| location: HistoryLocation, action: Action |}) => void): Unregister,
     block(
       blocker: (transition: {|
         action: Action,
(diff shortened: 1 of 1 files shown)


definitions/npm/history_v5.x.x/flow_v0.104.x-/history_v5.x.x.js
```diff
@@ -22,7 +22,7 @@ declare module 'history' {
     go(n: number): void,
     goBack(): void,
     goForward(): void,
-    listen((location: HistoryLocation, action: Action) => void): Unregister,
+    listen(({ location: HistoryLocation, action: Action }) => void): Unregister,
```
Brianzchen, line 25: Does `listen` accept it only an object or could accept two args? Asking if this breaks current implementation and if this should be an overload method instead?

mikeldking: The `listen` callback function is passed in `location` and `action` via an object. See https://github.com/remix-run/history/blob/main/docs/api-reference.md#history.listen

Brianzchen (approved):
LGTM

pascalduez (approved):

pascalduez:
Failing tests not related...

Issue #3833, opened 2020-05-13
peerDependencies and devDependencies shouldn't warn


Brianzchen:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When running `flow-typed install` and it scans my `package.json` for dependencies it also considers `peerDependencies` and throws warnings

<img width="664" alt="Screen Shot 2020-05-13 at 10 55 46 am" src="https://user-images.githubusercontent.com/12436524/81761202-ad6d7780-950c-11ea-9985-643089cf87ca.png">

**If the current behavior is a bug, please provide the steps to reproduce.**
1. Have `devDependencies` and `peerDependencies` with some of the same packages
2. Run `flow-typed install`
3. Get warnings

**What is the expected behavior?**
Ideally no warnings because there's nothing wrong with the package.json

**Local Environment Information**
```
System:
    OS: macOS Mojave 10.14.6
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 188.45 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.13.0 - ~/.nvm/versions/node/v12.13.0/bin/node
    Yarn: 1.22.0 - ~/.yarn/bin/yarn
    npm: 6.12.0 - ~/.nvm/versions/node/v12.13.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
```

Brianzchen:
This tells me about duplicates for a number of reasons.. want to think of a better idea first

PR #4157, opened 2021-11-09
[cli] Fix installing lib def package dep uses `>=`

### LINKED ISSUE #4067: dependencies listed as `>=` can't resolve lib def
Bug

I have a dependency listed as `"react-dom": ">=16.8.0"` and it flow-typed will not able to find the lib and instead just install a stub

Have dependencies listed like this
```
  "peerDependencies": {
    "react": ">=16.8.0",
    "react-dom": ">=16.8.0"
  },
```

When I run `flow-typed install` it will install a stub for my `react-dom`

It it should install `react-dom_v16.x.x.js`

```
System:
    OS: macOS 10.15.7
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 598.48 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.17.0 - ~/.nvm/versions/node/v12.17.0/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.17.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    Cargo: 1.46.0 - ~/.cargo/bin/cargo
    Homebrew: 3.0.8 - /usr/local/bin/brew
    Maven: 3.8.0 - /usr/local/bin/mvn
    pip3: 20.1.1 - /usr/local/bin/pip3
    RubyGems: 3.0.3 - /usr/bin/gem
  Utilities:
    Make: 3.81 - /usr/bin/make
    GCC: 4.2.1 - /usr/bin/gcc
    Git: 2.28.0 - /usr/local/bin/git
    Clang: 11.0.3 - /usr/bin/clang
  Servers:
    Apache: 2.4.41 - /usr/sbin/apachectl
  SDKs:
    iOS SDK:
      Platforms: iOS 13.7, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
  IDEs:
    Nano: 2.0.6 - /usr/bin/nano
    VSCode: 1.49.0 - /usr/local/bin/code
    Vim: 8.1 - /usr/bin/vim
    Xcode: 11.7/11E801a - /usr/bin/xcodebuild
  Languages:
    Bash: 3.2.57 - /bin/bash
    Java: 11.0.9 - /usr/bin/javac
    Perl: 5.18.4 - /usr/bin/perl
    PHP: 7.3.11 - /usr/bin/php
    Python: 2.7.16 - /usr/bin/python
    Python3: 3.8.5 - /usr/local/bin/python3
    Ruby: 2.6.3 - /usr/bin/ruby
    Rust: 1.46.0 - /Users/brianchen/.cargo/bin/rustc
  Databases:
    SQLite: 3.28.0 - /usr/bin/sqlite3
  Browsers:
    Chrome: 90.0.4430.85
    Firefox: 86.0
    Safari: 14.0.3
```

Brianzchen:
Closes #4067

- Links to documentation: 
- Link to GitHub or NPM: 
- Type of contribution: fix

Other notes:


GAntoine:
@pascalduez does this change make sense to you? Specifically

> But in the case of flow types, an import of a lib should probably match the lowest version that matches the range to assume backwards compatibility usage

If that sounds right we can merge this in (with a test or two).

Brianzchen:
> with a test or two

I can look into this thanks

pascalduez:
Feels a little bit contrived, but why not. It's probably filling a gap we have.  
Can't the `semver` package be used for that?

Brianzchen:
> Can't the `semver` package be used for that?

Exactly what I thought initially, but if you think about semver ^ !== >= so they would never be compatible hence the transform in this PR

> Though there is something unanswered in the issue, your use case if for peerDependencies, right?

No actually that's something different, despite deps, peer deps, dev deps flow-typed will install in the same way. I do want to make a separate PR maybe in the future that will solve my other problem though not a problem that causes dev issues and just a logging annoyance 

> Cause I guess we have another issue with those, I often get a warning "dependency listed twice in package.json". I guess we should skip a peerDep if present as dep or devDep.

Yep I have one already [here](https://github.com/flow-typed/flow-typed/issues/3833)

Brianzchen:
@GAntoine Test added!

PR #4159, opened 2021-11-11
[redux-saga_v1.x.x] Allow for any return value on Subpattern functions

romellem:
- Links to documentation: https://redux-saga.js.org/docs/api/#takepattern
- Link to GitHub or NPM: https://github.com/redux-saga/redux-saga/blob/1ecb1bed867eeafc69757df8acf1024b438a79e0/packages/types/types/index.d.ts#L12
- Type of contribution: fix

Other notes:

Function patterns don't require returning a boolean, but instead just returning _any_ value, which is checked for "thruthiness" or "falsiness."

> If `pattern` is a function, the action is matched if `pattern(action)` is true (e.g. `take(action => action.entities)` will match all actions having a (truthy) `entities` field.)


Brianzchen (approved):
Sounds fair

pascalduez (approved):

PR #4164, opened 2021-11-21
[DOCS] Improve issues template

Brianzchen:
Previous github issues we had were the default set by github which doesn't necessarily apply to the information we need to help our user base.

This adds new issue templates to cater for 4 areas
- CLI Bug reporting
- CLI Feature requests/discussions
- Definition requests
- Definition issues

cc: @GAntoine @pascalduez @AndrewSouthpaw


.github/ISSUE_TEMPLATE/definition_request.yml
```diff
@@ -0,0 +1,49 @@
...
+      required: true
+
+  - type: input
+    id: flow_version
+    attributes:
+      label: Flow Version
+      description: What minimum flow version do you need this working on
+      placeholder: 0.83.0
+    validations:
+      required: true
```
AndrewSouthpaw, line 30: I don't think this should be required, some people won't know what is the significance of this question. I'd rather it not be a barrier to entry, and force people to put in a random value.

AndrewSouthpaw, line 30:
```suggestion
  - type: input
    id: flow_version
    attributes:
      label: Flow Version
      description: What minimum flow version do you need this working on
      placeholder: 0.83.0
    validations:
      required: false
```

AndrewSouthpaw (changes requested):
This is great! I didn't know this was a feature of GitHub, really neat. Thanks for the work, @Brianzchen! Just one comment, and then I think this is good to go.

.github/ISSUE_TEMPLATE/definition_request.yml
Brianzchen: Yep I'm not fussy there. Done!

GAntoine (approved):

PR #4170, opened 2021-11-23
[cli] install --skipCache option

### LINKED ISSUE #2581: Add option to skip cache update on install
Please add option to skip cache update when running `flow-typed install`.

I want to run `flow-typed install` in parallel across few packages. Updating cache multiple times is not just a waste of resources, but runs into problem of git lock.

Brianzchen:
Closes #2581

- Links to documentation: 
- Link to GitHub or NPM: 

Other notes:
cc: @jirutka

### DIFF
--- a/cli/src/lib/npm/__tests__/npmLibDefs-test.js
+++ b/cli/src/lib/npm/__tests__/npmLibDefs-test.js
@@ -186,10 +186,9 @@ describe('npmLibDefs', () => {
         const pkgName = 'jest-test-npm-package';
         const pkgVersion = 'v1.0.0';
         const flowVersion = {kind: 'all'};
-        const skipCache = false;
 
         await expect(
-          findNpmLibDef(pkgName, pkgVersion, flowVersion, undefined, skipCache),
+          findNpmLibDef(pkgName, pkgVersion, flowVersion),
         ).resolves.toEqual(null);
       });
     });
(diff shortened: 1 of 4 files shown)


cli/src/lib/npm/npmLibDefs.js
```diff
@@ -181,8 +181,10 @@ async function extractLibDefsFromNpmPkgDir(
   return libDefs;
 }
 
-async function getCacheNpmLibDefs(cacheExpiry) {
-  await ensureCacheRepo(cacheExpiry);
+async function getCacheNpmLibDefs(cacheExpiry, skipCache) {
```
GAntoine, line 184: just to be explicit about it
```suggestion
async function getCacheNpmLibDefs(cacheExpiry, skipCache = false) {
```

cli/src/lib/npm/npmLibDefs.js
```diff
@@ -374,8 +376,9 @@ export async function findNpmLibDef(
   pkgVersion: string,
   flowVersion: FlowVersion,
   useCacheUntil: number = CACHE_REPO_EXPIRY,
+  skipCache: boolean,
```
GAntoine, line 379:
```suggestion
  skipCache: boolean = false,
```

GAntoine (changes requested):

cli/src/lib/npm/__tests__/npmLibDefs-test.js
```diff
@@ -197,11 +199,18 @@ describe('npmLibDefs', () => {
...
         let error;
         try {
-          filtered = await findNpmLibDef(pkgName, pkgVersion, flowVersion);
+          filtered = await findNpmLibDef(
+            pkgName,
+            pkgVersion,
+            flowVersion,
+            undefined,
+            skipCache,
+          );
```
GAntoine, line 213: Since `skipCache` now has a default value, I don't think you need to explicitly pass it here (same goes for the previous test)

GAntoine (approved):
LGTM, one small thing I noted, but feel free to merge this in after that's changed (or not, it's still correct as-is)

cli/src/lib/npm/__tests__/npmLibDefs-test.js
Brianzchen: Yeah makes sense. Let me fix that up

PR #4171, opened 2021-11-23
[cli] New outdated command

### LINKED ISSUE #2077: New CLI command: `outdated`
Since it is recommended [here](https://github.com/flowtype/flow-typed/wiki/FAQs#why-do-i-need-to-commit-the-libdefs-that-flow-typed-installs-for-my-project) to commit the libdef files for a project to be stable, it would be nice to have a command `flow-typed outdated` in order to detect and warn when we should update the files!

[`Yarn` has that feature](https://yarnpkg.com/en/docs/cli/outdated) and it is quite convenient.

Brianzchen:
Closes #2077. User can run `flow-typed outdated` to render a cli table letting the user know which deps are outdated.

- Either the lib def was a stub but is now typed
- Or it was typed but there's been a change checking whether the libdef signature has changed

Screen Shot 2022-02-04 at 6 30 47 am

Notes on extra changes:
- Added `test:watch` script to easily tweak changes
- `determineFlowVersion` renamed to `determineFlowSpecificVersion` to avoid name conflict and shift to `lib` for reuse with `outdated` command


GAntoine (approved):

AndrewSouthpaw:
Seems pretty useful! What else needs to be done for this to not be a WIP?

Brianzchen:
> Seems pretty useful! What else needs to be done for this to not be a WIP?

The comment blocks if you read the command basically outlines all the places I still haven't figured out yet. It's definitely a fun change imo but I've just been preoccupied with other stuff. Given the complexity vs the value I've just put a much lower priority on this so far.

Also it'll need tests which shouldn't be too difficult once the change is done.

Brianzchen:
Only have tests remaining!

GAntoine (approved):

PR #4172, opened 2021-11-25
[lodash] Remove dissoc

### LINKED ISSUE #1862: lodash dissoc type
I couldn't find `libdef` for lodash on their repo so apologies in advance if this is irrelevant here. I think the return type of `_.dissoc` should be an `object` rather than `boolean`.

See [lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js][1]

`dissoc` is an [alias][2] for `unset` but it doesn't mutate the object and rather it returns a new one. 

```js
// I think this is wrong
dissoc(path: Array<string> | string): (object: Object) => boolean;

// This might be more accurate
dissoc(path: Array<string> | string): (object: Object) => Object;
```

[1]: https://github.com/flowtype/flow-typed/blob/0289401daee889dfee9232c21512462b689f7395/definitions/npm/lodash_v4.x.x/flow_v0.63.x-/lodash_v4.x.x.js#L2984

[2]: https://github.com/lodash/lodash/wiki/FP-Guide#aliases

Brianzchen:
Closes #1862. I was trawling through issues and found this one. But when searching documentation, what TS does, even lodash code itself, `dissoc` doesn't exist so I assume it must be old and we can cleanup

- Links to documentation: https://lodash.com/docs/4.17.15
- Link to GitHub or NPM: https://www.npmjs.com/package/lodash

Other notes:


GAntoine:
`dissoc` was/is an alias for `unset`, whether its still in use is another question.

definitions/npm/lodash_v4.x.x/flow_v0.153.x-v0.154.x/lodash_v4.x.x.js
```diff
@@ -591,7 +591,7 @@ declare module "lodash" {
     filter:
       & (<T>(array?: ?$ReadOnlyArray<T>, predicate?: ?Predicate<T>) => Array<T>)
       & (<A, T: ReadOnlyIndexerObject<A>>(
-        object: T,
+        object?: ?T,
```
pascalduez, line 594: Si t's a bit wider thant removing dissoc?

Brianzchen: Yeah so lodash never had all it's tests ever passing so I took the opportunity to do some cleanup work while I was in the area. `filter()` and `filter(null)` both return empty array.

Should I just split them into separate PR's though?

PR #4176, opened 2021-11-26
[cicd] Run builds on windows and linux (WIP)

Brianzchen:
- Links to documentation: 
- Link to GitHub or NPM: 

Other notes:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,0 +1,40 @@
+name: Linux Build
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the master branch
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw:
Sifting through some old notifications... is this still a wip?

Brianzchen:
> Sifting through some old notifications... is this still a wip?

Yeah... It doesn't actually work and is giving a false positive. Seems like from another issue recently that people that use flow-typed are using WSL anyways so maybe it's not necessary anymore and it's hard for me to test myself without a windows machine :(

AndrewSouthpaw:
Yeah at my work we don't support actual Windows machines, we expect people to be on WSL.

Brianzchen:
Will close in that case, seems like a change that isn't necessary right now

PR #4180, opened 2021-11-28
[cli] Replace unzipper with node-stream-zip

GAntoine:
[node-stream-zip](https://github.com/antelle/node-stream-zip) is more actively maintained and required fewer dependencies. Performance should be exactly the same.


cli/src/lib/cacheRepoUtils.js
```diff
@@ -105,7 +105,6 @@ export async function ensureCacheRepo(
         }
 
         if (lastUpdated + cacheRepoExpiry < Date.now()) {
-          console.log('• rebasing flow-typed cache...');
```
Brianzchen, line 108: Why remove this? Because we have options like skip cache users may want to see this

GAntoine: Good catch, that was removed by accident

Brianzchen (approved):
LGTM

cli/flow-typed/npm/next_v9.x.x.js
```diff
@@ -0,0 +1,250 @@
+// flow-typed signature: 514646ec2cd38bf48c399f8b3ac0ea8c
+// flow-typed version: 071a03ea49/next_v9.x.x/flow_>=v0.104.x
+
+declare module "next" {
```
pascalduez, line 4: Are we using Next.js on the CLI? :)

GAntoine: It must be a dependency of a dependency, as per @Brianzchen recent work.

Brianzchen: I had a look, flowgen which is ships with flow has a dependency on `next`

PR #4182, opened 2021-11-29
[cli] Add a published readme

Brianzchen:
When the cli is published it's currently a little barren right now
https://www.npmjs.com/package/flow-typed

other notes:
- Should I add anything else?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ cli/dist/*
 cli/.test-dir/*
 cli/.flow-bins-cache/*
 cli/package-lock.json
+cli/README.md
 
 # Editors
 .vscode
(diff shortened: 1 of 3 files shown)


GAntoine:
I couldn't figure out why it was empty for the longest time 🤦🏼 
Could we copy over the existing README into cli, or symlink it somehow? I think having the full readme on npm would be useful.

Brianzchen:
> I couldn't figure out why it was empty for the longest time 🤦🏼 Could we copy over the existing README into cli, or symlink it somehow? I think having the full readme on npm would be useful.

Makes sense! Updating some script naming too, it would make sense from reading this https://docs.npmjs.com/cli/v8/using-npm/scripts#prepare-and-prepublish

GAntoine (approved):
Yeah that's perfect 👏🏼

Issue #929, opened 2017-06-07
Add `// @flow` annotation to the flow-typed definitions


NameFILIP:
Currently, `eslint-plugin-flowtype` linter fails for flow-typed definitions (See https://github.com/gajus/eslint-plugin-flowtype/issues/224). The reason is the way it decides if a file has flowtype definitions (See https://github.com/gajus/eslint-plugin-flowtype/blob/v2.32.1/src/utilities/isFlowFile.js#L17).

The question here is: 
Should flow-typed files have a `// @flow` on the first line of the file?
Alternatively, the plugin can be changed to check for something like `// flow-typed signature ...`

Brianzchen:
Added docs for how to handle eslint errors from flow-typed

PR #4183, opened 2021-11-29
[cli] Install will create eslintrc

### LINKED ISSUE #779: Improve flow-typed and eslint compatibility
After installation of flow-typed in a project in the `/flow-typed` folder there are often style conflicts with .eslintrc rules at the root of the project.
It is normal because the definitions are defined with a style that comes from the github repository.

There are many possible solutions : 
* Ask ESLint to ignore 'flow-typed' folder by default
* `flow-typed install` can configure `.eslintignore` to ignore flow-typed folder
* `flow-typed install` can create a `.eslintrc` in the flow-typed directory that disable all rules (root=true and empty rules)

Brianzchen:
Closes #779, #929

When starting a new project `flow-typed install` should be seamless and with zero config. With eslint it's not really the case and at least for me I'd add `flow-typed` into `.eslintignore`. There's basically no reason to run eslint rules at all in libdefs coming from `flow-typed` so this change adds an automatic `.eslintrc.js` that disables rules.

It can be overridden though for some unique use cases.

### DIFF
--- a/cli/src/lib/fileUtils.js
+++ b/cli/src/lib/fileUtils.js
@@ -18,6 +18,18 @@ export function copyDir(srcPath: string, destPath: string): Promise<void> {
   });
 }
 
+export function writeFile(destPath: string, content: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    fs.writeFile(destPath, content, (err: Error) => {
+      if (err) {
+        reject(err);
+      } else {
+        resolve();
+      }
+    });
+  });
+}
+
 export function copyFile(
   srcPath: string,
   destPath: string,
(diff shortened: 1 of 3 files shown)


cli/src/lib/fileUtils.js
```diff
@@ -18,6 +18,18 @@ export function copyDir(srcPath: string, destPath: string): Promise<void> {
   });
 }
 
+export function writeFile(destPath: string, content: string): Promise<void> {
```
pascalduez, line 21: We should probably start using Node's `fs/promises` at some point.

pascalduez:
IMO this is going a bit too far, it's not `flow-typed` responsibility.   
Also:
-  in the current state it will add the ESLint config file even if the project isn't using ESLint.
- enforce the `.eslintrc.js` config file type whereas others exists
- enforce setting that kind of ignore/config into the `flow-typed` dir, whereas users might prefer dealing with this at the project root

Brianzchen:
Good points... Let me come back to this

GAntoine:
I agree with pascal here. Is this something we can add to the docs, with specified steps on how to make flow-typed work with eslint?

Issue #2881, opened 2018-10-24
Why does it take so long to install new libdefs?


callumlocke:
In my client's project (admittedly large, with 140 direct dependencies in `node_modules`), I often run `flow-typed install --skip` to update or add any new libdefs.

This command generally takes 60-80 seconds to run.

What is it doing? My understanding is it refreshes the local clone of the `flow-typed` repo, looks at my project's `package.json`, and then just copies over the appropriate libdefs (in my case about 25 files). Am I missing some necessary, expensive operation? Or is the long command runtime due to inefficiencies that we could improve?

I am not complaining here, I want to help improve it, I just want to know first if there is some extra aspect to this command that I am missing.

PR #4185, opened 2021-12-01
[cli] Install will restart flow bin

### LINKED ISSUE #4184: restart flow after install command
### CLI Version

3.5.0

When I run `flow-typed install` sometimes large diffs will cause flow to crap itself and essentially be broken but flow is still running. I end up with flow errors such as module not found or cannot find `describe` for jest definitions and such.

Some of my team who are not as experienced will be completely stuck but myself understanding the problem knows I can solve the issue with `yarn flow stop && yarn flow`

After running `flow-typed install` I would like the cli to also run `npm run flow stop && npm run flow` in the dir that executed the installation.

Choosing npm over yarn because not everyone uses yarn.

### Anything else?

I feel like this is very important a DX improvement, aimed at continuing to create a seamless experience for our users

Yes

Brianzchen:
Closes #4184.

User when running `flow-typed install` will now also have flow stop and start which will have more guarantee of a stable install process.

Added an option `skipFlowRestart` that will tell flow-typed to not run this post install process but it being false by default seems better for every day user.

This does slow down install by a bit, so will need to look into performance of flow-typed install later anyways as mentioned here #2881

Screen Shot 2021-12-01 at 7 25 01 pm

### DIFF
--- a/cli/flow-typed/npm/lodash_v4.x.x.js
+++ b/cli/flow-typed/npm/lodash_v4.x.x.js
@@ -1,5 +1,5 @@
-// flow-typed signature: bda9729ed82a5e508dab97c2976a44c6
-// flow-typed version: ae5138a380/lodash_v4.x.x/flow_>=v0.155.x
+// flow-typed signature: fd88436625196dbb8bf813a1f1684e58
+// flow-typed version: 179ae71d58/lodash_v4.x.x/flow_>=v0.155.x
 
 declare module "lodash" {
   declare type Path = $ReadOnlyArray<string | number> | string | number;
@@ -177,7 +177,7 @@ declare module "lodash" {
   declare type NestedArray<V> = Array<Array<V>>;
   declare type Collection<V, K = Key> = $ReadOnlyArray<V> | ReadOnlyIndexerObject<V, K>;
 
-  declare type matchesIterateeShorthand = { [Key]: any, ... };
+  declare type matchesIterateeShorthand = { [key: string]: any, ... };
   declare type matchesPropertyIterateeShorthand = [string, any];
   declare type propertyIterateeShorthand = string;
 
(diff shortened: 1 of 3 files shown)


GAntoine:
Implementation-wise this looks good. What happens to users who didn't need flow to restart? Is it just that the install time will be longer, but they won't see any different behaviour?

Brianzchen:
> Implementation-wise this looks good. What happens to users who didn't need flow to restart? Is it just that the install time will be longer, but they won't see any different behaviour?

That's right, we have no way of knowing whether or not a user **needs** flow to restart, but it's a common case that after install flow needs a kick to work properly again. So just best to run it for everyone.

https://github.com/facebook/flow/issues/8695 here's an issue I also raised a while back related to this

GAntoine (approved):

pascalduez:
This should have been set as opt-in not opt-out.   
What's the current mood? Forcing more and more opinionated things by default to our users?

GAntoine:
@pascalduez My understanding was that this won't actually change the behaviour from the user's perspective, but if that's not the case then there's a good argument for putting this behind a flag.

Brianzchen:
My justification for opt-in by default on restarts is because the default on any option should cater to the simple use cases that help non-advanced users of flow-typed get the most effective usage. flow-typed isn't an effective tool if running the most common command can often cause flow to break, I understand that this is a flow problem but it's triggered by the large rapid diff created by flow-typed.

I want to encourage more people to use flow and for that I need a seamless experience I feel bad telling them, `Flow is a great tool that sometimes breaks but don't worry if it does you can try running X or Y and it should be fine again`, people would immediately opt for TS which from what I understand has a very nice DX.

Furthermore, updates to library definitions folders already cause flow to restart so this is just making sure it gets one more clean restart that doesn't cause some definitions to not load correctly. Therefore I don't think it would cause any extra load time. For those that have special use case they can turn this off of course and the flag exists.

pascalduez:
I've never add issues with installing or modifying libdefs, Flow's updating just fine.   
Furthermore I'm worried how this might interfere with VSCode extension (depending on settings), seems okay after a quick test, but still. 
So this seems like again opinionated conclusions based on a unique experience without much of the community consensus (granted that there's not much of a community left...). 
Looks like it's the path currently taken by flow-typed.

Brianzchen:
Just curious @pascalduez what was your quick testing method? I've found this tends to happen frequently for me and my team where computer specs vary but our flow setup is nothing special, we run small repos 20-30 deps/devDeps. I've often found this issue to happen after we upgrade flow though and then upgrade our libdefs which if there's a lot of stubs would cause lots of rewrites

PR #4187, opened 2021-12-01
[cli] Fix definitions not running in ci env

Brianzchen:
For some reason the tests stopped working in ci. This is related how the checkout action only pulls a minimal clone of the repo and then can't diff between the current changes vs master/previous commit.

As part of this I found that if you made a commit and try to run tests with `onlyChanged` it won't find any tests kind of like how `jest --watch` works. But since we're always interested in comparing to master it will now always find tests to run if there are changes regardless if they're committed or not.

Please check out this build [c907934](https://github.com/flow-typed/flow-typed/runs/4401191310?check_suite_focus=true) to prove that tests run as expected when there was a change

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
         with:
-          fetch-depth: 2
+          fetch-depth: 0
       - name: Setup Node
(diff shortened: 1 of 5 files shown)


Brianzchen:
Improved the logic generally and works well locally. something must be wrong in the CI env itself

GAntoine (approved):

PR #4192, opened 2021-12-14
Add  `res` to express$Request

### LINKED ISSUE #4190: Add `res` to express$Request definition in express_v4.17.x.js
### Library definition file name

npm/express_v4.17.x.js

### Your dependent version

4.17.x

### Flow Version

0.133.0

### Description

I'd like to add `res` to Request definition in https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/express_v4.17.x/flow_v0.104.x-/express_v4.17.x.js#L51.

1. The def already exists in typescript defs: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-serve-static-core/index.d.ts#L648
2. This is where `res` it's being assigned in express https://github.com/expressjs/express/blob/master/lib/middleware/init.js#L28
3. Docs added to expressjs recently: https://github.com/expressjs/expressjs.com/pull/1291

Now I have to have my own intersection type override to fix it, but if approved, I'll send a PR.

### Do you want to submit a pull request to fix this issue?

Yes

micmmakarov:
Closes #4190.

### DIFF
--- a/definitions/npm/express_v4.17.x/flow_v0.104.x-/express_v4.17.x.js
+++ b/definitions/npm/express_v4.17.x/flow_v0.104.x-/express_v4.17.x.js
@@ -63,6 +63,7 @@ declare class express$Request extends http$IncomingMessage mixins express$Reques
   path: string;
   protocol: "https" | "http";
   query: { [name: string]: string | Array<string>, ... };
+  res?: express$Response | void;
   route: string;
   secure: boolean;
   signedCookies: { [signedCookie: string]: string, ... };
--- a/definitions/npm/express_v4.17.x/flow_v0.104.x-/express_v4.17.x.js
+++ b/definitions/npm/express_v4.17.x/flow_v0.104.x-/express_v4.17.x.js
@@ -63,7 +63,7 @@ declare class express$Request extends http$IncomingMessage mixins express$Reques
   path: string;
   protocol: "https" | "http";
   query: { [name: string]: string | Array<string>, ... };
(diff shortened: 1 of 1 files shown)


definitions/npm/express_v4.17.x/flow_v0.104.x-/express_v4.17.x.js
```diff
@@ -63,6 +63,7 @@ declare class express$Request extends http$IncomingMessage mixins express$Reques
   path: string;
   protocol: "https" | "http";
   query: { [name: string]: string | Array<string>, ... };
+  res?: express$Response | void;
```
Brianzchen, line 66: This should be fine, `?:` already represents optional (undefined), unless you did `void` because you meant to handle `null` in that case you'd want `?express$Response` or `express$Response | null`
```suggestion
  res?: express$Response;
```

micmmakarov:
Although typescript does both: `res?: Response<ResBody, Locals> | undefined;`
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-serve-static-core/index.d.ts#L648

That's where I took the original idea.

Brianzchen (approved):

Brianzchen:
> Although typescript does both: `res?: Response<ResBody, Locals> | undefined;` 

Thanks, Don't know if that's really necessary though... Happy to have this fixed with `| void` later if we come across some kind of issue

PR #4193, opened 2021-12-16
[cli] Dramatically improve install time

### LINKED ISSUE #2881: Why does it take so long to install new libdefs?

Brianzchen:
Closes #2881

### The issue
When looking into the above issue, the install command is not that slow but tbh it could be faster because it's really simple, it just copies a bunch of files from one place to another. After digging into the code what I found was that the reason it's so slow is because the base function it uses to install a single lib def will always update the cache and then loop through every cached libdef to find the one to install to a users project. This is extremely ineffective especially the `getNpmLibDefs` which bogs down the process as you increase the amount of dependencies.

### Solution
What I did to fix this was just make the cache update and cache lookup conditional in the base function but for the install command it will only ever update cache once (unless user wants to skipCache) and get all cached libdefs only once.

### Other notes:
From my testing I got performance improvements of around `6979` ms on average to `2252` ms on an average sized repo. I tested this by basically adding a ton of log statements

cc: @callumlocke @AndrewSouthpaw @pascalduez @GAntoine


GAntoine (commented):
Change wise I think this makes sense. Could you add a test to cover the new `extLibDefs`?

GAntoine (approved):

PR #4197, opened 2021-12-18
[cli] upgrade octokit/rest version

### LINKED ISSUE #4195: create-stub not working with osx monterey
### CLI Version

3.5.0

### Operating system

macOS Monterey 12.x

### Bug description

I'm getting a JS error from a deep dependency being outdated while trying to run `create-stub` on monterey.

```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at macosRelease (/<path>/node_modules/macos-release/index.js:26:26)
    at osName (/<path>/node_modules/os-name/index.js:21:18)
    at getUserAgent (/<path>/node_modules/@octokit/rest/node_modules/universal-user-agent/dist-node/index.js:11:53)
    at parseOptions (/<path>/node_modules/@octokit/rest/lib/parse-client-options.js:80:57)
    at new Octokit (/<path>/node_modules/@octokit/rest/lib/constructor.js:23:31)
    at Object.<anonymous> (/<path>/node_modules/flow-typed/dist/lib/github.js:10:16)
```

Seems like the Octokit version you use is somewhat outdated. They have some dependencies that eventually lead to the following block of code in `macos-release` package.

```
const nameMap = new Map([
	[20, ['Big Sur', '11']],
	[19, ['Catalina', '10.15']],
	[18, ['Mojave', '10.14']],
        ...
```
The newer versions of Octokit seem to not even have those deps anymore, so updating should fix this issue I assume (but don't know if it will cause any new issues for you 😄)

### Steps to reproduce

Run `create-stub` for some package while on osx monterey

### Error output/screenshots

_No response_

### Flow configuration

n/a

### Anything else?

_No response_

### Do you want to submit a pull request to fix this bug?

No (hope someone else can)

Brianzchen:
Closes #4195. Upgrade a dependency that's causing issues on newer versions of macos, this dependency is in fact only used in the [test runner](https://github.com/flow-typed/flow-typed/blob/master/cli/src/commands/runTests.js#L5) command but because of code that's executed because of `require()`'s it breaks unrelated functionality when using flow-typed.

octokit/rest major release notes:
- https://github.com/octokit/rest.js/releases/tag/v17.0.0
- https://github.com/octokit/rest.js/releases/tag/v18.0.0

Test strategy: 
- `yarn test` in cli
- `./quick_run_def_tests.sh` pulls latest 15 versions of flow to test

Screen Shot 2021-12-18 at 11 09 53 pm

Other notes: @GAntoine because this one is a blocker we should consider preparing a new minor release soon?

cc: @ansonlouis


GAntoine:
> because this one is a blocker we should consider preparing a new minor release soon?

Yeah I'm hoping to have one out this weekend while I'm on vacation.

GAntoine (approved):

PR #4200, opened 2021-12-21
[redux-devtools-extension] fix toplevel-library-import errors

grit96:
Move the imports and type declarations into the `declare module 'redux-devtools-extension'` block, and duplicate the type declarations into the other module blocks.

### DIFF
--- a/definitions/npm/redux-devtools-extension_v2.x.x/flow_v0.104.x-/redux-devtools-extension_v2.x.x.js
+++ b/definitions/npm/redux-devtools-extension_v2.x.x/flow_v0.104.x-/redux-devtools-extension_v2.x.x.js
@@ -1,5 +1,32 @@
-import type { ActionCreator, StoreEnhancer } from 'redux';
-import typeof { compose } from 'redux';
+declare type DispatchAPI<A> = (action: A) => A;
+
+declare type Dispatch<A: { type: *, ... }> = DispatchAPI<A>;
+
+declare type Store<S, A, D = Dispatch<A>> = {
+  dispatch: D,
+  getState(): S,
+  subscribe(listener: () => void): () => void,
+  replaceReducer(nextReducer: Reducer<S, A>): void,
+  ...
+};
+
+declare type Reducer<S, A> = (state: S | void, action: A) => S;
+
+declare type StoreCreator<S, A, D = Dispatch<A>> = {
+  (reducer: Reducer<S, A>, enhancer?: StoreEnhancer<S, A, D>): Store<S, A, D>,
+  (
+    reducer: Reducer<S, A>,
+    preloadedState: S,
+    enhancer?: StoreEnhancer<S, A, D>
+  ): Store<S, A, D>,
+  ...
+};
+
+declare type StoreEnhancer<S, A, D = Dispatch<A>> = (
+  next: StoreCreator<S, A, D>
+) => StoreCreator<S, A, D>;
+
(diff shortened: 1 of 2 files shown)


Brianzchen:
So this is against our guidelines outlined [here](https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#dont-import-types-from-other-libdefs) as it would cause our tests suite to fail which you can see [here](https://github.com/flow-typed/flow-typed/runs/4600424963?check_suite_focus=true#step:5:337) being unable to resolve the `redux` module. 

For now your best bet is to copy the definitions from redux and add them into your libdef like [so](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/redux-thunk_v2.x.x/flow_v0.83.x-/redux-thunk_v2.x.x.js#L6) but there is some ideas on how to solve this problem to improve scale such as https://github.com/flow-typed/flow-typed/pull/4116 and the main issue tracked here https://github.com/flow-typed/flow-typed/issues/1857

grit96:
@Brianzchen Thanks for the info, I've pushed some updated changes attempting to follow the guidelines.

definitions/npm/redux-devtools-extension_v2.x.x/flow_v0.104.x-/redux-devtools-extension_v2.x.x.js
```diff
@@ -45,7 +72,7 @@ declare type $npm$ReduxDevtoolsExtension$DevToolsOptions = {
 };
 
 declare function $npm$ReduxDevtoolsExtension$composeWithDevTools<A, B>(ab: A => B): A => B;
-declare function $npm$ReduxDevtoolsExtension$composeWithDevTools(options: $npm$ReduxDevtoolsExtension$DevToolsOptions): compose;
+declare function $npm$ReduxDevtoolsExtension$composeWithDevTools(options: $npm$ReduxDevtoolsExtension$DevToolsOptions): Function;
```
Brianzchen, line 75: The function type here is aliased to `any` so it's not very useful, maybe we can use something more specific or even `(...args: Array<any>) => any` would be better?
ref: https://flow.org/en/docs/types/functions/#toc-function-type

grit96: I will switch it to `(...args: Array<any>) => any`. I did try to use `$Compose` which is what compose is typed as in redux, but there were lots of errors in the tests.

Brianzchen (approved):

Ashoat:
I was able to get this to work like so: https://gist.github.com/Ashoat/63fb2d95351711019e2fe7787a3a54e9

It's not entirely clear to me if the imports actually work, but `flow type-at-pos` gave me a name declared from the module from which the types are being imported... so it seems like it's working?

Is it actually possible to import types across libdefs now?

Brianzchen:
> Is it actually possible to import types across libdefs now?

Flow has always supported this but flow-typed the cli tooling doesn't support it in a way that allows for libdefs to be tested properly, it's documented in [contributing](https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md#dont-import-types-from-other-libdefs) docs though looks like not in detail so I'll try improve that...

But I do want to support imports between libdefs which is why I started a discussion RFC a week ago and I'd love your input @Ashoat as someone who has been asking for it for a long time! https://github.com/flow-typed/flow-typed/discussions/4242

PR #4201, opened 2021-12-21
[redux] fix Reducer state type

grit96:
Fix the state argument type in the reducer type


Brianzchen (approved):
LGTM, I'll leave this open for a couple days if that's alright with you to see if anyone else has an opinion

PR #4203, opened 2021-12-21
[react-redux] export ConnectedComponent

grit96:
Export the `ConnectedComponent` class to allow for connected component exports to be annotated.

### DIFF
--- a/definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/react-redux_v7.x.x.js
+++ b/definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/react-redux_v7.x.x.js
@@ -64,16 +64,19 @@ declare module "react-redux" {
     // and provide the DispatchProps type to the DP type parameter.
     | ((dispatch: D, ownProps: OP) => (dispatch: D, ownProps: OP) => DP);
 
-  declare class ConnectedComponent<OP, +WC> extends React$Component<OP> {
+  declare class ConnectedComponentClass<OP, +WC> extends React$Component<OP> {
     static +WrappedComponent: WC;
     getWrappedInstance(): React$ElementRef<WC>;
   }
+
+  declare export type ConnectedComponent = typeof ConnectedComponentClass;
+
   // The connection of the Wrapped Component and the Connected Component
   // happens here in `MP: P`. It means that type wise MP belongs to P,
   // so to say MP >= P.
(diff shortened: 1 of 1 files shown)


Brianzchen (approved):
LGTM, error are unrelated and I'll merge a fix soon
Screen Shot 2021-12-22 at 9 41 56 am

Brianzchen:
@grit96 I have a question, you're exporting this class which is a variable which should match the api of the real `react-redux` exports. Does `react-redux` actually have a named export as `ConnectedComponent` or are you more looking for the type of the class in which case, maybe we can export an interface or a type that is a `typeof ConnectedComponent`?

In any case, I'm thinking we should add a test case which shows how you plan to use this or how it's useful

grit96:
@Brianzchen `react-redux` exports a type called [`ConnectedComponent`](https://github.com/reduxjs/react-redux/blob/master/src/types.ts#L90) which I was trying to match here. Either an interface or a `typeof ConnectedComponent` would work.

Brianzchen:
> @Brianzchen `react-redux` exports a type called [`ConnectedComponent`](https://github.com/reduxjs/react-redux/blob/master/src/types.ts#L90) which I was trying to match here. Either an interface or a `typeof ConnectedComponent` would work.

Thanks, in that case I'm pretty sure we shouldn't export the class because it's not real runnable code.

```
  declare class ConnectedComponentClass<OP, +WC> extends React$Component<OP> {
    static +WrappedComponent: WC;
    getWrappedInstance(): React$ElementRef<WC>;
  }

  declare export type ConnectedComponent = typeof ConnectedComponentClass;
```

Just a suggestion how about instead, we rename `ConnectedComponent` to `ConnectedComponentClass` which shouldn't have impacts because it's not exposed and then we export a new type called `ConnectedComponent` which is `typeof ConnectedComponentClass`? By doing this rename we can retain the naming scheme just like TS.

Brianzchen (approved):

grit96:
I realise now exporting the type like this doesn't quite work the same as exporting the class because the type is not polymorphic.

Can I suggest doing the following to preserve the polymorphism?

Replace this line
`declare export type ConnectedComponent = typeof ConnectedComponentClass;`
with
`declare export type ConnectedComponent<OP, +WC> = ConnectedComponentClass<OP, WC>;`

Brianzchen:
For me this is fine. We added this to suit your use case but I'd strongly recommend this time we add a test to showcase the usage along with this change

PR #4210, opened 2021-12-29
rxjs: Update types for rxjs 7.x.

djcsdy:
- Links to documentation: https://rxjs.dev/6-to-7-change-summary
- Link to GitHub or NPM: https://github.com/ReactiveX/rxjs
- Type of contribution: new definition (update for new library version)

I have tested these types in my own project. However I have not run the tests in flow-typed because when I follow the instructions to do so, I get the message "No libdef file found!".


Brianzchen:
@djcsdy Thanks for the PR. It's unfortunate that tests couldn't run locally, now that you've done a push if you try `./quick_run_def_tests.sh` is it able to run? Alternatively you could run `node cli/dist/cli.js run-tests rxjs` once you have a locally built cli.

But looks like there are tests that need fixing, rxjs v6 didn't have any errors throwing so your additions may have caused that. Also I imagine you copied the initial definitions from v6 and then made some tweaks, it would be great if those could have been split into two commits so reviewers can see the copy process and then see the addition otherwise we really need to go through 80+ with an also understanding of rxjs before this can be merged.

djcsdy:
I'll force push to split the commits unless you tell me not to.

I ran `./build.sh && node cli/dist/cli.js run-tests rxjs`. The build runs ok, but the second command fails with "No libdef file found!". This is on Ubuntu/WSL with the code checked out on the Linux filesystem. I don't think I'm doing anything wrong, but maybe...

Brianzchen:
> I'll force push to split the commits unless you tell me not to.

If you can that would be REALLY appreciated. Otherwise I'm basically merging and trusting you it works, which I no doubt believe it works for you but there just needs to be a process in case in the future someone else uses this and complains something doesn't work and it's just a blank void as to what work you did.

> I ran `./build.sh && node cli/dist/cli.js run-tests rxjs`. The build runs ok, but the second command fails with "No libdef file found!". This is on Ubuntu/WSL with the code checked out on the Linux filesystem. I don't think I'm doing anything wrong, but maybe...

ok thanks. Maybe this is a windows problem since i'm on macos I can't test it myself. I'll investigate the code and hopefully I'll get you to try somethings for me but I suppose until then you'll have to reference back to the github builds to see what your errors are because those still need fixing

Brianzchen:
> "No libdef file found!"

Ok! After digging into this area I can confirm it's probably not a flow-typed issue but something on your end. This error is throwing basically because somewhere along your journey you've created a package folder and then a flow version folder, but inside the flow version folder there is no corresponding libdef file, eg if you have `rxjs_v1.x.x` package folder, later on you must also have a `rxjs_v1.x.x.js` file. Because it can't be found it's throwing. I imagine there's probably a rogue folder sitting somewhere without any files and because it's not checked in it only throwing errors on your local.

What I can do is make a quick fix so that you get better logs so you don't need to scroll through almost a thousand libdefs

djcsdy:
Thanks. I still had an empty redux-observable_v1.2.x folder left over from some previous work.

djcsdy:
I force pushed to split into two commits. Please forgive the extra force-pushes. It seems I am half asleep and somehow managed to get the commits in reverse order on the first attempt :-).

Fixes for the tests to follow but probably not today as it's almost beer o'clock. Happy New Year! :-).

djcsdy:
Fixed tests.

Brianzchen (approved):
LGTM Will give a couple days to see if there's other feedback

PR #4211, opened 2021-12-29
@sentry/browser root captureException & captureMessage arg types update

emcelyea:
captureMessage & captureException exist on the root of Sentry and also at the Hub & Scope level. We currently type all instances as if they were Hub.captureException instead of Sentry.captureException.

Hub captureException docs:
https://getsentry.github.io/sentry-javascript/interfaces/types.hub-1.html#captureexception

root captureException docs:
https://getsentry.github.io/sentry-javascript/modules/minimal.html#captureexception

This updates the root captureMessage/Exception arguments:


Brianzchen:
Looks like some tests have failed, they could be fixed but also some additional tests to secure your work would be great!

PR #4213, opened 2021-12-30
[cli] Provide better logs related to dir structure when running tests

Brianzchen:
In https://github.com/flow-typed/flow-typed/pull/4210 a contributor was unable to run tests when create a new lib def because they may have made a mistake when initially trying to create the dir structure. This throws an error but it's not one that's descriptive enough to help them pinpoint where to go to solve the issue.

Other notes, when I was building this fix I opened the `definitions/npm` dir in my finder and then it started throwing a lot of errors related to `DS_Store` I think instead of throwing we should just exit in that case so I've added that also.

### DIFF
--- a/cli/src/lib/libDefs.js
+++ b/cli/src/lib/libDefs.js
@@ -168,6 +168,8 @@ export async function getLibDefs(defsDir: string): Promise<Array<LibDef>> {
   const defsDirItems = await fs.readdir(defsDir);
   await P.all(
     defsDirItems.map(async item => {
+      if (item === '.DS_Store') return;
+
       const itemPath = path.join(defsDir, item);
       const itemStat = await fs.stat(itemPath);
       if (itemStat.isDirectory()) {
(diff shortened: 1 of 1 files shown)


cli/src/lib/libDefs.js
```diff
@@ -168,6 +168,8 @@ export async function getLibDefs(defsDir: string): Promise<Array<LibDef>> {
   const defsDirItems = await fs.readdir(defsDir);
   await P.all(
     defsDirItems.map(async item => {
+      if (item === '.DS_Store') return;
```
GAntoine, line 171: Could we instead get rid of the DS_Store files and add a test to make sure they don't come back?

Brianzchen: How do we do that? Mine randomly appeared because I opened my finder, it's not like you can gitignore it because the issue is when we run it locally and because it creates a `DS_Store` in every single dir means that when cleaning up I was trying to go into every `definitions/npm/@scope` to delete this.

Unless you're saying when we run this loop we should delete it instead of returning? Not sure which approach is best here, I imagine deleting isn't good because `DS_Store` is used to retain folder positioning if someone does open the repo in their finder which has some usage.

GAntoine: Right yeah makes sense, I see how it can be an issue when running it locally. 👍🏼

GAntoine (approved):

cli/src/lib/libDefs.js
Brianzchen: Actually I can see how this code makes no sense when reading this later. I'll add a comment to describe why this issue needs to be exited before I merge

PR #4216, opened 2022-01-02
[ci/cd] Post eslint warnings for defintions

### LINKED ISSUE #49: [Test CI] Automatically comment on PRs that add `any` types

Brianzchen:
Closes #49

With the introduction of github actions, eslint will automatically add inline errors to the PR to give really good visual feedback for when a change has linting errors


Brianzchen:
node_setup which automatically annotates github prs doesn't annotate warnings, so looking into that

Brianzchen:
@GAntoine I'm quite close to getting this action to work as I expect but I'm missing a key part where the `GITHUB_TOKEN` doesn't have permission to submit annotations, you can see the error [here](https://github.com/flow-typed/flow-typed/runs/4733011475?check_suite_focus=true#step:7:9) could we enable it to see if I'm able to get it working?

Brianzchen:
Cannot write checks using annotate tool because all forked public repos can only have read permission
ref: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

GAntoine:
@Brianzchen Can you create a branch in this repo directly and try again? We can delete the branch after it's been merged.

Brianzchen:
This will be handled by flow-typed-pr-bot instead

PR #4223, opened 2022-01-05
Add missing trailing comma from VerifyOptions type in jsonwebtoken definition

gthouret:
This fix is for a definition that was added earlier today and breaks our production builds as we are using flow-bin < 0.103.

Your automated tests only cover the last 15 flow-bin versions published to npm.


Brianzchen (approved):

Brianzchen:
So sorry, very careless of me and the tests didn't run in old versions. But thanks for fixing this up for me!

PR #4230, opened 2022-01-10
Update humps v2.x.x camelizeKeys function definition

bruno-b-martins:
Co-authored-by: David Coelho <davidcoelho02@outlook.com>

- Links to documentation: https://github.com/domchristie/humps/tree/050b1fd852cf0847bed9513655b2afbb20d89859#converting-object-keys
- Link to GitHub or NPM: https://github.com/domchristie/humps
- Type of contribution: fix

Other notes:

We were working with hump's `camelizeKeys` and hitting an error due to its definition.
We got it to work properly with this change, but feel free to tell us if we should have done this differently.

Thanks for your review 🙇

### DIFF
--- a/definitions/npm/humps_v2.x.x/flow_v0.104.x-/humps_v2.x.x.js
+++ b/definitions/npm/humps_v2.x.x/flow_v0.104.x-/humps_v2.x.x.js
@@ -18,9 +18,9 @@ declare module "humps" {
   declare export function depascalize(string, options?: Options): string;
 
   declare export function camelizeKeys(
-    Object | Array<Object>,
+    { +[string]: mixed, ... },
     options?: Options
-  ): Object;
+  ): { [string]: mixed, ... };
   declare export function pascalizeKeys(
     { +[string]: mixed, ... },
     options?: Options
@@ -33,6 +33,11 @@ declare module "humps" {
     { +[string]: mixed, ... },
     options?: Options
   ): { [string]: mixed, ... };
+
+  declare export function camelizeKeys(
+    Array<{ +[string]: mixed, ... }>,
+    options?: Options
+  ): Array<{ [string]: mixed, ... }>;
 
   declare export default {|
     camelize: typeof camelize,
--- a/definitions/npm/humps_v2.x.x/flow_v0.104.x-/humps_v2.x.x.js
+++ b/definitions/npm/humps_v2.x.x/flow_v0.104.x-/humps_v2.x.x.js
@@ -18,9 +18,9 @@ declare module "humps" {
   declare export function depascalize(string, options?: Options): string;
 
   declare export function camelizeKeys(
-    { +[string]: mixed, ... },
+    Object | Array<Object>,
     options?: Options
-  ): { [string]: mixed, ... };
+  ): Object;
   declare export function pascalizeKeys(
     { +[string]: mixed, ... },
     options?: Options
--- a/definitions/npm/humps_v2.x.x/flow_v0.104.x-/test_humps_v2.x.x.js
+++ b/definitions/npm/humps_v2.x.x/flow_v0.104.x-/test_humps_v2.x.x.js
@@ -31,5 +31,8 @@ describe("tests", () => {
     (pascalizeKeys({ foo: 1 }): { [string]: mixed, ... });
     (decamelizeKeys({ foo: 1 }): { [string]: mixed, ... });
     (depascalizeKeys({ foo: 1 }): { [string]: mixed, ... });
+
+    (camelizeKeys([{ foo: 1 }]): Array<{ [string]: mixed, ... }>);
+    (camelizeKeys([{ foo: "bar" }]): Array<{ [string]: mixed, ... }>);
   });
 });


Brianzchen:
Thanks for submitting this change, it looks like `camelizeKeys` should accept array but previously didn't. Are the errors you're getting only happening if you try an pass an array or also as a plain object? I'm asking because you've changed `{ +[string]: mixed, ... }` to `Object` which we really shouldn't because `Object` is just an alias for `any` meaning you've said `camelizeKeys` takes `any` or `Array<any>`.

Side note you're getting some minor build errors because of flow suppression codes, do you mind fixing those while you're in the area 🙏

bruno-b-martins:
Thanks for having a look @Brianzchen .

`camelizeKeys` accepts either an object or an array like:
```javascript
var object = { attr_one: 'foo', attr_two: 'bar' }
humps.camelizeKeys(object); // { attrOne: 'foo', attrTwo: 'bar' }
```

```javascript
var array = [{ attr_one: 'foo' }, { attr_one: 'bar' }]
humps.camelizeKeys(array); // [{ attrOne: 'foo' }, { attrOne: 'bar' }]
```

What would be a proper rule for asserting we're passing the right stuff then?

Brianzchen:
[here](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/humps_v2.x.x/flow_v0.104.x-/test_humps_v2.x.x.js#L30) is the test case showing it working with objects, you can add a test for arrays also to verify and then your change should look something like this

```
declare export function camelizeKeys(
    { +[string]: mixed, ... },
    options?: Options
  ): { [string]: mixed, ... };
  declare export function camelizeKeys(
    Array<{ +[string]: mixed, ... }>,
    options?: Options
  ): Array<{ +[string]: mixed, ... }>;
```

This is basically like method overloading to allow multiple inputs.

bruno-b-martins:
Thank you very much for taking the time to explain us this.

I've updated the definition but I'm still getting an error on our project:
```javascript
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/_common/components/Component.js:115:23

Cannot call map because mixed [1] is incompatible with string [2] in property activeEmail of array element.
[incompatible-call]

     src/_common/components/Component.js
      99| ...
     100|    suggestedUsers = camelizeKeys(filteredUsers);
     114| ...
     115│    suggestionsList = map(suggestedUsers, (user: User) => {
     116| ...

     flow-typed/npm/humps_v2.x.x.js
 [1]  43│   ): Array<{ [string]: mixed, ... }>;

     flow-typed/project_vx.x.x.js
 [2]  32│       activeEmail: string,
```

```javascript
// flow-typed/project_vx.x.x.js
...
declare export type User = {
    activeEmail: string,
    ...
};
...
```

Can you help us understand why `mixed` is incompatible with `string`?

We noticed that if we define the return of `camilizeKeys` as `Array<{ [string]: any, ... }>`, it works 😅 but we're avoiding using `any` according to your guidelines.

Brianzchen:
So you problem is related to how it's being used after you've run the function which returns values as mixed. In a strict sense mixed is good because after you transform `hi_there` to `hiThere` if you mistype your code and write `hiTher` it would be mixed and you can't use it immediately, you're forced to first [type refine](https://flow.org/en/docs/lang/refinements/) it to ensure that yes this is a string and now I can use it. It creates more predictable code when flow of course can't figure out that some property using snake case would have the same type as a property with camel casing. 

When you previously had it as `Object` (which is basically `any`) or `{ [string]: any }` because those are more high level type structures they're castable into your object type of `User`.

The way to use this properly in it's mixed form would be,
```
suggestedUsers = camelizeKeys(filteredUsers);
// ...
suggestionsList = map(suggestedUsers, (user) => { // I'm not casting it here
  const {
    activeEmail,
  } = user;

  if (typeof activeEmail === 'string') {
    // do whatever you want
  }

  // or ----------------
  if (typeof user.activeEmail === 'string') {
    // do whatever you want
  }
});
```

Flow's generally a lot better when you rely on the power of inference to protect you from runtime errors while using annotations to create boundaries where if two types mismatch to then throw errors. When you `camilize` your `filteredUsers` var it won't necessarily be typeof `User` so makes sense to throw an error.

---

Now the other side is to say hey it's not a big deal and changing this libdef's value type from `mixed` to `any` would be lot better which could be a counter argument that I'll leave here if you want to fight for it 😄

bruno-b-martins:
Thank you once again for taking the time to explain this.

To be honest, I have "mixed" feelings about this 😅 
In one hand, I like that we're casting our `suggestedUsers` to `User`, so I'd like to keep that behavior.
On the other hand, I don't want to make the libdef too permissive by changing `mixed` to `any`. But I do think this is the best approach...

- - - - -

About the "flow suppression codes" on the build that is failing. What exactly can I do about them?
I'm new to this project and, therefore, don't know how to proceed here. Can you point me in the right direction?

Brianzchen:
> About the "flow suppression codes" on the build that is failing. What exactly can I do about them?
I'm new to this project and, therefore, don't know how to proceed here. Can you point me in the right direction?

Have a read of this https://flow.org/en/docs/errors/

It's failing on [this](https://github.com/flow-typed/flow-typed/pull/4230/files#diff-80b75c70e7156c2ec4762e2c536cbd48350ef5de3e56eb9a4fea9e6959e88aefR18) line. Basically after a certain version flow recommended adding suppression error codes and as people maintain definitions it's good to update them along the way if it's not too much trouble. You basically need to update the line to `// $FlowExpectedError[incompatible-call]` judging by your test failure

> I don't want to make the libdef too permissive by changing mixed to any

Yep agree, I'd say if I made this libdef from scratch I would have chosen `any` but because someone else made this there must've been a thoughtful reason so I'm reluctant to change it if that works out for you. If anything flow's intention makes you write more code for stability so it's sure doing that job.

PR #4249, opened 2022-02-02
[cli] Turn off validate-def checks against npm

Brianzchen:
Hear me out, this is the beginning of a series of changes. After having a read of #87 and then testing `validate-def` I found that the testing against npmjs doesn't actually work. It uses `got` to ping npmjs.com but because it's hitting a website even if the website says 404 the website itself it a 200 response. Furthermore hitting a website like crazy reaches a `HTTPError: Response code 429 (Too Many Requests)` error. I've changed this to use `api.npms.io/v2/` instead which can support the load and will reliably return with whether the request succeeds or fails.

But the code using a mix of then/catch + async/await doesn't actually work properly, even if the `got` reaches a 404 it still ends saying `All libdefs are named and structured correctly`.

Finally there is one library called `core.es5` that was added 2 years ago and isn't actually an npm library, because of this it is meant to fail but it's perfectly valid given the reason it's there but it's not in the right place (`definitions/npm`) and as a result I cannot have the npmjs check working until we've shifted the definition somewhere else (I'm thinking `definitions/core` but I'll raise a discussion around this first). 

## What this PR does
Turn off the npmjs part of the validation to improve speed as we improve other areas to allow it turn back on.

### DIFF
--- a/cli/src/lib/npm/npmLibDefs.js
+++ b/cli/src/lib/npm/npmLibDefs.js
@@ -1,5 +1,4 @@
 // @flow
-
 import {
   ensureCacheRepo,
   getCacheRepoDir,
@@ -31,7 +30,7 @@ import {
 
 import semver from 'semver';
 
-import got from 'got';
+// import got from 'got';
 
 import {ValidationError} from '../ValidationError';
 import {TEST_FILE_NAME_RE} from '../libDefs';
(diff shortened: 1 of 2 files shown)


AndrewSouthpaw (approved):
LGTM, thanks for the context!

GAntoine (approved):
Yeah seems fine to me, this was originally a change I proposed waaay back in the day, and the implementation is lacking. Glad to see you're working on improving it. 👍🏼

Brianzchen:
https://github.com/flow-typed/flow-typed/discussions/4250 discussion on core defs here. We can catchup later to discuss whether core defs is something worth doing after an initial read.

PR #4255, opened 2022-02-08
[cli] Core definitions

Brianzchen:
Implements #4250

## Test Plan
- [x] runTests work to find the initially added core defs
- [x] when `ft-config.json` exists in `libdefDir` (commonly `flow-typed`) it will install array of env
- [x] `installs` into `flow-typed/core` dir with code signature
- [x] `update` command works
- [x] outdated command will output if it's outdated
- [x] install command will not update if core def has been modified
- [x] docs updated with core definitions
- [x] contributing docs updated with core defs
- [x] test outdated
- [x] test install

## Sample usage

1. Create a`flow-typed.config.json` inside project root
Screen Shot 2022-02-20 at 6 40 03 pm
2. Run `flow-typed install`

**Installing fresh or with overwrite or updating**
Screen Shot 2022-02-20 at 6 43 49 pm
**If it already exists and modified**
Screen Shot 2022-02-20 at 6 45 06 pm
**If a env is referenced that does not exist or not compatible with flow version**
Screen Shot 2022-02-20 at 6 46 23 pm
3. Run `flow-typed outdated`

**If there has been an update in the registry**
Screen Shot 2022-02-20 at 6 48 09 pm
**If nothing compatible**
Screen Shot 2022-02-20 at 6 49 29 pm


Brianzchen:
I'm thinking that a solution like this is necessary regardless to solve some of flow's problems with missing core definitions and my use case is centred around `jsx` usage in particular. 

I wonder how a library that ships flow or a libdef could take advantage of core definitions. It would be very useful if a libdef of a component library was able to type it's `Input` component with all the props that are natively supported by html and it would make sense if we said that a library like `material-ui` must run on a `jsx` environment.

This seems duplicative of the base definitions solution though with the major difference being that it doesn't import changes and I feel like they could co-exist solving different problems but equally could be consolidated to keep one solution for solving multiple problems.

Brianzchen:
Ok after having a day to think this through here are my thoughts.

Imo an environment served as a core definition and base definition are two different things, one runs globally the other is imported, redux types shouldn't be exposed globally as they are interfaced through a library for example and core definitions are a means to add more types that should be available as part of the tool depending on the application usage.

To support the lib defs and flow shipped libraries respectively,

I believe the first can be supported with libdefs owning a package.json just like for base definitions. If a libdef needs a core def installing it shouldn't cause conflict and won't overwrite any other core def since they are only versioned by flow version, we can pull all installed libdef's core defs and flatten the array with `env` var in `ft-config`.

The second I believe automation is not necessary and a requirement on a libraries readme should be sufficient imo.

Brianzchen:
I might rename everything to `env` instead of `core`. It just doesn't make sense during usage

definitions/core/es5-1/README.md
```diff
@@ -0,0 +1,9 @@
+An alternative JS core library definition for ES5.1
+
+The Flow core library definitions cover the latest ES versions. That has a lot of additional methods and classes compared to ES5 and so code-completion and type checking do not give errors about using those new methods and classes. Use this core library definition to write pure ES5.1 Javascript without having to add pollyfills or compile down to ES5 via Babel.
+
+Requires the `no_flowlib` option.
```
AndrewSouthpaw, line 5: I'm not sure this would be clear to someone unfamiliar with Flow, maybe clarify which option this is referring to, or link to the docs?

definitions/core/jsx/flow_v0.83.x-/jsx.js
```diff
@@ -0,0 +1,84 @@
+declare type jsx$HTMLElementProps = {|
```
AndrewSouthpaw, line 1: Can you link to the source here?

docs/core-definitions.md
```diff
@@ -0,0 +1,37 @@
...
+
+Flow is a static analysis tool that exposes a range of global definitions to help you type your application. But flow only supports for the most part, two types of environments, node and browser and sometimes you want something a little extra that flow just doesn't support and isn't quite right to have flow support as a core system.
+
+That's where core definitions from flow-typed come into play, a home for all those global definitions to share and reuse and standardize for specific environments.
+
+To make it more concrete, lets take a look at some examples of when you might want to create to leverage a core definition.
+
+### jsx
+
+Flow is built by Meta and ships with react definitions it does not type or provide a list of available jsx properties. These can live as a core definition instead.
```
AndrewSouthpaw, line 11: This sentence reads a little unclearly, try tweaking?

docs/ft-config.md
```diff
@@ -0,0 +1,17 @@
+# ft-config.json
```
AndrewSouthpaw, line 1: How about `.flow-typed-config.json`? I'm not sure `ft` would be meaningful to most people. I don't feel strongly here, but it even took me a second to realize `ft` mean "flow-typed".

docs/core-definitions.md
AndrewSouthpaw: I think it'd be worth clarify what we are lacking that we get when adding `jsx` as an env. Most people use Flow with React and think everything is fine, so it might be unclear why it's valuable to add it.

AndrewSouthpaw (approved):
Really excited to see this land, thanks for your hard work @Brianzchen!

I don't dig much into the implementation, overall it looks good, and the tests seem reasonable. I mostly focused on the docs and overall implementation strategy.

definitions/core/es5-1/README.md
Brianzchen: Actually we should ignore everything inside `core/es5-1` that just copied from an existing definition that was in `definition/npm` but isn't actually a npm package

docs/ft-config.md
AndrewSouthpaw: Sounds perfect.

Brianzchen:
@GAntoine 🙏

AndrewSouthpaw:
This looks fine I'll go ahead and merge, everything is green.

GAntoine (approved):
Sorry it took so long! I'm really happy with this approach. 👍🏼 

Before we merge this in, I'd like to do another minor deploy (if we have any pending cli changes).

And once it's merged and deployed, nothing will change for users until we write the first core definition right? In which case we can plan and deprecate accordingly.

cli/src/commands/outdated.js
```diff
@@ -181,6 +183,67 @@ export async function run(args: Args): Promise<number> {
     }),
   );
 
+  const ftConfig = getFtConfig(cwd);
+  const {env} = ftConfig ?? {};
+
+  if (Array.isArray(env)) {
```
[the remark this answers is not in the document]
Brianzchen: Correct, `Array.isArray(env)` evaluates to `false`. I only wrote it in this way with spread for when we extend config file in the future

Brianzchen:
> And once it's merged and deployed, nothing will change for users until we write the first core definition right? In which case we can plan and deprecate accordingly.

Nothing will change for users until we ship a new version which is also non-breaking. No users will get this new feature unless they also implement a config in their project also. 

In terms of deprecation it's just that with `es5-1`, that we should remove from `npm` after we publish the latest version.

This PR also includes the beginnings of the first real env def I plan to use extensively though I didn't want this PR that builds the CLI to have that reviewed so I added a minimal version just to prove it's functionality. 

---

Feel free to merge @GAntoine when you're ready but it looks like we haven't done any more work on CLI since last 3.7.0 release.

PR #4268, opened 2022-02-19
[cli] Create test harness

### LINKED ISSUE #4267: Repo test harness
### CLI Version

3.6.1

So I was talking to @meandmax today and he was telling me about how contributing definitions is hard or it's not easy to get into. One problem is that the `definitions` isn't actually running flow and it basically can't, if you do you'll get millions of errors because some defs may not be maintained or be incompatible with the current version running.

This ends up without a good place to build definitions while you're working in this repo. This poses problems for a contributor who wants to build a type definition not necessarily tied to a project and wanting quick type checking feedback from their IDE

Create a test harness dir in the root directory that can be quickly spun up to run any flow version of choice and build the definition with tests which can then be easily transferred into `definitions` to run the full test suite and submit.

Whatever custom settings used for the test harness should be git ignored so that it doesn't cause random changes during commits.

Two options I can think of
1. root command that runs to create a new dir and group of files like a basic flow config, package.json, base file, test file, etc
2. There being an existing test harness dir that is preconfigured to run latest flow version but can be swapped to run a different flow version whenever we want

I'm impartial to either but I think solution 1 can be easier and should be achievable with a simple bash script.

### Anything else?

Should also add some docs as part of this change

Yes

Brianzchen:
Closes #4267

Run a sh script that will create the test harness for faster local development of definitions

# Examples

## Running it to create harness
ezgif-1-5e419daa8a

## Running with a specific flow version
ezgif-1-ae01255ff5

## Coding in the harness
ezgif-1-84b09c0c1b

## Test harness file structure
Screen Shot 2022-02-19 at 9 14 15 pm

Other notes:
@meandmax @GAntoine @AndrewSouthpaw

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,9 @@ cli/.flow-bins-cache/*
 cli/package-lock.json
 cli/README.md
 
+# Test harness
+harness/
+
 # Editors
 .vscode
 .idea
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,6 +88,8 @@ and now you'd like to contribute it.
 >
 > This will cover steps 1 - 4, of course without real definitions or tests written for you.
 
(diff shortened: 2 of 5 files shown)


GAntoine (approved):
💯

meandmax:
LGTM, haven’t tested it yet but will do soon. 👌

Issue #4272, opened 2022-02-22
React Router hooks generic type support


bottd:
Hello, I am trying to work with the flow types for react-router v5. I am running into issues with the typings for `useParams`. I believe `useParams` needs support for generics in order to support accurate typings. Currently, the type for `useParams` is:
```js
declare export function useParams(): $PropertyType<$PropertyType<ContextRouter, 'match'>, 'params'>;
```
This results in the following return type:
```
{ [key: string]: ?string, ... },
```
This causes issues because we cannot define what the keys should be within 'params'. Due to how `useParams` is used, you can guarantee that a key will not be `null` or `undefined`. Due to these types I end up with errors because my key that is always string is seen as possibly `null` or `undefined`. 

The typings within [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-router/index.d.ts) allow for providing a type for params:
```
export function useParams<Params extends { [K in keyof Params]?: string } = {}>(): Params;
```

If it is okay to make this update I would be happy to make the PR happen.

Brianzchen:
Hi @bottd that sounds great, I've personally used this and found it quite annoying too (I tended to just move on with my life as opposed to improving it 😅 ). But I agree params for url slugs makes sense to be populated if consumer is sure of the route passed in.

If you're happy to make the change yourself please go ahead and make the changes remembering to add some tests to verify this usage as well as if a user opts not to use generics.

Let me know if you need any help along the way 🚀

bottd:
@Brianzchen thanks for the help! I opened a PR for this, let me know if you have any thoughts on my proposed change.

PR #4273, opened 2022-02-22
[react-router] Update type for useParams to allow for generics

bottd:
I made an update to the `useParams` types after opening #4272. I tried to add a bound to the type to require that the provided type has string values only, but ran into issues with making that work with the default type that is provided since it has the values as `?string`. If there is a quick fix for this I'd love to add it, but these types would work for me!


Brianzchen (approved):
I'm pretty happy with this. As long as tests pass I'm ok to merge

bottd:
🎉 thanks!

PR #4274, opened 2022-02-23
[react-router-dom] Update type for useParams to allow for generics

bottd:
Hello, sorry for the near duplicate PR. I realized this morning I only updated the type in `react-router` but forgot `react-router-dom` 😅. This is the same changes from #4273 applied to the `react-router-dom` types.


Brianzchen (approved):
At least you helped to maintain both libs then 👍

PR #4286, opened 2022-03-17
[base64-arraybuffer] Add base64-arraybuffer v1

ngyikp:
- Links to documentation: https://github.com/niklasvh/base64-arraybuffer/blob/5339d8f0abed59c893e87116a785bf52acdc17aa/src/index.ts
- Link to GitHub or NPM: https://github.com/niklasvh/base64-arraybuffer

### DIFF
--- a/definitions/npm/base64-arraybuffer_v1.x.x/flow_v0.83.x-/base64-arraybuffer_v1.x.x.js
+++ b/definitions/npm/base64-arraybuffer_v1.x.x/flow_v0.83.x-/base64-arraybuffer_v1.x.x.js
@@ -1,0 +1,5 @@
+declare module 'base64-arraybuffer' {
+  declare function encode(arraybuffer: ArrayBuffer): string;
+
+  declare function decode(base64: string): ArrayBuffer;
+}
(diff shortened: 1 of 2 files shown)


definitions/npm/base64-arraybuffer_v1.x.x/flow_v0.83.x-/test_base64-arraybuffer_v1.x.x.js
```diff
@@ -0,0 +1,20 @@
+import { describe, it } from 'flow-typed-test';
```
Brianzchen, line 1: I'm pretty sure you need to add this otherwise it doesn't typecheck
```suggestion
// @flow
import { describe, it } from 'flow-typed-test';
```

definitions/npm/base64-arraybuffer_v1.x.x/flow_v0.83.x-/base64-arraybuffer_v1.x.x.js
```diff
@@ -0,0 +1,5 @@
+declare module 'base64-arraybuffer' {
+  declare function encode(arraybuffer: ArrayBuffer): string;
+
+  declare function decode(base64: string): ArrayBuffer;
```
Brianzchen, line 4: I don't think these should be usable unless you decided to export them
```suggestion
  declare export function encode(arraybuffer: ArrayBuffer): string;

  declare export function decode(base64: string): ArrayBuffer;
```

Brianzchen (changes requested):
At first glance this feels off. Can you let me know if I've found a bug. I haven't tested it personally though so I'm happy to be wrong too :)

Brianzchen (approved):
Apologies. I just tested this and it works. I'm very surprised it does but it's good enough to ship imo as it's well tested

PR #4290, opened 2022-03-29
[react] Add new environment to override react definitions

Brianzchen:
React components actually allow for extra properties in the react component once defined, there are many real world cases where this occurs but flow libs don't allow for this.

I'm adding a new flow-typed `environment` that overrides right now a single piece of the react definitions that doesn't require users to upgrade to the latest flow version to take advantage of this. I believe def can have future use cases where commonly flow libs are not backwards compatible.

- Links to documentation: N/A
- Link to GitHub or NPM: https://www.npmjs.com/package/react

Other notes: This use case came from @meandmax when we were pairing and found that flow blocked adding extra properties to a higher order component.


GAntoine (approved):

AndrewSouthpaw (approved):
LGTM!

PR #4295, opened 2022-04-02
feat: add bn.js libdefs (v2, v3, v4, v5)

Sekhmet:
- Links to documentation:
  - [v5](https://github.com/indutny/bn.js/blob/v5.2.0/README.md)
  - [v4](https://github.com/indutny/bn.js/blob/v4.12.0/README.md)
  - [v3](https://github.com/indutny/bn.js/blob/v3.3.0/README.md)
  - [v2](https://github.com/indutny/bn.js/blob/v2.2.0/README.md)
- Link to GitHub or NPM: [GitHub](https://github.com/indutny/bn.js)

Other notes:
- v5 and v4 are exactly the same
- v3 is missing some util functions compared to v4
- v2 is missing unsigned operations

### DIFF
--- a/definitions/npm/bn.js_v2.x.x/flow_v0.83.x-/bn.js_v2.x.x.js
+++ b/definitions/npm/bn.js_v2.x.x/flow_v0.83.x-/bn.js_v2.x.x.js
@@ -1,0 +1,126 @@
+declare module 'bn.js' {
+  declare type Endianness = 'le' | 'be';
+  declare type Primes = 'k256' | 'p224' | 'p192' | 'p25519';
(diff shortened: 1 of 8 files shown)


Brianzchen:
Thank you and really awesome effort! I will review this tonight, but at a first glance, do you think you could add a couple negative tests here and there to verify that it doesn't accept what isn't expected? Just in case something in the future is faulty and accepts any we can catch that.

Point `3.` [here](https://flow-typed.github.io/flow-typed/#/contributing?id=_4-write-a-test-file-whose-name-starts-with-test_-in-the-flow-version-directory) describes that.

PR #4301, opened 2022-04-15
Reactdom18createroot

noahehall:
- Links to documentation: 
  - [createRoot annotation](https://github.com/facebook/react/blob/726ba80298dad46b331aca1bcb71ca9b2004ed91/packages/react-dom/client.js#L25)
  -  [createRootOptions annotation](https://github.com/facebook/react/blob/726ba80298dad46b331aca1bcb71ca9b2004ed91/packages/react-dom/src/client/ReactDOMRoot.js#L26)
- Link to GitHub or NPM: https://github.com/facebook/react/blob/main/packages/react-dom/
- Type of contribution:  update existing

Other notes:

hopefully we can squash and merge the arbitrary commits so i dont have to manually drop them


definitions/npm/react-dom_v18.x.x/flow_v0.127.x-/react-dom_v18.x.x.js
```diff
@@ -22,6 +22,27 @@ declare module 'react-dom' {
     container: Element
   ): React$Portal;
 
+  declare type CreateRootOptions = {
```
Brianzchen, line 25: Doesn't this need to be in `react-dom/client`? Docs say so [here](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis)

Brianzchen:
Hey I'm pretty happy if you wanna do the react-18 changes cause honestly I have other things going on. But probably a couple things here,

- Some tests would be great
- The large type properties you've just replaced with `any` just want to check was that intentional so that this was easier to build, I'm not sure how impactful that would be? Personally I don't think I'd ever use those properties anyways though how type accurate do we wanna be?
- And mentioned above but I thought `createRoot` only comes from the new `react-dom/client`

Brianzchen:
I've just added `createRoot` to `react-dom/client`. It looks like you're right that it does get exported from the main module, do you want to make a change so that we reuse and it imports def from `react-dom/client` and exports in `react-dom`?

PR #4309, opened 2022-04-30
[react-dom] Export createRoot, hydrateRoot from main index

Brianzchen:
The new functions `createRoot` and `hydrateRoot` are actually available in `react-dom` but throw errors when called from `react-dom` instead of `react-dom/client`. Deprecation JS docs have been added suggest against usage though usage is completely valid from a soundness perspective.

This change was originally proposed by @noahehall in #4301 and I'm simply completing it with proper test verification.

- Links to documentation: https://github.com/facebook/react/blob/main/packages/react-dom/index.js#L15
- Link to GitHub or NPM: https://www.npmjs.com/package/react-dom
- Type of contribution: fix

Other notes: Refactor react-dom tests to match describe/it format

### DIFF
--- a/definitions/npm/react-dom_v18.x.x/flow_v0.127.x-/react-dom_v18.x.x.js
+++ b/definitions/npm/react-dom_v18.x.x/flow_v0.127.x-/react-dom_v18.x.x.js
@@ -99,9 +99,19 @@ declare module 'react-dom_shared-types' {
     _currentlyRenderingFiber?: any,
     _initialVersionAsOfFirstRender?: MutableSourceVersion | null,
   |};
+
+  declare opaque type FiberRoot;
 }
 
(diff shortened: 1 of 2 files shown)


pascalduez:
**question:** What's the point of adding support for something that wasn't implemented (so no breaking change nor deprecation needed), and should not be used? It seems it's only available on the main export for internal (tests) needs.  
A user migrating to React 18 should see only references to the proper imports in docs.

Brianzchen:
Fair, I only added it because looked like @noahehall wanted this and it's a completely sound usage despite not being proper from the maintainers of react. I think I'll leave this PR open unless there's overwhelming requirement to merge

Brianzchen:
Will not be implemented

Issue #4307, opened 2022-04-28
RXJS v6.6.7


fedgrant:
### Library definition file name

rxjs_v6.x.x.js

### Your dependent version

6.6.7

### Flow Version

0.142.0

### Description

I believe there is a duplicate type
```
declare export function multicast<T>(
    subjectOrSubjectFactory: rxjs$FactoryOrValue<rxjs$Subject<T>>
): rxjs$UnaryFunction<rxjs$Observable<T>, rxjs$ConnectableObservable<T>>;
  
declare type rxjs$FactoryOrValue<T> = T | (() => T);
```

```
declare export function multicast<T>(
    rxjs$SubjectFactory: () => rxjs$Subject<T>
): rxjs$UnaryFunction<rxjs$Observable<T>, rxjs$ConnectableObservable<T>>;
```

link for reference [  declare export function multicast<T>(](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/rxjs_v6.x.x/flow_v0.104.x-/rxjs_v6.x.x.js#L2525)

### Do you want to submit a pull request to fix this issue?

Yes

fedgrant:
I can submit a PR if this is actually a bug, I'm not 100% that it is.

Brianzchen:
Hey thanks for your submission, though I'm not super familiar with rxjs I believe the duplicates are overloaded methods for the same function so that they can be used in a multitude of ways. Or are you suggesting that one/some of these definitions are exactly the same and therefore we should clean some of them up? 

### Screenshot of overload in action
<img width="960" alt="Screen Shot 2022-04-29 at 7 24 41 am" src="https://user-images.githubusercontent.com/12436524/165848436-e0a9743e-2da6-49a5-aa56-b35633a4aa18.png">

fedgrant:
I think they are the same.

`rxjs$FactoryOrValue<rxjs$Subject<T>>` would be  `rxjs$Subject<T> | (() => rxjs$Subject<T>);`

which is the same as `() => rxjs$Subject<T>` in the second case

Brianzchen:
Yes as I keep staring at this they are the same as far as I can tell. I'm ok if you'd like to remove them. As long as tests continue to pass which I imagine they should it's no trouble with me

PR #4313, opened 2022-05-09
Removed dup typing. Modified the typing for the rest of multicast

fedgrant:
- Links to documentation: https://github.com/ReactiveX/rxjs/blob/6.x/src/internal/operators/multicast.ts#L38
- Link to GitHub or NPM: 
- Type of contribution:  fix  | refactor

Other notes:
Modified the rest of the multicast typing to follow the official code. You are always allowed to pass in a function that returns a subject or just a subject.


Brianzchen:
Closes https://github.com/flow-typed/flow-typed/issues/4307

Brianzchen:
Thanks applying the change you raised before. Just one thing is that I realised that there's actually zero tests around `multicast` do you mind adding some so that we can exercise that your changes or the type itself works the way users intend?

Brianzchen (approved):
Gonna move ahead with this PR, it looks good and we can revert if issues are raised.

PR #4315, opened 2022-05-11
[fuse.js] feat: add v6.6.x

seawind543:
- Links to documentation: https://fusejs.io
- Link to GitHub or NPM: https://www.npmjs.com/package/fuse.js

Other notes:
Generated by Flowgen from `fuse.js` **v6.6.2**'s Typescript Definition: https://github.com/krisk/Fuse/blob/master/dist/fuse.d.ts

Flowgen: https://www.npmjs.com/package/flowgen


Brianzchen:
Hey there, thanks for putting in this pull request. Generation from flowgen may give you the right type defs but seems like you'll to need to do a few manual tasks before this will be mergable.

- Firstly if you haven't already having a read over the [CONTRIBUTING](https://github.com/flow-typed/flow-typed/blob/master/CONTRIBUTING.md) docs will give you an idea of the structure and our expectations
- You may want to use the `./create_def.sh` which is also mentioned on the docs on how to create the folder struct for your new definition. Right now you have an issue where your top level dir is `6.x.x` while module is `6.6.x` which is causing tests to break. If you were to align them, I'd recommend `6.x.x` as that gives you a wider range to play with and easier maintainability
- You'll need to wrap your definition around a `declare module ''` block which `./create_def.sh` would create that structure for you
- Tests are recommended to be wrapped in `describe/it` blocks so that it's easier to maintain and the test cases are human readable
- Once you're ready with your def you should at least run `./quick_run_def_tests.sh` to make sure your tests are passing locally since they're broken in CI

Goodluck and if you need more help reach out!

seawind543:
Hi @Brianzchen 

Thank you for the helpful information. I had re-write the definition and test cases. Could you please help me to take a look again? 🙏 Thank you. Let me know if any issues/misunderstandings.

Brianzchen (approved):
Nice one thanks! It looks likes you've tested the usecase, feel free to raise more prs if you need to define the types and or add more stuff

PR #4331, opened 2022-06-08
[cli] Definitions depend on other definitions

Brianzchen:
This adds the capability for one definition to depend on and thus import from other definitions in flow-typed. Details mapped [here](https://github.com/flow-typed/flow-typed/discussions/4242#discussioncomment-2838850)

A definition's flow dir can now contain a `package.json` that can hold a `deps` property to hold a list of dependency definitions and their supported versions.

```
{
  "deps": {
    "dep":  ["v1.x.x", "v2.x.x"],
    "dep2": ["v2.x.x"],
    "dep3": ["v2.1.x", "2.2.x", "2.x.x"]
  }
}
```

**A major caveat is that with this implementation mostly imposed by the original architecture, all defs and the flow versions that want to depend on other defs need to use the same flow version, you will see the changes that I've made to various redux packages as I've tested my changes.**

## Tasks
- [x] Depending on a definition can run tests
- [x] Changing a definition will trigger all dependants to also run when not part of test pattern
- [x] Changing a definition will trigger all dependants to run when `onlyChanged` is active
- [x] When `flow-typed install` if I have a valid dependency version already installed use same logic to install that specific dep (eg if new version of definition exists)
- [x] When `flow-typed install` if I do no have valid dependency version installed, install the last version referenced in the dependency's array
- [x] if `flow-typed install` installs a dependency that is not in the version list it should override to install a supported version
- [x] `flow-typed install` will prioritise `deps_` prefix libdefs if it finds both
- [x] When installing, if a dep has another dep it should recursively install
- [x] Running install twice on a `deps_` prefix definition doesn't trigger warning of overridden definition
- [x] Add documentation
- [x] Add tests to install command

## Adoption Strategy
Definitions can be created in duplicated definition package dirs prefixed with `deps_` that supports all the same functionality as regular dirs. This is to support 2 concurrent package dirs at the same time because the repo is used by all versions of flow-typed cli.

`deps_` prefixed definitions will contain new package.json functionality as we give a warranty period to older version users. When we're ready we can simply delete all the `deps_` prefixes and replace the old defs and users will not need to upgrade flow-typed cli as both versions support dependency features

Screen Shot 2022-06-26 at 7 29 37 pm


Brianzchen:
This is now ready for review. A few points to consider

- My strategy of adoption can work for future enhancements that users on old versions won't have support for, it can be reused but of course for new implementations we'd need to use a different keyword so that current stable version users don't receive new changes until they have the new CLI version to match.
- I've added tests to install as well as manually with local definition changes and tested run tests throughly manually only
- Definitions depending on `redux` has been my first use case. I've broken `redux` into two additional version ranges and then added new `deps_` temp definitions that can now depend on `redux` (could have been done as a separate PR but I think showcasing test support functioning as expected is important)
- If we assume there are regressions created, the new changes should have no effect on existing definitions as long as we never leverage `deps` feature in case there's some uncertainty around the change

cli/src/commands/install.js
```diff
@@ -494,6 +498,18 @@ async function installNpmLibDefs({
...
           }
+
+          // If this libdef has dependencies let's first add it to a giant list
+          if (libDef.depVersions) {
+            Object.keys(libDef.depVersions).forEach(dep => {
+              if (libDef.depVersions) {
+                // This may result in overriding other libDef dependencies
+                // but we don't care because either way a project cannot have the
+                // same module declared twice.
+                defDepsToInstall[dep] = libDef.depVersions[dep];
```
GAntoine, line 509: 👍🏼

cli/src/commands/install.js
```diff
@@ -543,6 +559,49 @@ async function installNpmLibDefs({
...
+                // If no supported version warn it'll be overridden and continue
+                listItem(
+                  colors.yellow(
+                    `One of your definitions has a dependency to ${dep} @ version(s) ${defVersions.join(
+                      ', ',
+                    )}`,
+                  ),
+                  `You have version ${colors.yellow(libDef.version)} installed`,
+                  `We're overriding to a supported version to fix flow-typed ${colors.red(
+                    'but you may experience other errors',
```
GAntoine, line 584:
```suggestion
                    'but you may experience other type errors',
```

GAntoine (changes requested):
Overall looks really solid!

Any chance using `package.json` for the deps file could cause confusion? Seeing as its structure isn't typical.

Brianzchen:
> Any chance using `package.json` for the deps file could cause confusion? Seeing as its structure isn't typical.

Yep, `package.json` was a recommendation naming convention and I'm not attached to it. `config.json` or...?

cli/src/commands/__tests__/install-test.js
```diff
@@ -1488,6 +1488,295 @@ declare type jsx$HTMLElementProps = {||}`;
...
+                  'src',
+                  'flow-typed',
+                  'npm',
+                  'a-override_v1.x.x.js',
+                ),
+              ),
+            ]),
+          ).toEqual([true, false, true, false]);
+        });
+      });
```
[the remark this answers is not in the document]
Brianzchen: Yep as the name of the test suggests 😅 `installs deps_ prefix definition as priority if it exists`

GAntoine:
yeah config.json is a good choice, or flowdeps.json if you want to be specific to this project?

Brianzchen:
> yeah config.json is a good choice, or flowdeps.json if you want to be specific to this project?

maybe just config.json then unless more suggestions, it's a file that exists in the repo but people running install would never see this. They would only receive the final flowtyped definitions

cli/src/commands/__tests__/install-test.js
```diff
@@ -1488,6 +1488,295 @@ declare type jsx$HTMLElementProps = {||}`;
...
+
+          // Installs libdef
+          expect(
+            await fs.exists(
+              path.join(
+                FLOWPROJ_DIR,
+                'src',
+                'flow-typed',
+                'npm',
+                'foo_v1.x.x.js',
```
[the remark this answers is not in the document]
Brianzchen: Yep that's right, so we're checking that even though it's not declared in the projects package.json it will still appear in the flow-typed dir after install. It's references here in the fixtures https://github.com/flow-typed/flow-typed/pull/4331/files#diff-1037410880ad59c3a8d21833993d5e979a72f0d42a96c97ea3a84dc0d0c67e7d

GAntoine (approved):

Brianzchen:
Removed redux use case and moved to https://github.com/flow-typed/flow-typed/pull/4339 but keeping all testing logic here for continuity. Should make it easier if there's fixes that need to be implemented.

Also targeted `v4` branch and installing definitions from `main` instead `master` so should be safe to merge.

cli/src/commands/runTests.js
```diff
@@ -578,34 +616,101 @@ async function runTestGroup(
...
+            const tempPath = `${depBasePath}${scope}deps_${packageName}_${version}/${flowDirVersion}/${packageName}_${version}.js`;
+            const path = `${depBasePath}${scope}${packageName}_${version}/${flowDirVersion}/${packageName}_${version}.js`;
+            if (fs.existsSync(tempPath)) {
+              return tempPath;
+            }
+
+            if (!fs.existsSync(path)) {
+              throw new Error(
+                colors.red(
+                  `${depName}@${version} cannot be a dependency of ${testGroup.id} because either the dependency@version does't exist or they do not have matching flow version ranges`,
```
AndrewSouthpaw, line 678: Maybe link to the relevant bit in README to guide the person to resolution. I'm imagining a contributor bumping into this error and being unclear on how to resolve it. It'd be great to give some suggestions in the docs.

cli/src/commands/runTests.js
```diff
@@ -578,34 +616,101 @@ async function runTestGroup(
...
+            return path;
+          });
+        });
+
+        const longestDep = mappedDepPaths.reduce((acc, cur) => {
+          if (cur.length > acc) {
+            return cur.length;
+          }
+          return acc;
+        }, 0);
```
AndrewSouthpaw, line 692: I think `Math.max(...mappedDepPaths.map(x => x.length))` would do this too 😄

cli/src/commands/runTests.js
```diff
@@ -578,34 +616,101 @@ async function runTestGroup(
...
+
+        const depGroup = [];
+        for (let i = 0, len = longestDep; i < len; i++) {
+          const newGroup = [];
+          mappedDepPaths.forEach(o => {
```
AndrewSouthpaw, line 697: Rename `o` to a full name, not exactly clear what it is.

cli/src/commands/runTests.js
```diff
@@ -636,6 +741,13 @@ async function runTests(
       if (testGroup.id.match(pattern) != null) {
         return true;
       }
+
+      const depsList = Object.keys(testGroup.deps);
+      if (depsList.length > 0) {
+        return depsList.some(dep => {
+          return testGroup.deps[dep].some(o => `${dep}_${o}`.match(pattern));
+        });
+      }
```
AndrewSouthpaw, line 750: It's not exactly clear what this is doing, add a quick comment?

AndrewSouthpaw (approved):
Amazing work!

cli/src/commands/runTests.js
Brianzchen: Good point! Took me a while to recall but it's pretty important

PR #4343, opened 2022-07-16
fix: HTMLElement for element passed to MatcherFunction

magus:
Type of contribution: **fix**

The change below changed the `Matcher` function type to what it is today from the following

```
export type MatcherFunction = (content: string, element: HTMLElement) => boolean
```
> https://github.com/testing-library/dom-testing-library/commit/accb6cc60628cb6b5bd4d9be2ead41724995e5ae

The previous type above seems more correct since the matcher function is not called with `null` elements as far as I can tell? The type being `HTMLElement` also allows the match function to correctly interact with the DOM properties of the element, whereas `Element` is too general and throws an error when you try to access fields such as `element.style`.

== Tests

Confirmed tests failed without my changes and now pass with my changes via `./quick_run_def_tests.sh`

> Screen Shot 2022-07-16 at 12 02 16 AM

### DIFF
--- a/definitions/npm/@testing-library/react_v12.x.x/flow_v0.104.x-/react_v12.x.x.js
+++ b/definitions/npm/@testing-library/react_v12.x.x/flow_v0.104.x-/react_v12.x.x.js
@@ -184,7 +184,7 @@ declare module '@testing-library/react' {
 
   declare type MatcherFunction = (
     content: string,
-    element: ?Element
+    element: HTMLElement
   ) => boolean;
 
   declare type Matcher = MatcherFunction | RegExp | string | number;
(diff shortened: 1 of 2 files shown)


Brianzchen:
This seems legit... If react-testing-library is the one looping through elements why would it give you a null one to check over.

Do you have any additional proof of this because, I would have hoped if the library is literally written in TS the TS defs would be accurate.

magus:
Yeah I'm not too sure but I have no run into any issues in a fairly large codebase.

Poking around the `testing-library/dom-testing-library` source

> https://github.com/testing-library/dom-testing-library/tree/main/src/queries

It looks like most queries use `Array.from(element.querySelectorAll(...))` so ... I suppose if that could return `null` for some reason? Guess we'd need someone like @marcosvega91 or @kentcdodds to confirm why the types are setup that way.

Brianzchen (approved):

Brianzchen:
Happy to merge this as is. If at a later point we get a reply from RTL maintainers and this isn't the case anymore we can easily revert and improve the def

PR #4344, opened 2022-07-16
Relocate @@aria-query types into module

magus:
Type of contribution: **fix**

Declaring types under `@@aria-query` is only valid if you depend on `@@aria-query`, since these are meant to be used locally within this type file we should relocate the types down into the `@testing-library/react` module instead.

If you try to use this file in a local `flow-typed` file without `@@aria-query` installed you will encounter `eslint` failures for an unresolved `@@aria-query` import (`import type { ARIARole } from '@@aria-query';`)

### DIFF
--- a/definitions/npm/@testing-library/react_v12.x.x/flow_v0.104.x-/react_v12.x.x.js
+++ b/definitions/npm/@testing-library/react_v12.x.x/flow_v0.104.x-/react_v12.x.x.js
@@ -1,168 +1,164 @@
-/**
- * A local copy from:
- * https://github.com/A11yance/aria-query/blob/2e6a3011a0d8987655f3a14853934fe3df38a8d8/flow/aria.js
- */
- declare module '@@aria-query' {
-  declare export type ARIAAbstractRole =
-    | 'command'
-    | 'composite'
-    | 'input'
-    | 'landmark'
-    | 'range'
-    | 'roletype'
-    | 'section'
-    | 'sectionhead'
-    | 'select'
-    | 'structure'
-    | 'widget'
-    | 'window';
-
-  declare export type ARIAWidgetRole =
-    | 'button'
-    | 'checkbox'
-    | 'gridcell'
-    | 'link'
-    | 'menuitem'
-    | 'menuitemcheckbox'
-    | 'menuitemradio'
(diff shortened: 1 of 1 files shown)


Brianzchen:
Fair change though I believe these were in a separate module initially because it's a separation of concerns to the types of react-testing-library. The goal I suppose was that eventually it could be moved into a dedicated definition module.

But regarding your concern, are you ok with [just disabling eslint in flow-typed/](https://flow-typed.github.io/flow-typed/#/faq?id=my-flow-typednpm-dir-is-throwing-lots-of-eslint-errors-after-installing-definitions). Mainly because you really shouldn't be updating flow typed defs with your project standards and as you have more strict rules you'll probably find more errors.

magus:
Yup we do ignore `flow-typed/npm` but since I've ejected to a local version this ignore no longer applies. We could of course ignore this file as a one-off but it felt more correct to just fix it. The usage of `ARIARole` is only in one spot, and some are actually not used (e.g. `ARIAAbstractRole`) so it felt like a safe change to move them.

```
declare type ByRoleMatcher = ARIARole | MatcherFunction;
```

Brianzchen:
Just found the [pr](https://github.com/flow-typed/flow-typed/pull/4002) that added the declaration, maybe if @pascalduez can give some insight and judge if this change is fair

pascalduez:
This was intended and following a common pattern we have in flow-typed, unfortunately not documented.   
Hence the kinda weird `@@` (but apparently it's not weird enough :). 
The main goal is only about code structure, and also showing the intent of waiting for cross lib defs imports.   

So please keep it as is. 
Lib defs should be considered as node_modules and not linted.

Brianzchen:
Are these types meant to from [aria-query](https://www.npmjs.com/package/aria-query). If so we can start building these in [main branch](https://github.com/flow-typed/flow-typed/blob/main/CONTRIBUTING.md#importing-types-from-other-libdefs) getting ready for the v4 CLI to be released.

@magus I'd be interested in submitting this change but will be sometime in August since I have some other priorities. What do you think?

magus:
This isn’t urgent so feel free to close or delay that’s fine with me it’s already fixed in our local case!

PR #4350, opened 2022-07-23
[lodash.shuffle] Add definition

### LINKED ISSUE #2488: Support modularized NPM packages

Brianzchen:
Closes #2488 as an example of how to reexport definitions for scoped versions of large packages.

- Links to documentation: https://www.npmjs.com/package/lodash.shuffle
- Link to GitHub or NPM: https://www.npmjs.com/package/lodash.shuffle

Other notes: cc @AndrewSouthpaw wanna make sure you're happy we continue this approach now that it's low barrier to entry 🙏 

**This is targeting `main` branch**


AndrewSouthpaw (approved):
Yeah this looks great to me! It makes sense to create the smaller scoped package versions for people who use those instead. 

Was there something in particular you were concerned on my end about?

GAntoine (approved):

Brianzchen:
> Was there something in particular you were concerned on my end about?

Not at all, just that you commented on the original issue and wanted to make sure you were across the solution before we cont.

PR #4354, opened 2022-07-25
[cli] Checkout into mainline cache if not already

Brianzchen:
When a user uses flow-typed for the first time or existing. Upon downloading the definition cache they'll most likely be on `master` because that's the default branch when we clone.

v4 cli relies on `main` branch for new definitions so when cloning or rebasing we'll now first check the mainline branch and if it's not correct the cli will run a fetch and checkout prior to pulling.

> We can keep this logic even after making `main` the default branch as a nice safeguard

ezgif-1-aa00332aa7


GAntoine (approved):

Brianzchen:
@GAntoine v4 ready for another round of publishing!

PR #4357, opened 2022-07-26
Add @mapbox/mapbox-gl-supported libdef

NameFILIP:
- Link to GitHub or NPM: https://github.com/mapbox/mapbox-gl-supported
- Type of contribution: new definition

PR #4358, opened 2022-07-27
Add @mapbox/mapbox-gl-supported v1 libdef

NameFILIP:
- Links to documentation: https://github.com/mapbox/mapbox-gl-supported/tree/v1.5.0
- Link to GitHub or NPM: https://github.com/mapbox/mapbox-gl-supported/blob/v1.5.0/index.js

Other notes: Adding the definition for v1 which has slightly different exports format compared to v2


NameFILIP:
This PR is similar to the previously merged one: https://github.com/flow-typed/flow-typed/pull/4357

Brianzchen (approved):
i see this is 1.x.x support as opposed to 2.x.x before

PR #4359, opened 2022-07-27
[dom-testing-library] MaybeIntersectionHTMLElement

magus:
Type of contribution: **fix**

The type returned as a maybe `IntersectionHTMLElement` is `?IntersectionHTMLElement` however this is imprecise and includes `undefined` so the full type is `undefined | null | IntersectionHTMLElement`, this breaks when used with a very common scenario such as the type on the `JSDOM` `toContainElement` matcher which requires a `HTMLElement | SVGElement | null`. 

Something worth noting is #4135 introduced this union type (`IntersectionHTMLElement`) which is handling the fact that various HTML elements have different properties we may want to access during testing but we cannot use `HTMLElement` and refine into a more specific shape so we use a union to accomplish that.

We can confirm the type is `HTMLElement | null` here in `dom-testing-library`

> https://github.com/testing-library/dom-testing-library/blob/73a5694529dbfff289f3d7a01470c45ef5c77715/types/queries.d.ts#L5
> https://github.com/testing-library/dom-testing-library/blob/73a5694529dbfff289f3d7a01470c45ef5c77715/src/query-helpers.ts#L144

This work builds upon #4343 which is an ongoing effort at my company to get our `flow` types for this library correct

== Tests

Confirmed tests failed without my changes and now pass with my changes via `./quick_run_def_tests.sh`

> Screen Shot 2022-07-27 at 1 04 17 PM

### DIFF
--- a/definitions/npm/@testing-library/react_v12.x.x/flow_v0.104.x-/react_v12.x.x.js
+++ b/definitions/npm/@testing-library/react_v12.x.x/flow_v0.104.x-/react_v12.x.x.js
@@ -276,10 +276,12 @@ declare module '@testing-library/react' {
     & HTMLSelectElement;
   // End mixed html types
 
+  declare type MaybeIntersectionHTMLElement = null | IntersectionHTMLElement;
+
   declare type QueryByBoundAttribute = (
     text: Matcher,
     options?: MatcherOptions
-  ) => ?IntersectionHTMLElement;
+  ) => MaybeIntersectionHTMLElement;
 
   declare type AllByBoundAttribute = (
     text: Matcher,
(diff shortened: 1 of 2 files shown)


Brianzchen (approved):
I thought `?IntersectionHTMLElement` was maybe which is `null | type` already.. Change here seems fair and if it makes us more accurate I'm all for it

Brianzchen:
Regarding your concerns over https://github.com/flow-typed/flow-typed/pull/4135. Since I implemented it I understand that RTL types are actually just HTMLElement, but without deviating from it, RTL + flow was extremely tedious to use, maybe it's easier with TS in the way they can just type refine but with flow we had to write explicit code paths to refine the type which just made my team either write redundant code or not flow type their tests at all.

I don't think aligning back to HTMLElement just because RTL has those types would be a very valuable change.

magus:
I agree after trying it out I realize having to retype each instance of `HTMLElement` for the different kinds (such as `HTMLAnchorElement`) would be tedious.

PR #4382, opened 2022-09-16
[tempy] Create definitions for v3.x.x

yungsters:
- Links to documentation: https://github.com/sindresorhus/tempy/blob/main/index.d.ts

Other notes:
Created these by referencing the latest TypeScript definitions for `tempy@3.0.0` and added tests.

### DIFF
--- a/definitions/npm/tempy_v3.x.x/test_tempy.js
+++ b/definitions/npm/tempy_v3.x.x/test_tempy.js
@@ -1,0 +1,111 @@
+// @flow
+
+import { describe, it } from 'flow-typed-test';
+import {
+  temporaryFileTask,
+  temporaryDirectory,
+  temporaryDirectoryTask,
+  temporaryWrite,
+  temporaryWriteTask,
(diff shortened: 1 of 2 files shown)


definitions/npm/tempy_v3.x.x/test_tempy.js
```diff
@@ -0,0 +1,58 @@
+// @flow
```
Brianzchen, line 1: Regarding test would it be possible to add in the `describe/it` format? Would make it easier to differentiate use cases in the future for extending

Brianzchen (approved):
Just one comment regarding test format

definitions/npm/tempy_v3.x.x/test_tempy.js
yungsters: Thanks for the feedback. I've updated the test file with `describe`/`it`.

Brianzchen (approved):

PR #4385, opened 2022-10-03
[CLI] merge v4 branch into `main`

Brianzchen:
v4 installation code previously lived in a separate branch to the definitions and test runner code. Since installs are quite stable and we've tested adding a various range of definitions it's quite safe to merge `v4` into `main` and make it the standard rc branch prior to v4's official release.


cli/package.json
```diff
@@ -10,7 +10,7 @@
   "bugs": {
     "url": "https://github.com/flow-typed/flow-typed/issues/new/choose"
   },
-  "version": "3.8.0",
+  "version": "4.0.0-1",
```
GAntoine, line 13: Don't bump these by hand, I do it on release with a script.

GAntoine (changes requested):

cli/package.json
Brianzchen: I think this was bumped during prev releases. current RC is `4.0.0-1`

cli/src/lib/git.js
```diff
@@ -1,8 +1,33 @@
 // @flow
-
 import whichCb from 'which';
+import simpleGit from 'simple-git';
+
 import {child_process} from './node';
 
+const mainlineBranch = 'main';
```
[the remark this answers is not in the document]
Brianzchen: this was more so that I didn't need to repeat `main` keyword all over the codebase. But it's quite locked in.

How do you mean by config though? Should I shift this var to another utility module do you think?

GAntoine (approved):

PR #4389, opened 2022-10-12
Update Context object in next v10 to match next's Typescript types more closely

elischutze:
- Links to documentation: 
- Link to GitHub or NPM: 
- Type of contribution: **addition**

Other notes:

### DIFF
--- a/definitions/npm/next_v10.x.x/flow_v0.104.x-/next_v10.x.x.js
+++ b/definitions/npm/next_v10.x.x/flow_v0.104.x-/next_v10.x.x.js
@@ -49,10 +49,16 @@ declare module "next" {
   declare export type Context = {
     +pathname: string,
     +query: any,
-    +req?: any,
-    +res?: any,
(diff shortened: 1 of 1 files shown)


definitions/npm/next_v10.x.x/flow_v0.104.x-/next_v10.x.x.js
```diff
@@ -49,10 +49,16 @@ declare module "next" {
...
     +query: any,
-    +req?: any,
-    +res?: any,
+    +req?: http$IncomingMessage<>,
+    +res?: http$ServerResponse,
     +xhr?: any,
-    +err?: any,
+    +err?: (Error & {
+      statusCode?: number;
+  }) | null;,
```
Brianzchen, line 57: This is causing tests to fail. Could you make a fix here then from the root dir run `./quick_run_def_tests.sh` the contributing docs should have more details if you run into setup issues.

PR #4392, opened 2022-10-20
Add type argument default and bound to useDispatch

SamChou19815:
As part of the local type inference project of Flow, we will require all the type arguments to be inferable from local context. In the react-redux libdef, `useDispatch` has an under-constrained type argument D, that will require a lot of annotation burden for users.

In this PR, I added default and bound to useDispatch so that:

1. You will now actually get a dispatch function instead of anything you want, and only can dispatch an action instead of arbitrary stuff.
2. If you don't care about limiting the actions you can dispatch, you don't need to annotate anything. This is necessary to upgrading Meta's internal codebase.

@jbrown215 first created this fix and tested it our internal codebase, which works extremely well.

- Links to documentation: https://medium.com/flow-type/introducing-local-type-inference-for-flow-6af65b7830aa
- Type of contribution: fix

### DIFF
--- a/definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/react-redux_v7.x.x.js
+++ b/definitions/npm/react-redux_v7.x.x/flow_v0.142.x-/react-redux_v7.x.x.js
@@ -203,7 +203,29 @@ declare module "react-redux" {
   // Typings for Hooks
   // ------------------------------------------------------------
 
-  declare export function useDispatch<D>(): (
+  /* Action and AnyAction are taken from the redux TypeScript types defined here:
+   * https://github.com/reduxjs/redux/blob/d794c56f78eccb56ba3c67971c26df8ee34dacc1/src/types/actions.ts
+   *
+   * We turn them into objects so that they can be spread.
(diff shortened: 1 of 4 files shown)


SamChou19815:
Unfortunately the build is already broken before this PR, due to Flow change documented here: https://medium.com/flow-type/requiring-more-annotations-to-functions-and-classes-in-flow-e8aa9b1d76bd

Brianzchen:
Hey @SamChou19815 I've made the fixes to react-redux in #4393 so you can pull master to have your changes run in CI.

Also, I don't fully understand this change much, do you think some tests could be added to secure your change so that we don't accidentally break it in the future for whatever your use case is?

SamChou19815:
Rebased onto master. Tests WIP

SamChou19815:
I was only able to add tests for supplying invalid type argument. It's impossible to test the default argument case because other returns of useDispatch are horribly unsound.

Brianzchen:
I'll look at reflecting these changes back to `main` branch maybe tomorrow when I'm not working

PR #4397, opened 2022-10-23
Fix jest mockResolveValue typing

agalatan:
- Links to documentation: https://jestjs.io/docs/mock-function-api/#mockfnmockresolvedvaluevalue
- Type of contribution: Fix

**Problem:**
Jest definitions for **mockResolvedValue** (and **mockResolvedValueOnce**) are incorrect:
- current definition forces you to use a Promise as input, which is incorrect
- the fact that there are no tests for **mockResolvedValue** also hints to a problem in the typing

**Solution:**
- Use the following util to "depromisify" the return value of a mocked function, and use that as input for **mockResolvedValue**

```
type PromiseMatcher = (<T>(Promise<T>, any) => T) & (<T>(any, T) => T);
type Depromisify<X> = $Call<PromiseMatcher, X, X>;
```

### DIFF
--- a/definitions/npm/jest_v25.x.x/flow_v0.104.x-v0.133.x/jest_v25.x.x.js
+++ b/definitions/npm/jest_v25.x.x/flow_v0.104.x-v0.133.x/jest_v25.x.x.js
@@ -1,3 +1,6 @@
+type PromiseMatcher = (<T>(Promise<T>, any) => T) & (<T>(any, T) => T);
+type Depromisify<X> = $Call<PromiseMatcher, X, X>;
+
 type JestMockFn<TArguments: $ReadOnlyArray<*>, TReturn> = {
  (...args: TArguments): TReturn,
  /**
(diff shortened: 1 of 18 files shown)


Brianzchen:
Hey thanks for the submission, looks like tests are failing. If you can fix them first we can look at this change further

agalatan:
On it, thx.

agalatan:
- Just FYI, the ./build.sh fails (at least for me) with error `EPERM: operation not permitted, unlink '/Users/agalatan/workspace/flow-typed/cli/.flow-bins-cache/TMP-flow-v0.190.0'` making local testing quite unpleasant.
- I was able to hack it, by replacing the unlink with a brute-force `execSync('rm -rf blabla')`

Brianzchen:
I'm not sure what would cause that since I don't use windows I can't test, but do you use WSL or something, and also do you know what you get if you run `os.type()`?

PR #4398, opened 2022-10-25
[environments] Add a few prefixed node core libs

pascalduez:
- Links to documentation: https://nodejs.org/api/esm.html#node-imports

Other notes:
Following a discussion on Discord, adding a few common prefixed node core libs aliases. 
More will be added as needed I guess? Or should we dump the whole thing at once?
`flow_v0155.x-` because of indexed access.


GAntoine (approved):
yeah makes sense

PR #4400, opened 2022-10-26
[environments:jsx] Add a few more HTMLElement attributes

pascalduez:
- Links to documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#list_of_global_attributes

Other notes:
Add a few more `HTMLElement` attributes.

Would have been nice to be able to use `csstype` for the style prop but did not seemed to work in my tests.   
Same for the `role` prop and `aria-query`.   

NB: those files does not seem to be using prettier, which makes it quite tedious to contribute to.

### DIFF
--- a/definitions/environments/jsx/flow_v0.83.x-/jsx.js
+++ b/definitions/environments/jsx/flow_v0.83.x-/jsx.js
@@ -29,19 +29,41 @@
    /**
     * Specifies that an element is not yet, or is no longer, relevant
     */
-   hidden?: string,
+   hidden?: boolean | '' | 'hidden' | 'until-found',
    /**
     * Specifies a unique id for an element
     */
    id?: string,
    /**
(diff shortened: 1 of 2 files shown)


Brianzchen (approved):
Feel free to add the prettier tooling. I personally don't use it so never noticed

pascalduez:
> Feel free to add the prettier tooling. I personally don't use it so never noticed

Well, reminds me of old discussions :) Let's move this t o another issue.  
Fact is, we have a `.prettierrc` file on the repo, this is why my editor applies formatting by default.

pascalduez:
And thanks for booting the `environments` feature @Brianzchen, finally starting to use it.

Brianzchen:
No worries! Let me know if we can improve it in any way. Happy to build the tooling to support more use cases

PR #4401, opened 2022-10-31
[docs] Fix jsx env code sample

pascalduez:
- Type of contribution: fix

Other notes:
Within the `jsx` environment there's no `$Attributes`  type associated with `jsx$HTMLElement`. 
Depending on the initial implementation idea should we:
- remove it from the docs?
- Add it to the env?


Brianzchen (approved):
Oh yep oops! I was playing around with naming before the final implementation

PR #4406, opened 2022-11-24
[cli] Tests now run against changed flow version

### LINKED ISSUE #4384: Test older flow version ranges
### CLI Version

4.0.0-1

Currently when a definition is modified we blindly just test the last few versions flow versions against the definition. We don't want to test all because that would take forever but we should ideally test against versions in the range that the changes effected.

When a flow definition is modified, check what versions it actually affected such as if it modified a test or definition file only related to a flow range of `0.100.0` - `0.124.0`. Even though current version might be in the `0.200.x`+ range, testing the last 5 versions in CI has no value and instead we should test the last 5 versions before `0.124.0`

### Anything else?

I could make the change but I'm tied in up in other interests at the moment so I'm happy for others to take a stab.

Otherwise this is an interesting feature to solve regardless and helps with flow-typed'd overall test reliability. 

No (hope someone else can)

Brianzchen:
Closes #4384.

Previously, tests would always run against the latest x number of flow versions (default 15). But this made it rather and difficult and unreliable when trying to maintain definitions against older versions as those would never be tested in CI.

With this change, the flow is a bit different:

1. The test runner will attempt to download all flow releases
2. Then if the definition has no upper boundary (`v0.125.x-`) will take the latest 15 versions up to the lower boundary (`0.125.x`)
3. Otherwise if upper boundary is defined (`v0.125.x-v0.135.x`) we will start tests against the upper boundary (`0.135.x`) and take the latest 15 versions starting from there up to the lower boundary (`0.125.x`)

> Caveat here is that startup times of running will be much slower as we're now downloading pages of release notes instead of just the top 15 but they'll be cached for the duration of the run against multiple definitions as shown below

Screen Shot 2022-11-25 at 5 50 02 am
Screen Shot 2022-11-25 at 5 50 12 am


cli/src/commands/runTests.js
```diff
@@ -147,160 +148,182 @@ function printSkipMessage(flowVersion, githubUrl) {
...
-let _flowBinVersionPromise = null;
 async function getOrderedFlowBinVersions(
-  numberOfReleases: number = 15,
+  numberOfReleases: number,
+  flowVersion: FlowVersion,
 ): Promise<Array<string>> {
-  if (_flowBinVersionPromise !== null) {
-    return _flowBinVersionPromise;
-  }
-  return (_flowBinVersionPromise = (async function() {
```
Brianzchen, line 162: This blocks main change is the removal of an anonymous function causing a big diff, main changes are addition of calling `selectFlowTestVersions` and the while loop that pulls all release versions

cli/src/lib/github.js
```diff
@@ -1,13 +1,12 @@
 // @flow
-
 import {Octokit} from '@octokit/rest';
 
-const CLIENT = new Octokit();
-
-if (process.env.GH_TOK) {
-  CLIENT.authenticate({type: 'oauth', token: process.env.GH_TOK});
```
Brianzchen, line 8: This is very old code that isn't actually correct. Updated with new API and reflected in docs

PR #4430, opened 2023-04-07
Drop duplicate declaration of lodash-es/create module

whut:
In current versions of Flow this duplicate cause flow to crash with "Should have been excluded: .$module__lodash-es/create"

I found the log entry when running flow server with --temp-dir flag and looking at the log file generated in this temp dir:

Unhandled exception: (Failure "Should have been excluded: .$module__lodash-es/create") Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33 Called from Env.init_builtins_from_libdef.(fun) in file "src/typing/env.ml", line 917, characters 6-35

- Type of contribution: fix


Brianzchen:
Oops! That's my bad, will merge once green

PR #4432, opened 2023-04-15
Updates defs to not use deprecated `*` type, and rename `$Partial` to…

gkz:
The `*` has been deprecated and just `any` since version 0.163. It became an error in all cases in version 0.204
`$Partial` was renamed to `Partial` in version 0.201, it was completely removed in version 0.203: https://medium.com/flow-type/announcing-partial-required-flow-utility-types-catch-annotations-3a32f0bf2a20

Update various definitions for this change. I've updated definitions to start a new `0.201.x-` folder, and change the previous one to `xxx-0.200.x`. 0.201 was also the point LTI became the only option, so good point for a cutoff.

Used this transform to update the definitions:
```
function transformer(file, api) {
  const j = api.jscodeshift;

  return j(file.source)
    .find(j.ExistsTypeAnnotation)
    .forEach(path => {
      j(path).replaceWith(
        j.anyTypeAnnotation
      );
    })
    .toSource();
}
transformer.parser = 'flow';
module.exports = transformer;
```
Some commands I used (notes for myself later):
```
# `*` type
jscodeshift --parser=flow -t transform.js .
git st -s | cut -c '4-' | ag --only '^.*\-/' | sed 's/.$//' | sort | uniq > files
# revert changes
cat files | while read line; do cp -r "$line" "$(echo $line | sed -E 's/flow_v0.[0-9]{,3}/flow_v0.201/')"; git mv "$line" "$(echo $line)v.200.x"; done
ag -g 201 | jscodeshift --stdin --parser=flow -t transform.js

# partial
ag -l '\$Partial' | ag --only '^.*\-/' | sed 's/.$//' | sort | uniq > partial-files
cat partial-files | while read line; do cp -r "$line" "$(echo $line | sed -E 's/flow_v0.[0-9]{,3}/flow_v0.201/')"; git mv "$line" "$(echo $line)v.200.x"; done
ag -g 201 | sed -i 's/\$Partial</Partial/g'
```


Brianzchen (approved):

pascalduez:
Sorry for not reviewing earlier, but there's an error in the file naming. `flow_v0.104.x-v.200.x` 

```diff
-flow_v0.104.x-v.200.x
+flow_v0.104.x-v0.200.x
```

gkz:
Working on a fix

PR #4438, opened 2023-04-25
Update defs to use `Partial` instead of deprecated `$Shape`

gkz:
Codemod `$Shape` to `Partial` in the definitions, after the deprecation of `$Shape`: https://medium.com/flow-type/announcing-partial-required-flow-utility-types-catch-annotations-3a32f0bf2a20

Sometimes however, when used in an input position, it could work better to turn `$Shape<T>` into `$ReadOnly<Partial<T>>` rather than `Partial<T>`. I don't know this is the case, so if people have new issues with their defs after this codemod, they can try adding the `$ReadOnly` and see if it works better.

```
ag -l '\$Shape<' | ag --only '^.*\-/' | sed 's/.$//' | sed '/v0.201.x\-/ d' | sort | uniq > files
cat files | while read line; do cp -r "$line" "$(echo $line | sed -E 's/flow_v0.[0-9]{,3}/flow_v0.201/')"; git mv "$line" "$(echo $line)v0.200.x"; done
ag -l '\$Shape<' | ag 'flow_v0.201.x' | xargs sed -i 's/\$Shape</Partial</g'
```


Brianzchen:
Will look into this after I finish merging the previous `.200.x` -> `0.200.x` into main branch. Something I have yet to finish up.

Brianzchen:
Sorry it's taken a while. I don't have internet at home after moving house.... If all goes well will cont working on this after next weekend.

Brianzchen:
Found something odd, `definitions/npm/idb_v5.x.x/flow_v0.201.x-/idb_v5.x.x.js` created `0.83.x-0.200.x` & `0.201.x-` but both files have no diffs. The def does have usage of `$Shape` though and wasn't transformed into `Partial`.

Upon expanding more `0.201.x-` defs [eg](definitions/npm/hoist-non-react-statics_v2.x.x/flow_v0.201.x-/hoist-non-react-statics_v2.x.x.js) they also aren't transformed into `Partial`, might be worth checking if the codemod is correctly applying.

gkz:
I'll take a look, thanks for looking

PR #4440, opened 2023-05-13
[cli] Fix running definition tests locally on MacOS

Brianzchen:
If you look at the list of [releases](https://github.com/facebook/flow/releases), previously up to version [v0.200.1](https://github.com/facebook/flow/releases/tag/v0.200.1) flow used to release only a x86 binary

Screen Shot 2023-05-14 at 9 19 44 am

After this, from [v0.201.0](https://github.com/facebook/flow/releases/tag/v0.201.0) onward flow started releasing osx arm binaries.

Screen Shot 2023-05-14 at 9 20 41 am

The original flow-typed fetching binary logic never anticipated arm, so the logic was never handled and when it found 2 binaries for macos it crashed silently.

## What was fixed

- More logic to binary selection
- Refined regex expression so that it will never find multiple binaries

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -23,12 +23,28 @@ export type Args = {
   ...
 };
 
-// Used to decide which binary to fetch for each version of Flow
-const BIN_PLATFORM = (_ => {
+/**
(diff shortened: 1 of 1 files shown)


GAntoine (approved):
love it

PR #4423, opened 2023-03-15
Fix runTests command for flow@^0.200.0 set `exact_by_default` option

iamdey:
from 0.200.0, `exact_by_default` option is required in flowconfig and from 0.202.0, this option is set to `true` by default

cf. https://flow.org/en/docs/config/options/#toc-exact-by-default-boolean

- Type of contribution: fix

Fixes #4422

PR #4441, opened 2023-05-14
[aria-query] Add new definition ranges to keep compatibility with RTL

Brianzchen:
Due to the recent deprecation of `*` type `@testing-library/react` was broken into two version ranges at `0.200.x`.

`aria-query` is a definition that is used by `@testing-library/react` in the new `@next` version of flow-typed and needs to be updated to also have the same version ranges to maintain definition compatibility.

---

On top of this, the version ranges had to be slightly tweaked from `0.200.x` to `0.197.x` due to differences in flow errors being thrown in old/new versions. `@next` version of flow-typed has more fine grained version testing which is why when the change was merged into `master` this wasn't caught and is now fixed.


Brianzchen:
Additionally fix a test that was broken by https://github.com/flow-typed/flow-typed/pull/4423. This wasn't caught in `master` because `master` branch doesn't have as many tests as `main`

PR #4443, opened 2023-05-16
[react-router-dom] update MemoryRouter props

iamdey:
- Links to documentation: 
- Link to GitHub or NPM: https://github.com/remix-run/react-router/blob/853a9382b748466f9bf1af5c4c7e5571670f8b97/packages/react-router/lib/components.tsx#L139-L144
- Type of contribution:  fix & refactor

Other notes:
- fix props for MemoryRouter
- export props the same way as the ts def
- create a new flow range to support `Partial` utility

Commit are atomic, I can easily revert the new flow range if it's not recommended. Note that an unofficial `LocationShape` exists that suppose to solve the issue for other types that should changed for `Partial<Location>` if I'm correct (but I don't have enough time to make sure of it, so I prefer to keep it simple).


Brianzchen (approved):
Great thanks I've checked each commit and does what you say.

We can improve LocationShapes definition later as you say if there's further improvements

Brianzchen:
FYI I'm making a change to add `redirect` function and I got this.

```
Tests for react-router-dom_v6.x.x/flow_v0.203.x- ran successfully on flow v0.201.0.
        Consider setting v0.201.0 as the lower bound!
```

I may change the flow range in a future PR

iamdey:
Aw, I got this message but didn't understand it correctly, that's strange because of the usage of `Partial` that shouldn't work on 0.201.0

PR #4450, opened 2023-05-25
[env:jsx] Add HTMLButtonElement,  HTMLAnchorElement definitions

pascalduez:
- Links to documentation: 
  * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button
  * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

Note: sorry for the formatting diff, but that's what we get at not enforcing our prettier config...

All I needed was the `type` property :sweat_smile:


GAntoine (approved):

Brianzchen (approved):
No worries. thanks for adding :)

PR #4459, opened 2023-06-19
Add libdef for openai-node v3.3.0

joshuanapoli:
- Links to documentation: https://platform.openai.com/docs/api-reference
- Link to GitHub or NPM: https://github.com/openai/openai-node

Other notes:
- The flow executable version range is split only because of depending on axios types, which are split. At the moment, I still need the older range :-/
- I generated the libdef using dts-bundle-generator and flowgen from the original typescript. I had to change the Axios type names, since typescript and flow use different names for the same types.
- The build failed, but I don't think that this PR caused the problem:
   > Package `tua-body-scroll` does not exist on npm, is this meant to be an environment instead?


Brianzchen:
> The build failed, but I don't think that this PR caused the problem:

Thanks, geez my bad it's meant to be this https://www.npmjs.com/package/tua-body-scroll-lock.

Will look into it tonight and get back to you

Brianzchen:
Also this is quite a humongous definition file. Do you think more tests would be helpful? Or is the part you've tested all you need for it to be considered secure?

Brianzchen:
Pull from `main`, only your tests should impact now

joshuanapoli:
> Also this is quite a humongous definition file. Do you think more tests would be helpful? Or is the part you've tested all you need for it to be considered secure?

Well, it's covering the most common use-case of the library, which is pretty simple. (It's also the only case that I'm using.)


> Pull from main, only your tests should impact now

Ok, I'll add test to cover another piece of the library and pull from main.

joshuanapoli:
@Brianzchen thanks for your review. This PR is updated now and has passing tests.

Brianzchen:
Thanks I’ll do some sanity check tonight 🤞

Brianzchen (approved):

Brianzchen:
Looks fine overall. Just one question, did you want to make the def `3.x.x` instead of `3.3.x`? Would make it easier to maintain with future minor releases, it's generally the standard unless you believe minors might cause dramatic breaking changes or you really want to maintain them separately.

joshuanapoli:
@Brianzchen thanks; I pushed c2221f151935fcb0d8d3ae3e30be9e8548f9826e renaming with more general versioning v3.x.x.

PR #4466, opened 2023-07-14
match semantics for expect methods

meop:
These functions pairs should match in their type restrictions and semantics (they are just inversions of each other):

expect.arrayContaining, expect.not.arrayContaining
expect.objectContaining, expect.not.objectContaining

The Flow types for the 'not' versions were already correct: allowing typed instances of objects to be passed into the methods, not just inlined literal objects.

This pull request updates the normal versions of the functions to match.

### DIFF
--- a/definitions/npm/jest_v29.x.x/flow_v0.201.x-/test_jest_v29.x.x.js
+++ b/definitions/npm/jest_v29.x.x/flow_v0.201.x-/test_jest_v29.x.x.js
@@ -15,11 +15,16 @@ jest.atoMockOff();
 const mockFn = jest.fn();
 mockFn.mock.calls.map(String).map((a) => a + a);
 
-type Value = {
+type ExactType = {|
+  value: number
+|};
+
+type InexactType = {
   value: number,
+  ...
 };
 
-const valueArray: Array<Value> = [
+const valueArray: Array<ExactType | InexactType> = [
   {
     value: 0,
   },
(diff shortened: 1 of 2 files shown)


definitions/npm/jest_v29.x.x/flow_v0.201.x-/test_jest_v29.x.x.js
```diff
@@ -15,6 +15,19 @@ jest.atoMockOff();
 const mockFn = jest.fn();
 mockFn.mock.calls.map(String).map((a) => a + a);
 
+type Value = {
+  value: number,
+};
```
Brianzchen, line 20: Please update type to either, `{ value: number, ... }` or `{| value: number |}`

Right now it's ambiguous, though that highlights that maybe you should add a test for both instances 

ref: https://github.com/flow-typed/flow-typed/actions/runs/5558120502/jobs/10173104682?pr=4466#step:5:1146

Brianzchen (changes requested):

Brianzchen:
You have failing tests, that's because jest_29 has some flow tests that need updating with local annotations mostly. It would be nice to fix them but not necessary for this PR.

definitions/npm/jest_v29.x.x/flow_v0.201.x-/test_jest_v29.x.x.js
meop: Thank you for review. Updated.

PR #4479, opened 2023-07-28
[cli] Bump semver solves VULN

Brianzchen:
Closes #4460 by fixing the lower and upper bound semver functions which previously relied on logic from `semver` where it's lower bound was returned as an undefined object if the lower bound was anything but in newer versions the lower bound is simply missing.

This allows us to now update semver and keep it updated.


GAntoine (approved):
Looks good!

PR #4491, opened 2023-08-12
[node] Add process module definitions

Brianzchen:
- Links to documentation: https://nodejs.org/api/process.html
- Link to GitHub or NPM: https://nodejs.org/api/process.html

Other notes:

### DIFF
--- a/definitions/environments/node/flow_v0.155.x-/node.js
+++ b/definitions/environments/node/flow_v0.155.x-/node.js
@@ -4,6 +4,14 @@ declare module 'fs/promises' {
   declare module.exports: $Exports<'fs'>['promises'];
 }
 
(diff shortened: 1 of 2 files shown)


definitions/environments/node/flow_v0.155.x-/node.js
```diff
@@ -4,6 +4,14 @@ declare module 'fs/promises' {
   declare module.exports: $Exports<'fs'>['promises'];
 }
 
+declare module 'process' {
+  declare module.exports: typeof process;
```
pascalduez, line 8: Maybe try using the type directly? 

```suggestion
  declare module.exports: Process;
```

pascalduez (approved):

PR #4535, opened 2023-10-12
[cli] Third party libraries can be configured into config.json

### LINKED ISSUE #4534: react-native loaded into test runner
### CLI Version

4.x.x

As shown here https://github.com/flow-typed/flow-typed/pull/4533#discussion_r1355912499

There are cases where definitions want to use the types directly from react native just like we would with react. But since react-native is not part of the core lib this can't be possible

I think with a definitions `config.json` we can add a new property for third party libraries to install that are typed for the definition to depend upon during testing. 

We won't use this config during the install phase as we'll assume this is already catered for in a libraries package.json `dependencies` field. 

### Anything else?

Alternatively for `react-native` we could just load it into the test function regardless but I think that's not the right approach and would misleading

Yes

Brianzchen:
Closes #4534

## Working list
- [ ] Added docs
- [ ] Change `deps` to `defDeps` to distinguish between the two kinds

### DIFF
--- a/cli/src/commands/runTests.js
+++ b/cli/src/commands/runTests.js
@@ -76,6 +76,15 @@ type TestGroup = {
   deps: {
     [key: string]: Array<string>,
   },
+  /**
+   * Object of dependency version pairs for dependencies to install
+   * and load for testing so that definitions can be tested against
+   * typed libraries they depend as they would when installed
+   * in a real project
+   */
+  npmDeps: {
+    [key: string]: string,
+  },
 };
 
 /**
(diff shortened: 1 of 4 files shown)


Brianzchen:
Thanks @joshuanapoli for that context. I've tried to make this work but while doing so made a shocking discovery...
Screenshot 2023-10-18 081544

Seeking more input from the flow team but at the moment limited by core flow itself.

Interestingly importing at the top level of a library definition file doesn't give an unresolved error but instead is more of a lint rule saying it's not good practice and still errors

PR #4540, opened 2023-10-22
[cli] pnpm support

### LINKED ISSUE #4484: Install command cannot detect workspace usage with pnpm
### CLI Version

4.0.0-3

### Operating system

Windows

### Bug description

When I run flow-typed install with a monorepo workspace using pnpm the nested dependencies do not get installed

### Steps to reproduce

In a repo that uses pnpm there will be a `pnpm-workspace.yaml` file that declares a list of packages such as
```
packages:
  - 'packages/*'
```

Then when I run `flow-typed install` I should expect to see nested dependencies installed but I do not.

### Error output/screenshots

_No response_

### Flow configuration

[ignore]
.*node_modules/.*
!.*node_modules/react-native/.*
.*node_modules/react-native/Libraries/polyfills/.*
!.*node_modules/@.*
.*/malformed_package_json/package.json
!.*node_modules/bzc-ui/.*
!.*node_modules/csstype/.*
!.*node_modules/date-fns/.*
!.*node_modules/tiny-invariant/.*
!.*node_modules/key-commander/.*
!.*node_modules/react-ld/.*
!.*node_modules/recoil/.*
.*dist/.*
.*lib/.*

[include]

[untyped]

[libs]

[lints]
deprecated-type=error
unnecessary-optional-chain=error
ambiguous-object-type=error

[options]
enums=true
format.bracket_spacing=true
format.single_quotes=true
exact_by_default=true
include_warnings=true

[strict]

[version]

[declarations]

### Anything else?

Currently I can use https://github.com/flow-typed/flow-typed/blob/main/docs/flow-typed-config.md#workspaces as a workaround which works as expected but ideally flow-typed should handle pnpm use cases

### Do you want to submit a pull request to fix this bug?

Yes

Brianzchen:
Closes #4484. This change does 2 key things

- Scan `pnpm-workspace.yaml` for package globs to find workspaces when creating a dependency list to map to flow-typed definitions
- And not stubbing dependencies when a library ships with type definitions but the dependency sits in a workspace's `node_modules` instead of in the root `node_modules`. This is default behaviour for pnpm so previously would cause stubs to constantly get installed even if a workspace dependency shipped type defs. This could also be an issue for yarn or any other package manager if they used the `noHoist` option.


GAntoine (approved):
nice

PR #4541, opened 2023-10-24
[cli] Install a dependency's environment definitions

### LINKED ISSUE #4445: Install the environments needed for node_module packages
### CLI Version

4.x.x

If I have a package that I publish to npm and also ship flow definitions, my package can depend on definitions from flow-typed.

Currently it correctly installs dependency package definitions but environment definitions that the npm package uses are not installed.

This caused issues with resolving autocompletion in my app using the npm package until I dug into node_modules and realised that the `.js.flow` files were throwing errors and I was missing a global type definition.

When installing type definitions used by a package, a package can define a field in their package.json like the following so that when flow-typed scans deps for defs to install it knows.

```
{
  "flow-typed": {
    "envs": ["jsx"]
  }
}
```

This can get repetitive as it means they would equally need this defined in `flow-typed.config.json` and package.json

Another option would be that flow-typed analyses the package for a `flow-typed.config.json` to look for any extra environments it may depend on.

I believe the latter is a much better solution with the only caveat that `flow-typed.config.json` is recommended to be published but I think it's ok as metadata for flow-typed analysis going forward

### Anything else?

I plan to implement this just for v4.x.x (@next) and will implement this when I get a chance unless there's any comments from others

Yes

Brianzchen:
Closes #4445. For any dependency that ships flow type definitions check if they have a `flow-typed.config.json` and if so, read their `env` to check if they depend on any environment definitions.

If so install them along side any environment definitions the project installing might need. This allows a library to be developed with flow-typed knowing it will behave the same when consumed by a consumer using flow.

Not very common for open source, but very useful when developing a stack of internal npm libraries all shipping flow types.

### DIFF
--- a/cli/src/commands/__tests__/__install-fixtures__/end-to-end/fakeCacheRepo/definitions/environments/react/flow_v0.83.x-/react.js
+++ b/cli/src/commands/__tests__/__install-fixtures__/end-to-end/fakeCacheRepo/definitions/environments/react/flow_v0.83.x-/react.js
@@ -1,0 +1,3 @@
+declare type jsx$HTMLElementProps = {|
+  test: boolean,
+|};
--- a/cli/src/commands/__tests__/install-test.js
+++ b/cli/src/commands/__tests__/install-test.js
@@ -1504,6 +1504,118 @@ declare type jsx$HTMLElementProps = {||}`;
           ).not.toEqual(installedDef);
         });
       });
+
+      it('installs envs from dependency flow-typed.config even if no local flow-typed.config', () => {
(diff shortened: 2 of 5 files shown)


GAntoine (approved):
lgtm, if i understand it right there's no change in behaviour (outside of this exact one) right?

Brianzchen:
> lgtm, if i understand it right there's no change in behaviour (outside of this exact one) right?

Yep thats right! Test prove that at least :D

AndrewSouthpaw (commented):
Generally lgtm, thanks for putting this together! (I'm OOO and didn't review closely but overall direction is a win.)

PR #4547, opened 2023-10-28
[cli] Update flow-bin to 0.200.0

Brianzchen:
Update flow-bin to a newer version, at this version local type annotations becomes a requirement and a more checks such as if a variable is declared as `string` it cannot be reassigned as `number`.

Current version is 0.220.0 and will upgrade to that incrementally, the changes are quite large


pascalduez (approved):
Tests where failing, but I couldn't spot an issue from the diff, I've re-run them, they seem highly volatile.

EDIT: and now they pass...

Brianzchen:
> Tests where failing, but I couldn't spot an issue from the diff, I've re-run them, they seem highly volatile.

yeah we used to have that issue where flow definition runs were flakey.. Now issue with unit test suite but usually passes after first run.

PR #4550, opened 2023-11-04
[graphql] [@apollo/client] add graphql libdef and integrate with @apollo/client

joshuanapoli:
- Type of contribution: add libdef

Other notes:
- Add a module definition for "@apollo/client/utilities". This module has a CJS export and is meant as a public interface of the package, though not thoroughly documented. Moreover, not all of its symbols are exported at the root level; importing the sub-module in an application is necessary in some cases. In my case, I need `getMainDefinition`, which is shown in [apollo client example code](https://www.apollographql.com/docs/react/api/link/introduction/#directional-composition). 
- Fix ApolloLink and Operation class inheritance, which were incorrectly extending a global class. This didn't create the correct class inheritance and caused wrong flow errors in correctly written application code.
- Move some hack global types back into their proper module, to match the typescript sources.

There is a limitation in flow libdefs, which makes it awkward to define a class hierarchy that spans a module boundary. The following is an error:
```
declare module "@apollo/client/link/core" {
  declare export class ApolloLink {};
}

declare module "@apollo/client" {
  import type { ApolloLink } from "@apollo/client/link/core";
  declare export class HttpLink extends ApolloLink {}; // value-as-type error
```

I tried combinations of `import typeof` and `extends Class<Base>`, but these don't seem to solve the problem.

In this PR I left the error suppression `$FlowFixMe[value-as-type]` in the libdef. This lets the libdef source express the class hierarchy and module membership correctly to a human reader, and there are no incorrect flow diagnostics in application code that uses these types.


definitions/npm/@apollo/client_v3.x.x/flow_v0.142.x-/client_v3.x.x.js
```diff
@@ -70,13 +70,24 @@ declare interface GraphQL$ExecutionResult<
   data?: TData | null;
   extensions?: TExtensions;
 }
-
-// @apollo/client/link/core/types.d.ts
-// Globals to support global ApolloLink class
-declare type ApolloClient$Path = $ReadOnlyArray<string | number>;
+declare type GraphQL$VariableNode = Object;
```
Brianzchen, line 73: Shouldn't use `Object` as this resolves to `any`. Use `{ ...  }` or `{ [key: string]: any }` instead

definitions/npm/@apollo/client_v3.x.x/flow_v0.142.x-/client_v3.x.x.js
```diff
@@ -2777,3 +2700,130 @@ declare module "@apollo/client/testing" {
...
+  declare export function getDefaultValues(definition: OperationDefinitionNode | void): { [key: string]: any };
+
+  // @apollo/client/utilities/graphql/storeUtils.d.ts
+
+  declare export interface Reference {
+    +__ref: string;
+  }
+  declare export function makeReference(id: string): Reference;
+  declare export function isReference(obj: any): boolean;
+  declare export type StoreValue = number | string | string[] | Reference | Reference[] | null | void | void | Object;
```
Brianzchen, line 2767: Should tidy this type up a bit. `void | void` and `Object` should be `{ ... }` as mentioned above

definitions/npm/@apollo/client_v3.x.x/flow_v0.142.x-/client_v3.x.x.js
```diff
@@ -2777,3 +2700,130 @@ declare module "@apollo/client/testing" {
...
+  declare export interface StoreObject {
+    __typename?: string;
+    [storeFieldName: string]: StoreValue;
+  }
+  declare export function isGraphQL$DocumentNode(value: any): boolean;
+  declare export function valueToObjectRepresentation(
+    argObj: any,
+    name: GraphQL$NameNode,
+    value: GraphQL$ValueNode,
+    variables?: Object
```
Brianzchen, line 2777: Object

Brianzchen (changes requested):

Brianzchen:
Re `ApolloLink` issue. Could you declare the class in the top level and then export/use in the two module definitions?

definitions/npm/@apollo/client_v3.x.x/flow_v0.142.x-/client_v3.x.x.js
joshuanapoli: Fixed in 93a521257c7eb1afd802f07067c1b16e20cbbfa4

definitions/npm/@apollo/client_v3.x.x/flow_v0.142.x-/client_v3.x.x.js
joshuanapoli: Fixed in a7631809327d6a1d516410919564b2e7aa94f6f5

definitions/npm/@apollo/client_v3.x.x/flow_v0.142.x-/client_v3.x.x.js
joshuanapoli: Fixed in b3c6310a

joshuanapoli:
> Re `ApolloLink` issue. Could you declare the class in the top level and then export/use in the two module definitions?

The ApolloLink class depends on some other types, which leads to pulling a fair number of types into the global scope. The type dependencies are important in the application code, when we define application-specific links. I tried this in the original libdef, but I thought that it didn't turn out well. For example, there were challenges with name collisions. Anyway somehow I didn't end up with a correct inheritance tree, which caused flow errors in correctly written application code (the problem is now covered in the libdef test).

Another approach is to move ApolloLink and all of the classes the extend it to a single "private" module. Then re-export the classes from the correct module definitions. This would make the flow type checking "work" while correctly modeling the class inheritance. It would avoid creating new problems with symbol name collisions in the global scope. However, it does define a fictitious module definition, which might be frustrating to anyone reading the types or maybe when the IDE auto-detects a wrong include path.

joshuanapoli:
> Re `ApolloLink` issue. Could you declare the class in the top level and then export/use in the two module definitions?

One problem with with this approach is that `import type` is an error at the top level.

Brianzchen:
> > Re `ApolloLink` issue. Could you declare the class in the top level and then export/use in the two module definitions?
> 
> One problem with with this approach is that `import type` is an error at the top level of a library. I would like to use imported types (from "graphql" and "@apollo/client/...") in the ApolloLink definition.

Right I see

Brianzchen (approved):

PR #4567, opened 2024-02-15
[flow-enums-runtime] Add definition

Brianzchen:
Makes sense a core flow library to be typed, though is probably never expected to be used directly :) 

- Links to documentation: https://www.npmjs.com/package/flow-enums-runtime
- Link to GitHub or NPM: https://www.npmjs.com/package/flow-enums-runtime

Other notes:


gantoine (approved):
reasonable, approved 👍🏼

PR #4600, opened 2025-03-04
Add `rel` to `ReactDOM$linkProps`

mwiencek:
- Links to documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel


SamChou19815 (approved):

mwiencek:
I noticed it said merging was blocked due to the commit not being signed, so I force-pushed a new signed commit.

SamChou19815:
I will retry the job after we finished our releases. You have hit a rare occasion when the CI jobs that downloads flow binaries are run in the middle of new flow release

SamChou19815:
I will land for now. We get errors because the latest Flow binary removes most of the builtins. We can cleanup later by linking together dom, bom, and cssom

PR #4614, opened 2025-05-18
[flow|environment] Use mapping instead of overrides for createElement

dunnbobcat:
Rather than list out every tag name and element for every method, we can do it once with an object type. This is more flexible / reusable for other methods, and can be extended in the future for things like SVG. This also changes a few instances of `HTMLElement` to the more generic `Element`, which is all the spec guarantees.

Test Plan:
- `./quick_run_def_tests.sh`

- Links to documentation:
  - https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement
  - https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS
  - https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByTagName
  - https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByTagNameNS
  - https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector
  - https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll
  - https://developer.mozilla.org/en-US/docs/Web/API/Element/getElementsByTagName
  - https://developer.mozilla.org/en-US/docs/Web/API/Element/getElementsByTagNameNS
  - https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector
  - https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelectorAll
  - https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection

- Type of contribution: fix | refactor

### DIFF
--- a/definitions/environments/dom/flow_v0.261.x-/dom.js
+++ b/definitions/environments/dom/flow_v0.261.x-/dom.js
@@ -1059,7 +1059,7 @@ declare class Attr extends Node {
   localName: string;
 }
 
-declare class HTMLCollection<+Elem: HTMLElement> {
+declare class HTMLCollection<+Elem: Element> {
   @@iterator(): Iterator<Elem>;
   length: number;
   item(nameOrIndex?: any, optionalIndex?: any): Elem | null;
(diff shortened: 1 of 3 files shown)


definitions/environments/dom/flow_v0.261.x-/dom.js
```diff
@@ -4622,3 +4083,121 @@ declare class Notification extends EventTarget {
 
   close(): void;
 }
+
+
+type HTMLElementMap = {
```
SamChou19815, line 4088: This is also the approach taken by TS. Since we are already on this route, let's just align the name with TS as well. TS uses `HTMLElementTagNameMap`

SamChou19815 (approved):

PR #4621, opened 2025-07-26
[flow-typed|cli] Add environmentOnly option to run-tests

dunnbobcat:
Summary:
It would be helpful to have an option to only run the `environments` definition tests, so this adds one. I don't know if there's a better way to detect "environment" tests outside of inspecting the path, however.

Test Plan:
- `node cli/dist/cli.js run-tests --environmentOnly`

- Links to documentation:
- Link to GitHub or NPM:

Other notes:


gantoine (approved):
👍🏼

PR #4622, opened 2025-07-26
[flow|environment] Actually fix environment tests

dunnbobcat:
Summary:
This is my fault, but any time we break out definitions from an existing
module, we need to list the new dependencies alongside the original module. My
previous commits only tested the changed files, but that isn't sufficient for
environment changes.

This fixes all the environment tests, with the exception of `react.js`, which
complains that we are overwriting an existing definition within flow itself.

Test Plan:
- `node cli/dist/cli.js run-tests --environmentOnly`


pascalduez (approved):
Probably needs a rebase after https://github.com/flow-typed/flow-typed/pull/4621 was merged. 
Failing tests does not seem related at first glance?

dunnbobcat:
Looks like these got scooped up #4624 — my bad with the stacked diffs, I'm not well versed in the GitHub workflows yet!

